PR #7, opened 2022-04-02
Bug 570760 - Option to automatically add requirements to product-launch

HannesWell:
This PR resolves [Bug 570760](https://bugs.eclipse.org/bugs/show_bug.cgi?id=570760) 

It was first addressed in [Eclipse Gerrit Change 186834](https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/186834) but abandoned because it was not yet compelted at the time of the move.


HannesWell:
Last comment from @laeubi at the mentioned Gerrit:

> PDE is the consumer. If I create a product and check it into my code repository I don't like to tell the user he should set something in his launch config before launching the product ... This is an important part of the configuration and should be persisted.

I have not yet considered that point, but I agree with Christoph that this alone is a reason to persist the newly added attribute in the product, even tough we are not yet 100% certain that P2/Tycho will be able to handle this subsequently (but I think it is likely that this is possible).

Besides that I wonder if `autoIncludeRequirements=true` should be used as default for Plug-in and Feature based launches, because this is the ProductActions default for both cases, isn't it? I have to check that.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
```diff
@@ -681,6 +686,25 @@ public void setUseFeatures(boolean use) {
...
+		boolean old = fAutoIncludeRequirements;
+		fAutoIncludeRequirements = includeRequirements;
+		if (isEditable()) {
+			firePropertyChanged(P_INCLUDE_REQUIREMENTS_AUTOMATICALLY, Boolean.toString(old),
+					Boolean.toString(fAutoIncludeRequirements));
+		}
+	}
+
+	private boolean getAutoAddRequirementsDefault() {
+		return fUseFeatures; // true if features are used, else false
```
jhonnen, line 705: in my experience, tycho does not include requirements for feature based products

jhonnen (commented):
I'm not sure... Persisting that option is a good point, but that's basically what launch configurations are for.
To me the primary consumer for the product file is the product build, not the developer launching it locally. So I still would not add any option, that is not used by the product build in the near future.

But since PDE doesn't have a product build anymore and I'm not involved with Tycho, I'll leave that decision to you.
The code looks fine though (except for the unclear default).

laeubi:
> Persisting that option is a good point, but that's basically what launch configurations are for.

Why having products at all then? We could simply use "OSGi Framework Launch" config then

> To me the primary consumer for the product file is the product build  not the developer

I literally don't know anyone writing product files but never launching them in the IDE, for my daliy development launching products is what I do all day, and I almost never change/edit run configs for that.

HannesWell:
I checked how Tycho/P2 is behaving and verified that even for plug-in based launches requirements are included automatically.
The code was updated accordingly.
Consequently the default is changed for plug-in based products.

I'm in favor to submit this change now. I already created an issue to support that flag in p2: https://github.com/eclipse-equinox/p2/issues/76

The only thing I'm thinking about to change is the location of the `Include requirements` check-box. At the moment it is at the overview page next to the radio-buttons if the product is based on plug-ins or features. But I think it would fit better into the bottom of the contents tab.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
HannesWell: My experiments showed the opposite: Tycho includes requirements for feature-based products as well as for plugin-based products. But I have not yet followed it in the P2 source-code.

HannesWell:
> The only thing I'm thinking about to change is the location of the `Include requirements` check-box. At the moment it is at the overview page next to the radio-buttons if the product is based on plug-ins or features. But I think it would fit better into the bottom of the contents tab.

I have done that. The Overview-page is now unchanged and the `Contents` page looks like the following:
- For products based on Features:
grafik
- For products based on Plug-ins:
grafik

Please not that I have updated the label of the check-box to include optional dependencies when adding required plug-ins to make it clear that this check box is only considered by the button on the right. Not when automatically including requirements.

The check-box `Include required Plug-ins/Features automatically` is enabled by default (if not attribute is present).
For Plug-in based Products that changes how products are launched from the IDE (but reflects how Tycho builds such products).
For feature based products this is also changed compared to the behavior of the 2022-03 release, not because of this change but because Feature-based launches are now created for feature based products.

Please let me know what you think about this change.
I plan to submit this latest tomorrow evening so that it will be part of M3 (which I again expected one week later, somehow the milestones are always earlier than expected).

HannesWell:
No objections were raised, the build looks good and today is last day of development. So this is finally ready for submission.

HannesWell:
PR for the N&N entry:
https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/24

PR #8, opened 2022-04-03
Bug 571101: ApiAnalysisApplication - add support for p2-repo baseline and clean up API-Application

HannesWell:
This PR resolves [Bug 571101](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571101)

It was first addressed in [Eclipse Gerrit Change 176109](https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/176109) and [Gerrit Change 176108](https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/176108), both were abandoned because they were not completed at the time of the move to GH.

The first commit has to be tested if the new way of printing the API analysis result works well and for the second commit a corresponding UI to defined a P2 repo as API baseline directly is still missing.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
@@ -13,15 +13,20 @@
  *******************************************************************************/
 package org.eclipse.pde.api.tools.internal;
 
+import static java.util.function.Predicate.not;
+
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


vogella:
That feature will be nice as it  will remove the need to download the SDK for setting up the API baseline. Thanks for working on this @HannesWell

merks:
> That feature will be nice as it will remove the need to download the SDK for setting up the API baseline. Thanks for working on this @HannesWell

The Oomph setups already remove that need.  And because they reuse the shared bundle pool, setting up the API baseline takes just seconds.

HannesWell:
> > That feature will be nice as it will remove the need to download the SDK for setting up the API baseline. Thanks for working on this @HannesWell
> 
> The Oomph setups already remove that need. And because they reuse the shared bundle pool, setting up the API baseline takes just seconds.

That's right. I also mentioned this in the discussion at [the initial Gerrit](https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/176109).
The associated Bug was created with the intention to only improve the `ApiAnalysisApplication` with the goal to simplify the setup in the Maven/Tycho build, nevertheless Mickael suggested to add corresponding UI support even if Oomph provides similar/better support for this.

HannesWell:
> @HannesWell is this still interesting?

Not anymore.

PR #14, opened 2022-04-06
improve ClasspathUtilCore.getPath performance eclipse-pde#13

jukzi:
isAbsolute() is cheap, exists() is I/O. fail fast!

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathUtilCore.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathUtilCore.java
@@ -221,7 +221,7 @@ public class ClasspathUtilCore {
 				return new Path(file.getAbsolutePath());
 			}
 			file = new File(libraryName);
-			if (file.exists() && file.isAbsolute()) {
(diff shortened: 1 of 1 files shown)


HannesWell:
> Looks good. Will merge after m1 has been released

I can only agree with this.

> isAbsolute() is cheap, exists() is I/O. fail fast!

Thank you Jörg. I/O is indeed very often surprisingly slow and I already noticed it multiple times that devlopers are not aware of it (at least the write code like the one you have corrected).
But when you have to do I/O it can also make a big difference how you do it.
For example `File.copy()` is in my experience magnitudes faster (in my case, it was about 200 times) than transferring a `FileInputStream` to an `FileOutputStream`.

jukzi:
superseeded by 
https://github.com/eclipse-pde/eclipse.pde.ui/pull/15

mickaelistria:
> superseeded by https://github.com/eclipse-pde/eclipse.pde.ui/pull/15

So should we close this one?

EclipseNewb:
> So should we close this one?

if you agree to merge #15 instead- yes!

mickaelistria:
#15 looks good to me to be merged as soon as M1 is complete. So let's close this one.

PR #31, opened 2022-04-15
Fix #30: fix check on missing requirements in Equinox preLaunchCheck()

### LINKED ISSUE #30: EquinoxLaunchConfiguration (RAP launcher) always fails with "The org.eclipse.osgi plug-in is missing from this configuration."
From: [Bugzilla 579708](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579708) by @ifurnadjiev 

RAP launcher (based on org.eclipse.pde.launching.EquinoxLaunchConfiguration) is broken after the change https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/186860

Before the change in org.eclipse.pde.launching.EquinoxLaunchConfiguration.preLaunchCheck(ILaunchConfiguration, ILaunch, IProgressMonitor) the missing OSGi bundle has been added to fModels and fAllBundles automatically. After the change the code:
    if (!RequirementHelper.addApplicationLaunchRequirements(fModels, configuration)) {
			throw new CoreException(Status.error(PDEMessages.EquinoxLaunchConfiguration_oldTarget));
		}
always throws. For normal operation RequirementHelper.addApplicationLaunchRequirements returns false (as the missingRequirements there is false ).

For RAP launcher (RAP Tools) this issue is a blocker!

HannesWell:
This fixes #30.
`RequirementHelper.addApplicationLaunchRequirements()` now returns true if
all requirements are satisfied and false if requirements are missing.

Instead of reverting the return value of `addApplicationLaunchRequirements()` the if-condition in the callers could be reverted, but I think it is more meaningful and does align better e.g. with Java-standard Collection if true is returned in case all requirements are satisfied.

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/RequirementHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/RequirementHelper.java
@@ -88,7 +88,7 @@ public class RequirementHelper {
 		String pluginResolution = isFeatureBasedLaunch ? configuration.getAttribute(IPDELauncherConstants.FEATURE_PLUGIN_RESOLUTION, IPDELauncherConstants.LOCATION_WORKSPACE) : IPDELauncherConstants.LOCATION_WORKSPACE;
 
 		List<String> appRequirements = getApplicationLaunchRequirements(configuration);
-		boolean missingRequirements = false;
+		boolean allRequirementsSatisfied = true;
 		for (String requiredBundleId : appRequirements) {
 			ModelEntry entry = PluginRegistry.findEntry(requiredBundleId);
 			if (entry != null) {
@@ -99,10 +99,10 @@ public class RequirementHelper {
 					addPlugin.accept(plugin);
 				}
 			} else {
-				missingRequirements = true;
+				allRequirementsSatisfied = false;
 			}
 		}
-		return missingRequirements;
+		return allRequirementsSatisfied;
 	}
 
 	public static List<String> getProductRequirements(ILaunchConfiguration config) throws CoreException {


ifurnadjiev:
If the missing requirement is not the osgi bundle, but other... the error message presented to the user is wrong ("The org.eclipse.osgi plug-in is missing from this configuration.")

ifurnadjiev:
Also the osgi bundle is not added to fModels and fAllBundles in EquinoxLaunchConfiguration. As a result org.eclipse.pde.launching.EquinoxLaunchConfiguration.saveConfigurationFile(ILaunchConfiguration) fails on line 84

ifurnadjiev:
For me the fix is not good.

HannesWell:
> If the missing requirement is not the osgi bundle, but other... the error message presented to the user is wrong ("The org.eclipse.osgi plug-in is missing from this configuration.")

A EquinoxLaunchConfiguration only has `org.eclipse.osgi` as requirement (see the static initializer at the top) and it is not intended until now to give sub-classes the ability to define own requirements. Nevertheless technically it would be possible but you would then relay on non API that can change any time.
If this is what you want to do, please open a dedicated issue for that.


> Also the osgi bundle is not added to fModels and fAllBundles in EquinoxLaunchConfiguration. As a result org.eclipse.pde.launching.EquinoxLaunchConfiguration.saveConfigurationFile(ILaunchConfiguration) fails on line 84

For an Equinox-launch `org.eclipse.osgi` is added to `fModels` in `RequirementHelper.addApplicationLaunchRequirements()` and subsequently added to `fAllBundles` in `EquinoxLaunchConfiguration.preLaunchCheck()`.

If this is not added for a subclass of `EquinoxLaunchConfiguration` this is probably because you use another launchConfigurationType-Id for your subclass and the requirement of `org.eclipse.osgi` is therefore not found?
Please check if in `RequirementHelper.addApplicationLaunchRequirements()` `getApplicationLaunchRequirements()` returns a non-empty List (in line 90). If this is the case that should be handled in a dedicated issue too.

HannesWell:
> For me the fix is not good.

This PR itself is good, since it fixes a bug that happens for all Equinox based launches. The other points you mentioned are likely due to other issues.

ifurnadjiev:
Yes... RAP launcher has it's own type. Your recommendation is to add static initializer like:
static {
    RequirementHelper.registerSameRequirementsAsFor( "org.eclipse.rap.ui.launch.RAPLauncher", //$NON-NLS-1$
                                                     IPDELauncherConstants.OSGI_CONFIGURATION_TYPE );
}
Isn't it? The only small API problem is that RequirementHelper is internal class.

HannesWell:
> Yes... RAP launcher has it's own type. Your recommendation is to add static initializer like:
>static { RequirementHelper.registerSameRequirementsAsFor( "org.eclipse.rap.ui.launch.RAPLauncher", //$NON-NLS-1$ IPDELauncherConstants.OSGI_CONFIGURATION_TYPE ); } 
> Isn't it? 

Yes exactly, that workaround should fix the problem.

> The only small API problem is that RequirementHelper is internal class.

That's right, but I that problem is not too small. Given that previously the `org.eclipse.osgi` requirement was added by default even for subclasses of `EquinoxLaunchConfiguration`, this should now happen too. The mentioned workaround should not be necessary.

Would you mind to open a new issue that the `org.eclipse.osgi` requirement is not added for subclasses of `EquinoxLaunchConfiguration` anymore?

ifurnadjiev:
Done! Issue #32

PR #33, opened 2022-04-18
Add required o.e.osgi bundle for subclasses of EquinoxLaunchConfig too

HannesWell:
This PR aims to fix https://github.com/eclipse-pde/eclipse.pde.ui/issues/32.

My first attempt was to add the requirements associated to a launch-configuration-type to all sub-classes of the corresponding LaunchConfigurationDelegate. But besides the the launch-configuration this would also require the launche's mode to obtain the correct delegate and this is not always trivial and would require a medium amount of changes.
Furthermore it is not correct because just because a launch-delegate extends EclipseApplicationLaunchConfiguration it does not mean that the launch-type of the sub-class has the same requirements.
So I have limited this fix to the previous specialty of the Equinox-Launch to add the required o.e.osgi and ensured that it is added.

If there is a demand to register own requirements for custom Launch-Delegates we have to decide how to provide an API to do that. But this was not yet requested.

@ifurnadjiev can you test if this PR really satisfies your needs?

@jhonnen what do you think about this?

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
@@ -111,7 +111,8 @@ public class BundleLauncherHelper {
 
 	private static void addRequiredBundles(Map<IPluginModelBase, String> bundle2startLevel, ILaunchConfiguration configuration) throws CoreException {
 
-		RequirementHelper.addApplicationLaunchRequirements(bundle2startLevel, configuration);
+		List<String> appRequirements = RequirementHelper.getApplicationLaunchRequirements(configuration);
+		RequirementHelper.addApplicationLaunchRequirements(appRequirements, configuration, bundle2startLevel);
 
 		boolean includeOptional = configuration.getAttribute(IPDELauncherConstants.INCLUDE_OPTIONAL, true);
 		Set<BundleDescription> requiredDependencies = includeOptional //
@@ -167,7 +168,8 @@ public class BundleLauncherHelper {
 
 		if (addRequirements) {
 			// Add all missing  plug-ins required by the application/product set in the config
-			RequirementHelper.addApplicationLaunchRequirements(configuration, launchPlugins, launchPlugins::add);
+			List<String> appRequirements = RequirementHelper.getApplicationLaunchRequirements(configuration);
+			RequirementHelper.addApplicationLaunchRequirements(appRequirements, configuration, launchPlugins, launchPlugins::add);
 
 			// Get all required plugins
 			Set<BundleDescription> additionalBundles = DependencyManager.getDependencies(launchPlugins);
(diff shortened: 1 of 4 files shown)


ifurnadjiev:
Our tests are green again with this change and the RAP launcher is back to life.

HannesWell:
> Our tests are green again with this change and the RAP launcher is back to life.

Great. Thanks for testing.

> The change is OK I think. It's not great, that there are two different modes of requirement calculation, but since that was the previous behavior, we can't help that...

Indeed. But I did not found a better solution.

> 
> Is the `registerLaunchTypeRequirements` call in `EquinoxLaunchConfiguration` still used or can that be removed now?

It can be removed if we are sure every Plug-in that is launched will always in any case be wired to the `o.e.osgi` bundle.
If this is the case `BundleLauncherHelper.addRequiredBundles()` respectively `getMergedBundleMapFeatureBased()` will add a missing o.e.osgi when computing the requirements-closure if it has not been added before via `RequirementHelper.getApplicationLaunchRequirements()`.

I tried really hard to create a Plug-in that is not wired to `o.e.osgi` but failed, although I stripped everything of and the Manifest only contained the BSN, Version and the `Manifest-Version + Bundle-ManifestVersion` headers.
Nevertheless I'm not sure if it would be possible to wire a Plug-in to another OSGi-Systembundle, for example if one has Felix in his target-platform and not o.e.osgi?

@tjwatson could you help here?

HannesWell:
> Is the `registerLaunchTypeRequirements` call in `EquinoxLaunchConfiguration` still used or can that be removed now?

I tried really hard to create a scenario where a Plug-in is not wired to `org.eclipse.osgi` but as long as `org.eclipse.osgi` is present even a totally stripped down bundle is wired to it via an `osgi.ee` wire. Even when I added (what I think is) another OSGi-System bundle like Apache-Felix Framework (Maven Coordinates org.apache.felix:org.apache.felix.main) a plug-in is wired to `org.eclipse.osgi`. Only when I removed `org.eclipse.osgi` from the TP the wiring vanished, but then the enforced addition of `org.eclipse.osgi` in EquinoxLaunchConfiguration.preLaunchCheck() fails because o.e.osgi is missing.

With that I think there is indeed no need to call `registerLaunchTypeRequirements` anymore because `org.eclipse.osgi` is added anyways via the requirements-closure if requirements are added and it does not have to be specified separately as application requirement.

HannesWell:
Thank you Tom for the explanation.

> Note that PDE did have support for allowing other "system bundle" implementations such as Felix to be used by the resolver. 

So PDE does not offer this possibility anymore?
To me it looks like since [Bug 283731](https://bugs.eclipse.org/bugs/show_bug.cgi?id=283731) one can not specify another System-Bundles than org.eclipse.osgi anymore.

The only way I see at the moment to modify the MinimalState.systemBundle field is to have a bundle in the TP that specifies a header `Eclipse-SystemBundle=true` in its MANIFEST.MF. At least the `org.apache.felix.framework/main` bundle (not sure which is the correct one) does not do that, but others maybe do?

HannesWell:
Since it is in general possible to have another System-bundle (although it looks complicated to me, but it could be become simple in the future) I reverted the removal of the requirements registration.
If we are absolutely certain it is not necessary in any case we still can remove it in the future, but I would like to complete this change.

BTW. the build failure is likely due to some flaw in the SDK-prereq-TP. The I build is failing with the same error.

Issue #42, opened 2022-04-22
APi tools should import package ASM instead of require bundle it...


laeubi:
Currently API tools can become unresolvable if ASM 9.2 and ASM 9.3 are present see:
https://github.com/eclipse/xtext-eclipse/issues/1842

laeubi:
`org.eclipse.jdt.launching.javaagent` is another bundle in platform that uses require bundle ...

laeubi:
See also https://github.com/eclipse-jdt/eclipse.jdt.core/issues/36

PR #43, opened 2022-04-22
APi tools should import package ASM instead of require bundle it

laeubi:
Fix https://github.com/eclipse-pde/eclipse.pde/issues/42

FYI @bjhargrave @jonahgraham

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
@@ -12,10 +12,8 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.25.0,4.0.0)",
  org.eclipse.core.resources;bundle-version="[3.4.0,4.0.0)",
  org.eclipse.text;bundle-version="[3.4.0,4.0.0)",
  org.eclipse.core.filebuffers;bundle-version="[3.4.0,4.0.0)",
- org.objectweb.asm;bundle-version="[9.3.0,10.0.0)",
  org.eclipse.equinox.frameworkadmin;bundle-version="[2.0.0,3.0.0)",
- org.eclipse.core.variables;bundle-version="[3.2.0,4.0.0)",
- org.objectweb.asm.tree;bundle-version="[9.2.0,10.0.0)"
+ org.eclipse.core.variables;bundle-version="[3.2.0,4.0.0)"
 Export-Package: org.eclipse.pde.api.tools.internal;x-friends:="org.eclipse.pde.api.tools.tests,org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.generator",
  org.eclipse.pde.api.tools.internal.builder;x-friends:="org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.tests",
  org.eclipse.pde.api.tools.internal.comparator;x-friends:="org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.tests",
@@ -35,7 +33,10 @@ Export-Package: org.eclipse.pde.api.tools.internal;x-friends:="org.eclipse.pde.a
  org.eclipse.pde.api.tools.internal.util;x-friends:="org.eclipse.pde.api.tools.tests,org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.generator",
  org.eclipse.pde.api.tools.internal.util.profiles;x-internal:=true
 Bundle-RequiredExecutionEnvironment: JavaSE-11
-Import-Package: com.ibm.icu.util
+Import-Package: com.ibm.icu.util,
+ org.objectweb.asm;version="[9.3.0,10.0.0)",
+ org.objectweb.asm.signature;version="[9.3.0,10.0.0)",
+ org.objectweb.asm.tree;version="[9.3.0,10.0.0)"
 Bundle-Activator: org.eclipse.pde.api.tools.internal.provisional.ApiPlugin
 Eclipse-LazyStart: true
 Bundle-ActivationPolicy: lazy


apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
```diff
@@ -35,7 +33,10 @@ Export-Package: org.eclipse.pde.api.tools.internal;x-friends:="org.eclipse.pde.a
  org.eclipse.pde.api.tools.internal.util;x-friends:="org.eclipse.pde.api.tools.tests,org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.generator",
  org.eclipse.pde.api.tools.internal.util.profiles;x-internal:=true
 Bundle-RequiredExecutionEnvironment: JavaSE-11
-Import-Package: com.ibm.icu.util
+Import-Package: com.ibm.icu.util,
+ org.objectweb.asm;version="[9.2.0,9.4.0)",
```
jonahgraham, line 37: Make these all `[9.3.0,9.4.0)`?

laeubi: Don't know, it previously has required something from 9.2 and 9.3 so maybe it was meant to be compatible with both... so if upgrading from 9.2 > 9.3 does not required any code changes the lower bound would be fine here.

akurtakov: 9.3 is required for Java 19 work.

bjhargrave: The intersection of the require bundle versions is `[9.3.0,10.0.0)` and this is what you want for the import package ranges of the asm packages (since ASM seems to version their bundles and packages identically). `[9.2.0,9.4.0)` or `[9.3.0,9.4.0)` is excessively narrow.

jonahgraham: I think the 9.3 lower bound update on asm.tree was simply missed in https://github.com/eclipse-pde/eclipse.pde/commit/404eee00f455b45956705070906f40726c2219a2

laeubi: For example `ReferenceExtractor` extends `ClassVisitor` what is an ASM class (there are more) so we need to use provider version range I think here what would be `[9.3.0,9.4.0)`?

apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
```diff
@@ -35,7 +33,10 @@ Export-Package: org.eclipse.pde.api.tools.internal;x-friends:="org.eclipse.pde.a
  org.eclipse.pde.api.tools.internal.util;x-friends:="org.eclipse.pde.api.tools.tests,org.eclipse.pde.api.tools.ui,org.eclipse.pde.api.tools.generator",
  org.eclipse.pde.api.tools.internal.util.profiles;x-internal:=true
 Bundle-RequiredExecutionEnvironment: JavaSE-11
-Import-Package: com.ibm.icu.util
+Import-Package: com.ibm.icu.util,
+ org.objectweb.asm;version="[9.3.0,9.4.0)",
```
iloveeclipse, line 37: Please use [9.3.0, 10.0.0) also below

laeubi: as mentioned before I think `[9.3.0, 10.0.0)` is the wrong version range if we extend classes of the `org.objectweb.asm` package.

> Providers implementing an API package must import with a version range that specifies the floor of the version used for compilation and the ceiling of the next minor part.

http://docs.osgi.org/whitepaper/semantic-versioning/060-importer-policy.html

apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
HannesWell: > For example `ReferenceExtractor` extends `ClassVisitor` what is an ASM class (there are more) so we need to use provider version range I think here what would be `[9.3.0,9.4.0)`?

In that case you are right. Even if it requires more version bumps, but it has to be done.

apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
iloveeclipse: ... and as mentioned before, this is too narrow range. ASM plans to make breaking API changes in 10 version, not in 9.x, see https://mail.ow2.org/wws/arc/asm/2022-04/msg00000.html

HannesWell (approved):

apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
bjhargrave: > For example `ReferenceExtractor` extends `ClassVisitor` what is an ASM class (there are more) so we need to use provider version range I think here what would be `[9.3.0,9.4.0)`?

But no one cared about this before since the bundle-version was [9.3.0,10.0.0) :-)

You are not a provider of the ASM API. You are a consumer of the API and you can rely upon the provider of the API (ASM) to not break consumer types like ClassVisitor without a major version bump. See <https://docs.osgi.org/whitepaper/semantic-versioning/060-importer-policy.html>.

apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
bjhargrave: @laeubi you are misreading the whitepaper. Here, you are a consumer of the API using a ConsumerType class in the way expected by the API. The API provider will not break you without a major version bump.

HannesWell: Do I understand it right, that the only scenario where it's reasonable to use a minor version upper-bound is when you are a 'full' implementer of an API (e.g. like Equinox implements the OSGi API)?
Implementing/extending classes that are intended to be subclassed by consumers does not require the restrict to minor upper-bounds? But a provider of a class intended to be extended/implemented by consumers has to obey the same to SemVer rules when adding new methods to that type like when removing methods from a 'referenced' only API (meaning major-version change in case of addition)?

laeubi: > Here, you are a consumer of the API using a ConsumerType class in the way expected by the API. The API provider will not break you without a major version bump.

I was just a bit curious as the do not use the `@ConsumerType` and `@ProviderType` in their code but the javadoc of `@ConsumerType` clearly states:

> A type is assumed to be [ConsumerType](http://docs.osgi.org/javadoc/r6/annotation/org/osgi/annotation/versioning/ConsumerType.html) if it is not marked either ConsumerType or [ProviderType](http://docs.osgi.org/javadoc/r6/annotation/org/osgi/annotation/versioning/ProviderType.html)

laeubi:
Is there no CI checks to be triggered here?

akurtakov:
Jenkins job needed a change after rename of the repo - done now.

Issue #53, opened 2022-04-28
Build slow while MANIFEST.MF editor open


jukzi:
I have a MANIFEST.MF file open in the PDE Editor. Then i started Project/Clean/"Clean all Projects" in a big workspace. After that the progress was stuck for ~20min at 100% Cleaning all projects.
![image](https://user-images.githubusercontent.com/51790620/165737834-4e3af20d-306a-4ee4-a521-f95bf3531896.png)

I see the open PDE editor is informed about every single ".class" file changed during the build "display.asyncExec"

```
 Thread [Worker-72: Cleaning all projects] (Suspended (breakpoint at line 234 in org.eclipse.pde.internal.ui.editor.context.InputContextManager)) 
 org.eclipse.pde.internal.ui.editor.plugin.PluginInputContextManager(org.eclipse.pde.internal.ui.editor.context.InputContextManager).asyncStructureChanged(org.eclipse.core.resources.IFile, boolean) line: 234 
 org.eclipse.pde.internal.ui.editor.plugin.PluginInputContextManager(org.eclipse.pde.internal.ui.editor.context.InputContextManager).lambda$0(org.eclipse.core.resources.IResourceDelta) line: 218  
 org.eclipse.pde.internal.ui.editor.context.InputContextManager$$Lambda$1254.1386710085.visit(org.eclipse.core.resources.IResourceDelta) line: not available  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 67  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
 org.eclipse.core.internal.events.ResourceDelta.accept(org.eclipse.core.resources.IResourceDeltaVisitor, int) line: 77  
[...]

PR #54, opened 2022-04-28
Build slow while MANIFEST.MF editor open

merks:
Improve the processing of the resource delta to avoid an asyncExec call
for each added and removed file but rather use a single asyncExec call
to process all added and removed files at once.

https://github.com/eclipse-pde/eclipse.pde/issues/53


vogella:
@[jukzi](https://github.com/jukzi) please review.

merks:
The failing test

https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/PR-54/1/testReport/junit/org.eclipse.pde.core.tests.internal/WorkspaceModelManagerTest/testDelete_projectWithChangedBundleRoot/

works locally and org.eclipse.pde.internal.ui.editor.context.InputContextManager.resourceChanged(IResourceChangeEvent) is not called during that test, so it seems unrelated.

Any ideas?

iloveeclipse:
There are some asynchronous issues with PDE that shows up as PDE test fails. I've retriggerred the build.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,30 +208,42 @@ public void resourceChanged(IResourceChangeEvent event) {
...
 
-	private void asyncStructureChanged(final IFile file, final boolean added) {
+	private void asyncStructureChanged(List<IFile> added, List<IFile> removed) {
 		if (editor == null || editor.getEditorSite() == null)
 			return;
 		Shell shell = editor.getEditorSite().getShell();
 		Display display = shell != null ? shell.getDisplay() : Display.getDefault();
 
-		display.asyncExec(() -> structureChanged(file, added));
+		display.asyncExec(() -> {
```
iloveeclipse, line 239: I would propose here is to schedule UIJob, which shows progress, checks for cancellation and so can be terminated by user in case the loop never ends here (100000 changes may take a while to process - and we are on UI thread).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,30 +208,42 @@ public void resourceChanged(IResourceChangeEvent event) {
...
 			PDEPlugin.logException(e);
 		}
 	}
 
-	private void asyncStructureChanged(final IFile file, final boolean added) {
+	private void asyncStructureChanged(List<IFile> added, List<IFile> removed) {
 		if (editor == null || editor.getEditorSite() == null)
 			return;
 		Shell shell = editor.getEditorSite().getShell();
 		Display display = shell != null ? shell.getDisplay() : Display.getDefault();
```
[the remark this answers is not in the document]
merks: Again, it was that way before...

But yes, the code is rather awful and  it strikes me that PlatformUI.getWorkbench().getDisplay() would be a much simpler way to get a display.   The existing logic doesn't check isDisposed on the Display, but I suppose that's not possible while processing a resource delta...

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
iloveeclipse: Well, the bug says: there is a huge number of UI tasks, and so moving from 100_000 single tasks that block because of the inefficient scheduling to a single task that blocks because of the work done by editor is not a big improvement.

So if we would give user a chance to cancel, so why not? However, I've not tried, may be not a UI job but ProgressService would be better here - because cancellation of a job if UI is busy is surely not possible. 

Something like `PlatformUI.getWorkbench().getProgressService().runInUI(PDEPlugin.getActiveWorkbenchWindow(), runnable, null);`

Ideally I would check if the `structureChanged()` need to run in UI at all, or to which extent it depends on UI and move as much as possible work from UI thread.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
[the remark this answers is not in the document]
iloveeclipse: Ed, dispose checks are possible from non UI thread, but if it is not disposed, it doesn't mean anything, because it has to be checked again at the task start time in UI thread.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
merks: How much baggage do I need for a UIJob?  E.g., can it be like this:

https://github.com/eclipse-pde/eclipse.pde/blob/96d307ba51b304cf51bd245e5f00a08657050c6c/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/target/TargetDefinitionContentPage.java#L271

Or would I need a new job name that's translatable?

And again, what's the effect of cancel on the state displayed to the user?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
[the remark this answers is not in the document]
merks: That makes you wonder how it's ever safe to call asyncExec because at any point the display can be disposed
```
public void asyncExec (Runnable runnable) {
	synchronized (Device.class) {
		if (isDisposed ()) error (SWT.ERROR_DEVICE_DISPOSED);
		synchronizer.asyncExec (runnable);
	}
}
```
But digressing into theoretical problems that can't really happen in practice won't serve us well, will it?   After all, once the display itself is disposed, the user won't see anything ever again.  So let's not go there.

For now I force pushed a change that is not error prone.  I'm not so convinced a UIJob will serve us well, because I don't think not updating the visible state will be good for the user...

Please comment on the newer changes if you still feel strongly about that.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,30 +206,38 @@ public void resourceChanged(IResourceChangeEvent event) {
...
-		if (editor == null || editor.getEditorSite() == null)
-			return;
-		Shell shell = editor.getEditorSite().getShell();
-		Display display = shell != null ? shell.getDisplay() : Display.getDefault();
-
-		display.asyncExec(() -> structureChanged(file, added));
+	private void asyncStructureChanged(List<IFile> added, List<IFile> removed) {
+		Display display = PlatformUI.getWorkbench().getDisplay();
+		display.asyncExec(() -> {
+			for (IFile file : added) {
```
iloveeclipse, line 234: Right before this for loop please add check if the editor is still alive.

iloveeclipse (changes requested):

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,30 +208,48 @@ public void resourceChanged(IResourceChangeEvent event) {
...
+		if (monitoredFiles == null)
 			return;
-		Shell shell = editor.getEditorSite().getShell();
-		Display display = shell != null ? shell.getDisplay() : Display.getDefault();
-
-		display.asyncExec(() -> structureChanged(file, added));
+		try {
+			IProgressService progressService = PlatformUI.getWorkbench().getProgressService();
+			progressService.runInUI(progressService, monitor -> {
+				for (IFile file : added) {
```
iloveeclipse, line 239: This looks better now (haven't tested), but **please** as the first task in the UI thread, check if the editor is still there.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,30 +208,48 @@ public void resourceChanged(IResourceChangeEvent event) {
...
 
 		try {
+			List<IFile> added = new ArrayList<>();
+			List<IFile> removed = new ArrayList<>();
 			delta.accept(delta1 -> {
-				int kind = delta1.getKind();
 				IResource resource = delta1.getResource();
-				if (resource instanceof IFile) {
+				if (resource.getType() == IResource.FILE) {
+					int kind = delta1.getKind();
```
iloveeclipse, line 216: Please filter files (match against actually monitored) before pushing everyone to UI thread (see this is currently done here: https://github.com/eclipse-pde/eclipse.pde/blob/fdbfcab2f6d2013bad76c187b5e47fcc6324ecd8/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java#L256 )

With that we will only have ~5 files to be processed in UI, not 100_000.

iloveeclipse (changes requested):
I've finally had a time to look at the code in the IDE, it seems we can skip most of the files from very beginning, before pushing them to UI thread.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -208,44 +207,55 @@ public void resourceChanged(IResourceChangeEvent event) {
 		IResourceDelta delta = event.getDelta();
 
 		try {
+			List<IFile> added = new ArrayList<>();
+			List<IFile> removed = new ArrayList<>();
 			delta.accept(delta1 -> {
-				int kind = delta1.getKind();
 				IResource resource = delta1.getResource();
-				if (resource instanceof IFile) {
+				if (monitoredFiles.contains(resource)) {
```
iloveeclipse, line 214: This is almost perfect, just we do not need to perform 100_000 times lookup in a map if the delta isn't interesting for us (which is true in most cases).
So consider this slight modification:
```
diff --git a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
index c14d606..55e5538 100644
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
@@ -212,9 +212,13 @@
 			delta.accept(delta1 -> {
+				int kind = delta1.getKind();
+				if (kind != IResourceDelta.ADDED && kind != IResourceDelta.REMOVED) {
+					return true;
+				}
 				IResource resource = delta1.getResource();
 				if (monitoredFiles.contains(resource)) {
-					int kind = delta1.getKind();
-					if (kind == IResourceDelta.ADDED)
+					if (kind == IResourceDelta.ADDED) {
 						added.add((IFile) resource);
-					else if (kind == IResourceDelta.REMOVED)
+					} else if (kind == IResourceDelta.REMOVED) {
 						removed.add((IFile) resource);
+					}
 					return false;

```

iloveeclipse (approved):
Next time please use "{ }" for if/else blocks. 
If this would be gerrit, I would fix this in a second, but it is not gerrit, so let merge that PR.

merks:
> Next time please use "{ }" for if/else blocks. If this would be gerrit, I would fix this in a second, but it is not gerrit, so let merge that PR.

I agree.   I prefer to use braces always too, so I aborted the build and did that too.

There's a bunch of other ugly stuff in this class too, like a HashTable and I have no idea why the value type for it Object and not InputContext.  Yuck!  But that is a different topic...

vogella:
> There's a bunch of other ugly stuff in this class too, like a HashTable

While you add it, please send another PR for this, otherwise it will stay like this forever.

merks:
The build failure looks unrelated so I will ignore it.

SudhindraKKulkarni:
The changes made as part of this fix have resulted in the issue https://github.com/eclipse-pde/eclipse.pde/issues/769

Issue #55, opened 2022-04-29
Test failure in I-build


akurtakov:
In the last few I-builds we have test failure:
https://download.eclipse.org/eclipse/downloads/drops4/I20220428-1800/testresults/html/org.eclipse.pde.ui.templates.tests_ep424I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

```
Generated product fails validation: [Missing Constraint: Import-Package: org.osgi.util.function; version="[1.0.0,2.0.0)", Missing Constraint: Require-Bundle: org.osgi.util.promise; bundle-version="[1.2.0,1.3.0)", Missing Constraint: Import-Package: org.osgi.util.promise; version="[1.0.0,2.0.0)", Missing Constraint: Require-Bundle: org.osgi.util.position; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.measurement; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.xml; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.function; bundle-version="[1.2.0,1.3.0)"]

java.lang.AssertionError: Generated product fails validation:
[Missing Constraint: Import-Package: org.osgi.util.function; version="[1.0.0,2.0.0)", Missing Constraint: Require-Bundle: org.osgi.util.promise; bundle-version="[1.2.0,1.3.0)", Missing Constraint: Import-Package: org.osgi.util.promise; version="[1.0.0,2.0.0)", Missing Constraint: Require-Bundle: org.osgi.util.position; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.measurement; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.xml; bundle-version="[1.0.0,1.1.0)", Missing Constraint: Require-Bundle: org.osgi.util.function; bundle-version="[1.2.0,1.3.0)"]
at org.junit.Assert.fail(Assert.java:89)
at org.eclipse.pde.ui.templates.tests.TestPDETemplates.validateProduct(TestPDETemplates.java:196)
at jdk.internal.reflect.GeneratedMethodAccessor45.invoke(Unknown Source)
[...]

PR #56, opened 2022-04-29
Minor changes to TestPDETemplates

akurtakov:
* Stop using assertThat(message, variable equalTo(constant). It's just
complicated way to write assertEquals(message, constant, variable) which
even calls out to yet another library.
* Remove needless throws declaration.

Seen while investigating #55

Change-Id: If4b7a6937a9ec9acac88e435c21c52f7e7e6f1d8


akurtakov:
Same test failure as in the open issue so pushing this one as first step towards a fix.

PR #57, opened 2022-04-29
Avoid the use of HashTable in InputContextManager

merks:
Use Map<IEditorInput, InputContext> instead of Hashtable<IEditorInput,Object> because the value type is known and then all iterations over the map's values can be simplified.


iloveeclipse:
Note, Hashtable is synchronized, HashMap not. So if you want migrate from Hashtable you probably want to check if the code can be used in multiple threads (I think so here), so probably you want use ConcurrentHashMap or guard access via some locks etc.

merks:
@iloveeclipse I did confirm that modifications to the map happen only on the UI thread so I think there is no concern then, right?

iloveeclipse:
Sure.

vogella:
Thanks, Ed

Ed Merks ***@***.***> schrieb am Fr., 29. Apr. 2022, 17:32:

PR #71, opened 2022-05-03
Fully reflect merge of pde.build in PDE's Oomph setup

HannesWell:
This PR inlines the `PDE Core` Project and adds a `PDE Build` Working-Set.
Furthermore it removes the 'core' element from ids.

Just like for the renaming of the project from PDE UI to PDE Core we have to provide a way to inform users that they have to update the Project-stream referenced in their workspace setup. Otherwise the setup is broken.
@merks can you please review this?

### DIFF
--- a/releng/org.eclipse.pde.setup/PDE.setup
+++ b/releng/org.eclipse.pde.setup/PDE.setup
@@ -58,7 +58,7 @@
   <setupTask
       xsi:type="setup:EclipseIniTask"
       option="-Doomph.redirection.pde"
-      value="=https://raw.githubusercontent.com/eclipse-pde/eclipse.pde/master/releng/org.eclipse.pde.setup/->${github.clone.pde.core.location|uri}/releng/org.eclipse.pde.setup/"
+      value="=https://raw.githubusercontent.com/eclipse-pde/eclipse.pde/master/releng/org.eclipse.pde.setup/->${github.clone.pde.location|uri}/releng/org.eclipse.pde.setup/"
       vm="true">
     <description>Set an Oomph redirection system property to redirect the logical location of this setup to its physical location in the Git clone.</description>
   </setupTask>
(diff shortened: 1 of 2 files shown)


merks:
FYI, I had to commit this as well because the SDK configuration references the refactored project too.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/41ba8b732b83db85f7ce54f37b74b4c474211da7

HannesWell:
Thanks for taking care of that!

merks:
BTW, with my recent changes to the aggregator setup, these things are visible now within projects:

image

Making it easier to notice and to change.

HannesWell:
That's handy! Until now I had to import them manually.

What would be even more useful to test changes of the eclipse-sdk-prereqs.start would be if the content of that file would also be included into the activated modular target-platform if present. At the moment I manually added a 'target'-type target referring to that file (but interestingly that additional content stays even when I re-run the setup, which is good in my case).

merks:
In the past such a thing was not possible , but now in principle,  it  could be composed though some people have trouble getting that to work and I don't have time to make such a thing work:

https://www.eclipse.org/forums/index.php/mv/msg/1110731/1852217/#msg_1852217

It's currently a simple matter of switching the active target in the preferences (and then wait and wait for all the bundles that you already have downloaded to be downloaded yet again)...

HannesWell:
> In the past such a thing was not possible , but now in principle, it could be composed though some people have trouble getting that to work and I don't have time to make such a thing work:
> 
> https://www.eclipse.org/forums/index.php/mv/msg/1110731/1852217/#msg_1852217
> 

Since I have a relativly simple workaround for my problem, I'm fine at the moment. :)

> It's currently a simple matter of switching the active target in the preferences (and then wait and wait for all the bundles that you already have downloaded to be downloaded yet again)...

Yep, I think we should consider to use the shared Bundle pool in PDE for the Target-Platform resolution, if one is available.
I have the intention to try that in the future (did not have the time yet) or do you think there are general things that make it impossible?

Issue #88, opened 2022-05-10
NoSuchElementException being thrown when showing old launch configs


martinlippert:
I am trying to open and edit old launch configs with the latest Eclipse 4.24 M2 bits and pieces, but I am getting this exception and it looks like the selection of plugins on the "Plug-ins" tab is gone (the list is full of un-checked plugins), so there is some data getting lost here.

```
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.orElseThrow(Optional.java:377)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getSelectedModels(BundleLauncherHelper.java:435)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getBundleMap(BundleLauncherHelper.java:417)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getWorkspaceBundleMap(BundleLauncherHelper.java:380)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getAllSelectedPluginBundles(BundleLauncherHelper.java:107)
	at org.eclipse.pde.internal.ui.launcher.PluginBlock.initializePluginsState(PluginBlock.java:91)
	at org.eclipse.pde.internal.ui.launcher.PluginBlock.initializeFrom(PluginBlock.java:61)
	at org.eclipse.pde.internal.ui.launcher.BlockAdapter.initializeFrom(BlockAdapter.java:66)
	at org.eclipse.pde.ui.launcher.PluginsTab.initializeFrom(PluginsTab.java:158)
	at org.eclipse.pde.ui.launcher.AbstractPDELaunchConfigurationTabGroup.lambda$0(AbstractPDELaunchConfigurationTabGroup.java:63)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
	at org.eclipse.pde.ui.launcher.AbstractPDELaunchConfigurationTabGroup.initializeFrom(AbstractPDELaunchConfigurationTabGroup.java:55)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupWrapper.initializeFrom(LaunchConfigurationTabGroupWrapper.java:182)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupViewer.displayInstanceTabs(LaunchConfigurationTabGroupViewer.java:868)
[...]

PR #93, opened 2022-05-12
Don't throw NSEE for missing workspace-plugins in Launch-configs

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/88


vik-chand:
@HannesWell  Can you please verify this fix. Thanks

HannesWell:
> @HannesWell Can you please verify this fix. Thanks

Already done and reported to the issue: https://github.com/eclipse-pde/eclipse.pde/issues/88#issuecomment-1128465966

Since with Bugzilla+Gerrit we always reported the verification I assumed I' will do it in the issue instead of the PR. Not sure what's the definite place to do this.

HannesWell:
Understand. Then for completeness :)

Verified with Eclipse SDK

Version: 2022-06 (4.24)
Build id: I20220516-1800

PR #95, opened 2022-05-12
Log the exception when failing to update the target profile cache

pzi-dsa:
An error during this step causes the number of available artifacts to be
empty. So the error log should reflect that something went wrong.


vogella:
LGTM, thanks @pzi-dsa 

@HannesWell any concerns here?

HannesWell (approved):
Looks good to me as well.
I see no reason to not log this serious failure.

HannesWell:
Thank you @pzi-dsa for your contribution.

PR #99, opened 2022-05-12
Only single build per branch/PR.

akurtakov:
And kill the previous if there is new one scheduled.


akurtakov:
Merge as this is non-code change improving infrastructure usage.

Issue #49, opened 2022-04-26
Bug 579742 Unable to start Eclipse application from older target platform since 2022-03 due to security manager issue


SarikaSinha:
As part of [bug #575210](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575210), a PDE launch configuration automatically adds -Djava.security.manager=allow to the command line as soon as the target JRE is Java 17. However, this breaks launching Eclipse applications that are based on an older version of Eclipse (such as 2021-03) because they try to load the class "allow" for the security manager but that obviously doesn't exist (org.eclipse.osgi.internal.framework.SystemBundleActivator.installSecurityManager(EquinoxConfiguration)):

org.osgi.framework.BundleException: Exception in org.eclipse.osgi.internal.framework.SystemBundleActivator.start() of bundle org.eclipse.osgi.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:835)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:763)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1012)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule.initWorker(EquinoxBundle.java:198)
	at org.eclipse.osgi.container.SystemModule.init(SystemModule.java:99)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle.init(EquinoxBundle.java:281)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle.init(EquinoxBundle.java:265)
	at org.eclipse.osgi.launch.Equinox.init(Equinox.java:204)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.startup(EclipseStarter.java:316)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:251)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
[...]

PR #100, opened 2022-05-12
Use PDEState target models to compare version to set security manager allow flag

SarikaSinha:
fixes https://github.com/eclipse-pde/eclipse.pde/issues/49

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
@@ -26,11 +26,11 @@ import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.launching.*;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.plugin.TargetPlatform;
-import org.eclipse.pde.internal.core.ICoreConstants;
-import org.eclipse.pde.internal.core.TargetPlatformHelper;
+import org.eclipse.pde.internal.core.*;
 import org.eclipse.pde.internal.core.builders.PDEMarkerFactory;
 import org.eclipse.pde.internal.launching.*;
 import org.eclipse.pde.internal.launching.launcher.*;
+import org.osgi.framework.Version;
 
 /**
  * An abstract launch delegate for PDE-based launch configurations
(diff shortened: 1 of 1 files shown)


SarikaSinha:
@vik-chand Please review

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
```diff
@@ -155,21 +155,24 @@ private String[] updateVMArgumentWithAdditionalArguments(String[] args, boolean
...
-			vmInstall = VMHelper.getVMInstall(configuration);
-			if (vmInstall instanceof AbstractVMInstall) {
-				AbstractVMInstall install = (AbstractVMInstall) vmInstall;
-				String vmver = install.getJavaVersion();
-				if (vmver != null && JavaCore.compareJavaVersions(vmver, JavaCore.VERSION_17) >= 0) {
-					if (!argumentContainsAttribute(args, allowSecurityManager)) {
-						addAllowSecurityManager = true;
-						argLength++; // Need to add the argument
+
+		if (isEclipseBundleGreaterThanVersion(4, 24)) { // Don't add allow flags for eclipse before 4.24
```
HannesWell, line 159: The int-arguments are not used. Furthermore I suggest to rename this method to something like `isSecurityManagerAllowSupported()` and to make the local variable `Version comparedVersion = new Version(4, 24, 0);` a static constant with an expressive name like `SECURITY_MANAGER_ALLOW_MIN_VERSION`.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
```diff
@@ -189,6 +192,29 @@ private String[] updateVMArgumentWithAdditionalArguments(String[] args, boolean
 		return args;
 	}
 
+
+	private boolean isEclipseBundleGreaterThanVersion(int major, int minor) {
+		PDEState pdeState = TargetPlatformHelper.getPDEState();
+		if (pdeState != null) {
+			try {
+				Optional<IPluginModelBase> platformBaseModel = Arrays.asList(pdeState.getTargetModels()).stream().filter(x -> Objects.nonNull(x.getBundleDescription())).filter(x -> ("org.eclipse.platform").equals(x.getBundleDescription().getSymbolicName()))//$NON-NLS-1$
```
HannesWell, line 200: In reply to the conversation at the same place in the previous PR:

> ok, we will still need one not null filter(x -> Objects.nonNull(x.getBundleDescription()) before running through x.getBundleDescription().getSymbolicName().

Yes. Furthermore you could use `Arrays.stream()` instead of `Arrays.asList().stream()`.

> I also prefer to go the normal for each loop way for read and debug but this might be faster I guess and if we provide normal for loop codes, we get suggestions to use new ways to attract contributions.

As somebody who contributes for not very long I think that always using the newest thing just because it is newer does not attract more contributors.
IMHO clean code that uses the 'best' (fastest and especially most readable) style attracts more contributors. Sometimes this is a stream, sometimes a classic for-loop and sometimes a nested block should be refactored into a separate method.
But what's readable is of course also a matter of preference. :)

Besides that: Just checking all target-models is not the 100% right thing to do. Actually you should check the list of plug-ins that participate in the launched application since somebody could also have "org.eclipse.platform" in its workspace and if "org.eclipse.platform" is not a singleton there could be multiple versions in the active TP.

Unfortunately only the sub-classes have a List/Map/Set of actually launched plug-ins. So an abstract method to get the List<IPluginModelBase> of launched o.e.platform plug-ins has to be added and that list can be checked (e.g. if all plug-ins are above version `4.24`.

The question is if we want to make this method abstract and protected so that all sub-classes of `AbstractPDELaunchConfiguration` have to implement it or if we not make it abstract and only package and return a default value (that always results either in a true or false result?) only `EquinoxLaunchConfiguration` and `EclipseApplicationLaunchConfiguration` can overwrite it with the really launched plug-ins.

HannesWell (changes requested):
Thank you for your work.

I really appreciate the new commit-message style. Having a long URL as head line is IMHO not very expressive and I already wanted to suggest the style you already have applied.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
SarikaSinha: I have updated the PR to use incoming arguments and avoid List method.
Method name wise I feel this is good enough and there may be other checks in future.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
SarikaSinha: Current change is a good workaround for the problem the user is facing,  We can discuss further in a new issue and take the best approach.
@vik-chand WDYT ?

SarikaSinha: Created the followup issue https://github.com/eclipse-pde/eclipse.pde/issues/105

PR #106, opened 2022-05-14
Update launch-config when switching from feature to plug-in product

HannesWell:
If a user has created a launch-configuration for a feature-based product and changes that feature to be based on plug-ins and then launches the product from the product-editor, the attribute to use features or plug-ins is not updated so the launch-config is still based on feature.

This PR fixes that and also makes the order in which the launch-config attributes are set more meaning full.


HannesWell:
The actual freeze-time usually starts on Monday night, so this should be OK to be submitted.

HannesWell:
Verified with Eclipse SDK for feature -> plug-in based and the other way round (which worked before already)

Version: 2022-06 (4.24)
Build id: I20220516-1800

PR #107, opened 2022-05-14
Ignore optional-requirements by default for plugin-product launches

HannesWell:
To create similar applications like when build with Tycho, optional-requirements should be ignored by default when assembling the bundles that are included in a plug-in based product-application launched from the IDE.
If a user want's to include optional dependencies it can be re-enabled in the launch-config dialog by checking the corresponding box. 
The attribute is only set for newly created launch-configs to not change the value of existing configs that were set manually.

@jhonnen what do you think? This changes the default behavior for plug-in based products and should then be mentioned in the N&N.


jhonnen (approved):
sounds good to me

HannesWell:
Thank you Julian.
Since the actual freeze period usually should start this evening this is good to go now.

HannesWell:
Verified with Eclipse SDK

Version: 2022-06 (4.24)
Build id: I20220516-1800

HannesWell:
PR for the N&N entry:
https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/24

PR #110, opened 2022-05-15
Minor speed-up of test if launch requires SWT

HannesWell:
FYI @jhonnen, when I adapted the implementation of this test to the improved `DependencyManager`, we agreed that this test is not ideally implemented. I think this is a better solution. :)
Although I expect no noticeable performance improvement because the call will still be dominated by `DependencyManager.getSelfAndDependencies()`.


HannesWell:
Code-Freeze is usually Monday evening, so this should be good to go now.

HannesWell:
Verified with Eclipse SDK

Version: 2022-06 (4.24)
Build id: I20220516-1800

PR #111, opened 2022-05-15
Use static Collection factories

HannesWell:
Searched for ".unmodifiable(" and "Arrays.asList(" to find suitable applications.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.api.tools.ui; singleton:=true
-Bundle-Version: 1.2.800.qualifier
+Bundle-Version: 1.2.900.qualifier
 Bundle-Localization: plugin
 Eclipse-LazyStart: true
 Bundle-ActivationPolicy: lazy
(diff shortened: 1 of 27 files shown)


vogella (approved):
LGTM

vogella:
IMHO calling contains(null) is an error in the usage of this API. So doing this commit early in 4.25 should help us identify bad users and fix them.

merks:
> IMHO calling contains(null) is an error in the usage of this API. So doing this commit early in 4.25 should help us identify bad users and fix them.

That sounds like a scary comment to me.   I don't want to be a bad user, but do I have time to figure out what might be impacted by this so that I can add some special handling for null in cases that are just fine right now?

vogella:
> > IMHO calling contains(null) is an error in the usage of this API. So doing this commit early in 4.25 should help us identify bad users and fix them.
> 
> That sounds like a scary comment to me. I don't want to be a bad user, but do I have time to figure out what might be impacted by this so that I can add some special handling for null in cases that are just fine right now?

Yes, Java API can be scary ;-) See https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#contains-java.lang.Object- It is specified that you get an NPE is you call contains with null.

mickaelistria:
@HannesWell looks like some other version bumps are necessary.

akurtakov:
What is the plan for this one? At the very minimum it needs manual rebase now.

HannesWell:
> What is the plan for this one? At the very minimum it needs manual rebase now.

I want to complete this one. But IIRC there were some new test failures with the current state that needs to be fixed first.

HannesWell:
For me this is ready now. Does anybody want to review this?

akurtakov:
I see 350 failed tests not the usual 349 :) . I wonder which one would that be.

HannesWell:
> I see 350 failed tests not the usual 349 :) . I wonder which one would that be.

It was `configureProjectAndCheckMarkers[20: OSGi EventAdmin Service Example] (org.eclipse.pde.ui.templates.tests.TestPDETemplates)` and I fixed it.
The previously used HashSet contained null as valid value and since the statically created sets don't permit null it has to be checked separately. When checking the code again yesterday I didn't remember that `null` used to be a value contained in the set and assumed the null check was only to guard against passing null to the contains method of the static-set. But that was a false assumpption and I reverted that.

Nevertheless I wonder why this test did not fail in Jenkins?

HannesWell:
I plan to submit this once the master branch opens again.

HannesWell:
Besides the Windows GH workflow (which seems to have some FS issues), all builds are green, thus submitting.

PR #112, opened 2022-05-15
Include package-providing fragments when computing requirements closure

HannesWell:
In rare cases fragments provide new packages to their host wiring.
In such case those fragments have to be included into the requirements closure, otherwise the resolution of the final set of bundles will fail.

One real world example from Eclipse is `o.e.egit.core`.
This bundle imports the package `o.e.jgit.gpg.bc` that is provided by the fragment with the same name. Its host is `o.e.jgit`, which does not export such package.
If one creates a product or launch config that includes (or requires) `o.e.egit.core` but not `o.e.jgit.gpg.bc` and enables automatic inclusion of requirements the pre-launch validation fails because for `o.e.egit.core` the package-import `o.e.jgit.gpg.bc` is not satisfied.

When a corresponding product is build with Tycho, the fragment that is the only provider/exporter of a required package is included automatically as well.

Adding the capability's revision of a required wire to the closure instead of the provider wiring's revision fixes the issue descried above because it adds the 'real' resource wired to the requirement (which is preferred the host, but if not otherwise possible a fragment). A wiring's revision is always the host-bundle.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -202,7 +202,9 @@ public class DependencyManager {
 					// not included into the closure.
 					continue;
 				}
-				BundleRevision provider = getProvider(wire);
+				BundleRevision provider = wire.getCapability().getRevision();
+				// Use revision of required capability to support the case if
+				// fragments contribute new packages to their host's API.
 				if (provider instanceof BundleDescription && (includeOptional || !isOptional(wire.getRequirement()))) {
 					BundleDescription requiredBundle = (BundleDescription) provider;
 					addNewRequiredBundle(requiredBundle, closure, pending);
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
```diff
@@ -202,7 +202,9 @@ public static Set<BundleDescription> findRequirementsClosure(Collection<BundleDe
 					// not included into the closure.
 					continue;
 				}
-				BundleRevision provider = getProvider(wire);
+				BundleRevision provider = wire.getCapability().getRevision();
+				// Use revision of required capability to support the case if
+				// fragments contribute new packages the the host's API.
```
jhonnen, line 207:
```suggestion
				// fragments contribute new packages to the host's API.
```

jhonnen (approved):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
HannesWell: Thank you!

HannesWell:
Build looks good and today is last day of development so this is ready to go.

HannesWell:
Verified with Eclipse SDK

Version: 2022-06 (4.24)
Build id: I20220516-1800



@vik-chand FYI I created an `fix verified` label for PDE. Please adjust it if you are not fine with it.

vik-chand:
> Verified with Eclipse SDK
> 
> Version: 2022-06 (4.24) Build id: I20220516-1800
> 
> @vik-chand FYI I created an `fix verified` label for PDE. Please adjust it if you are not fine with it.

We can probably have a label called "verified" across all repos. What do you think @sravanlakkimsetti ? Is there a script to do the same?

Issue #90, opened 2022-05-11
Unable to load p2 dependencies from nested target files


pzi-dsa:
I'm currently trying to create a modular target file where I use a base file, which is shared among all projects, and then add additional dependencies for each project individually.
However, I noticed that there seems to be issues resolving p2 dependencies from the base file.
Example:
![image](https://user-images.githubusercontent.com/92990747/167854896-9d3b0f5f-3ecb-436f-85e9-17053f234a27.png)

Both nested target files are identical and contain log4j, once referenced via the Orbit update site and once via Maven Central. The only difference is that one is named _child.target_, while the other is named _nested.target_.
Child1/2:
![image](https://user-images.githubusercontent.com/92990747/167854790-baf7ec17-d826-4d39-b9d5-aa1e14ee54ce.png)

Those errors also show up on the latest integration build for 4.24.

vogella:
@pzi-dsa does it work with Tycho? Can you share an example target file which shows the problem?

vogella:
@laeubi you implemented both, can you have a look?

laeubi:
First it would help to attach the target file that creates the problem, second I don't fully understand what the error is e.g. what is the expected behaviour?
Just in case please note that you need to reload the target whenever one of the childs change, this will not be detected automatically.

pzi-dsa:
> what is the expected behaviour?

I would assume that the p2 artifacts from both embedded target files are added to the classpath. It's not a problem when both are identical. But effectively, all bundles provided by Child 2 are unavailable.

Reloading the embedded files, the parent file and any other combination doesn't resolve the problem.

laeubi:
Does this also happen if they *are* different? It might be that pde is filtering stuff when it is already there?

laeubi:
As it works with the maven type it seems not be directly specific to m2e or the nested target, I fear one needs to debug this to get more details, @pzi-dsa do you like to take a look?
The interesting part should be at org.eclipse.pde.internal.core.target.TargetReferenceBundleContainer.getTarget(IProgressMonitor) that is where it loads the target that is referenced and simply return its content.

pzi-dsa:
It seems like the problem shows up while trying to load the artifacts from the target file. The error message is normally ignored, but this is the corresponding stacktrace:
```
org.eclipse.core.runtime.CoreException: Problems loading repositories
	at org.eclipse.pde.internal.core.target.IUBundleContainer.cacheIUs(IUBundleContainer.java:304)
[...]

PR #116, opened 2022-05-17
Delete the profile of nested target files on reload.

pzi-dsa:
Otherwise Eclipse will use the old, (potentially outdated) profiles to resolve their artifacts.

If one or more artifacts are not contained in the old profile, resolving the file fails until the metadata folder is cleared.

I've tried out some ideas to reuse the code for deleting the profiles in the TargetEditor class, but so far without success.

Resolves issue #90

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetReferenceBundleContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetReferenceBundleContainer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Christoph Läubrich and others.
+ * Copyright (c) 2021, 2022 Christoph Läubrich and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


jhonnen:
Are you're thinking of https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/165209?
Back then, `refreshRepository` did not refresh the children of composite repositories. But I believe that has been fixed in P2 in the meantime, so you could give it another go ;)

HannesWell:
> Are you're thinking of https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/165209? Back then, `refreshRepository` did not refresh the children of composite repositories. But I believe that has been fixed in P2 in the meantime, so you could give it another go ;)

Yes, exactly.
Do you think we should consider this for 2022-06 RC1 or wait until the master opens for the next release-cycle?
I'm not familiar enough with that part of p2 yet to tell if this change is a safe one or not.

jhonnen:
Me neither, so let's wait for the next release.

HannesWell:
The master is now open for 4.25.

@pzi-dsa thank you for your contribution!

Issue #121, opened 2022-05-23
The Create target reference page doesn't work well on Windows


merks:
Attempting to use `file:${project_loc:/<project-name>}/` results in a URI syntax error:

![image](https://user-images.githubusercontent.com/208716/169828071-ad580082-7e7f-429b-a035-d442ccc5c181.png)

The logic implemented by org.eclipse.pde.internal.core.target.RemoteTargetHandle.get(String), which calls org.eclipse.pde.internal.core.target.RemoteTargetHandle.convertRawToUri(String) should be reused.

It would also be nice if the dialog simply listed the available *.target files in the workspace.

merks:
This is what  I've implemented.  I.e., the logic is reused so that there is no validation error and I populated the dropdown with the available *.target files in the workspace (excluding the target definition itself of course).

![image](https://user-images.githubusercontent.com/208716/169830604-ab465476-2214-44c0-b671-b05de8e1a4f1.png)

PR #122, opened 2022-05-23
Ensure that file:${project_loc:/project}/my.target validates on Windows

merks:
as a location URI in  the 'Create target reference' page

Also provide choices of available workspace *.target files in the
drop-down of the 'Create target reference' page

https://github.com/eclipse-pde/eclipse.pde/issues/121

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/RemoteTargetHandle.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/RemoteTargetHandle.java
@@ -108,11 +108,17 @@ public class RemoteTargetHandle implements ITargetHandle {
 		return definition;
 	}
 
+	public static URI getEffectiveUri(String uri) throws CoreException, URISyntaxException {
+		Objects.requireNonNull(uri);
+		IStringVariableManager manager = VariablesPlugin.getDefault().getStringVariableManager();
+		URI resolvedUri = new URI(convertRawToUri(manager.performStringSubstitution(uri)));
+		return resolvedUri;
+	}
+
 	public static RemoteTargetHandle get(String uri) throws CoreException {
 		Objects.requireNonNull(uri);
 		try {
-			IStringVariableManager manager = VariablesPlugin.getDefault().getStringVariableManager();
-			URI resolvedUri = new URI(convertRawToUri(manager.performStringSubstitution(uri)));
+			URI resolvedUri = getEffectiveUri(uri);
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@merks : we are in RC1 freeze, merge withoutr module owner approval & review is not allowed.

merks:
Sorry, the check passed without that being highlighted!

merks:
@iloveeclipse I receive so much email (often well over a hundred) that I easily overlook things.  Given M3 is done I assumed things were open.  Sorry.

HannesWell:
> I honestly have no idea which checks are currently enabled in which repos, and which one is working or not. If there was no warning, please open bug for PDE.

The freeze period check is enabled for PDE, but since there is not date that contains "stabilization" in its name the check did not fail.
I created a PR with small improvements and to discuss further improvements: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/272
Please participate if you have suggestions.

vik-chand:
I tested this fix. The fix looks good. However I am able to add a target twice. Either the added target shouldn't appear in the dropdown or while adding it the 2nd time, it should say that it is already added.

merks:
> I tested this fix. The fix looks good. However I am able to add a target twice. Either the added target shouldn't appear in the dropdown or while adding it the 2nd time, it should say that it is already added.

Actually there are many things that can be done wrongly.  Firstly you aren't restricted by what's in the combo box of choices; you can type anything and even when it shows an error, you can still hit Finish.    So yes, you can add duplicates, but is that actually wrong rather than just completely pointless?  (The latter because I don't see errors when this is done.) Granted it would be better for the logic to be even smarter and that's simple to implement.  Also bad would be to introduce circularity, but I believe somewhere that is checked because one would likely get into an infinite loop or stack overflow if that's the case...

Shall I add logic to avoid duplicates in the offered choices?   Shall we all check that it's not a duplicate URI, for the came even if the user entered the URI manually?  Would it be good to check for circularity too?   That's a more expensive check and in general the thing can be a URI that is remote and we don't want to have to load the URI to do checking, so probably that is not a good idea?

vik-chand:
We can improve a bit in early 4.25. At least a duplicate check wont be expensive. I won't be too keen on circularity check. For RC1 timeframe, the current implementation is good and makes inclusion of target within a target really usable. 

The scope is limited to just edit of target, hence the risk is extremely low.  +1 from me for RC1 as it is.

merks:
@vik-chand  Thanks.  I agree the current change is not risky and makes the feature more usable (and less confusing) for Windows. Further changes, while nice to have, are really not urgent.

Issue #126, opened 2022-05-27
Exceptions thrown by IUBundleContainer.cacheIUs and recorded by IUBundleContainer.synchronizerChanged are lost


merks:
In this debug session you can see that the resolve method expects exceptions to be thrown and records those in the multi-status in the catch clause below the selected new logic that I've added:

![image](https://user-images.githubusercontent.com/208716/170686005-076ad2c0-f519-49a8-a4c5-a1a58db51313.png)

But the exception thrown by IUBundleContainer.cacheIUs 

![image](https://user-images.githubusercontent.com/208716/170680768-e410a9e2-9881-4fcb-8671-8a25dc8cadde.png)

Is recorded

![image](https://user-images.githubusercontent.com/208716/170680860-f718159a-e66c-4ac1-8ecf-b9a10f99c686.png)

So we should check the status of each container for this recorded status and record it the same was as if it were thrown as is done by the selected logic in the first picture.

PR #127, opened 2022-05-27
Exceptions thrown by IUBundleContainer.cacheIUs are lost

merks:
These exceptions are recorded by IUBundleContainer.synchronizerChanged
but TargetDefinition.resolve(IProgressMonitor) only notices exceptions
that are thrown and does not consider the status that recorded in the
IContainerLocation.

No logging is needed because the exception was already logged when it
was recorded.

https://github.com/eclipse-pde/eclipse.pde/issues/126


akurtakov:
@merks , do you target RC2 for this patch?

merks:
It would be awfully nice yes!    I'm extending Oomph's targlets to support composing other target definitions.  This is especially relevant because it allows targlets to reuse m2e's target location extension for direct consumption of Maven artifacts, which is becoming more and more important.  The loss of information about failures provides for poor feedback to the user...

akurtakov:
The change looks really safe so consider it PMC approved.

merks:
@akurtakov Thanks so much for the fast and pro-active review!!

Issue #128, opened 2022-05-28
The Reload button in the target platform Edit... wizard is a no-op


merks:
This button does nothing:

![image](https://user-images.githubusercontent.com/208716/170811117-1b56d0ad-062b-4345-b36b-13702761b692.png)

It's implemented by org.eclipse.pde.internal.ui.shared.target.TargetLocationsGroup which has methods addTargetChangedListener and addTargetReloadListener.  

The target editor calls org.eclipse.pde.internal.ui.editor.targetdefinition.LocationsSection.createClient(Section, FormToolkit) and that does this:

![image](https://user-images.githubusercontent.com/208716/170811192-434b289f-8feb-46f4-883a-a3b75089633c.png)

The wizard calls org.eclipse.pde.internal.ui.wizards.target.TargetDefinitionContentPage.initializeListeners() which does this:

![image](https://user-images.githubusercontent.com/208716/170811206-e37a8868-7aea-49bf-bd17-db402f9721ae.png)

But it should listen for the reload button too:

![image](https://user-images.githubusercontent.com/208716/170811236-bc865ece-c2cc-41db-a91b-70e6d32a4930.png)

PR #129, opened 2022-05-28
The Reload button in the target platform Edit... wizard is a no-op

merks:
TargetDefinitionContentPage.initializeListeners() should also listen to reloads.

https://github.com/eclipse-pde/eclipse.pde/issues/128


merks:
@akurtakov  This is a nice-to-have for RC2...

The build failure is just one of those annoying random failures. :-(

merks:
@akurtakov Do you think we can complete this for RC2?

akurtakov:
Oops, sorry that I missed that one. It looks good to me (PMC approval) but we need one more committer in order to push it now. 
@mickaelistria , @vik-chand , @HannesWell - could any one of you give your +1 if its fine with you?

akurtakov (approved):

PR #136, opened 2022-06-04
Version bumps for 4.25 stream

akurtakov:
Fixes comparator issues due to lambda order jdt bug.
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/313


akurtakov:
Push as verification can't succeed until we have 4.25 I-build and this one prevents it.

PR #144, opened 2022-06-08
Cosmetics to UpdateUnitVersionsCommandTests

akurtakov:
Unstable test so first round of cleanups to it.
* Use Map.of
* Remove useless cast


akurtakov:
No, I don't plan such work. I monitor failing tests in I-builds and clean failing ones before I look deeper in what causes them, that's it.

PR #155, opened 2022-06-13
Stop promoting update site creation via the feature editor

vogella:
Publishing a feature can be done via multiple means, for command line
publishing Tycho is frequently used and the latest Tycho version stopped
supporting update sites. Therefore, remove the "advertisement" of using
update sites from the feature editor, an editor should not promote an
publishing alternative which is not supported anymore by the most
popular build tool.


mickaelistria:
Yes, I think it's fine to remove this outdated notice. Particularly since the page already contains a note about exporting.

PR #173, opened 2022-06-21
Sort product editor outline view (fixes #172)

### LINKED ISSUE #172: .product editor outline does not react to sort button
The outline of the product file editor does not correctly sort items when the sort button is pressed.
image

Bananeweizen:
The product editor outline has been unsorted since the very first commit
by overriding the sort() method as empty code. By removing the empty
override everything works well again, because the super class uses the
viewer comparator, and that uses the label provider.


HannesWell (approved):
Just verified that this works well. Thank you!

Just to be sure, I link this to the issue:
Fixes #172

Issue #167, opened 2022-06-20
PDE product editor does not retain plugin order in the configuration section


laeubi:
Currently plugins are reordered from the XML representation.

HannesWell:
Noticed that as well (a while ago).
The reason is that a HashSet is used to maintain them. I have already a half finished commit locally, so I can take care of this.
But there are a few more flaws in the editor. For example repo-references also sometimes get duplicated.

laeubi:
> But there are a few more flaws in the editor.

One flaw after the other ;-)

> The reason is that a HashSet is used to maintain them

Using a `LinkedHashMap` sound reasonable

> I have already a half finished commit locally, so I can take care of this.

:+1:

PR #174, opened 2022-06-21
Stabilize content of configuration- and repo-section in product-files

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/167

The list of fRepositories is changed to be a LinkedHashSet to avoid
multiple entries with the same URL. If multiple entries with the same
URL exists they are merged into one that is enabled if any of those
entries with that URL is enabled.

FYI @laeubi

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
@@ -22,8 +22,11 @@ import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import org.eclipse.pde.core.IModelChangedEvent;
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
```diff
@@ -66,7 +69,7 @@ public class Product extends ProductObject implements IProduct {
 
 	private TreeMap<String, IProductObject> fPlugins = new TreeMap<>();
```
laeubi, line 70: Can we use a LinkedMap her too?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
```diff
@@ -66,7 +69,7 @@ public class Product extends ProductObject implements IProduct {
...
 	private TreeMap<String, IProductObject> fPluginConfigurations = new TreeMap<>();
```
laeubi, line 71: Can we use a LinkedMap her too?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java
HannesWell: I think the motivation to use a TreeMap is to sort all plug-ins alphabetically according to their symbolic bundle name.
In contrast to Products based on features one cannot move plug-ins of plug-in based products up and down in the editor. So if that is changed there should be some buttons to move plug-ins too. Otherwise one can only add plug-ins at the end of the list (or modify the file directly in the Source tab). But if the order cannot be controlled by the user  having it sorted alphabetically is the best way to make the order arbitrary.

But if you want to be able specify a order of plug-ins I suggest to have that in a dedicated PR (since this is more than just changing the implementation) or discuss it in a separate issue. But I have to say that personally I don't have a great need for that.

HannesWell:
The test failure seem unrelated to this change, merging.

PR #176, opened 2022-06-22
Remove unused private method parameters in pde.ui

vogella:
Using the JDT cleanup to remove unused private method parameters in
PDE.UI.


HannesWell:
@vogella can you please rebase on top of the current master and force-push this change so we can see if the tests pass?

Issue #179, opened 2022-06-23
Launch validation shows "Missing constraint JavaSE 1.1" error for org.apache.xml.resolver


iloveeclipse:
1) Have latest SDK with Webtools installed additionally (so we have org.apache.xml.resolver in the platform).
2) Right click and run /org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/launching/ConfigurationEncodingTests.java as Eclipse JUnit test.
3) First execution works without validation errors, ignore test results.
4) Go to the created launch config, change any setting (I've just added one extra system argument) and apply.
5) Run test again => validation error dialog is shown.

**What is wrong?**
1) In the launch config **lot of unrelated** changes are done (not the system arguments only). I would expect all default values are set in advance, not after "Apply", but obviously this is not the case.
2) Because of some "additionally" added defaults, some of them causes validation to fail.
3) I don't know which setting causes that, and I don't know if validation error is appropriate.
4) That must be some recent change.

![image](https://user-images.githubusercontent.com/964108/175265955-97f1ba60-46aa-48d5-86c6-e9dbf62c48cb.png)

akurtakov:
Which Java version do you run on ?

iloveeclipse:
Note: org.apache.xml.resolver bundle has this attribute set in manifest: Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.1))"

akurtakov:
My guess would be that because in the past BREE was J2SE-1.1/1.2.... mapping it to JavaSE-1.1 is missing somewhere when coming from capability. This comes from https://git.eclipse.org/c/orbit/orbit-recipes.git/commit/?id=76a5833e3dd93960c84a3026a0f14d27bd623b09 . While it's clear bug somewhere in PDE/Equinox retrospectively mapping like this is quite a stretch . If it's up to me I would ask why is this jar still needed at all.

HannesWell:
> Note: `org.apache.xml.resolver` bundle has this attribute set in manifest: `Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.1))"`

The indicated error seems to be related to https://github.com/eclipse-pde/eclipse.pde/issues/130, which is an error I see in my m2e workspace for a while already.

The changes in the launch config are probably because it is created without writing default values into the config.
[...]

PR #180, opened 2022-06-23
Set AUTOMATIC_VALIDATE=false when creating JUnit Plug-in Test launches

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/179

For newly created JUnit Plug-in Test launches the automated Plug-in
validation prior launching is not performed, but when that
launch-configuration is modified in the 'Run/Debug configurations'
dialog and then launched the validation happens.

The reason is that previously the JUnitWorkbenchLaunchShortcut did not
set a value for AUTOMATIC_VALIDATE when creating a new test
launch-configuration. JUnitLaunchConfigurationDelegate then used the
default of false when a that new JUnit Plug-in Test launch config was
launched (usually immediately). But when that config was opened in the
'Run/Debug configurations' dialog in
AbstractPluginBlock.initializeButtonsFrom() the check-box for 'Validate
Plug-ins automatically prior to launching' is initialized to true in
case of an absent value.

With this change AUTOMATIC_VALIDATE=false is set when creating a JUnit
Plug-in Test launch via the usual shortcut so that the validation only
happens if one activly activates it in the launch-config dialog.


HannesWell:
@iloveeclipse, @merks is this what you want?
I could also set `AUTOMATIC_VALIDATE=true` but I suspect then there will be many complaints about failing validation as described in https://github.com/eclipse-pde/eclipse.pde/issues/179#issuecomment-1164979603.

iloveeclipse:
@HannesWell : as you like, I have no problem with either default, as long as it is reproducible.

Could you also check why all other values are set "later" and especially if they change they values "automatically" on first manual save? 

If there are such values, they ideally should persist initial values on creation too. It is highly surprising /time consuming to debug issues that aren't actuall application issue but the behavior change due the unattended changes in the launch config.

merks:
I'd avoid changing the behavior because, as you say, if it suddenly works differently then people will complain...

HannesWell:
> Could you also check why all other values are set "later" and especially if they change they values "automatically" on first manual save?

I'm pretty sure that this happens for the same reason, they are not set explicitly be the launch shortcut but the editor/dialog does set them when you open the config. And for the other values you/we were just lucky that the defaults are consistent.

> 
> If there are such values, they ideally should persist initial values on creation too. It is highly surprising /time consuming to debug issues that aren't actuall application issue but the behavior change due the unattended changes in the launch config.

I can try to collect the missing ones from your screenshot and add them as well. Hopefully I don't find more inconsistencies.

HannesWell:
> If there are such values, they ideally should persist initial values on creation too. It is highly surprising /time consuming to debug issues that aren't actuall application issue but the behavior change due the unattended changes in the launch config.

I looked into this in more detail and now think it is not a good idea to set all values in the `org.eclipse.pde.ui.launcher.JUnitWorkbenchLaunchShortcut` launch short cut because there are values that are not controlled by PDE, for example `org.eclipse.jdt.junit.TESTNAME`, which is handled by the `org.eclipse.jdt.junit.launcher.JUnitLaunchShortcut`. PDE is not in charge of that and should therefore not interfere.
Furthermore there are options like SHOW_SELECTED_ONLY that are actually setting how things are displayed in the dialog and not relevant for a launch. Therefore the launch-shortcut should not interfere here as well.
Additionally there are some attributes that are not considered for the  launch type selected by JUnit Plug-in tests. For example `AUTOMATIC_ADD` or `INCLUDE_OPTIONAL` is not considered by the `BundleLauncherHelper` when `USE_DEFAULT=true` is set. Because then just every bundle available in the Target-Platform and the Workspace is added to the launched Eclipse-Application. So it is basically irrelevant to add them unless you change your test in the launch-dialog to only include specific plug-ins/features.

So I think it is not reasonable feasible in the launch shortcut to add all attributes. It would be better to rework the dialog to not set attributes that weren't set before and were not changed by the user. But from my recent work in the editor I can just say that that part of the code is a big PITA.
I have to fix some flaws in that dialog in conjunction with the recently added automatic inclusion of requirements and can keep this in mind. Maybe I will find a good solution.

For now I just added some more attributes that I think are OK to set but there still will be some changes after you have opened and modified the launch-config the first time.

HannesWell:
Since no further objections were raised and at least all reasonable possible parts are addressed I think this is good to go.

Issue #68, opened 2022-05-03
Run as Plug-in test still adds the fragments to the runtime which are invalid for the platform


vogella:
I'm currently helping a bit with WindowBuilder trying to cleanup the repo and to activate more tests.

If I open org.eclipse.wb.tests.utils.CommonTests and run it as plug-ins tests I see the following errors:

org.osgi.framework.BundleException: Could not resolve module: org.eclipse.wb.os.linux [278]
  Unresolved requirement: Require-Capability: eclipse.platform; filter:="(& (osgi.ws=gtk) (osgi.os=linux))"

	at org.eclipse.osgi.container.Module.start(Module.java:463)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.lambda$1(ModuleContainer.java:1834)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1829)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1775)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1739)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1661)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:228)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:339)

!ENTRY org.eclipse.wb.os.macosx 4 0 2022-05-03 15:13:27.816
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.wb.os.macosx [279]
  Unresolved requirement: Require-Capability: eclipse.platform; filter:="(osgi.os=macosx)"

	at org.eclipse.osgi.container.Module.start(Module.java:463)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.lambda$1(ModuleContainer.java:1834)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1829)
[...]

PR #181, opened 2022-06-23
Exclude unresolved fragments when launching with all active plug-ins

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/68

For example JUnit Plug-in tests that usually include all available bundles (Plug-ins and Fragments) when being launched often suffer from messages about unresolved bundles due to unsatisfied environment constraints (e.g. when having platform specific fragments for multiple different platforms there are always some unsatisfied on the current platform).

Only exclude fragments because sometimes unresolved bundles (non-fragments) are wanted in the launched application.

I verified that unresolved fragments (may it be due to wrong environment or missing dependencies) are excluded, but I was not yet able to reproduce the failed resolution messages from the launched application in my minimal reproducer, although I have seen them before.

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
@@ -82,9 +82,10 @@ public class BundleLauncherHelper {
 
 			if (wc.getAttribute(IPDELauncherConstants.USE_DEFAULT, true)) {
 				Map<IPluginModelBase, String> map = new LinkedHashMap<>();
-				IPluginModelBase[] models = PluginRegistry.getActiveModels();
-				for (IPluginModelBase model : models) {
-					addBundleToMap(map, model, DEFAULT_START_LEVELS);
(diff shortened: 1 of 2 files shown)


HannesWell:
@merks, @vogella what do you think?
In general I could try to also check if an unresolved fragment also has a not matching platform filter (and assume that this is the reason for the resolution failure). But I'm not sure if this is possible and necessary.

merks:
Certainly the behavior you are fixing is annoying and I'll be happy when that's no longer annoying!  If a fragment don't resolve, I suppose it would be nice to know why it doesn't resolve, but almost always that will be a deliberate thing and it's hard to imagine the corner cases that such a deeper check might help cover...

HannesWell:
Right. I made the check more fine grained so that only unresolved fragments are excluded that have a Platform-Filter that does not match the target environment. Such fragments cannot resolve any way. Therefore I think it is save to exclude them.

HannesWell:
@merks are you fine with the current approach or do you see any potential problem? I also added a test-case that succeeds.
Btw. the all green check-marks are very nice 😄

HannesWell:
> @HannesWell This looks very good to me!

Great, then lets merge it now. :)

PR #184, opened 2022-06-24
Don't add duplicate locations in the target editor ( fixes #183)

### LINKED ISSUE #183: Don't add duplicate locations in the target editor
Right now it is possible to add duplicate directory container, feature container,target reference etc. If the location
is already added, it should not be added again

vik-chand:
Don't add duplicate locations in the target editor( fixes #183)
Change-Id: Icabab6136ddddf13b61eda762f2f9ef5a18e1db0

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/AddBundleContainerSelectionPage.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/AddBundleContainerSelectionPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2021 IBM Corporation and others.
+ * Copyright (c) 2009, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/AddBundleContainerSelectionPage.java
```diff
@@ -179,10 +179,12 @@ public boolean performFinish() {
 							if (oldContainers == null) {
 								fTarget.setTargetLocations(new ITargetLocation[] {container});
 							} else {
-								ITargetLocation[] newContainers = new ITargetLocation[oldContainers.length + 1];
-								System.arraycopy(oldContainers, 0, newContainers, 0, oldContainers.length);
-								newContainers[newContainers.length - 1] = container;
-								fTarget.setTargetLocations(newContainers);
+								if (!Arrays.asList(oldContainers).contains(container)) {
```
merks, line 182: If you moved the guard condition up into the `} else {` as `} else if (!Arrays.asList(oldContainers).contains(container)) {` I think the code would be cleaner and there would be fewer lines marked as changed.

vik-chand: Thanks Ed ! Done.

Issue #185, opened 2022-06-24
[product launch] element cannot be mapped to a null key


laeubi:
@HannesWell I try to build a product for m2e with just the m2e features using the new automatic added dependencies, but this currently fails with:

```
java.lang.NullPointerException: element cannot be mapped to a null key
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at java.base/java.util.stream.Collectors.lambda$groupingBy$53(Collectors.java:1127)
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getPrioritizedAvailableFeatures(BundleLauncherHelper.java:247)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getSelectedFeatures(BundleLauncherHelper.java:198)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getMergedBundleMapFeatureBased(BundleLauncherHelper.java:135)
	at org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper.getMergedBundleMap(BundleLauncherHelper.java:94)
	at org.eclipse.pde.launching.EclipseApplicationLaunchConfiguration.preLaunchCheck(EclipseApplicationLaunchConfiguration.java:192)
	at org.eclipse.pde.launching.AbstractPDELaunchConfiguration.launch(AbstractPDELaunchConfiguration.java:115)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:803)
[...]

PR #186, opened 2022-06-25
Skip and warn features without Id when launching

HannesWell:
Skipping features with null id prevents subsequent NPEs.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/185

@laeubi what do you think?


laeubi (approved):
Looks good

PR #188, opened 2022-06-28
If a file or directory doesnt exist while adding a target, make it an

### LINKED ISSUE #187: If a file or directory doesnt exist while adding a target, make it an error
While trying out scenarios, few times, I added them which later turned over to be an error in the target editor.
Disable finish button is probably good to have but just setting error may be good enough.

vik-chand:
If a file or directory doesnt exist while adding a target, make it an error (fixes #187)

Change-Id: I06216a81a29930086e2941f1cc73e940bb6c3f25

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/EditDirectoryContainerPage.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/EditDirectoryContainerPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2017 IBM Corporation and others.
+ * Copyright (c) 2009, 2012 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


HannesWell:
The obligation to create a dedicated issue for each change has been removed recently since on GH PRs and Issues are quite equivalent regarding the ability to discussion things. So especially for small changes like this you can just create a PR a describe the issue and the motivation for the applied solution in the commit and PR message instead:
https://www.eclipse.org/lists/eclipse-dev/msg12031.html

vik-chand:
> The obligation to create a dedicated issue for each change has been removed recently since on GH PRs and Issues are quite equivalent regarding the ability to discussion things. So especially for small changes like this you can just create a PR a describe the issue and the motivation for the applied solution in the commit and PR message instead: https://www.eclipse.org/lists/eclipse-dev/msg12031.html



> The obligation to create a dedicated issue for each change has been removed recently since on GH PRs and Issues are quite equivalent regarding the ability to discussion things. So especially for small changes like this you can just create a PR a describe the issue and the motivation for the applied solution in the commit and PR message instead: https://www.eclipse.org/lists/eclipse-dev/msg12031.html

Thanks Hannes. I prefer to create an issue and a PR even if change is trivial. Also I like to put milestones in all issues and mark it verified for all issues. ( for my changes as well as other changes in PDE). With just PR, those things are missing. Also many times, people just search issues and they duplicate work. Most of the labels, we put on the issue and not PR and all queries are based on that.

Issue #202, opened 2022-07-10
Allow contribution of external javadoc and source locations


laeubi:
Currently a plugin can contribute javadoc and/or source locations *inside* the plugin itself through the following extension points:

- org.eclipse.pde.core.source
- org.eclipse.pde.core.javadoc

this has the limitation that the javadoc+source must be embedded and known in advance. In some situations it would be required to provide external locations dynamically e.g.:

- maven has the concept of source/javadoc artifacts that could be queried
- if a bundle contains a Bundle-Source-Location one could checkout
- one could have a "source-repository"
- ...

FYI @merks @HannesWell

PR #203, opened 2022-07-10

laeubi:
Currently a plugin can contribute source locations inside the plugin itself through the extension point `org.eclipse.pde.core.source` this has the limitation that the source must be embedded and known in advance.

In some situations it would be required to provide external locations dynamically e.g. maven has the concept of sourceartifacts that could be queried, if a bundle contains a Bundle-Source-Location one could checkout one could have a "source-repository" and so on.

This adds the necessary parts to the extension point and a first implementation that search for a co-located source bundle, that is co-located with the bundle (e.g. in a bundle pool).

Part of https://github.com/eclipse-pde/eclipse.pde/issues/202

### DIFF
--- a/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.15.300.qualifier
+Bundle-Version: 3.16.0.qualifier
 Bundle-Activator: org.eclipse.pde.internal.core.PDECore
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.core/plugin.xml
+++ b/ui/org.eclipse.pde.core/plugin.xml
@@ -394,4 +394,8 @@
              type="Target">
        </targetLocation>
     </extension>
+    <extension
+          point="org.eclipse.pde.core.source">
+          <locator class="org.eclipse.pde.internal.core.ColocationPluginSourcePathLocator" />
+    </extension>
 </plugin>
--- a/ui/org.eclipse.pde.core/pom.xml
+++ b/ui/org.eclipse.pde.core/pom.xml
@@ -19,7 +19,7 @@
     <relativePath>../../</relativePath>
   </parent>
   <artifactId>org.eclipse.pde.core</artifactId>
-  <version>3.15.300-SNAPSHOT</version>
+  <version>3.16.0-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
   <properties>
     <defaultSigning-excludeInnerJars>true</defaultSigning-excludeInnerJars>
--- a/ui/org.eclipse.pde.core/schema/source.exsd
+++ b/ui/org.eclipse.pde.core/schema/source.exsd
@@ -21,9 +21,10 @@ In addition, it should contain any file or directory specified in the build.prop
          </appInfo>
       </annotation>
       <complexType>
-         <sequence>
-            <element ref="location" minOccurs="1" maxOccurs="unbounded"/>
-         </sequence>
+         <choice minOccurs="1" maxOccurs="unbounded">
+            <element ref="location"/>
+            <element ref="locator"/>
+         </choice>
          <attribute name="point" type="string" use="required">
             <annotation>
                <documentation>
@@ -66,14 +67,20 @@ In addition, it should contain any file or directory specified in the build.prop
       </complexType>
    </element>
 
-   <annotation>
-      <appInfo>
-         <meta.section type="apiInfo"/>
-      </appInfo>
-      <documentation>
-         No Java code is required for this extension point.
-      </documentation>
-   </annotation>
+   <element name="locator">
+      <complexType>
+         <attribute name="class" type="string" use="required">
+            <annotation>
+               <documentation>
+                  Provides the locator used to find the source-path for the given plugin model
+               </documentation>
+               <appInfo>
+                  <meta.attribute kind="java" basedOn=":org.eclipse.pde.core.IPluginSourcePathLocator"/>
+               </appInfo>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
 
    <annotation>
       <appInfo>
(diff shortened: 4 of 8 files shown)


mickaelistria (approved):
Definitely a simple and valuable improvement!

merks:
Just an observation.  If it's in a shared bundle pool, it's managed by one or more profiles, and if it's not in any profile, it can be, and will eventually be, garbage collected from that pool.

laeubi:
I have now enhanced the initial idea by implementing it as an extension that later can be used by m2e or others to even locate more sources as described in #202.

laeubi:
I could need a hint how to resolve the API errors, I tried bumping pde.ui as well but this seems not working...

mickaelistria (changes requested):
-1 for a new extension point. There is already `org.eclipse.jdt.launching.sourceContainerResolvers` which is exected to achieve that. This one should be used instead.

laeubi:
First this is not "new" it is an existing extension point, second `org.eclipse.jdt.launching` is about *launching* and can't be used for classpath source attatchments.

mickaelistria:
> First this is not "new" it is an existing extension point,

OK, my bad, I focused the public interface. Maybe it's better to not expose it at the moment as it's not part of fixing this particular isuse. Later, if extenders need it, then we may consider adding a new public API, but "maybe it will be useful later" is not a compelling reason to do it right now.

laeubi:
> Later, if extenders need it, then we may consider adding a new public API

Later is now as I plan to add source-lookup in m2e for maven derived items as soon as this is available :-)

laeubi:
Without "maven complexity" the file won't be there, as this requires you to at least request this artifact once with sources ...

laeubi:
> Isn't it default in m2e? There is a preference that tells to automatically download sources and which is turned on by default. This preference should be enough to control it.

This is not enough, as you need a project with a pom.xml that references this specific artifact and it need to be build before PDE kicks in.

> Also, I believe it's better to hardcode this strategy in PDE for eg cases where one wouldn't have m2e installed or wouldn't use a m2e container in their target but would build a .target referencing locations on filesystem.

If m2e is not there, the file will never be there but only for cases where you have run a maven commandline build and luckily have referenced this artifact, this does not sounds like something we could rely on. Also coding maven defaults into PDE does not sounds like a good idea.

Anyone, everyone is free to add such a strategy later on (even directly in PDE) thanks to the extension point that makes this possible now.

laeubi:
This really turns into one of this fruitless discussion, If I would have added this API first, it would have been asked to have a consumer in platform, now that I supply a consumer right with it it is asked for external user, and I supply one and then it is considered not useful at all to add API because API is "bad" (where in this case the API is already there)... on the other hand user complaining about missing sources.

I already prototypes the m2e counter-part and it works well with this API, so if you have any specific concerns about the API itself (e.g. you feel something is missing, incomplete or wrong) because you have investigated alternatives and/or different use-cases please raise them, but without that is completely useless to scatter things more just to have them rejected.

HannesWell:
I cannot review this change in detail in my phone, but what Christoph has described sounds reasonable to me and I support it.

> This is not enough, as you need a project with a pom.xml that references this specific artifact and it need to be build before PDE kicks in. Also keep in mind cases where the artifact is only pulled in by a P2 update-site...
> 
> > Also, I believe it's better to hardcode this strategy in PDE for eg cases where one wouldn't have m2e installed or wouldn't use a m2e container in their target but would build a .target referencing locations on filesystem.
> 
> If m2e is not there, the file will never be there but only for cases where you have run a maven commandline build and luckily have referenced this artifact, this does not sounds like something we could rely on. Also coding maven defaults into PDE does not sounds like a good idea.

These are two important points that IMHO justify this change. Especially since we want to consume more artifacts from maven-central directly but actually don't want to include them in the Features (and the availabilty of sourses ist one/the only reason for that).

And since this requires M2E, we should simply make it so well that it can be added to more standard packages.

Don't know if this already the case but can it be ensured to download the sources immediatly while TP Resolution? This avoids the need for a permanent Internet connection and improves the performance of the first look-up in a source.

laeubi:
As its is not possible to tag all committers I'm just adding a few I'm aware of so @akurtakov @merks @HannesWell @vogella @vik-chand @iloveeclipse any thoughts? Without a public extension point here the whole "source" story becomes useless for me as it makes the whole process of contributing sources always bound to PDE internal and e.g. m2e already uses too much of platform/pde internals already and given the many complains about missing sources I think we add a direct value for users, also m2e is part of many EPPs.

> The other forms implying extension point, new API and so on seem really overkill to me at the moment. As mentioned, downloading the sources is something m2e is supposed to take care of and PDE can just rely on this to find sources by name/path convention.

As described above this won't work, and is not flexible enough, the only "overkill" is all these discussion o **not** improve PDE just for the sake of some imaginary complexity.

> I'm actually willing to merge [02caaf1](https://github.com/eclipse-pde/eclipse.pde/commit/02caaf1cf97c74ded111c24064bafff8de1f615e) now, and we can keep discussing about the extensibility on this issue.

This has no value for me, so I'm not interested in that as it only covers a very narrow use-case and I'm not see there is anything that this adds to future discussion beside "see we have done it there the same way and we don't like to change" ... so my aim is not to add more hacks to PDE here...

mickaelistria:
> As described above this won't work, and is not flexible enough

Can you please rephrase the reasons why it won't work and won't be flexible enough? As m2e can efficiently manage source attachement for JDT in general, I'm really not certain about what makes it so different here that another (child of) extension point or API is needed.
m2e is supposed to download the sources locally when fetching the artifact as the m2e preference implies -if it does not, it's a bug in PDE-, and this is supposed to happen *before* the .target is resolved. On the other end, access to sources doesn't have to happen early in the lifecycle for PDE I believe. So the flow seems correct to me and a new API unnessary.
What do I get wrong?

laeubi:
> if we don't want to add a public API extension point, why not define an internal OSGi Service Interface and a corresponding implementimg DS component in M2E?

The extension point *is* already API, I just extended it, and if its an *internal* interface m2e can not (officially) use it.

> What do I get wrong?

You assume that there is a project or target resolved by m2e, but assume I have a p2 site that contains a maven artifact (but not a corresponding source bundle), I used that in a target file with a usual "Update Site location" and I create a plain Bundle Plugin, how should m2e ever know there is something to download?

mickaelistria:
> The extension point is already API, I just extended it,

Extending an API is still adding some API.

> You assume that there is a project or target resolved by m2e, but assume I have a p2 site that contains a maven artifact (but not a corresponding source bundle), I used that in a target file with a usual "Update Site location" and I create a plain Bundle Plugin, how should m2e ever know there is something to download?

OK, I now understand the use-case and see how it makes sense in some cases; and I now see how such extension can help and how this proposal makes sense.
I'm not against it any more, but would prefer to wait for the m2e consumer to be ready to give a change to review the API in practice.

In the meantime, I will probably merge https://github.com/eclipse-pde/eclipse.pde/commit/02caaf1cf97c74ded111c24064bafff8de1f615e and an addition that hardcodes Maven source location rules.

Issue #69, opened 2022-05-03
Support bundle-annotations


laeubi:
PDE currently supports DS annotations, ist would be usefull to support https://blog.osgi.org/2018/07/osgi-r7-highlights-bundle-annotations.html as well.

laeubi:
R8 also defines the `org.osgi:org.osgi.service.feature` annotations:

- https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.feature.html#org.osgi.service.feature.annotation.RequireFeatureService
- https://blog.osgi.org/2019/07/new-osgi-work-features.html
- https://projects.eclipse.org/projects/technology.osgi/releases/1.0-feature-service

iloveeclipse:
Note: PR #234 causes a huge regression for compiling workspaces with plugins, 
I'm mostly unable to work now using I20220802-1800 build because compiler needs ages now compiling a single bundle.
Compilation of entire workspace with lot of SDK bundles takes ~10 minutes instead of 1-2.

To reproduce: import jdt.core bundle in the workspace and do a clean build. It takes ages.

iloveeclipse:
Looking on stack traces I assume it is `org.eclipse.pde.internal.ui.annotations.OSGiAnnotationsCompilationParticipant` that causes trouble. Consider to either disable that until performance issues are identified or make optional via preference, or fix performance issue.

Since I'm using nightly SDK builds for my daily work ("eat my own dog food"), I can't continue to do this anymore since I20220802-1800, so if there will be no real fix soon, I plan to disable new compilation participant today (there are no tests for that one anyway).

laeubi:
@iloveeclipse I'll take a look at this, as you are more familiar with JDT, is there anything "wrong" in the usage or anything that could be performed better here? I assume it is common for code to inspect the Java-Files so I'm a bit astonished that this is causing such slow down?

iloveeclipse:
@laeubi : sorry, I have no time to work on this. 

Briefly looking on the code I assume [isActive() ](https://github.com/eclipse-pde/eclipse.pde/blob/883fe148dbe465e38fd9512ec87d5dd02a949986/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/annotations/OSGiAnnotationsCompilationParticipant.java#L34) doesn't work, as it seem to slowdown compilation of ALL projects, independently if there is any annotation setup (I assumed there is no changes in JDT so far to use new annotations).

[...]

PR #209, opened 2022-07-14
Add support for the OSGi Bundle and Versioning Annotations

laeubi:
OSGi has support for annotations that are only retained at the class
file (but not on the runtime) to help tools generating meta-data or
perform several checks.

This adding them to the compile-classpath without requiring the bundle
developer to import them (what would create an undesired runtime binding
to those packages) if they are found in the target platform.

This is part of https://github.com/eclipse-pde/eclipse.pde/issues/69

### To test this I did the following, because even though the annotations are part of the SDK target they are not available in the SDK-IDE Target platform state:

1. Start an Eclipse with this patch
2. download https://mvnrepository.com/artifact/org.osgi/org.osgi.annotation.versioning/1.1.2 and https://mvnrepository.com/artifact/org.osgi/org.osgi.annotation.bundle/2.0.0
3. create a simple project named "target", create a folder lib, put both files in and add the following target
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="annotations">
	<locations>
		<location path="${eclipse_home}" type="Profile"/>
		<location path="${project_loc:/targets}/lib" type="Directory"/>
	</locations>
</target>
```

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/OSGiAnnotationsClasspathContributor.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/OSGiAnnotationsClasspathContributor.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.core;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.jdt.core.IAccessRule;
+import org.eclipse.jdt.core.IClasspathAttribute;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.osgi.service.resolver.BundleDescription;
+import org.eclipse.pde.core.IClasspathContributor;
+import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.core.plugin.PluginRegistry;
+
+/**
+ * This makes the <a href=
+ * "https://docs.osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.bundle">OSGi
+ * Bundle Annotations</a> and <a href=
+ * "https://docs.osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.versioning">OSGi
+ * Versioning Annotations</a> available to plugin projects if they are part of
+ * the target platform.
+ *
+ */
+public class OSGiAnnotationsClasspathContributor implements IClasspathContributor {
+
+	private Collection<String> OSGI_ANNOTATIONS = List.of("org.osgi.annotation.versioning", //$NON-NLS-1$
+			"org.osgi.annotation.bundle"); //$NON-NLS-1$
+
+	@Override
+	public List<IClasspathEntry> getInitialEntries(BundleDescription project) {
+		IPluginModelBase model = PluginRegistry.findModel(project);
+		if (model != null) {
+			return OSGI_ANNOTATIONS.stream().map(PluginRegistry::findModel).filter(Objects::nonNull)
+					.filter(IPluginModelBase::isEnabled).map(IPluginModelBase::getInstallLocation)
+					.filter(Objects::nonNull).map(Path::new).map(path -> JavaCore.newLibraryEntry(path, path, Path.ROOT,
+							new IAccessRule[0], new IClasspathAttribute[0], false))
+					.collect(Collectors.toList());
+		}
+		return Collections.emptyList();
+	}
+
+	@Override
+	public List<IClasspathEntry> getEntriesForDependency(BundleDescription project, BundleDescription addedDependency) {
+		return Collections.emptyList();
+	}
+
+}
--- a/ui/org.eclipse.pde.core/plugin.xml
+++ b/ui/org.eclipse.pde.core/plugin.xml
@@ -394,4 +394,10 @@
              type="Target">
        </targetLocation>
     </extension>
+    <extension
+          point="org.eclipse.pde.core.pluginClasspathContributors">
+       <contributor
+             class="org.eclipse.pde.internal.core.OSGiAnnotationsClasspathContributor">
+       </contributor>
+    </extension>
 </plugin>


laeubi:
> Wouldn't it be better/simpler to make it a specific classpath container?

PDE specifically offers this extension point and the DS Annotations Plugin also uses this to contribute the annotations so it seems the correct way to do it.

> Or to add them to build.properties/additional.bundles ?

1. `additional.bundles` is already misused for other things and is obviously not suitable for this purpose as we don't want PDE to add anything to the manifest here (what is the documented purpose of `additional.bundles`).
2. one needs to add it to each bundle instead of getting support automatically (what is what I expect from a tool to help me not doing everything manually).

laeubi:
Build is green, so anything seems settled maybe @HannesWell do you want to review/test this change also?

HannesWell:
> Build is green, so anything seems settled maybe @HannesWell do you want to review/test this change also?

Yes I will at the weekend.

HannesWell:
This change works well and I think it is a good addition to PDE. This would for example allow to remove the `osgi.annotation.jar` from `org.eclipse.osgi`.

But instead of obtaining the jars from the Target-Platform I think it would be better to embed them into a PDE plug-in. This would be more convenient for the user (no need to adjust the TP and everything works auto-magically). And not having them in the TP would avoid the risk that those annotation bundles are referenced in a Manifest and become a runtime dependency 'accidentally'.

I'm not yet sure where to place them but I think for the beginning it would be ok to place them just in pde.core. Or we create a dedicated plug-in for that story?
On the long run I would like to simplify the embedding of annotation bundles in PDE any way (namely the the ds-annotations).
Currently I have an approach similar to the maven-runtime or lemminx in m2e in my mind. I think this could simplify it a lot and the annotation-jars are so small that I think it is ok to just place all versions in one plug-in.
Since PDE is usually installed as one monolith anyways even a single plug-in that carries all annotation jars would be possible. But on the other hand a split per topic doesn't hurt either or simply place them in the consuming bundle.

Another think to consider is how we deal with multiple versions of those annotation bundles. But maybe we could simply add the latest version.

laeubi:
> But instead of obtaining the jars from the Target-Platform I think it would be better to embed them into a PDE plug-in. This would be more convenient for the user (no need to adjust the TP and everything works auto-magically).

But it won't be less flexible (updates), more complicates (IP clearance) and duplicates content that might differ between build and IDE

> I'm not yet sure where to place them but I think for the beginning it would be ok to place them just in pde.core. Or we create a dedicated plug-in for that story?

As said, I don't see any advantage in embedding them anywhere. With the new m2e maven targets its a matter of seconds for users to add them to the target.

> Since PDE is usually installed as one monolith anyways even a single plug-in that carries all annotation jars would be possible.

PDE-core already is badly shaped, don't make it worse by adding more unrelated stuff to it... for platform itself I think we can add a p2.inf so PDE includes the annotations when installed in the profile (I think @merks can best give some hints to how its included but not installed).

> Another think to consider is how we deal with multiple versions of those annotation bundles. But maybe we could simply add the latest version.

Maybe it is more simple to simply pull things from the target so users has a choice instead of blaming PDE ;-)

HannesWell:
Your arguments are well and I don't want to mitigate them. Actually that would be very handy because the user can simply control the version via the TP. And if there is some tooling that processes them it can check if it supports that version. E.g. this could also be used for PDE DS and we would not have to embedded the bundles?

Nevertheless I have the concern that with requiring to add those annotation-bundles to the TP we somehow encourage (or at least increase the change that) users to reference those annotation-bundles in the Manifest.
Do you know how other tools for example BND handles this?
@mickaelistria what do you think about that?

laeubi:
> E.g. this could also be used for PDE DS and we would not have to embedded the bundles?

PDE does not "embedd" the DS annotation bundles, it provides a separate plugin (that's part of the target platform) that then contains a copy of the official jars, maybe because it was hard those days to get it from other sources, and it uses the plugin name to distinguish between versions (in contrast to use the package version).

> Nevertheless I have the concern that with requiring to add those annotation-bundles to the TP we somehow encourage (or at least increase the change that) users to reference those annotation-bundles in the Manifest.

If people like to do so, it is not forbidden and embedding them won't prevent that, it just complicates things as then either Tycho also needs to embeds stuff or people would need to add it to their target as well to make it work in Tycho...

> Do you know how other tools for example BND handles this?

If you generate the Manifest by BND it ignores those, but users still can force to import that if they like to ...

HannesWell (approved):
> > E.g. this could also be used for PDE DS and we would not have to embedded the bundles?
> 
> PDE does not "embedd" the DS annotation bundles, it provides a separate plugin (that's part of the target platform) that then contains a copy of the official jars, maybe because it was hard those days to get it from other sources, and it uses the plugin name to distinguish between versions (in contrast to use the package version).

Yes that's what I mean. But you are probably right that it was only done this way due to the limited capabilities back then compared to now.
Nevertheless this separate annotation-bundles usually come into the TP with the pde-feature and are not added explicitly unlike now. And from my experience users tend to add things to the Manifest (or what ever), 'just-in-case' even if it is not necessary. And if those bundles come into their mind because they have added them to TP, this chances probably gets higher.
But I think we cannot protect users from everything.

> 
> > Nevertheless I have the concern that with requiring to add those annotation-bundles to the TP we somehow encourage (or at least increase the change that) users to reference those annotation-bundles in the Manifest.
> 
> If people like to do so, it is not forbidden and embedding them won't prevent that, it just complicates things as then either Tycho also needs to embeds stuff or people would need to add it to their target as well to make it work in Tycho...
> 

That's right.
OK the extra handling in Tycho is another point. Btw. doesn't Tycho also need support for that or have you done that already?
And having multiple places to adjust when updating to new versions is indeed bad.

> > Do you know how other tools for example BND handles this?
> 
> If you generate the Manifest by BND it ignores those, but users still can force to import that if they like to ...

OK.

I'm not 100% happy with the approach referencing the annotation bundles from the TP, but you convinced me that it is the best of all approaches. So I'm fine with this.

Nevertheless I would appreciate @mickaelistria's consent as well.

@laeubi since this requires a non-trivial set up for the user to make it work, I think a N&N entry should be added for PDE mentioning this new feature and that a user has to add the annotation bundles to the TP (as you descried e.g. via Maven-Targets) but does not have to add them to the Manifest.

laeubi:
> And from my experience users tend to add things to the Manifest (or what ever), 'just-in-case' even if it is not necessary. And if those bundles come into their mind because they have added them to TP, this chances probably gets higher.
> But I think we cannot protect users from everything.

If anyone like, he can provide a PDE plugin that checks for useless imports ...

> OK the extra handling in Tycho is another point. Btw. doesn't Tycho also need support for that or have you done that already?

I'll add support for Tycho once PDE has support, there are enough features that only work in Tycho alone...

> since this requires a non-trivial set up for the user to make it work,

Anyone finding it "non trivial" to add content to their target should probably better not use that feature :-D 
Neverless we should inform users about the opportunity to do so.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/OSGiAnnotationsClasspathContributor.java
```diff
@@ -0,0 +1,63 @@
...
+public class OSGiAnnotationsClasspathContributor implements IClasspathContributor {
+
+	private Collection<String> OSGI_ANNOTATIONS = List.of("org.osgi.annotation.versioning", //$NON-NLS-1$
+			"org.osgi.annotation.bundle"); //$NON-NLS-1$
+
+	@Override
+	public List<IClasspathEntry> getInitialEntries(BundleDescription project) {
+		IPluginModelBase model = PluginRegistry.findModel(project);
+		if (model != null) {
+			return OSGI_ANNOTATIONS.stream().map(PluginRegistry::findModel).filter(Objects::nonNull)
```
HannesWell, line 49:
```suggestion
	private static Set<String> OSGI_ANNOTATIONS_PACKAGES = Set.of("org.osgi.annotation.versioning", //$NON-NLS-1$
			"org.osgi.annotation.bundle"); //$NON-NLS-1$

	@Override
	public List<IClasspathEntry> getInitialEntries(BundleDescription project) {
		IPluginModelBase model = PluginRegistry.findModel(project);
		if (model != null) {
			Stream<ExportPackageDescription> allExportedPackages = Arrays
					.stream(model.getBundleDescription().getContainingState().getExportedPackages());
			return allExportedPackages.filter(e -> OSGI_ANNOTATIONS_PACKAGES.contains(e.getName()))
					.map(ExportPackageDescription::getExporter).map(PluginRegistry::findModel).filter(Objects::nonNull)
```

HannesWell (commented):
It looks like @mickaelistria has no objections.

One last thing that just came into my mind, due to the work it required to move to bundles from Maven-Central: I think it would be better if the annotation-bundles added to the classpath were identified by the package, meaning that we add the bundles that export the corresponding packages.

But I have not yet found a good way to do that with the old equinox-resolver that PDE is currently using. The only way I see is to iterate over all Exported-Packages of the TP-state.

Does anybody have a better idea?

Additionally the suggested code should be improved to only include the latest bundle for each package, e.g. if multiple versions are available.

> > since this requires a non-trivial set up for the user to make it work,
> 
> Anyone finding it "non trivial" to add content to their target should probably better not use that feature :-D Neverless we should inform users about the opportunity to do so.

That's true. :P I more meant that we should add a little hint into the N&N, just in case somebody didn't notice the Maven-targets yet. :D

laeubi:
> Does anybody have a better idea?

Just use the official name from the official artifacts and don't complicate things too much, I think there is no reason that someone packs his own ... Nerveless we might later further improve this code if things are changing or PDE allows efficient query based on filters like P2 does.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/OSGiAnnotationsClasspathContributor.java
laeubi: We should keep it simple and fast here...

HannesWell:
> > Does anybody have a better idea?
> 
> Just use the official name from the official artifacts and don't complicate things too much, I think there is no reason that someone packs his own ... Nerveless we might later further improve this code if things are changing or PDE allows efficient query based on filters like P2 does.

OK, yes probably not worth the effort and it can still be fixed later if necessary and better alternatives are available.

@laeubi can you please add a N&N entry to inform about this new feature, with a little hint how to obtain the annotation-bundles.

Issue #74, opened 2022-05-05
DS Annotations processor updates the MANIFEST.MF file in a unfriendly way, in regards to a source versioning system.


nport:
The attached picture shows the result of a build. Nothing is changed in the source code, just the Eclipse version. On the right Eclipse Oxygen, on the left Eclipse 2022-03, both on Windows 10.

![DSManifestCompare](https://user-images.githubusercontent.com/6675853/166956401-352a379a-7d78-4704-add6-861111f5806c.PNG)

The ServiceComponent manifest header has been moved inside the MANIFEST.MF file, but the header value is the same. So the diff tool shows a difference, but there are no semantic changes in the manifest. 
Even if the ServiceComponent header value was not the same it would be better not to change the header position, but just update its content.

This is a general rule that should be applied while updating the MANIFEST.MF file.

This second picture shows that even in the case the ServiceComponent manifest header position is not changed, it is the order of the DS xml files that is changed after the build. On the left (Eclipse 2022-03) the order is alphabetical. 

![DSManifestCompare1](https://user-images.githubusercontent.com/6675853/166956521-6c9e2d90-3e6f-4225-92be-23e259646498.PNG)

As a general rule the pre-existing order should be preserved.

I am using tycho to build my workspace from the command line but these changes appeared after a rebuild inside eclipse.

nport:
Just noticed that one of the above issues is the result of an explicit bugfix of bug 573557, as described in class [DSAnnotationCompilationParticipant.java](https://github.com/eclipse-pde/eclipse.pde/blob/b8fdc05299682d27de59e466ff74d2e7d05b47d3/ds/org.eclipse.pde.ds.annotations/src/org/eclipse/pde/ds/internal/annotations/DSAnnotationCompilationParticipant.java#L14).

In my opinion the xml files ordering must apply only when the Service-Component header is not present in the MANIFEST.MF. If this header is already present and the update must preserve the current xml files ordering, in particular if no changes in the xml file list is necessary.

akurtakov:
@laeubi as you did the bugfix mentioned, what do you think?

vogella:
After the initial change from @laeubi the order should be stable afterwards, right? Or does it still changes after the intial sorting?

laeubi:
This is a mixture of two issues:

[...]

PR #212, opened 2022-07-14
Enforce a deterministic Manifest-Header order when inserting new keys

laeubi:
Currently, when adding a manifest header the new value is always placed
at the bottom. This has the disadvantage, that the order of manifest
headers is determined by the order of inserts.

Especially if headers are added/removed (e.g. DS) automatically this can
lead to modified manifest that are semantically equivalent.

To prevent this, but still keep user-order of manifest headers, the
following algorithm is applied to have always the same insertion place:

1) All headers of the manifest are fetched
2) Special headers, that need to be retained at the top are ignored
3) All remaining headers are sorted ignoring their case
4) The header that is before the header to insert in this alphabetical
list is chosen as the insertion point
5) if the is no predecessor, the Bundle-ManifestVersion header is chosen
6) if nothing matches the header is inserted at the end

This ensures that (given this algorithm was applied at least once) there
is always a deterministic insertion point even in the case where a
header is removed and later added back.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/74

### DIFF
--- a/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/AbstractEditingModel.java
+++ b/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/AbstractEditingModel.java
@@ -46,7 +46,7 @@ public abstract class AbstractEditingModel extends PlatformObject implements IEd
 	private Charset fCharset;
 	private IResource fUnderlyingResource;
 	private String fInstallLocation;
-	private boolean fStale;
+	private volatile boolean fStale;
 
 	public AbstractEditingModel(IDocument document, boolean isReconciling) {
 		fDocument = document;
--- a/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/BundleModel.java
+++ b/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/BundleModel.java
@@ -61,7 +61,7 @@ public class BundleModel extends AbstractEditingModel implements IBundleModel {
 	}
 
 	@Override
-	public void adjustOffsets(IDocument document) {
+	public synchronized void adjustOffsets(IDocument document) {
 		((Bundle) getBundle()).clearOffsets();
 		((Bundle) getBundle()).adjustOffsets(document);
 	}
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContext.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContext.java
@@ -230,6 +230,10 @@ public abstract class InputContext {
 				if (fModel instanceof IEditingModel)
 					((IEditingModel) fModel).setStale(true);
 				edit.apply(doc);
+				if (fModel instanceof IEditingModel) {
+					IEditingModel editingModel = (IEditingModel) fModel;
+					editingModel.reconciled(doc);
+				}
 				fEditOperations.clear();
 			} catch (MalformedTreeException | BadLocationException e) {
 				PDEPlugin.logException(e);
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/BundleInputContext.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/BundleInputContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2017 IBM Corporation and others.
+ * Copyright (c) 2003, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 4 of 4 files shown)


laeubi:
Does anyone know if the failure is related to the change? Is the test sensitive to Header ordering in manifest files? Anyone likes to review this?

iloveeclipse:
> Does anyone know if the failure is related to the change?

If you mean 
```
java.lang.AssertionError: 

Expecting empty but was: [Marker [on: /org.eclipse.pde.core.tests.internal.core.builders.BundleErrorReporterTest/META-INF/MANIFEST.MF, id: 442, type: org.eclipse.pde.core.problem, attributes: [categoryId: fatal, compilerKey: compilers.p.unresolved-import, lineNumber: 6, message: No available bundle exports package 'javax.xml.ws', packageName: javax.xml.ws, problemId: 4103, severity: 2], created: 7/14/22, 1:28 PM]]
	at org.eclipse.pde.core.tests.internal.core.builders.BundleErrorReporterTest.testErrorOnUnresolvedJrePackage(BundleErrorReporterTest.java:74)
```

no, no idea, I haven't seen this test failing recently. 

Note: you can always check what the official nightly SDK tests say, here latest one: https://download.eclipse.org/eclipse/downloads/drops4/I20220714-1800/testResults.php

HannesWell:
> Does anyone know if the failure is related to the change? Is the test sensitive to Header ordering in manifest files? Anyone likes to review this?

I will review this at the weekend.


Doesn't this also fix [Bug 576798](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576798)?


>     3. All remaining headers are sorted ignoring their case
> 

What do you think about instead of sorting header only alphabetical to specify a order for prominent headers manually that reasonable sorts and groups them and only sort other headers alphabetically? For example:
```
Bundle-ManifestVersion
Bundle-SymbolicName
Bundle-Version
Bundle-ActivationPolicy: lazy
Automatic-Module-Name
Bundle-Name
Bundle-Vendor
Bundle-RequiredExecutionEnvironment
Import-Package
Require-Bundle
Require-Capability
Export-Package
Provided-Capability
...
```
The list could contain all headers relevant for PDE. The order groups the headers that describe 'the bundle', its requirements, its capabilities. Of course the order is arbitrary and there are no strong reasons to have one specific header before another.
I created a company-internal plug-in a while a ago that does exactly that (I had the intention to contribute it one day, but now you were quicker) and IMHO it helps if the headers are grouped by their 'topic' in a reasonable manner.

Furthermore it would be handy to have a mean to sort fully sort all headers according to the order we finally impose. Maybe via the `Plug-in Tools` Context Menu from Package-Explorer or a 'In-Text' Button. In general it would be handy if this could be applied to multiple plug-in projects at once. But this is out of the scope of this change, so if you are not interested in that this can be done in a follow-up PR.

laeubi:
> Doesn't this also fix [Bug 576798](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576798)?

Yeah the Bugzilla is related, but I don't really search Bugzilla for duplicates/related things ...

> What do you think about instead of sorting header only alphabetical to specify a order for prominent headers manually that reasonable sorts and groups them and only sort other headers alphabetically?

I don't think that it is worth anything and don't want to judge what belongs together nor like to maintain any lists as these will always be incomplete, also see the discussion on the Bugzilla.

Actually Headers have no order and one should never need to edit the manifest by hand beside some rare cases so I don't really care the actual order as long as I get nice diffs and no unrelated reordering.

Also when should the reordering happen (ignoring that it will become very cumbersome to calculate a nice sequence of edit actions), e.g when it happens on next insert, people will complain that PDE "suddenly" reorders their manifest and ask for a way to disable it and all the hassle...

> Furthermore it would be handy to have a mean to sort fully sort all headers according to the order we finally impose. Maybe via the `Plug-in Tools` Context Menu from Package-Explorer or a 'In-Text' Button. In general it would be handy if this could be applied to multiple plug-in projects at once. But this is out of the scope of this change, so if you are not interested in that this can be done in a follow-up PR.

Could be a simple STRG+SHIFT+F keybinding, but I'm not sure its worth the effort.

laeubi:
I could fix the test by adding a sleep between edits, its seems that flushing the edits do not update the positions in sync in the ManifestHeaders ...

laeubi:
I debugged this and there are multiple concurrency issues:

1. `org.eclipse.pde.internal.core.text.bundle.BundleModel.adjustOffsets(IDocument)` is calling clearOffsets / adjustOffsets and of course between these calls some race can happen
2. Each header is iterated and thus also here a race can happen where one header is updated while the other is not
3. While `org.eclipse.pde.internal.ui.editor.context.InputContext.flushModel(IDocument)` is called in the main thread but the update of the positions is called `Thread [org.eclipse.jface.text.reconciler.MonoReconciler] `
4. Even though updates happen from different threads the fields are not marked as volatile

laeubi:
Just the usual https://github.com/eclipse-pde/eclipse.pde/issues/232 ... GH build even succeds...

PR #214, opened 2022-07-15

laeubi:
Currently a plugin can contribute source locations inside the plugin
itself through the extension point

org.eclipse.pde.core.source

this has the limitation that the source must be embedded and known in
advance. In some situations it would be required to provide external
locations dynamically e.g. maven has the concept of source artifacts
that could be queried, if a bundle contains a Bundle-Source-Location one
could checkout one could have a "source-repository" and so on.

This adds the necessary parts to the extension point and a extracts the
implementation that search for a Eclipse/Maven co-located source
bundles.

Part of https://github.com/eclipse-pde/eclipse.pde/issues/202

### DIFF
--- a/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.15.300.qualifier
+Bundle-Version: 3.16.0.qualifier
 Bundle-Activator: org.eclipse.pde.internal.core.PDECore
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.core/plugin.xml
+++ b/ui/org.eclipse.pde.core/plugin.xml
@@ -394,4 +394,9 @@
              type="Target">
        </targetLocation>
     </extension>
+    <extension
+          point="org.eclipse.pde.core.source">
+          <locator class="org.eclipse.pde.internal.core.ColocationPluginSourcePathLocator" />
+          <locator class="org.eclipse.pde.internal.core.LocalMavenPluginSourcePathLocator" />
+    </extension>
 </plugin>
--- a/ui/org.eclipse.pde.core/pom.xml
+++ b/ui/org.eclipse.pde.core/pom.xml
@@ -19,7 +19,7 @@
     <relativePath>../../</relativePath>
   </parent>
   <artifactId>org.eclipse.pde.core</artifactId>
-  <version>3.15.300-SNAPSHOT</version>
+  <version>3.16.0-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
   <properties>
     <defaultSigning-excludeInnerJars>true</defaultSigning-excludeInnerJars>
--- a/ui/org.eclipse.pde.core/schema/source.exsd
+++ b/ui/org.eclipse.pde.core/schema/source.exsd
@@ -21,9 +21,10 @@ In addition, it should contain any file or directory specified in the build.prop
          </appInfo>
       </annotation>
       <complexType>
-         <sequence>
-            <element ref="location" minOccurs="1" maxOccurs="unbounded"/>
-         </sequence>
+         <choice minOccurs="1" maxOccurs="unbounded">
+            <element ref="location"/>
+            <element ref="locator"/>
+         </choice>
          <attribute name="point" type="string" use="required">
             <annotation>
                <documentation>
@@ -66,14 +67,20 @@ In addition, it should contain any file or directory specified in the build.prop
       </complexType>
    </element>
 
-   <annotation>
-      <appInfo>
-         <meta.section type="apiInfo"/>
-      </appInfo>
-      <documentation>
-         No Java code is required for this extension point.
-      </documentation>
-   </annotation>
+   <element name="locator">
+      <complexType>
+         <attribute name="class" type="string" use="required">
+            <annotation>
+               <documentation>
+                  Provides the locator used to find the source-path for the given plugin model
+               </documentation>
+               <appInfo>
+                  <meta.attribute kind="java" basedOn=":org.eclipse.pde.core.IPluginSourcePathLocator"/>
+               </appInfo>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
 
    <annotation>
       <appInfo>
@@ -97,6 +104,15 @@ In addition, it should contain any file or directory specified in the build.prop
 &lt;/pre&gt;
 
 In the example above, the source location &lt;code&gt;src&lt;/code&gt; in the contributing plug-in has been registered.
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appInfo>
+         <meta.section type="apiInfo"/>
+      </appInfo>
+      <documentation>
+         No Java code is required for this extension point.
       </documentation>
    </annotation>
 
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IPluginSourcePathLocator.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IPluginSourcePathLocator.java
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.core;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.pde.core.plugin.IPluginBase;
+
+/**
+ * A plugin source path locator is capable of locating the path of the source
+ * for a given plugin.
+ * <p>
+ * A plugin source path locator is declared as an extension
+ * (<code>org.eclipse.pde.core.source</code>).
+ * </p>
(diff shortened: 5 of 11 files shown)


merks:
This is the error right?
```
07:02:40  [INFO] --- tycho-p2-plugin:3.0.0-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.pde.build.tests ---
07:02:40  [WARNING] MavenProject: org.eclipse.pde:org.eclipse.pde.build.tests:1.3.400-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.pde.ui_PR-214/build/org.eclipse.pde.build.tests/pom.xml: baseline and build artifacts have same version but different contents
07:02:40     classifier-sources: not present in baseline
07:02:40  
07:02:40  [INFO] MavenProject: org.eclipse.pde:org.eclipse.pde.build.tests:1.3.400-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.pde.ui_PR-214/build/org.eclipse.pde.build.tests/pom.xml
07:02:40      The main artifact has been replaced with the baseline version.
07:02:40      The following attached artifacts are not present in the baseline and have been removed: [sources]
```
I don't understand that either, but given it's just a test plugin so "who really cares about the version", one could try to modify this file such that there is a newer commit in the plugin to force the qualifier to change:

image

And if it's still not happy, you can increment the 1.3.400.qualifier to 1.3.500.qualifier in the pom.xml and the MANIFEST.MF and hope it doesn't complain that this change is unnecessary...

laeubi:
The problem is the API analysis:

> Marker [on: /org.eclipse.pde.ui/META-INF/MANIFEST.MF, id: 378, type: org.eclipse.pde.api.tools.compatibility, attributes: [apiMarkerID: 2, charEnd: 1, charStart: -1, lineNumber: 1, message: The type org.eclipse.pde.ui.launcher.AbstractPDELaunchConfiguration has been removed from org.eclipse.pde.ui_3.14.0, messagearguments: org.eclipse.pde.ui.launcher.AbstractPDELaunchConfiguration#org.eclipse.pde.ui_3.14.0, problemTypeName: org.eclipse.pde.ui.launcher.AbstractPDELaunchConfiguration, problemid: 305324134, severity: 2, sourceId: API Tools], created: 7/29/22, 10:55 AM]

merks:
I'm just stabbing in the dark too, but what if there are no changes in org.eclipse.pde.ui because there don't appear to be any necessary changes there...

laeubi:
> I'm just stabbing in the dark too, but what if there are no changes in org.eclipse.pde.ui because there don't appear to be any necessary changes there...

Yep it is a mystery :-) I even tried to bump the version of the UI, but the problem remains, it must be some strange PDE stuff here maybe something reexported or ...? But whatever I have tried the issue remains.

laeubi:
> but what if there are no changes in org.eclipse.pde.ui because there don't appear to be any necessary changes there

I rebased and removed any changes in the ui part so see if it makes any difference here.

laeubi:
What I now tried (as an extra commit) is bump the version of PDE-UI and raise the reexported minimum version of PDE-CORE, in the IDE it still complains, lets see what the build says here...

merks:
I'm going to reveal my complete ignorance.  I have no clue how I can fetch this PR (or your branch of your fork) into an IDE using EGit.  Assume that I am stupid and likely won't understand without pictures. 😛

laeubi:
1. copy URL (that is: `https://github.com/eclipse-pde/eclipse.pde/pull/214`)
2. Right click on the PDE repo and choose grafik
3. The make sure you select this repository as a source if you have multiple remotes grafik

merks:
In the IDE I can make **all** the errors in org.eclipse.pde.ui's MANIFEST.MF go away by replacing every occurrence of 3.13.600 with 3.14.0 

image

I just guessed at that solution and did it manually...

But no quickfix will fix this problem (the 61 variants of it):

image

One can try the quick fix, but nothing is saved and the problem does not go away.  So something seems fundamentally broken here in the API tools...

I think I would incrementally need to add the changes you made to see at which point does this error start to show.

laeubi:
Thanks Ed I applied as you suggested for the UI plugin, lets see if it makes a difference (the bad thing is that one gets other errors in the IDE than in the build sometimes...)

merks:
Oh, 503 errors, that's super helpful.

So I tried making the changes in a separate (older) IDE, and there were no problems.  Then I made the IDE refresh the API baseline so it rebuilds all; the 61 problems showed up.  Quick fix doesn't do anything yet again.  Then I replaced all the changed files with the originals, so the problems went away.  I put back the changes, but no problems showed at all this time.  Even refreshing the API baseline, no problems.  And now for the life of me I can't make errors show up in that IDE.  It's a mystery to me.  

But I also don't know why errors would be reported against a *.java file about an issue in some other **downstream** plugin's exports; it makes no sense...

So I try to remote debug this. Indeed the quick fix is a noop because it returns early here where it can't find the component corresponding to the org.eclipse.core.pde plugin:

image

What a nightmare.

laeubi:
> Oh, 503 errors, that's super helpful.

Eclipse Infra seem down: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1636

> And now for the life of me I can't make errors show up in that IDE. It's a mystery to me.

I also somehow got the errors went away in the IDE... hm... lets wait until the infra is back ...

merks:
> Error is still there but at least we see now nice GH annotation so one don't need to search the logfile ... ;-)
> 
> @eclipse-pde/eclipse-pde-committers any API tools expert her that can give a hint?

It took me a while to figure out to click on summary:

https://github.com/eclipse-pde/eclipse.pde/actions/runs/2816433739

What's strange here is that those problems in the IDE are reported on the *.java file not on the MANIFEST.MF or the org.eclipse.pde.ui bundle (where they belong):
```
Description	Resource	Path	Location	Type	ID
The re-exported type org.eclipse.pde.core.IBaseModel has been removed from org.eclipse.pde.ui_3.13.500	IBaseModel.java	/org.eclipse.pde.core/src/org/eclipse/pde/core	line 30	Compatibility Problem	105722
```
I think that's why quickfix in the IDE is not able to fix this problem because it's trying to fix in in org.eclipse.pde.core but that's not where it needs to be API filtered... 

I'm not sure how we can get this working without fixing whatever problem API tools is having reporting the problem in the wrong place...  Maybe it's not even a real proper problem to begin with...

merks:
Remote debugging I can see that the builder for org.eclipse.pde.ui is resolving to the *.java file when creating a marker:

image

It just seems wrong to me for a builder on a project to create markers on separate upstream project...

laeubi:
> It took me a while to figure out to click on summary:

These are now also show on the file tab:
https://github.com/eclipse-pde/eclipse.pde/pull/214/files#diff-3051fd10ea51ed839cc8645cca99a1ee59490e0163d08472c7fab6a567795fe3



> What's strange here is that those problems in the IDE are reported on the *.java file not of the MANIFEST.MF or the org.eclipse.pde.ui bundle (where they belong):

As far as I know the `ApiChecksApplication` uses the same markers as the IDE, (@mickaelistria might better tell about this), but probably because we only analyze one project here the `ApiChecksApplication` do not "see" the other project so results might vary here depending what projects are open in the workspace.

> It just seems wrong to me for a builder on a project to create markers on separate upstream project...

I think the problem, is that PDE-UI is reexporting content from PDE-CORE, so the builder might see the classes but they actually belong to a different project and thus are created for the containing one.

mickaelistria:
> As far as I know the ApiChecksApplication uses the same markers as the IDE, (@mickaelistria might better tell about this)

I confirm that.

> , but probably because we only analyze one project here the ApiChecksApplication do not "see" the other project so results might vary here depending what projects are open in the workspace.

It is expected that the API Tools will use the current project from the workspace and creates a target platform containing all the dependencies as listed by tycho-dependency-plugin.

Can this issue be reproduced with a local build?

laeubi:
@mickaelistria Yes if you use the Platform SDK and then checkout this PR you should get the markers as well but as @merks noticed they might be reported on the `*.java` file not of the `MANIFEST.MF`

laeubi:
@HannesWell do you maybe have an idea what could be causing this? I'll try rebase this once again and it would really be useful to include it so we can at least deliver this with m2e at 2023-03 release!

laeubi:
To ease analysis I just added one interface that is never used anywhere here:
- https://github.com/eclipse-pde/eclipse.pde/pull/374

lets see if APi tools complains there as well...

HannesWell:
> @HannesWell do you maybe have an idea what could be causing this? I'll try rebase this once again and it would really be useful to include it so we can at least deliver this with m2e at 2023-03 release!

I can have a Look this evening when back from Work and also review the change itself.

But I agree that this should be in included in 2022-12. We could also deliver the M2E counterpart for 2022-12 If we release 2.1 not as latest using I-builds and only change the latest pointer after the SimRel.

HannesWell:
Interestingly I only the many errors like `The re-exported type org.eclipse.pde.core.build.IBuildModelFactory has been removed from org.eclipse.pde.ui_3.13.600`, when I raise the minor version segment of pde.core. If the minor is not raised I only see one error for pde.ui
`The major version should be incremented in version 3.13.700, since API breakage occurred since version 3.13.600`

HannesWell:
Besides the strange API-tools errors this change looks good in general, but I can try it out with the corresponding M2E PR tomorrow.

laeubi:
But with this API error we obviously cant merge this.... its completely strange as only internal classes are changed

HannesWell:
> But with this API error we obviously cant merge this.... its completely strange as only internal classes are changed

That's right and it is completely strange. But as you said above the reexport probably contributes to this problem.
Probably one has to debug the API tools to see what happens there. But I won't have the time to do that today.

laeubi:
@merks already has done some debugging, and I have created 
- https://github.com/eclipse-pde/eclipse.pde/pull/374

with smaller commits and one can see that everything is fine (including adding new API) unless I add the actual implementation...

laeubi:
I reimplemented this now as a new extension point:
- https://github.com/eclipse-pde/eclipse.pde/pull/379

I'll keep this open for tracking bug 
- https://github.com/eclipse-pde/eclipse.pde/issues/378

merks:
@laeubi  I assume we could close this now, right?

HannesWell:
> @laeubi I assume we could close this now, right?

Lets close it. If really necessary it can be re-opened.

Issue #194, opened 2022-07-01
Validation error when launching child eclipse


sravanlakkimsetti:
We are getting this error
![image](https://user-images.githubusercontent.com/77106773/176875646-0fdfb7cd-251b-42f0-a8b9-a64d67f17281.png)

All these are for JDK packages. 

Build information:
Eclipse SDK
Version: 2022-06 (4.25)
Build id: I20220630-1800
OS: Windows 10, v.10.0, x86_64 / win32
Java vendor: Eclipse Adoptium
Java runtime version: 17.0.3+7
Java version: 17.0.3

Notifications: @SarikaSinha @akurtakov

sravanlakkimsetti:
These packages got added in Import-package directive in latest Jetty (version 10.0.11)

java.io,
java.lang,
java.lang.invoke,
java.net,
java.nio,
java.nio.channels,
java.nio.charset,
java.text,
java.util,
java.util.concurrent,
java.util.concurrent.atomic,
java.util.function,
java.util.regex,
java.util.stream,
java.util.zip,

vik-chand:
Happens 1st on 28th June build

gitlog of 28th 

2022-06-28	[Build input for build I20220628-1800](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/91f2e35dc3387bb556a1d7ff26694e726ad9226e)	Eclipse Releng Bot
2022-06-28	[Build input for build I20220628-1040](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/1099db163dcc053d65ac432e952a5cfde3469030)	Eclipse Releng Bot
2022-06-28	[Update to Jetty 10.0.11](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/bdf0a0b44e8f31fa2b38b6695367103fced40514)	Александър Куртаков

vik-chand:
<img width="916" alt="image" src="https://user-images.githubusercontent.com/99390647/177259941-c86c6b8d-cc0e-4fc9-9196-9ccd7132d123.png">

vik-chand:
I removed the import packages of  comment#2 (by Sravan) in org.eclipse.jetty.http and that particular no longer is in the list of validation error.

Another issue is why org.eclipse.osgi.internal.resolver.UserState@492d0748 cant resolve import of java.io etc

vik-chand:
[...]

PR #221, opened 2022-07-18
Fix system package launch validation

HannesWell:
Use the actual VM system packages for Launch-Validation.

The list of system-packages in `o.e.pde.api.tools/system_packages/<EE>-systempackages.profile` files is not suitable for Launch-validation.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/194.

Besides that I wonder if `ApiBaseline.getJavaProfilePropertiesForVMPackage()` could be replaced with the same method? This would also make creating the `system_packages/<EE>-systempackages.profile` files obsolete.


HannesWell:
@vik-chand do you want to review this?

vik-chand:
> @vik-chand do you want to review this?

I will review it by EOD today.

vik-chand (approved):
LGTM

vik-chand:
>Besides that I wonder if ApiBaseline.getJavaProfilePropertiesForVMPackage() could be replaced with the same method? >This would also make creating the system_packages/<EE>-systempackages.profile files obsolete.

I'll try removing the system-package in Y build and assuming launch with Java 19 works with the new code and adding a API baseline works without an error, then we could replace that code too in master stream.

HannesWell:
Thank you for reviewing Vikas.

flaviodonze:
Does this fix also apply to the PDE product-export validation?

grafik

HannesWell:
> Does this fix also apply to the PDE product-export validation?
> 
No it doesn't. But I strongly suspect that this has the same cause and can be fixed in the same way. When looking a bit if the API-profiles can be removed I noticed that PDE-Build uses them just like the launch validation did. So I think the same fix can be applied there.
I can probably provide a PR in the next days, unless you want to do that. :)

HannesWell:
> Does this fix also apply to the PDE product-export validation?
> 
Created https://github.com/eclipse-pde/eclipse.pde/pull/284 can you test the PR if that also works in your use case?

PR #227, opened 2022-07-20
Don't unconditionally dereference ClasspathJrt.getModule()

iloveeclipse:
While fixing https://github.com/eclipse-jdt/eclipse.jdt.core/pull/245
I've noticed that TargetPlatformHelper is the only caller of
ClasspathJrt.getModule() that does not check for null return, which is
however fully valid result.

Simply added a NP check to avoid NPE in
TargetPlatformHelper.querySystemPackages()

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
@@ -422,7 +422,11 @@ public class TargetPlatformHelper {
 			String path = new File(vm.getInstallLocation(), jrtPath).toString(); // $NON-NLS-1$
 			var jrt = org.eclipse.jdt.internal.core.builder.ClasspathLocation.forJrtSystem(path, null, null, release);
 			for (String moduleName : jrt.getModuleNames(null)) {
-				for (var packageExport : jrt.getModule(moduleName).exports()) {
+				var module = jrt.getModule(moduleName);
+				if (module == null) {
+					continue;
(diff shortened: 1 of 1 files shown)


HannesWell:
> @HannesWell : I don't know which part of SDK is broken now, but this tiny change fails on gerrit with obscure errors. Do you know which change causes now PDE Gerrit to fail? Is this the Orbit -> Maven transition, is this recent "pomless" change, is this something else?

There are indeed many API-tools tests failing at the current master:
https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/master/100/#showFailuresLink

Probably you encountered them too. My PR build from tonight succeeded without an error.
Since I did not (yet) used more of Tycho-pomless for PDE I suspect that it is related to the change to Maven-Central bundle.
My first guess is that somewhere in the test-setup some bundle names are hard-coded.

laeubi:
Th master build fails since merging this issue :-(

iloveeclipse:
> Th master build fails since merging this issue :-(

But if you look at the code, there is nothing that wouldn't be broken anyway. The patch only fixes NPE. Or do you mean, for the correct PDE tests the code **must** fail?

laeubi:
It would not wonder me if PDE sometimes relies on a NPE be thrown ;-)

I just opened #232 to better track this, just wanted to note that not Gerrit/PR verification has an issue here...

iloveeclipse:
> It would not wonder me if PDE sometimes relies on a NPE be thrown ;-)

Just try to revert this commit. Note: the official SDK tests are OK, therefore I would guess it's some dependency issue.

PR #231, opened 2022-07-25
Require the bundle/version annotation bundles when installing PDE

laeubi:
If found in the repository then include the optional bundle/version annotation bundles so they are available when choosing the "The running platform" as a target, this will also include them in update-sites so people can install them from the SDK.

### DIFF
--- a/features/org.eclipse.pde-feature/feature.xml
+++ b/features/org.eclipse.pde-feature/feature.xml
@@ -21,6 +21,8 @@
 
    <requires>
       <import feature="org.eclipse.jdt" version="3.15.0" match="compatible"/>
+      <import plugin="org.osgi.annotation.versioning"/>
+      <import plugin="org.osgi.annotation.bundle"/>
    </requires>
 
    <plugin


laeubi:
@merks What do you think? Any objections?

merks:
I thought there were concerns about installing the annotations.  But I'm not sure what harm they could cause though...

laeubi:
> I thought there were concerns about installing the annotations.

Just from P2 side, do you think it would be fine? Or is there a better way, e.g. requiring the IU (or artifact?) instead to not trigger an install in the IDE but still include it?

> But I'm not sure what harm they could cause though...

From my side they are just a bit of theoretical concern as even if I install them in the IDE, if no one use them, they just take up a few kilobytes, still I wondering if we can include them in the P2 without executing the install touchpoint.

features/org.eclipse.pde-feature/p2.inf
```diff
@@ -0,0 +1,10 @@
+# Require the annotation bundles so they get included when installing PDE
+requires.0.namespace = osgi.bundle
+requires.0.name = org.osgi.annotation.versioning
+requires.0.greedy = true
+requires.0.optional = true
+
+requires.1.namespace = osgi.bundle
+requires.1.name = org.osgi.annotation.bundle
+requires.1.greedy = true
+requires.1.optional = true
```
mickaelistria, line 10: why not simply adding them into the feature.xml?

laeubi: We only want to include them if they are present and do not require them in general.

mickaelistria: I don't think it's worth the complexity. Those would better be always part of the feature IMO. It makes everything simpler for maintenance and usage, and adds a very small footprint.

laeubi: As long as we finally get those in I'm fine with it, can you provide a PR with adding them to the features? Last time i tried, they where not available in the SDK even though the are added to the prereq target... Or do we need to add/commit this and then wait for the next I-Build?

mickaelistria: The .target is deployed with https://ci.eclipse.org/releng/job/deploy-eclipse-sdk-target-pom/ ; so if the change is in an this build has run, then the bundes should be in and available for inclusion in features.

laeubi: It was added with this commit: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/420

mickaelistria: OK, so now it should be in, while it may not have been the case in the few minutes after .target change was merged.

laeubi: I did update my SDk to latest version and performed the setup task but still PDE does not offer me these bundles, I'll try to ad them by hand and the lets see if jenkins complains...

mickaelistria: I have them when looking at the "Content" tab of the eclipse-sdk-prereqs.target.

features/org.eclipse.pde-feature/build.properties
```diff
@@ -11,6 +11,6 @@
 # Contributors:
 #     IBM Corporation - initial API and implementation
 ###############################################################################
-bin.includes=\
-feature.xml,\
-feature.properties
+bin.includes = feature.xml,\
```
mickaelistria, line 14: Not sure if the build failure is related, but this file should probably be left unchanged.

mickaelistria:
Note that we're in a code-freeze period (the check is failing to detect that), so please refrain for further merges until M2 is complete.

Issue #232, opened 2022-07-25
org.eclipse.pde.api.tools.tests has 611 failing tests with Tycho Surefire, nightly build is fine


laeubi:
See https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/master/101/
and https://download.eclipse.org/eclipse/downloads/drops4/I20220724-1800/testresults/html/org.eclipse.pde.api.tools.tests_ep425I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

Looks like all org.eclipse.pde.api.tools.builder.tests.compatibility tests are failing on Jenkins after this PR https://github.com/eclipse-pde/eclipse.pde/pull/227.

Typical fail looks like:

```
junit.framework.AssertionFailedError: The project was not found in the workspace baseline: bundle.a
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleVersionTests.setupTest(BundleVersionTests.java:161)
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleVersionTests.test003(BundleVersionTests.java:223)
```

Reverting https://github.com/eclipse-pde/eclipse.pde/pull/227 fixes tests on Jenkins.
???

laeubi:
@mickaelistria @akurtakov any idea why API tools failing now?

akurtakov:
No idea. I pushed https://github.com/eclipse-pde/eclipse.pde/commit/b4ec281656556256554c6b4ab68870d014e8003e today so we are on the safe side but I haven't looked into more details.

merks:
Could this be related?

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/419

laeubi:
@merks it is not that the API **check** is failing but the API tool **test** itself.

laeubi:
I triggered a new build now jsut in case it was something temporal... https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/master/

laeubi:
The log currently shows a lot of
```

[INFO] --- tycho-compiler-plugin:3.0.0-SNAPSHOT:compile (default-compile) @ org.eclipse.pde.genericeditor.extension.tests ---
20:14:12  [INFO] Compiling 12 source files to /home/jenkins/agent/workspace/eclipse.pde.ui_master/ui/org.eclipse.pde.genericeditor.extension.tests/target/classes
[...]

PR #236, opened 2022-07-25
Trying to understand PDE test fails in Jenkins...

iloveeclipse:
The implicit types resolving for "var" seem to not work properly on
Jenkins, may be due different compilation settings / dependencies used?

See https://github.com/eclipse-pde/eclipse.pde/issues/232

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
@@ -43,6 +43,9 @@ import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.internal.compiler.env.IModule;
+import org.eclipse.jdt.internal.compiler.env.IModule.IPackageExport;
+import org.eclipse.jdt.internal.core.builder.ClasspathJrt;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.jdt.launching.environments.IExecutionEnvironment;
@@ -68,6 +71,7 @@ import org.osgi.framework.Constants;
 import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.Version;
 
+@SuppressWarnings("restriction")
 public class TargetPlatformHelper {
 
 	public static final String REFERENCE_PREFIX = "reference:"; //$NON-NLS-1$
@@ -403,7 +407,6 @@ public class TargetPlatformHelper {
 		return new Dictionary[] { TargetPlatformHelper.getTargetEnvironment(state) };
 	}
 
-	@SuppressWarnings("restriction")
 	public static String querySystemPackages(IExecutionEnvironment environment) {
 		IVMInstall vm = bestVmInstallFor(environment);
 		if (vm == null || !JavaRuntime.isModularJava(vm)) {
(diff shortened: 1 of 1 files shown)


mickaelistria:
Do a local build with `mvn clean verify -Pbuild-individual-bundles` reproduce the issue?

laeubi:
I have created https://github.com/eclipse/tycho/issues/1197 as it seems there is some check missing in Tycho...

mickaelistria:
One approach is to compare the list of bundles listed in <testModule>/target/work/configuration/config.ini with the ones that are part of the launch configuration, and progressively try to make the launch configuration as close as possible from Tycho one until it becomes clear which bundle is in the PDE launch that is missing in Tycho execution.

laeubi:
@iloveeclipse I also see a lots of:

```
 java.nio.file.NoSuchFileException: eclipse.pde/ui/org.eclipse.pde.core/target/ant_tasks/pde-ant.jar-classes
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
	at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1851)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1264)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:709)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:243)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:172)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:186)
	at org.eclipse.jdt.internal.compiler.batch.ClasspathJar.initialize(ClasspathJar.java:201)
	at org.eclipse.jdt.internal.compiler.batch.ClasspathMultiReleaseJar.initialize(ClasspathMultiReleaseJar.java:38)
	at org.eclipse.jdt.internal.compiler.batch.FileSystem.<init>(FileSystem.java:235)
	at org.eclipse.jdt.internal.compiler.batch.Main.getLibraryAccess(Main.java:3495)
	at org.eclipse.jdt.internal.compiler.batch.Main.performCompilation(Main.java:4741)
	at org.eclipse.jdt.internal.compiler.batch.Main.compile(Main.java:1802)
```


what is somewhat expected as this is a folder, I'm just wondering if JDT requires a jar file?

iloveeclipse:
> what is somewhat expected as this is a folder, I'm just wondering if JDT requires a jar file?

That's a bug. I have a fix, but that would most likely break the build...

iloveeclipse:
> eclipse.pde/ui/org.eclipse.pde.core/target/ant_tasks/pde-ant.jar-classes

We should fix that, where is this - in maven/tycho/build properties?

laeubi:
> That's a bug. I have a fix

In the compiler?

> but that would most likely break the build...

I'm already wondering why the build succeeds, I though it should fail the compilation in such cases?



> We should fix that, where is this - in maven/tycho/build properties?

Platform uses a crude combination of tycho, maven, and buildproperties declaring subjars to be build, and every time I ask it says **OH NO WE NEED THAT** this particular one is:
https://github.com/eclipse-pde/eclipse.pde/blob/6e1939b8018f40edb8c2c3d11493a734e9c1452a/ui/org.eclipse.pde.core/build.properties#L32

but there are man other places as well!

iloveeclipse:
> this particular one is:

I don't see ant_tasks/pde-ant.jar**-classes** anywhere. Where and how is that "-classes" thing passed to ECJ?

iloveeclipse:
I guess org.sat4j.core.jar is referenced in a MANIFEST.MF somewhere, but not as required bundle but in "Class-Path" attribute. I believe JDT scans that and automagically adds everything it can find there to the classpath.

laeubi:
Yes I created
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/261

laeubi:
I really don't get it...

- https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/PR-231/5/ (on top of master) **succeeds**
- https://ci.eclipse.org/pde/job/eclipse.pde.ui/job/master/107/ (now merged) **fails**

laeubi:
@merks do you know about the baseline stuff? The tests complain about:

> The project was not found in the workspace baseline: bundle.a

I also enabled stack traces so we can better see where things are going wrong:

- https://github.com/eclipse-pde/eclipse.pde/pull/241

merks:
@laeubi No, sorry, I've never looked at the implementation details around that.

mickaelistria:
As mentioned in https://github.com/eclipse-pde/eclipse.pde/issues/232#issuecomment-1197797747 , you don't need Jenkins to reproduce and debug the issue. Testing can more efficiently be done locally and you could even plug in the debugger.
So I think it'd be better to close this draft PR and keep discussing of progress (including "I tried X, it didn't work') on the ticket rather than scattering the discussion.

mickaelistria:
Now working as expected.

Issue #238, opened 2022-07-27
Cannot invoke "org.eclipse.pde.core.plugin.ModelEntry.getWorkspaceModels()" because "swtEntries" is null


fbertini:
Hello everyone,

I have installed the latest Eclipse IDE version (today it is eclipse-jee-2022-06-R-win32-x86_64.zip) and I always get the following error hovering the mouse on the RUN/debug button on the toolbar.

Cannot invoke "org.eclipse.pde.core.plugin.ModelEntry.getWorkspaceModels()" because "swtEntries" is null

Notice that this is happening only when I am working on a JUnit4 class.

PDE Version 4.24

References: [Eclipse Bugzilla](https://bugs.eclipse.org/bugs/show_bug.cgi?id=580456)

iloveeclipse:
@fbertini : please attach here Eclipse error **log** (you can find it under <your workspace>./metadata/.log).

fbertini:
Please find attached log file as requested.
[log.zip](https://github.com/eclipse-pde/eclipse.pde/files/9196425/log.zip)

Thank you!

iloveeclipse:
```
!ENTRY org.eclipse.core.jobs 4 2 2022-07-27 08:27:29.045
!MESSAGE An internal error occurred during: "Compute launch button tooltip".
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.pde.core.plugin.ModelEntry.getWorkspaceModels()" because "swtEntries" is null
	at org.eclipse.pde.internal.launching.launcher.LauncherUtils.requiresUI(LauncherUtils.java:331)
	at org.eclipse.pde.ui.launcher.JUnitWorkbenchLaunchShortcut.createLaunchConfiguration(JUnitWorkbenchLaunchShortcut.java:59)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchShortcut.findExistingLaunchConfigurations(JUnitLaunchShortcut.java:590)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchShortcut.getLaunchConfigurations(JUnitLaunchShortcut.java:558)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchShortcutExtension.getLaunchConfigurations(LaunchShortcutExtension.java:496)
	at org.eclipse.debug.internal.ui.contextlaunching.LaunchingResourceManager.getParticipatingLaunchConfigurations(LaunchingResourceManager.java:489)
[...]

PR #240, opened 2022-07-27
Don't fail with NPE if "org.eclipse.swt" is not in the target platform

iloveeclipse:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/238


ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LauncherUtils.java
```diff
@@ -328,6 +328,9 @@ public static boolean requiresUI(ILaunchConfiguration configuration) {
 					if (model != null) {
 						Set<BundleDescription> plugins = DependencyManager.getSelfAndDependencies(Set.of(model));
```
HannesWell, line 329: As little optimization this line can be moved below the null Check.

HannesWell (approved):
Thank you. The fix Looks good.

HannesWell:
Perfect. 👍🏽 Thanks!

Issue #249, opened 2022-07-29
Resolve classpath of required bundle during Tycho Surefire tests


gerking:
(Issue imported from https://bugs.eclipse.org/bugs/show_bug.cgi?id=578744)

When running a plug-in test with the Tycho Surefire plugin, the combination of PDE/JDT fails to resolve the classpath of a plug-in project if its required bundles include the bundle under test. In this case (i.e., if the bundle under test is among the required bundles), its classpath resolves to the bundle root instead of the `target/classes` directory (where the classes actually reside during a Tycho build).

The same test succeeds when running as a JUnit plug-in test inside a local Eclipse. Here, the classpath of the required bundle is successfully resolved to the `target/classes` directory.

A very similar issue has first been reported back in 2015 on the tycho-user mailing list: https://www.eclipse.org/lists/tycho-user/msg06725.html

mickaelistria:
PDE is not at all involved when running tests in Tycho. Please report it to https://github.com/eclipse/tycho/issues

gerking:
> Please report it to https://github.com/eclipse/tycho/issues

Done: eclipse-tycho/tycho#1274

PR #250, opened 2022-07-29
Add test case (#249)

gerking:


mickaelistria:
As mentioned, this should be reported to Tycho.

gerking:
> As mentioned, this should be reported to Tycho.

Yes, sorry. I was preparing a pull request for demonstration and didn't read your response on time.

PR #251, opened 2022-07-31
dogfooding test on pde build of latest enhanced for loop cleanup

carstenartur:
Don't be afraid. It will not be merged. It is just to test the 
Version: 2022-09 M1 (4.25.0 M1)
Build id: 20220714-1455
build of the enhanced for loop cleanup together with real world code.
This part contains code that does not contain any "red" error after applying the cleanup.
It still contains some issues (e.g. removed comments) but as some issues use to show up in a build this is a test run this build.


vogella:
@carstenartur did you find any issues with the clean-up? If not, we should review and merge if the change is good.

carstenartur:
> @carstenartur did you find any issues with the clean-up? If not, we should review and merge if the change is good.

Hi Lars,
I created it to check the latest version of the enhanced for each cleanup for bugs. As it still has a lot of hits it seems to be suitable. The comments removed by the cleanup are intentionaly (more or less) removed I just learned. So for this issue it is no longer needed. Regarding some of the known issues in the context of multithreaded access to collections and array the enhanced for loop cleanup still is not smart enough to detect them. This means you have to carefully evaluate otherwise a shit storm is garantied.
But my holiday just ends today and I do not want to keep it open for long time. Maybe in a few months...

Issue #217, opened 2022-07-16
[API tools] output the full path of the file?


laeubi:
Currently a message looks like this 

> * Marker [on: /org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF, id: 52, type: org.eclipse.pde.api.tools.version_numbering, attributes: [apiMarkerID: 4, charEnd: 132, charStart: 114, description: - The field org.eclipse.e4.core.contexts.IEclipseContext.PROPERTY_CONTEXT has been added in an interface that is tagged with '@noimplement', lineNumber: 4, message: The minor version should be incremented in version 1.10.100, since new APIs have been added since version 1.10.0, messagearguments: 1.10.100#1.10.0, problemid: 924844039, severity: 2, sourceId: API Tools, version: 1.11.0.qualifier], created: 7/16/22, 8:13 AM]

The file (here /org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF) is printed relative to the "workspace", for automatic tools detecting the issue an annotating files, it would be good to have a full path here, i'd like therefore suggest to include the full path in the attributes e.g. `filePath:  /full/path/to/plugins/org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF` in case the resource can be converted to a file.

Issue #216, opened 2022-07-16
[API Tools] extra line break at end of message?


laeubi:
I see the following in a build log
```
* Marker [on: /org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF, id: 52, type: org.eclipse.pde.api.tools.version_numbering, attributes: [apiMarkerID: 4, charEnd: 132, charStart: 114, description: - The field org.eclipse.e4.core.contexts.IEclipseContext.PROPERTY_CONTEXT has been added in an interface that is tagged with '@noimplement'
, lineNumber: 4, message: The minor version should be incremented in version 1.10.100, since new APIs have been added since version 1.10.0, messagearguments: 1.10.100#1.10.0, problemid: 924844039, severity: 2, sourceId: API Tools, version: 1.11.0.qualifier], created: 7/16/22, 8:13 AM]
```
It seems there is a trailing new line in the message, this currently breaks automatic detection of API problems in GH action, it would be good if the message could be trimmed.

PR #264, opened 2022-08-03
Format the output of markers of the ApiAnalysisApplication

laeubi:
Currently the ApiAnalysisApplication uses the toString representation of
the marker this has some drawbacks:

- it is unclear how it looks like as toString has only weak semantics
- it is very hard to read by the user
- it is very hard to reliable parse by tools

This changes the output to the following format:

`[SERVERITY] File FILENAME at line LINENUMBER: MESSAGE (location: PATH)`

This will fix https://github.com/eclipse-pde/eclipse.pde/issues/217 and fix https://github.com/eclipse-pde/eclipse.pde/issues/216

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
@@ -36,6 +36,7 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.ICoreRunnable;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
@@ -150,7 +151,7 @@ public class ApiAnalysisApplication implements IApplication {
 			if (allNonAPIErrors.length > 0) {
 				System.err.println("Some blocking (most likely link/compilation) errors are present:"); //$NON-NLS-1$
 				for (IMarker marker : allNonAPIErrors) {
-					System.err.println("* " + marker); //$NON-NLS-1$
+					printMarker(marker, "FATAL"); //$NON-NLS-1$
 				}
 				System.err.println("Some blocking (most likely link/compilation) errors are present ^^^"); //$NON-NLS-1$
 				return 10;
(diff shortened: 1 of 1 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
```diff
@@ -189,6 +190,31 @@ public Object start(IApplicationContext context) throws Exception {
...
 	}
 
+	private static void printMarker(IMarker marker, String type) {
+		String path = getFullPath(marker);
+		String file = marker.getResource().getName();
+		int lineNumber = marker.getAttribute(IMarker.LINE_NUMBER, -1);
+		String message = marker.getAttribute(IMarker.MESSAGE, "").trim(); //$NON-NLS-1$
+		String description = marker.getAttribute("description", "").trim(); //$NON-NLS-1$ //$NON-NLS-2$
+		if (!description.isEmpty()) {
+			message = String.format("%s %s", message, description); //$NON-NLS-1$
```
HannesWell, line 200: Why not use String concatenation directly here and below?
IMHO this is more readable and performance whise this should probably better since Java 9/10/11.

HannesWell (approved):
Look's good to me (but could only review on the phone).

Only one cosmetic suggestion.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
```diff
@@ -189,6 +190,31 @@ public Object start(IApplicationContext context) throws Exception {
 		}
 	}
 
+	private static void printMarker(IMarker marker, String type) {
+		String path = getFullPath(marker);
+		String file = marker.getResource().getName();
+		int lineNumber = marker.getAttribute(IMarker.LINE_NUMBER, -1);
+		String message = marker.getAttribute(IMarker.MESSAGE, "").trim(); //$NON-NLS-1$
```
HannesWell, line 197: strip() removes all kind of whitespaces.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
laeubi: I don't think it matters much here, but I found it often easier to read using String.format, that's not really a performance concern here and you only ever need one NON-NLS per line ;-)

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
laeubi: I don't think that it makes any difference here.

laeubi:
If there are no strong concern I'll merge this today as the current output makes me really mad and is hard to use, we really need better tools support here, I think if there are smaller style changes we can improve this later on.

HannesWell:
I'm fine with this as it is.
The suggestion are a matter of personal preferences so no need wait to discuss them. :)

PR #265, opened 2022-08-03
Temporarily disable OSGiAnnotationsCompilationParticipant (#69)

iloveeclipse:
The new OSGiAnnotationsCompilationParticipant is disabled because of
massive performance issues during compilation.

See https://github.com/eclipse-pde/eclipse.pde/issues/69

### DIFF
--- a/ui/org.eclipse.pde.ui/plugin.xml
+++ b/ui/org.eclipse.pde.ui/plugin.xml
@@ -2376,7 +2376,10 @@
       	</action>
       </objectContribution>
   </extension>
-      <extension
+  <!-- Compilation participant below disabled because of massive performance issues -->
+  <!-- See https://github.com/eclipse-pde/eclipse.pde/issues/69 --> 
+  <!-- 
+  <extension
          point="org.eclipse.jdt.core.compilationParticipant">
       <compilationParticipant
             class="org.eclipse.pde.internal.ui.annotations.OSGiAnnotationsCompilationParticipant"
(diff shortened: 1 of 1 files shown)


laeubi:
@iloveeclipse I would prefer to take a look first if we can improve here instead of disabling this by default, so any hint what could be done better here as there is similar code in PDE on other places.

laeubi:
I think the problematic part is:

```
ASTParser parser = ASTParser.newParser(AST.JLS3);
parser.setResolveBindings(true);
parser.setBindingsRecovery(true);
```

Two questions:

1. is there a better way to inspect the AST (e.g. is there an AST cached somewhere?)
2. is the ResolveBindings / BindingsRecover required if I only want to process annotations? (javadoc mention it might be costly to do so)

To give an example where the AST is used is here:
https://github.com/eclipse-pde/eclipse.pde/blob/883fe148dbe465e38fd9512ec87d5dd02a949986/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/annotations/CustomHeaderAnnotationProcessor.java#L36

iloveeclipse:
> @iloveeclipse I would prefer to take a look first if we can improve here instead of disabling this by default, so any hint what could be done better here as there is similar code in PDE on other places.

Sure, as said, I plan to merge that this evening **except** there is a proper solution found.
Using SDK now is impossible due slow compilation.

iloveeclipse:
Note: jenkins build failed on unrelated test `WorkspaceProductModelManagerTest.testProducts_productAddedAndRemoved`.

iloveeclipse:
Hopefully not needed after #266

PR #266, opened 2022-08-03
Only process package-info.java in OSGiAnnotationsCompilationParticipant

laeubi:
Currently the OSGiAnnotationsCompilationParticipant scans every source file but actually currently only processing of annotations on `package-info.java` is supported, thus we can short-circuit here to not slow down things to much in cases where it is know that nothing has to be done.

@iloveeclipse I think this should recover previous performance effectively without disabling this feature. Still need to investigate if there are more performant ways to investigate the annotations itself.


laeubi:
@iloveeclipse do you think its worth triggering a new Ibuild?

iloveeclipse:
> @iloveeclipse do you think its worth triggering a new Ibuild?

Yes, but more important because of https://github.com/eclipse-platform/eclipse.platform/pull/64

I will try to do that.

iloveeclipse:
> I will try to do that.

Murphy's law: guess what do we see there: https://ci.eclipse.org/releng/view/Builds/job/I-build-4.25/

image

laeubi:
> > I will try to do that.
> 
> Murphy's law: guess what do we see there: https://ci.eclipse.org/releng/view/Builds/job/I-build-4.25/
> 
> image

It is actually more because "infra" is considered alow-priority service without any special SLA if not requested by a "strategic member" :-\

PR #268, opened 2022-08-05
Remove m2e nature and builder from org.eclipse.pde.unittest.junit

merks:
This causes errors in the workspace's pom.xml when m2e is installed.


merks:
This change is relent for this issue:

https://github.com/eclipse-platform/.github/issues/10#issuecomment-1205398439

laeubi (approved):

merks:
I doubt the build is affected but let's see and hope there aren't random unrelated failures...

laeubi:
Thanks for taking care of this!

Issue #36, opened 2022-04-21
Support DS 1.4


laeubi:
Currently DS supports a maximum level of 1.3 for DS components.

HannesWell:
When enabling 1.4 I think it would be best to include 1.5 as well (unless there are major additional difficulties, I have not yet looked into the details):
https://search.maven.org/artifact/org.osgi/org.osgi.service.component.annotations

laeubi:
Sure I have not yet what has changed between 1.3 -> 1.4 -> 1.5 maybe @tjwatson or @bjhargrave can give a short summary.

bjhargrave:
DS 1.3 -> 1.4: https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e46608

DS 1.4 -> 1.5: https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e39795

laeubi:
thanks for the doc reference, I have one question regarding [constructor injection](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-constructor.injection) (maybe also just overseen that):

Ho is this supposed to work with annotation? Should a annotation processor simply assume that constructor injection is desired if there is a public non-default constructor?

bjhargrave:
The constructor must be annotated `@Activate` and the constructor arguments which are not activation object types must be annotated `@Reference`.

See <https://github.com/osgi/osgi/blob/b9f994ed76c755ebe2634465a27e46429ee0b836/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/ConstructorInjection.java#L49-L58> for a test example.

laeubi:
Sounds obvious, thanks for the example!

kwin:
Particularly useful are https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#org.osgi.service.component.propertytypes. Unfortunately there is no warning from PDE when those non-supported property type annotations are found. They are just silently ignored.

HannesWell:
As suggested in [#269](https://github.com/eclipse-pde/eclipse.pde/pull/269#issuecomment-1207167253) we could consider to use the PDE DS processor internally to generate the component XML.

But only if it is possible to get back errors/warnings from the BND-processor, so that they can be displayed in the java files.
This might be more work in the beginning but for future versions we can inherit the implementation work from BND.

[...]

PR #269, opened 2022-08-06
Always provide the official annotations if present in target

laeubi:
Currently PDE ships an (outdated) version of the DS annotations, like
with the bunde/version annotations it would be better to simply fetch
the latest version from the given users target.

### DIFF
--- a/features/org.eclipse.pde-feature/feature.xml
+++ b/features/org.eclipse.pde-feature/feature.xml
@@ -23,6 +23,7 @@
       <import feature="org.eclipse.jdt" version="3.15.0" match="compatible"/>
       <import plugin="org.osgi.annotation.versioning"/>
       <import plugin="org.osgi.annotation.bundle"/>
+      <import plugin="org.osgi.service.component.annotations"/>
    </requires>
 
    <plugin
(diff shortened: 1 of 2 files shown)


iloveeclipse:
> PDE ships an (outdated) version of the DS annotations,

Wouldn't be better to fix **that** too? Maybe the "better" version isn't in the target?

laeubi:
> Wouldn't be better to fix **that** too? Maybe the "better" version isn't in the target?

Yes that would be the next step to completely get rid of these old wrapper bundles (see https://github.com/eclipse-pde/eclipse.pde/discussions/252) but first we net to get the new version into PDE.
I think this will require one I-Build and thus if this one succeeds I can start tomorrow getting rid of the old stuff.

HannesWell:
To be able to fully use them the PDE DS processor has to support them to.
Do you Plan to work on that as well (https://github.com/eclipse-pde/eclipse.pde/issues/36)?

laeubi:
> To be able to fully use them the PDE DS processor has to support them to.

At least Tycho support that already as it uses BND under the hood :-)

> Do you Plan to work on that as well (#36)?

If I find the time, yes, especially the constructor injection is very useful.

HannesWell:
> > To be able to fully use them the PDE DS processor has to support them to.
> 
> At least Tycho support that already as it uses BND under the hood :-)

Nice!
I already wondered If we should use BND in PDE-DS as processor too. That would simplify future updates top.

> 
> > Do you Plan to work on that as well (#36)?
> 
> If I find the time, yes, especially the constructor injection is very useful.

Please let me know if you start working on it to prevent duplicated work.I have that on my List as well, but only with medium priority, so it could take longer.

laeubi:
> I already wondered If we should use BND in PDE-DS as processor too. That would simplify future updates top.

I think that generating the XML is not really the most-valuable thing here but the checking and feedback for the users about possible problems and that something BND probably can not do very good (but I haven't checked yet) as it runs as a plugin inside the analyzer so it seems hard to get useful messages back from that.

> Please let me know if you start working on it to prevent duplicated work.I have that on my List as well, but only with medium priority, so it could take longer.

Same here, at the moment there are more priority work like getting the API tools working better in CI-Builds and the general integration of the Bundle-Integration and so on...

Issue #272, opened 2022-08-09
ClasspathContributorTest::testAdditionalClasspathEntries fails


vik-chand:
```
e	Unexpected classpath entry found: /Users/genie.releng/workspace/ep425I-unit-mac64-java17/workarea/I20220808-0820/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.osgi.service.component.annotations_1.5.0.202109301733.jar[CPE_LIBRARY][K_BINARY][sourcePath:/Users/genie.releng/workspace/ep425I-unit-mac64-java17/workarea/I20220808-0820/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.osgi.service.component.annotations_1.5.0.202109301733.jar][rootPath:/][isExported:false]

java.lang.AssertionError: Unexpected classpath entry found: /Users/genie.releng/workspace/ep425I-unit-mac64-java17/workarea/I20220808-0820/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.osgi.service.component.annotations_1.5.0.202109301733.jar[CPE_LIBRARY][K_BINARY][sourcePath:/Users/genie.releng/workspace/ep425I-unit-mac64-java17/workarea/I20220808-0820/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.osgi.service.component.annotations_1.5.0.202109301733.jar][rootPath:/][isExported:false]
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.pde.ui.tests.classpathcontributor.ClasspathContributorTest.testAdditionalClasspathEntries(ClasspathContributorTest.java:53)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.Native
```

PR #273, opened 2022-08-09
Add org.osgi.service.component.annotations to the allowed Classpath

laeubi:
Fix https://github.com/eclipse-pde/eclipse.pde/issues/272

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathcontributor/ClasspathContributorTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathcontributor/ClasspathContributorTest.java
@@ -70,6 +70,10 @@ public class ClasspathContributorTest {
 			// osgi bundle annotations
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


iloveeclipse:
```
java.lang.AssertionError: 
Generated product fails validation: 
[Missing Constraint: Import-Package: org.apache.commons.io; version="1.3.1", Missing Constraint: Import-Package: org.apache.commons.io.output; version="1.3.1"]
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.pde.ui.templates.tests.TestPDETemplates.validateProduct(TestPDETemplates.java:193)
```

PDE tests miss now `org.apache.commons.io` after recent changes (commons.io symbolic name changed), unrelated to this PR.
@laeubi : would you like to fix that too?

laeubi:
@iloveeclipse I poked around a bit but it seems the template needs to be adjusted and I haven't yet find out how this exactly works I think this could be `/org.eclipse.pde.ui.templates/templates_3.5/E4Application/$pluginId$.product` but lets update this in a separate PR...

PR #275, opened 2022-08-09
Update commons.io in example product

laeubi:
Lets see if that works...


HannesWell:
Since the PDE Product launcher as well as Tycho when building products adds all transitive dependencies by default it is actually sufficient to add the roots of the dependency graph to the product. So I suspect that many plug-ins can actually be removed.

laeubi:
There is just one of the API tools failures so I think the test is fixed now, I leave it to @iloveeclipse if we should merge this...

akurtakov:
This is weird as this PR changes the commons-io symbolic name to the Orbit one, so doesn't sound correct to me.

laeubi:
@akurtakov I just started the SDK and used the commons IO that is provided by the target platform... and it seems the test itself is happy now...

akurtakov:
No longer needed as commons-io is no longer explicitly listed.

PR #277, opened 2022-08-10
Update *.target editor to support Maven artifacts

akurtakov:
Handles coloring and completion.
Tests are adjusted accordingly.

### DIFF
--- a/ui/org.eclipse.pde.genericeditor.extension.tests/src/org/eclipse/pde/genericeditor/extension/tests/Bug527084CompletionWithCommentsTest.java
+++ b/ui/org.eclipse.pde.genericeditor.extension.tests/src/org/eclipse/pde/genericeditor/extension/tests/Bug527084CompletionWithCommentsTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 Red Hat Inc. and others
+ * Copyright (c) 2017, 2022 Red Hat Inc. and others
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -29,7 +29,7 @@ public class Bug527084CompletionWithCommentsTest extends AbstractTargetEditorTes
 
 	@Test
 	public void testTagNameCompletion() {
-		confirmCompletionAtOffset(89, "unit");
+		confirmCompletionAtOffset(89, "dependencies", "unit");
 	}
 
 	@Test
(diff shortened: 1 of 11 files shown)


merks:
Given the maven location is provided by an extension to PDE it seems odd that built-in support is provided for it in PDE itself.  

If I contributed analogous built-in support for Oomph's targlets now, would that be okay too?

It would seem better if the extension mechanism itself provided a way to contribute such things, but I understand that would be much more work...

akurtakov:
I would say targlets support should be fine too :) . My driver for this one is that I edit eclipse-platform-parent/pom.xml with this editor but never set is target platform .

vogella:
Awesome @akurtakov, this will make the usage of Maven dependencies much easier.

akurtakov:
I thought a bit more about targlets and the fact that I never used them as I thought they are not direclty usable from build scripts (my main driver), which could be totally wrong. This just points smth I've been thinking a lot lately, we have too many silos and with limited manpower it's time to lower the barriers and move things higher in the stack when we see them as significantly improving/simplifying things.

laeubi:
I think @mickaelistria could maybe give some pointers here, last time I checked the Generic Editor was/is used and one could specify "regions" that have a special content type so maybe Ommph/m2e can the hook in?

merks:
> I thought a bit more about targlets and the fact that I never used them as I thought they are not direclty usable from build scripts (my main driver), which could be totally wrong. This just points smth I've been thinking a lot lately, we have too many silos and with limited manpower it's time to lower the barriers and move things higher in the stack when we see them as significantly improving/simplifying things.

The problem is that moving such a thing pulls in a deep stack of dependencies:

image

Probably, for similar reasons, moving m2e (or the PDE parts there of) to the Platform also doesn't make so much sense...

akurtakov:
It's that way if we try to move things as a whole. But e.g. with this patch we can support editing of target files with Maven deps without any extra dependency and releng work could be done without caring about m2e and it's dependencies.

akurtakov:
So is anyone strictly against merging this one?

merks:
@akurtakov I have no problem with these changes!  It was just an observation.  For editing targlets embedded in a *.target file the user can open a structured editor and then don't need to know the XML syntax...

laeubi:
> So is anyone strictly against merging this one?

No, if its not a problem for PDE to support "unknown" items.

I think this would then solve https://github.com/eclipse-m2e/m2e-core/issues/555

PR #284, opened 2022-08-11
Use the actual system-packages from the JVM for Product/Site export

HannesWell:
As reported in https://github.com/eclipse-pde/eclipse.pde/pull/221#issuecomment-1198011695 #194 can also occur when exporting Products.

This applies the same fix to PDE-build. In my little test-case that worked.

Since pde.core depends on pde.build I had to copy the code from `TargetPlatformHelper` to PDE build. Looking at other places this is not the first time the dependency issue was solved like that.

Additionally this applies a minor shortcut for `TargetPlatformHelper.querySystemPackages()`

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -14,14 +14,15 @@
 package org.eclipse.pde.internal.build.site;
 
 import java.io.*;
-import java.net.URL;
-import java.net.URLConnection;
 import java.util.*;
 import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import org.eclipse.core.runtime.*;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.launching.IVMInstall;
+import org.eclipse.jdt.launching.JavaRuntime;
+import org.eclipse.jdt.launching.environments.IExecutionEnvironment;
 import org.eclipse.osgi.service.resolver.*;
 import org.eclipse.osgi.service.resolver.VersionRange;
 import org.eclipse.osgi.util.ManifestElement;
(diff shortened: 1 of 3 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -462,6 +461,50 @@ public void resolveState() {
...
+			return null;
+		}
+		String release = environment.getProfileProperties().getProperty(JavaCore.COMPILER_COMPLIANCE);
+		try {
+			Collection<String> packages = new TreeSet<>();
+			String jrtPath = "lib/" + org.eclipse.jdt.internal.compiler.util.JRTUtil.JRT_FS_JAR; //$NON-NLS-1$
+			String path = new File(vm.getInstallLocation(), jrtPath).toString(); // $NON-NLS-1$
+			var jrt = org.eclipse.jdt.internal.core.builder.ClasspathLocation.forJrtSystem(path, null, null, release);
+			for (String moduleName : jrt.getModuleNames(null)) {
+				for (var packageExport : jrt.getModule(moduleName).exports()) {
```
iloveeclipse, line 478: Please don't dereference `jrt.getModule(moduleName)`  unconditionally. This doesn't guarantee to return non-null value.

See https://github.com/eclipse-pde/eclipse.pde/pull/286

iloveeclipse (dismissed):

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
HannesWell: Thanks for the hint. Applied the same patch.

flaviodonze:
Sorry for my late answer, I was on holiday. 
I just tested the export with the integration build _eclipse-SDK-I20220817-1800-win32-x86_64_ and I don't run into the validation problem anymore. Thanks for fixing!

PR #285, opened 2022-08-11
Remove now unused API-tools system-packages profiles

HannesWell:
With https://github.com/eclipse-pde/eclipse.pde/pull/279 and https://github.com/eclipse-pde/eclipse.pde/pull/284 the system-packages profiles embeded in the pde.api.tools plugin are not used anywhere. At least I was not able to find an occurrence.

Therefore I think they can be removed.


vik-chand:
There is a hack in org.eclipse.pde.internal.build.site.PDEState. I will recommend keeping the system packages for 4.25.

vik-chand (commented):
There is a hack in org.eclipse.pde.internal.build.site.PDEState. I will recommend keeping the system packages for 4.25.

HannesWell:
> There is a hack in org.eclipse.pde.internal.build.site.PDEState. I will recommend keeping the system packages for 4.25.

To what hack are you referring exactly? The one just removed in https://github.com/eclipse-pde/eclipse.pde/pull/284?
Nevertheless I have no problem with holding this back until the next release-cycle.

vik-chand:
Now, with https://github.com/eclipse-pde/eclipse.pde/commit/ea34245635594625ba7a6dbbd84fe35046009b4e
we can probably remove system packages just after M3 ( Saturday/Sunday). I tested export after removing system packages and it worked fine.

iloveeclipse:
> we can probably remove system packages just after M3

Why after M3? Better we get it into M3 if the change is safe, and remove from RC1 if it turned out it is not.
Otherwise we will run into troubles removing the code short before release.

HannesWell:
> > we can probably remove system packages just after M3
> 
> Why after M3? Better we get it into M3 if the change is safe, and remove from RC1 if it turned out it is not. Otherwise we will run into troubles removing the code short before release.

I agree with Andrey. If we want to have this in 2022-09 it is more reasonable to submit this sooner than later.

HannesWell:
I just adjusted the code-comments to this change.

vik-chand:
Ok, in that case, we should merge ASAP.

vik-chand:
My only concern was not merging on a Tuesday ( testing day) but since I've already tested the scenario ( by using the launch after taking this change), I think it should be fine to merge today. I will test again on the next I build post the merge.

vik-chand (approved):
Tested, it looks well.

iloveeclipse:
611 test fails here too... https://github.com/eclipse-pde/eclipse.pde/pull/286

HannesWell:
> Ok, in that case, we should merge ASAP.

Done. :)

> 
> #232 needs solution.

Yes definitely. I'll try to look into it this evening. Maybe I'm lucky.

vik-chand:
> > Ok, in that case, we should merge ASAP.
> 
> Done. :)
> 
> > #232 needs solution.
> 
> Yes definitely. I'll try to look into it this evening. Maybe I'm lucky.

The next I build is 20 min away

PR #286, opened 2022-08-12
Don't unconditionally dereference ClasspathJrt.getModule()

iloveeclipse:
While fixing https://github.com/eclipse-jdt/eclipse.jdt.core/pull/245
I've noticed that TargetPlatformHelper is the only caller of
ClasspathJrt.getModule() that does not check for null return, which is
however fully valid result.

Simply added a NP check to avoid NPE in
TargetPlatformHelper.querySystemPackages().


iloveeclipse:
@vik-chand : this change always triggers 611 test fails, like in https://github.com/eclipse-pde/eclipse.pde/issues/232.

But it can't be the root cause. I would really like to merge that in 4.25, because packages are now computed via JDT and the code in JDT can return null.

vik-chand (approved):
looks good

iloveeclipse:
@vik-chand : OK to merge for RC1?

vik-chand:
> @vik-chand : OK to merge for RC1?

yes

PR #287, opened 2022-08-13
Disable artifact upload for windows due to actions/upload-artifact#240

laeubi:
See https://github.com/actions/upload-artifact/issues/240

FYI @vogella


HannesWell:
Thanks for disabling this. I don't think the test-failures are related to this change.
https://github.com/eclipse-pde/eclipse.pde/issues/232

PR #288, opened 2022-08-13
Clean up product launching

HannesWell:
Clean up Product LaunchAction: format, use braces and minor clean-ups
+ Replace arrays by collections

Simplify Product LaunchAction and associated classes
- Make recursive collection of child features iterative
- Reuse code to collect plugin-models from LaunchAction in other places

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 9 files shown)


HannesWell:
Looks like https://github.com/eclipse-pde/eclipse.pde/issues/232 is also affecting this change. But as far as I can tell, I did not fix any NPE.

HannesWell:
As expected this change does not cause a real regression. The 611 tests just fail even with an empty change.

iloveeclipse:
Hannes, does it mean we can push whatever we want?
I mean, https://github.com/eclipse-pde/eclipse.pde/pull/286 is a trivial fix, but gerrit always fails.

laeubi:
They do not fail **always** I have also seen PRs where they don't fail but haven't yet found what the difference is...

HannesWell:
> Hannes, does it mean we can push whatever we want? I mean, #286 is a trivial fix, but gerrit always fails.

With special care, that the corresponding change is not really causing the test-failure and although this is unpleasant I would say it is OK in this case. Because otherwise it means we cannot continue developing. And this is a prerequisite for mixed products that I then would have to delay until the next release cycle because I don't think I will find a solution for the test failres this weekend.

And as Christoph said the failures are random. For this PR the 611 failures occurred even when I have reverted the two commits and this change was effectively empty. But then they only occurred on GH, Jenkins was ok (actually there was one other of the occasional failures). Now all 611 fail. In https://github.com/eclipse-pde/eclipse.pde/pull/290 only GH has 611 failures. Jenkins is completely fine. I cannot tell what's going on at the moment, but I'm relatively sure that those two changes are not causing it.

HannesWell:
The master build with this change succeeded on Jenkins and had 611 failures on GH.

Issue #189, opened 2022-06-29
SourceTests.testBug206679_247198 fails since I20220628-1800


iloveeclipse:
After Jetty update SourceTests.testBug206679_247198 fails on all platforms.
```
testBug206679_247198 | Error 
Processing inclusion from feature org.eclipse.help: Bundle org.eclipse.equinox.http.jetty_3.8.100.v20211021-1418 failed to resolve.: Unsatisfied import package org.eclipse.jetty.http_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.server_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.server.handler_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.server.session_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.servlet_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.util_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.util.component_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.util.log_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.util.ssl_[10.0.2,11.0.0). Unsatisfied import package org.eclipse.jetty.util.thread_[10.0.2,11.0.0).

/home/jenkins/agent/workspace/ep425I-unit-cen64-gtk3-java11/workarea/I20220628-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.pde.build_3.11.200.v20220503-0921/scripts/genericTargets.xml:112: Processing inclusion from feature org.eclipse.help: Bundle org.eclipse.equinox.http.jetty_3.8.100.v20211021-1418 failed to resolve.:
Unsatisfied import package org.eclipse.jetty.http_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.server_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.server.handler_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.server.session_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.servlet_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.util_[10.0.2,11.0.0).
Unsatisfied import package org.eclipse.jetty.util.component_[10.0.2,11.0.0).
[...]

PR #289, opened 2022-08-13
Improve error reporting of PDE's ANT-tasks

HannesWell:
This could help to solve https://github.com/eclipse-pde/eclipse.pde/issues/189.


akurtakov:
Reporting as much information about build errors as possible makes sense.

PR #290, opened 2022-08-13
Leverage Tycho's pom-less capabilities

HannesWell:
Remove now unnecessary 'skipAPIAnalysis' properties because the
api-check profile of Eclipse platform now automatically skips the
ananlysis for projects without api-nature.

Remove (then) trivial pom.xml that can be generated by tycho-pomless.

Remove the pom.xml that specify only properties besides the
other basic data and specify these properties in the build.properties as
pom.model.property.<property-name> instead. See:
https://github.com/eclipse/tycho/wiki/Tycho-Pomless#define-properties


HannesWell:
The number of modules, executed tests and warnings remained unchanged and Jenkins succeeded.
The 611 failures on GH are probably https://github.com/eclipse-pde/eclipse.pde/issues/232.

PR #291, opened 2022-08-13
Bug 325614 - Support mixed products (consisting of features and bundles)

HannesWell:
[Bug 325614](https://bugs.eclipse.org/bugs/show_bug.cgi?id=325614): Add support for mixed-type products (like P2 supports them for a while), in the Product-Editor and when launching Products from the editor.
The boolean-attribute 'fUseFeatures' is replaced by enum-attribute 'type' whoose allowed values are 'bundles', 'features' and 'mixed'. All names are based on the model in P2.

To launch a mixed product from the IDE a feature based launch-configuration is created and all product-plugins are simply added as 'additional plugins'.

In the product overview one can now select `plug-ins`, `features` and `mixed`:

grafik

The `Contents` tab then of the Product-Editor is adjusted to show the `Plug-ins` and the `Features` section side-by-side for mixed products so that one can modify both simultaneously:

grafik

Alternatively the split could be horizontally:

grafik

Unfortunately there is no (or I did not find a simple way) to add a slicer or similar. Another alternative would be keep this in two tabs. But for now I think a vertical or horizontal split is a acceptable, relatively simple solution. One can also change the product to `plug-ins` or `features` to get the full screen and change back after a modification.

@vogella you wanted to be notified when I have completed this feature.

Everybody else is invited to test this. I would like to submit this latest on Monday evening so that it will make it into M3.


laeubi:
> In the product overview one can now select `plug-ins`, `features` and `mixed`:

I think the term "mixed" is not very good here, what about making the radio button a checkbox, or simply use "both" as a label?

> The `Contents` tab then of the Product-Editor is adjusted to show the `Plug-ins` and the `Features` section side-by-side for mixed products so that one can modify both simultaneously:

Why not one big list that simply shows both and allows to add both? That would save a few duplicated buttons as well.

**One thing that came into my mind as well:**

Why not drop the "this product is based on" completely and simply compute the value from what the user is choosing?

merks:
I know people out there have products that have both lists populated but are only using one of the lists (features).  I'm not sure how they will be affected...

HannesWell:
> > In the product overview one can now select `plug-ins`, `features` and `mixed`:
> 
> I think the term "mixed" is not very good here, what about making the radio button a checkbox, or simply use "both" as a label or even "plugin and features" and chance the other labels to "only plugins" and "only features"?


Yes I thought about that as well. I now changed it to `plug-ins only`, `features only` and `both, plug-ins and feature`. Starting with 'only' to me sounds a bit colloquially and less technical. But so does the mixed label now. So I'm not yet fully happy with those names. Maybe we should just stick with the previous ones and only adjust the new one:
`plug-ins`, `features` and `plug-ins and features (combined)`? 
Finding good names is really the toughest part. 😅 

grafik


> 
> > The `Contents` tab then of the Product-Editor is adjusted to show the `Plug-ins` and the `Features` section side-by-side for mixed products so that one can modify both simultaneously:
> 
> Why not one big list that simply shows both and allows to add both? That would save a few duplicated buttons as well.

Not a bad idea, but that would require more coding work that I can't do until tomorrow evening (2022-09 M3). The charm of the current side by side approach is that it was simple to implement.

> 
> **One thing that came into my mind as well:**
> 
> Why not drop the "this product is based on" completely and simply compute the value from what the user is choosing?

That's an even better idea. The same could actually be done for launches. A product could then just have one `units` sections (similar to P2's `InstallableUnit` term). For features the `.feature.group` suffix would have to be added when serializing the product.
But that would require even more work, in PDE, P2 and Tycho. So this is nothing I can do for 2022-09. We should have a dedicated discussion for that since this is a major change how products are modeled and Eclipse apps are launched. But I would support it.

That being said I think the current approach is something one can work with, even it is not yet perfect. But besides the labels from my side this is ready.



> I know people out there have products that have both lists populated but are only using one of the lists (features). I'm not sure how they will be affected...

They are not affected because such products have the `useFeatures` attribute either set to `false` or `true` which is translated to `type=bundles` or `type=features`. In either case only one of both sets is used and the other one is ignored. Only if one selected `mixed`, both sets are considered. I added some tests to ensure existing products continue to work as they are.

laeubi:
> Yes I thought about that as well. I now changed it to `plug-ins only`, `features only` and `both, plug-ins and feature`.

I would simply strip the "both" part

> Not a bad idea, but that would require more coding work that I can't do until tomorrow evening (2022-09 M3). The charm of the current side by side approach is that it was simple to implement.

Is there any need to rush here? I don't think there is an immediate need for this given we already have had this missing for years now and I think a simpler an cleaner UI is always better.

> But that would require even more work, in PDE, P2 and Tycho.

Not if you keep the serialized format as-is, this ist just a matter of how you present things to the user. Lets say I only have added features = set to feature only, same for bundle.... if I add a bundle, then activate the mixed mode...

jhonnen (approved):
I've only done a cursory review, but that looks good. Nice bump in functionality with only minor changes 👍 

A unified list would be nice, but for now the (vertical) split is absolutely fine - we can always improve on that later.
Also a unified list would probably need functional adaptations, e.g. what does *add required plug-ins* do when a feature dependency is missing?

> Why not drop the "this product is based on" completely and simply compute the value from what the user is choosing?

That would break Ed's valid concern regarding backwards compatibility. We would have to implement some kind of migration first.

laeubi:
> That would break Ed's valid concern regarding backwards compatibility. We would have to implement some kind of migration first.

That won't break anything as the product would know on loading that this is "feature only" or "plugin only" and could act accordingly.

> Also a unified list would probably need functional adaptations, e.g. what does _add required plug-ins_ do when a feature dependency is missing?

The same question arises here... presenting two list do not solves this but makes it more confusing to the user, e.g. why ever should it only add required plugins of plugins but not of features? The whole purpose of a mixed product actually is to allow adding "missing" things without creating a full blown feature for that.

jhonnen:
> That won't break anything as the product would know on loading that this is "feature only" or "plugin only" and could act accordingly.

Example: you open an old product file with `useFeature=false`, but containing both plugins and features. The type toggle is computed to `plugins only` for backwards compatibility. How do you then switch to mixed mode?
I'm not saying this is huge issue, it's just needs to be thought out ;)

> The same question arises here... presenting two list do not solves this but makes it more confusing to the user, e.g. why ever should it only add required plugins of plugins but not of features? The whole purpose of a mixed product actually is to allow adding "missing" things without creating a full blown feature for that.

It's certainly a useability issue that should be solved in the near future, but I feel in it's current form this features brings a net improvement.

laeubi:
> Example: you open an old product file with `useFeature=false`, but containing both plugins and features. The type toggle is computed to `plugins only` for backwards compatibility. How do you then switch to mixed mode?
> I'm not saying this is huge issue, it's just needs to be thought out ;)

You simply add a feature to the product :-) PDE **might** ask to add the currently hidden ones as well, but honestly it is rather surprising that these data is stored at all, whenever I witch a product I have to delete this (then outdated/obsolete) stuff by hand what could be rather annoying...

> It's certainly a usability issue that should be solved in the near future, but I feel in it's current form this features brings a net improvement.

Why not then solve it now and release it in that near future? I think the UI is already a bit overwhelmed and just making it more complex do not seems sufficient. As a user not knowing much about this feature I would have several questions:

1. Why is features section smaller than plugins? Are features less important? Do plugins usually have longer names?
2. Why is there not just one `Add` button? Whats the difference between "Add requirements" and add "Required Plugins", are required plugins not requirements of features?
3. Will "Include required Plugins automatically" only apply to required plugins of items added on the right section? Or will it also add required plugins of features (from the left section)
4. Will "Include optional dependencies" (referencing the button on the right) only apply to items added on the right part or will it also add optional dependencies of features from the left?
5. Why I can/should sort features but not bundles

So I really wonder if we should introduce new UI with the hope that we will change it right away in the near future (so users need to adapt two times).

HannesWell:
> > Yes I thought about that as well. I now changed it to `plug-ins only`, `features only` and `both, plug-ins and feature`.
> 
> I would simply strip the "both" part

Yes the both part is not good.
Currently I think of three options:

- `plug-ins only`, `features only`, `plug-ins and features (combined)`
- `plug-ins`, `features`, `plug-ins and features (combined)`
- `plug-ins`, `features`, `plug-ins and features`

We have to keep in mind that those radio-buttons are limiting the minmal width of the first editor page so shorter ist better.
Currently I'm in favor of option two or three.
grafik


> > Not a bad idea, but that would require more coding work that I can't do until tomorrow evening (2022-09 M3). The charm of the current side by side approach is that it was simple to implement.
> 
> Is there any need to rush here? I don't think there is an immediate need for this given we already have had this missing for years now and I think a simpler an cleaner UI is always better.

Frankly I was surprised another time how early M3 is, I expected it to be next week. But yes I would like to use that in my products. But of course, just like everybody else I can continue to create a single-plug-in feature to add single plug-ins. So not having it is not a show stopper.

But on the other hand, I always get told that it is important that things get released frequently and that I should not wait until everything is perfect and that things can be done incrementally. So this would be the case here.

> I've only done a cursory review, but that looks good. Nice bump in functionality with only minor changes 👍
> 
> A unified list would be nice, but for now the (vertical) split is absolutely fine - we can always improve on that later. Also a unified list would probably need functional adaptations, e.g. what does _add required plug-ins_ do when a feature dependency is missing?

Thank you for your review. That is also my opinion.

> > Also a unified list would probably need functional adaptations, e.g. what does _add required plug-ins_ do when a feature dependency is missing?
> 
> The same question arises here... presenting two list do not solves this but makes it more confusing to the user, e.g. why ever should it only add required plugins of plugins but not of features? The whole purpose of a mixed product actually is to allow adding "missing" things without creating a full blown feature for that.

Actually with automatic-inclusion of dependencies enabled there are no missing requirements anymore (unless they have resolution errors like the `apache.xml.resolver` had). So that butten gets less important. I even think we should consider to remove those buttons because adding all dependencies results in over-defined products that are harder to maintain and understand. I think it is better to add the roots of the dependencies tree (usually the features/plugins you really want in your application) and let PDE/Tycho/P2 add all requirements. This also reduces the impact of changes in dependencies. But that is a general other question.


> > It's certainly a usability issue that should be solved in the near future, but I feel in it's current form this features brings a net improvement.
> 
> Why not then solve it now and release it in that near future? I think the UI is already a bit overwhelmed and just making it more complex do not seems sufficient. As a user not knowing much about the history of this feature I would have several questions:

These are all valid points.
I simplify don't have the time to address them today. And do you think we should postpone this feature therefore?
If you have time and want to address it, I can merge this now and you can apply them? That would of course be very welcome.

> Why is features section smaller than plugins? Are features less important? Do plugins usually have longer names?

If anybody can tell me how to break lines in label-texts then this issue can be addressed quickly, since the lower label is limiting the width of that section. Or a shorter but equivalent precise text would also help.

laeubi:
> Currently I think of three options:

I don't think "combined" adds anything useful, we don't really combine anything.

> Frankly I was surprised another time how early M3 is, I expected it to be next week. But yes I would like to use that in my products.

You can still use this, but not in the UI or simply use an I-Build of PDE :-)

> But on the other hand, I always get told that it is important that things get released frequently and that I should not wait until everything is perfect and that things can be done incrementally. So this would be the case here.

Still there is no value in rushing things in just because M3 is on the way, we are not talking about making it perfect, I think its more about how we generally handle this, and that could be much more effort than the coding itself.

> I simplify don't have the time to address them today. And do you think we should postpone this feature therefore?

Yes. Given there is no time to really try this out or more think about it I think we should not put this in.

> If you have time and want to address it, I can merge this now and you can apply them? That would of course be very welcome.

I don't think I will have time to work on these, and especially given today is last day for "changes" and as I said, these are more general concerns why I think this new feature is not yet ready to be merged.

HannesWell:
> > Frankly I was surprised another time how early M3 is, I expected it to be next week. But yes I would like to use that in my products.
> 
> You can still use this, but not in the UI or simply use an I-Build of PDE :-)
> 

My colleges are not so much into using I-builds. So unfortunately in my daily work I cannot use features that are only available in I-builds.

> > But on the other hand, I always get told that it is important that things get released frequently and that I should not wait until everything is perfect and that things can be done incrementally. So this would be the case here.
> 
> Still there is no value in rushing things in just because M3 is on the way, we are not talking about making it perfect, I think its more about how we generally handle this, and that could be much more effort than the coding itself.

OK then lets postpone this until this is better worked out. I will start corresponding discussions in the next days.

laeubi:
@HannesWell one thing I forget to mention:

All products carry a version `<?pde version="3.5"?>` I think if we change the format, this also needs to be adjusted to introduce a new product serialization version!

HannesWell:
> @HannesWell one thing I forget to mention:
> 
> All products carry a version `<?pde version="3.5"?>` I think if we change the format, this also needs to be adjusted to introduce a new product serialization version!

IIRC I have only seen checks for such version for *.target files, but I have not verified that explicitly and might be wrong.
But if that is considered the version should be incremented, yes.

HannesWell:
I started to improve the Contents tab for mixed products, but won't be able to complete that until tomorrow so I will miss M3 and consequently Eclipse 2022-12.
Nevertheless I would like to finish this then as soon as the master reopens for the spring release.

So what do you think about the following layout?
The Add Button will add Plugin or a Feature depending on the current selection.

It is basically a mixture of the Plugin and Feature section but without the Add required buttons. PDE-build probably can't export mixed products anyways and since the add required buttons should better be removed as discussed in https://github.com/eclipse-pde/eclipse.pde/discussions/331, new ones should not be added now

grafik

I'm also considering to not add the UP/DOWN buttons (which would only be enabled if one feature is selected) and just sort features alphabetically as well.

laeubi:
Looks promising, but I won't make a distinction when adding stuff and simply have a dialog where you can select both, it should be enough to have the icon to see if I'm adding a feature or a bundle

HannesWell:
> Looks promising, but I won't make a distinction when adding stuff and simply have a dialog where you can select both, it should be enough to have the icon to see if I'm adding a feature or a bundle

Personally I'm not always happen if I have to know from the icon what kind of element is presented to me. Therefore I personally like to have a little bit more context.

Nevertheless I'm thinking about keeping the table layout like in the feature-only or plugins-only content and just grouping them together and inserting the headline to the table as well (if there is an element of such kind). But I have to check if that looks good.
To not make the `Add` button context sensitive it could be replaced by a `Add Feature` plus a `Add Plug-in` button. That would have the advantage that the existing dialogs can just be re-used.
I'm also not sure if having a dialog to add both Features and Plugin gives the best user experience because there are usually much more plugins than feature the number of items one can select from is much probably much higher. Given the example, that I have the Eclipse SDK in my TP and want to add the `o.e.pde` feature to a product. If I have distinct buttons for each type I can hit `Add Feature`, type `pde` in the search and will likely end with one result. If Features and Plugins are available in the dialog the number of matches is much higher. Of course, if one want's to add a Plugin, having Features select-able as well, does not increase the number of items that much, but it still increases it.
So besides that it likely makes the implementation harder (but I can check that), I'm not sure if that would improve the workflow.

vogella:
UI looks great to me. Now that PDE uses Java 17 you may want to use the Java 17 switch statements.

Conflicts need to be resolved, once this is done I can test it with my examples to see if I find any issue. 

I actually would be nice UX simplification if the plug-in / feature / combined option could be removed. For example, if "combined" is not used in a product you could open a dialog asking the user, if he wants  to use only plug-ins -> delete all features in the product and set mixed or if he wants to use only features -> delete all plug-ins and set the mode to mixed. This way old product configuration files would be migrated once.

vogella:
@HannesWell can you fix the conflcts?

HannesWell:
@vogella I hope to find time to continue on this (and hopefully complete this) at the weekend.

> I actually would be nice UX simplification if the plug-in / feature / combined option could be removed. For example, if "combined" is not used in a product you could open a dialog asking the user, if he wants to use only plug-ins -> delete all features in the product and set mixed or if he wants to use only features -> delete all plug-ins and set the mode to mixed. This way old product configuration files would be migrated once.

There is much potential for improvements, in general I thought about removing the distinction between Plugin and Feature based products completely and only let users mix as they want it.
The same could actually also be done for Launch configs, which should make their use and implementation simpler.

But lets keep this focused on supporting mixed products and discuss further steps later.

HannesWell:
M3 again came earlier than assumed... I have reworked this draft to implement my suggestion from https://github.com/eclipse-pde/eclipse.pde/pull/291#issuecomment-1309307257. Currently it looks like this:

grafik

An empty product looks like

grafik

I think the buttons are self-explanatory and if not applicable they are disabled.
I'm also contemplating to only add the head-lines/labels if there is an element of that type present.

Does anybody know if it is possible to make the text in a single line in the table bold and if yes, how?

My plan is to test this as good as possible and then complete and submit this PR this evening just in time for M3.

iloveeclipse:
> Does anybody know if it is possible to make the text in a single line in the table bold and if yes, how?

I assume you are using JFace and so your `ILabelProvider` has to implement `IFontProvider` and return bold font on right element.

iloveeclipse:
> and submit this PR this evening just in time for M3.

Note that line `Monday - Last day of development, (and even then, no "big changes").`
in https://www.eclipse.org/lists/platform-releng-dev/msg39305.html

I assume what you plan here **is** a big change.

laeubi:
> I'm also contemplating to only add the head-lines/labels if there is an element of that type present.

I think actually the icon is enough to distinguish both types, if one absolutely wants, you can use a table with two columns, one with the ID and one with the type, the one can even make them sortable...

laeubi:
From a first debugging, the issue is mostly because

`https://sourceforge.net/projects/birt/files/release/4.16.0/p2.index`

redirects to

`https://sourceforge.net/projects/birt/files/release/4.16.0/p2.index/`

(please note the last `/` appended to the URL!) but Tycho tries to download a file and those maps both things to the cache file `.cache/tycho/https/sourceforge.net/projects/birt/files/release/4.16.0/p2.index` so on the second attempt it decides that there is already a cached variant of the URL and reuses the previous headers...

HannesWell:
> > Does anybody know if it is possible to make the text in a single line in the table bold and if yes, how?
> 
> I assume you are using JFace and so your `ILabelProvider` has to implement `IFontProvider` and return bold font on right element.

Thanks for the hint, I'll try that. 👍🏽 



> > and submit this PR this evening just in time for M3.
> 
> Note that line `Monday - Last day of development, (and even then, no "big changes").` in https://www.eclipse.org/lists/platform-releng-dev/msg39305.html
> 
> I assume what you plan here **is** a big change.

It is definitely not a small one, but I would have said it is a medium sized change. But ok, then lets postpone this.
This time I just should really finish this now so that it can be submitted as soon as the master opens for 4.29.



> > I'm also contemplating to only add the head-lines/labels if there is an element of that type present.
> 
> I think actually the icon is enough to distinguish both types, if one absolutely wants, you can use a table with two columns, one with the ID and one with the type, the one can even make them sortable...

Personally I'm not such a big fan of deriving the item type from an icon because there is no explanation for that (AFAIK) and you have to know this from experience. But I'll try out your suggestion with two columns.

HannesWell:
I just updated the PR on top of the master. It now also contains a commit to unify the `PluginSection` and `FeatureSection` class, that I want to submit in a separate PR once this is settled.

> > Personally I'm not such a big fan of deriving the item type from an icon because there is no explanation for that (AFAIK) and you have to know this from experience.
> 
> Strictly speaking then "Plugins and Fragments" should be splitted as well :-) so there would be actually four types:
> 
>     * Features
> 
>     * Bundles
> 
>     * Fragments
> 
>     * Patch Features

Well the intention of the split mainly reflects how the IUs are stored in the products respectively how one could choose before. Of course this is not a requirement but `Features` (projects with a feature.xml) and Bundles (projects with a MANIFEST.MF) are at least for me, the two 'main' categories, I'm interested in when assembling a product. Of course this could be split further but I think that's not necessary (and I also assume that's not what you want to argue for :) ). Of course others may have another preference on that.
Another reason for the current split is that different action are possible on features than on Plugins/Fragments. On Features the install mode can be toggled and they can be sorted manually, while both is not possible on Plugins (they are always sorted alphabetically). And the Properties button opens different dialogs for Features and for Plugins/Fragments.
Therefore I believe it makes sense to have that kind of split and group Features and to group Plugins.
All of that is simplify because it was like this in the Contents page of Feature-only or Plugin-only Products and I don't want to completely change that as well.


> > But I'll try out your suggestion with two columns.
> 
> You can even think about a third one for the version

If one specifies a version that is already displayed (via the label provider) in braces.

With the current state of this PR the contents page look like this:

grafik

Without labels it looks like
grafik

I have not completed my attempt to use a second column, but in general personally I like the current one with separating labels.

HannesWell:
Instead of using labels for Features respectively Plugins and Fragments, I could also add a tooltip that shows the type together with the element ID/BSN. Then the view would be simpler but it should still be clear what the elements represents, even if one does not know the icons.

Another point are the Up/Down buttons for features. In traditional feature-based products one can sort the features manually while in plug-in based products, the plugins are always sorted alphabetically. In order to simplify the UI I contemplate to remove the Up/Down buttons and sort features always alphabetically too in mixed products.
The only technical reason where the order of Features might be relevant are root features, because as far as I know/understand they are installed one after another when building a product. So theoretically with another order the final product could consist of different IUs. But on the other hand since they all use the same TP, I hardly believe this can really happen. The only way I could imagine are P2-Instructions that remove something that another feature adds.
@merks, @jonahgraham you have worked a lot on building products, did you ever have the need to adjust the order of Features or was this just eye-candy?

merks:
I've never thought the order could/would be semantically meaningful but it's true that touch points can modify things and the order of that modification might be relevant if they are modifying the same things, e.g., setting a system property in the eclipse.ini.  So no, I've never considered I should change the order for a semantic reason until you mentioned it.

jonahgraham:
> @merks, @jonahgraham you have worked a lot on building products, did you ever have the need to adjust the order of Features or was this just eye-candy?

I have never used it to semantically change anything. I have a minor use case where in the XML file I use the order to differentiate between "common" and "package specific" items in products:

- [common features across all epps](https://github.com/eclipse-packaging/packages/blob/150d889c5a596b51e01d57c2a846d630837d4c0d/packages/org.eclipse.epp.package.cpp.product/epp.product#L226)
- [product specific features](https://github.com/eclipse-packaging/packages/blob/150d889c5a596b51e01d57c2a846d630837d4c0d/packages/org.eclipse.epp.package.cpp.product/epp.product#L242)

vogella:
>It is definitely not a small one, but I would have said it is a medium sized change. But ok, then lets postpone this.
> This time I just should really finish this now so that it can be submitted as soon as the master opens for 4.29.

@HannesWell M1 is a good time for medium / big changes.

HannesWell:
Thanks Ed and Jonah. Then I think the capability to move up/down manually should stay for now, even though it is probably only rarely relevant.

I have again reworked the mixed content section now a bit and as requested it does not contain labels anymore. To help users to distinguish between Plugin/Fragments and Features, even if they don't know the icons a tool-tip is now shown that also mentions the elements type, when hovering over it:
grafik

With that I think every concern raised in between is now addressed and I'm now finally fine with this PR as it is.
From the involvement of others until now, I assume nobody wants do to a in-depth review of this.
Therefore I plan to merge this as soon as the CI succeeds and we can (hopefully) use this starting with tonight's I-builds.

> > It is definitely not a small one, but I would have said it is a medium sized change. But ok, then lets postpone this.
> > This time I just should really finish this now so that it can be submitted as soon as the master opens for 4.29.
> 
> @HannesWell M1 is a good time for medium / big changes.

It looks like this time is now. :)

PR #292, opened 2022-08-14
Test Launching from Product-Editor and fix handling of absent features

HannesWell:
- Fix handling of absent features when launching a Product from the Editor
- Add tests for Launching of Products from the Editor
- Move test-dummy-feature creation code to a more general place to make it re-usable


HannesWell:
Again the 611 failures from https://github.com/eclipse-pde/eclipse.pde/issues/232. This is again probably unrelated. https://github.com/eclipse-pde/eclipse.pde/pull/291, which also contains this change succeeds on Jenkins without these test failures.

PR #296, opened 2022-08-16
Clear workspace before each API-tools test class

HannesWell:
When looking closer at the failures of https://github.com/eclipse-pde/eclipse.pde/issues/232 I noticed that at least some tests like the `BundleVersionTests` fail with a missing `bundle.a` in the API baseline. When I run that test case in the IDE, in the `setupTest(String)` method there is no such project in the workspace and therefore it is not searched for such project in the API-baseline. From the test code I also can find a reason why a project called `bundle.a` should be present.

Therefore I suspect that `bundle.a` is leaking into the workspace from another test case. And maybe it helps to add an extra clean of the workspace.
But actually the workspace should be cleaned already before. But lets see, maybe we are lucky.


HannesWell:
Looks like this didn't help.

Issue #297, opened 2022-08-17
PDEPluginImages: NPE due to using SWT in background thread


jukzi:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Display.disposeExec(java.lang.Runnable)" because "toQuery" is null
	at org.eclipse.jface.resource.JFaceResources.getResources(JFaceResources.java:224)
	at org.eclipse.jface.resource.ImageRegistry.<init>(ImageRegistry.java:150)
	at org.eclipse.jface.resource.ImageRegistry.<init>(ImageRegistry.java:122)
	at org.eclipse.pde.internal.ui.PDEPluginImages.initialize(PDEPluginImages.java:325)
	at org.eclipse.pde.internal.ui.PDEPluginImages.get(PDEPluginImages.java:319)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.getImage(ManifestContentAssistProcessor.java:787)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.addPackageCompletions(ManifestContentAssistProcessor.java:692)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.handleExportPackageCompletion(ManifestContentAssistProcessor.java:444)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.computeValue(ManifestContentAssistProcessor.java:215)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.computeCompletionProposals(ManifestContentAssistProcessor.java:165)
	at org.eclipse.pde.internal.ui.editor.contentassist.ManifestContentAssistProcessor.computeCompletionProposals(ManifestContentAssistProcessor.java:97)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$10(AsyncCompletionProposalPopup.java:352)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$9(AsyncCompletionProposalPopup.java:351)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
[...]

PR #313, opened 2022-08-26
PDEPluginImages: fix NullPointerException #297

jukzi:
Images need to be initialized in Display tread.
ManifestContentAssistProcessor runs in background thread and use images.


jukzi:
611 failed tests? but it does not show which failed! what is this?

laeubi:
> 611 failed tests? but it does not show which failed! what is this?

https://github.com/eclipse-pde/eclipse.pde/issues/232

jukzi:
unrelated error:
```
org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testDelete_projectWithChangedBundleRoot
java.lang.AssertionError: expected null, but was:<plugin.a>
```

Issue #190, opened 2022-06-29
NPE in RequiresSection


jukzi:
i do not know how to reproduce. As far as i remember i edited MANIFEST.MF "Require-Bundle:" manually and started a full build. i had a little UI freeze before this error poped  up.

eclipse.buildId=4.23.0.I20220308-0310

```
java.lang.NullPointerException
	at org.eclipse.pde.internal.ui.editor.plugin.RequiresSection.createImportObjects(RequiresSection.java:651)
	at org.eclipse.pde.internal.ui.editor.plugin.RequiresSection$ImportContentProvider.getElements(RequiresSection.java:79)
	at org.eclipse.jface.viewers.StructuredViewer.getRawChildren(StructuredViewer.java:947)
	at org.eclipse.jface.viewers.ColumnViewer.getRawChildren(ColumnViewer.java:704)
	at org.eclipse.jface.viewers.AbstractTableViewer.getRawChildren(AbstractTableViewer.java:1038)
	at org.eclipse.jface.viewers.StructuredViewer.getFilteredChildren(StructuredViewer.java:850)
	at org.eclipse.jface.viewers.StructuredViewer.getSortedChildren(StructuredViewer.java:1031)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefreshAll(AbstractTableViewer.java:670)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:618)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:610)
	at org.eclipse.jface.viewers.StructuredViewer.lambda$2(StructuredViewer.java:1459)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1398)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1359)
	at org.eclipse.jface.viewers.StructuredViewer.refresh(StructuredViewer.java:1459)
	at org.eclipse.jface.viewers.ColumnViewer.refresh(ColumnViewer.java:526)
	at org.eclipse.jface.viewers.StructuredViewer.refresh(StructuredViewer.java:1418)
	at org.eclipse.pde.internal.ui.editor.plugin.RequiresSection.lambda$0(RequiresSection.java:632)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
[...]

PR #315, opened 2022-08-26
fix NPE in RequiresSection #190

jukzi:
getModel() can return null


jukzi:

HannesWell (approved):
Looks good.
Thanks for the contribution!

Issue #316, opened 2022-08-30
-nl parameter from product not corretly added to launch configuration


vogella:
If I add -nl de to the product, the -nl is not added to the launch configuration.

It becomes: -os ${target.os} -ws ${target.ws} -arch ${target.arch} -nl ${target.nl} -consoleLog de

Instead of 

-os ${target.os} -ws ${target.ws} -arch ${target.arch} -nl de -consoleLog 

To replicate:

1.) Create new plug-in with RCP and e4 template
2.) Add parameters -nl de to the Program Arguments
3.) Start product -> wrong value in launch

![image](https://user-images.githubusercontent.com/139910/187464436-c55670df-9398-4594-8dde-75c4d98e0007.png)

@vik-chand IIRC you worked on the product to launch configuration a while ago? Can you have a look?

vogella:
Reproducer in https://github.com/vogella/rcp-translations, clone and import and start via com.vogella.translation.product

HannesWell:
This is a regression from https://github.com/eclipse-pde/eclipse.pde/pull/288.

I just created a fix: https://github.com/eclipse-pde/eclipse.pde/pull/317

> If I add -nl de to the product, the -nl is not added to the launch configuration.
> 
> It becomes: -os target.os−ws{target.ws} -arch target.arch−nl{target.nl} -consoleLog de
> 
> Instead of
> 
> -os target.os−ws{target.ws} -arch ${target.arch} -nl de -consoleLog
> 

With the mentioned fix it is actually `-os ${target.os} -ws ${target.ws} -arch ${target.arch} -consoleLog -nl de`, but that should have the same effect.

vogella:
Verfied in Eclipse SDK
Version: 2022-09 (4.25)
Build id: I20220831-1800
OS: Linux, v.5.15.0-46-generic, x86_64 / gtk 3.24.33, WebKit 2.36.6
Java vendor: Eclipse Adoptium
Java runtime version: 17.0.3+7
Java version: 17.0.3

HannesWell:
> Verfied in Eclipse SDK Version: 2022-09 (4.25) Build id: I20220831-1800

Great! And thank you for the report.

PR #317, opened 2022-08-30
Fix regression in LaunchAction.concatArgs

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/316

This is a regression from https://github.com/eclipse-pde/eclipse.pde/pull/288.


HannesWell:
@vogella, @vik-chand I think we should merge this ASAP. Since this is a clear bug from me and the fix is trivial.

We need some approvals so late in the release cycle.

akurtakov:
Could you add some test case for that?

HannesWell:
In general yes, but I won't have time for that today.
So this should be merged without an extra Test, to have it available in 2022-09.

akurtakov:
PMC approved.

vogella:
Tested and fixes the reported issue.

Thanks @HannesWell for the fast fix.

vogella:
@akurtakov I think we still have one build today for RC2, if not please trigger a rebuild.

akurtakov:
There should be 2 more builds. Next one in 3h.

Issue #321, opened 2022-09-02
NPE in ClasspathHelper.getDevPaths()


jukzi:
randomly happened : 
eclipse.buildId=4.23.0.I20220308-0310
```
java.lang.NullPointerException
    at org.eclipse.pde.internal.core.ClasspathHelper.getDevPaths(ClasspathHelper.java:284)
    at org.eclipse.pde.internal.core.ClasspathHelper.getDevEntriesProperties(ClasspathHelper.java:117)
    at org.eclipse.pde.internal.core.ClasspathHelper.getDevEntriesProperties(ClasspathHelper.java:83)
    at org.eclipse.pde.launching.EclipseApplicationLaunchConfiguration.getProgramArguments(EclipseApplicationLaunchConfiguration.java:98)
    at org.eclipse.pde.launching.AbstractPDELaunchConfiguration.launch(AbstractPDELaunchConfiguration.java:127)
    at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:803)
    at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:716)
    at org.eclipse.debug.internal.ui.DebugUIPlugin.buildAndLaunch(DebugUIPlugin.java:1021)
    at org.eclipse.debug.internal.ui.DebugUIPlugin$2.run(DebugUIPlugin.java:1224)
    at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
after restarting eclipse the error was gone.

was already reported in https://bugs.eclipse.org/bugs/show_bug.cgi?id=437727

=> model.getPluginBase().getLibraries() returned a null element in the array

It's unclear how that can happen but it might help to make eager null checks at add(IPluginLibrary library) methods before adding to the list - to get the root reason somewhen.

jukzi:
The error could also happen if two launch configurations started at the same time.
org.eclipse.pde.internal.core.bundle.BundlePluginBase.getLibraries() is called in parallel, initializing BundlePluginBase.libraries with a non thread safe ArrayList and then adding elements. two add in parallel to an ArrayList can lead to an null element.

jukzi:
not planned to fix

Issue #300, opened 2022-08-18
Intermittent failures in WorkspaceModelManagerTest


vik-chand:
testBundleRootHandling_bundleRootChangedFromNoneToDefault
testBundleRootHandling_bundleRootChangedFromNoneToOther
testProducts_productAddedToDeriveFolderNestedInSrcFolder
- testProducts_productAddedAndRemoved - ( sometimes this fails)

These test fail only in one of the platforms and is intermittent

vik-chand:
testProducts_productAddedToDerivedSourceFolder fails on RC1 build on 1 platform

vik-chand:
testProducts_productAddedAndRemoved fails in RC2 build  ( on Mac) but the rest of the test pass

vik-chand:
testBundleRootHandling_bundleRootChangedFromNoneToDefault

testProducts_productAddedToDeriveFolderNestedInSrcFolder

failed on

I20221109-1850

Issue #34, opened 2022-04-19
Split API from implementation


laeubi:
currently there is a bundle `org.eclipse.pde.core` that contains API and implementation and even ant-task stuff.

That makes it quite hard to sue PDE api in other context without pulling in much dependencies. It would be good to have a new `org.eclipse.pde` containing all public API. bundles that previously contained that API could require bundle the new bundle and use the [visibility = reexport](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.requirebundle).

That way bundles currently requiring such a now splitted bundle will not need to change, bundles using import-package won't notice this anyways.

vogella:
+1

PR #322, opened 2022-09-02
eager detect NPE in ClasspathHelper.getDevPaths() #321

jukzi:


akurtakov:
Why have you made all these methods sycnronized?

jukzi:
> Why have you made all these methods sycnronized?

they all reference member "libraries" in different threads which results in unspecified behaviour like NPE

akurtakov:
Would we possibly benefit from using synchronizedList

jukzi:
> Would we possibly benefit from using synchronizedList?

that would not synchronize the ==null checks

HannesWell:
In the past I noted that many parts of PDE's internal model is not properly prepared for concurrent access although it is modified concurrently and from your recent PR's you obviously observed that as well. 
I also highly suspect that insufficient synchronization is the cause for intermittent test failures as reported in https://github.com/eclipse-pde/eclipse.pde/issues/300.

Therefore I wonder if PDE's internal model should be reworked in general to become threadsafe. One approach could be to make all model objects immutable and recreate them on each change. If the Eclipse Platform requires Java-17 in the not too far future records could be a good fit then.

jukzi:
> Therefore I wonder if PDE's internal model should be reworked in general to become threadsafe. One approach could be to make all model objects immutable and recreate them on each change. If the Eclipse Platform requires Java-17 in the not too far future records could be a good fit then.

That would be very welcome. or just add a lot of synchronize.

laeubi:
Please gracefully consider then also this:

- https://github.com/eclipse-pde/eclipse.pde/issues/34

the problem is that records are not binary compatible as far as I know, but maybe we just can have PDE 4.0 some time? :-)

jukzi:

akurtakov:
> > Therefore I wonder if PDE's internal model should be reworked in general to become threadsafe. One approach could be to make all model objects immutable and recreate them on each change. If the Eclipse Platform requires Java-17 in the not too far future records could be a good fit then.
> 
> That would be very welcome. or just add a lot of synchronize.

I would say it is fine to go Java 17 already so feel free to propose a patch using records .

vogella:
Platform requires Java 17 now, so you @jukzi could start using records if you desire. @HannesWell started this for PDE in https://github.com/eclipse-pde/eclipse.pde/pull/518

vogella:
@jukzi is this PR still relevant? Or is with Java 17 a better way desired / possible?

jukzi:
> @jukzi is this PR still relevant? Or is with Java 17 a better way desired / possible?

still relevant. i don't have the time for a complete rewrite and it does not look like anybody else is working on that.

jukzi:
closing as i don't find the time to resolve conflicts, test or major refactoring

PR #330, opened 2022-09-06
Bug 579310 - Remove broken "Enable/Disable" action from Plug-in Registry

HannesWell:
As described in [Bug 579310](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579310) the `Enable/Disable` action of the `Plug-in Registry` is broken because the compatibility implementation of the `org.eclipse.osgi.service.resolver.PlatformAdmin` does not support addition/removal of `DisableInfo`.

This is broken for a while already and since nobody has complained I suggest to simply remove those actions and the whole disable/enable story.

grafik

If the others I agree I have to check that I didn't miss to remove anything related and didn't remove too much.


laeubi:
I can't think where this could be useful and OSGi has better ways to ensure the right provider is chosen already. :+1:

vogella:
+1 for removal

PR #333, opened 2022-09-08
PDEPluginImages: fix NullPointerException #297

jukzi:
Images need to be initialized with a Display when used outside swt thread. (ManifestContentAssistProcessor runs in background thread and uses images)


jukzi:

akurtakov:
Thanks!

PR #340, opened 2022-09-21
Databinding update

vogella:


vogella:
@jensli  FYI, I "solved" it super quick and dirty to make the build work again. If you want, please provide a smarter solution

jensli:
@vogella We should have searched a bit for uses of the deleted classes... I'll do that tomorrow.

PR #353, opened 2022-10-02
Use ASM 9.4

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/606


akurtakov:
@iloveeclipse any reason not to merge this one now?

iloveeclipse:
> @iloveeclipse any reason not to merge this one now?

Yep, I wait for https://github.com/jacoco/jacoco/pull/1368 and ASM available on Orbit, because we don't ship all of 4 ASM jars in platform.
Otherwise bundles that need **all** ASM packages will either not work anymore or do not install etc.

iloveeclipse:
I mean this one https://git.eclipse.org/r/c/orbit/orbit-recipes/+/196151 and https://github.com/eclipse/xtext/issues/2104

akurtakov:
So what happens is Platform ships with 9.4 but doesn't strictly require it until others catch up?

iloveeclipse:
> So what happens is Platform ships with 9.4 but doesn't strictly require it until others catch up?

Currently we **ship already** but we don't **require** 9.4, so what I see in my SDK setup is that after I install my Bytecode Outline thing that **requires** all of ASM, the 9.3 versions from Orbit are "pulled in" and platform happily uses them.
With PDE **requiring** 9.4 ASM we will have a conflict and I will not be able to install my bundle. 
Also even if that would work, for Xtext requiring **some more but not all** ASM packages we will install few ASM 9.4 bundles additionally and in worst case (we had that in the product) the client that uses **both** PDE and Xtext in same class will have obscure runtime errors saying some ASM class != same ASM class because they were loaded from different jars by different code.

An "easy" solution would be to ship all 4 ASM bundles (they are tiny) in platform.

laeubi:
> Also even if that would work, for Xtext requiring **some more but not all** ASM packages we will install few ASM 9.4 bundles additionally and in worst case (we had that in the product) the client that uses **both** PDE and Xtext in same class will have obscure runtime errors saying some ASM class != same ASM class because they were loaded from different jars by different code.
> 
> An "easy" solution would be to ship all 4 ASM bundles (they are tiny) in platform.

As mentioned before, consumers that expose ASM through their API should use import-package with proper version ranges and export package with proper use-clauses, then OSGi will ensure consistent class-space across multiple consumer/provider.

iloveeclipse:
@HannesWell : I would appreciate if we wait with merge till Orbit bundles are available :
https://git.eclipse.org/r/c/orbit/orbit-recipes/+/196151

HannesWell:
> @HannesWell : I would appreciate if we wait with merge till Orbit bundles are available : https://git.eclipse.org/r/c/orbit/orbit-recipes/+/196151

Yes, as you prefer. :)
Actually I just rebased this to see if the 611 API-tools test failures still occur.

PR #354, opened 2022-10-05
611 failures

mickaelistria:


merks:
@iloveeclipse  @HannesWell 

I was revisiting the build failures with an eye to the API builder running a background job during tests.  Here's what I see in the debugger:

image

I've set a breakpoint in org.eclipse.pde.api.tools.builder.tests.ApiBuilderTest.beforeClass() and set a conditional breakpoint in org.eclipse.pde.api.tools.internal.builder.ApiAnalysisBuilder.isRunningAsJob() stopped if it returns true.  Here I have hit that conditional breakpoint but have not hit the beforeClass's breakpoint.  That class is in the type hierarchy:

image

That method is sometimes called much later during other tests but I don't understand why it's sometimes called and not other times.  In any case, it doesn't seem that we are reliably disabling the background builds.

If I hack org.eclipse.pde.api.tools.internal.builder.ApiAnalysisBuilder.isRunningAsJob() to return false always, the tests that are were failing then pass...

I think we need a better more reliable way to ensure the background job behavior is disabled...

laeubi:
It is just an idea, but probably one should not read the preferences always (what likley never change often in a real IDE anyways), but using a plain boolean field that can be set on the class?

One can then in the activator read the value from the prefs, set it in the class and update it on pref changes, if then the test sets the field also, one should be sure it is always in the right state (unless there is some code running that changes the preference value).

laeubi:
Also note that this particular testcase extends [BuilderTests](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests.java) what is a **Junit 3 Testcase** so annotations won't work here.

merks:
> Also note that this particular testcase extends [BuilderTests](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests.java) what is a **Junit 3 Testcase** so annotations won't work here. Instead of `@BeforeClass` one needs to override `protected void setUp()` (and call super method(!)).

Very insightful!

akurtakov:
This went nowhere and there are other efforts to actually fix the issues.

Issue #356, opened 2022-10-08
Error while resolving class file for ... via JavaSE-19


jukzi:
with JRE 19 i get errors in an oomphed workspace
```
eclipse.buildId=4.26.0.I20221007-1800
java.version=19
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE

org.eclipse.pde.api.tools
Error while resolving class file for: javax.security.auth.spi.LoginModule via JavaSE-19

org.eclipse.core.runtime.CoreException: Failed to open archive: C:\Users\JKubitz\eclipseDev\platform2022-10-08\eclipse\plugins\org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_19.0.0.v20221005-1658\jre\jmods\java.base.jmod
	at org.eclipse.pde.api.tools.internal.model.ApiElement.abort(ApiElement.java:104)
	at org.eclipse.pde.api.tools.internal.model.ArchiveApiTypeContainer.init(ArchiveApiTypeContainer.java:286)
	at org.eclipse.pde.api.tools.internal.model.ArchiveApiTypeContainer.findTypeRoot(ArchiveApiTypeContainer.java:228)
	at org.eclipse.pde.api.tools.internal.model.AbstractApiTypeContainer.findTypeRoot(AbstractApiTypeContainer.java:110)
	at org.eclipse.pde.api.tools.internal.util.Util.getClassFile(Util.java:599)
	at org.eclipse.pde.api.tools.internal.model.ApiType.resolveSuperType(ApiType.java:299)
	at org.eclipse.pde.api.tools.internal.model.ApiType.getSuperInterfaces(ApiType.java:258)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.collectAllInterfaces(ClassFileComparator.java:921)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.getInterfacesSet(ClassFileComparator.java:2073)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.checkSuperInterfaces(ClassFileComparator.java:299)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.getDelta(ClassFileComparator.java:1211)
	at org.eclipse.pde.api.tools.internal.provisional.comparator.ApiComparator$2.visit(ApiComparator.java:771)
[...]

PR #357, opened 2022-10-11
ArchiveApiTypeContainer should properly load classes from modules

merks:
Instead of assuming that modules are represented as *.jmod, which is only the case for a JDK but not for a JRE, the jrt file system should be used.  In addition, jar files should  use the jar file system, to uniformly handle both cases.  The JRT approach also allows loading all the modules of the JDK/JRE not just the java.base module.

https://github.com/eclipse-pde/eclipse.pde/issues/356

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
@@ -154,8 +154,8 @@ public class ArchiveApiTypeContainer extends ApiElement implements IApiTypeConta
 		Path path = Path.of(fLocation);
 		if (fLocation.endsWith("jrt-fs.jar")) {
 			AtomicReference<IOException> exception = new AtomicReference<>();
-			FileSystem jrtFileSystem = JRTS.computeIfAbsent(path, it -> {
-				Path jrePath = path.getParent().getParent();
+			FileSystem jrtFileSystem = JRTS.computeIfAbsent(path.toRealPath(), it -> {
+				Path jrePath = it.getParent().getParent();
 				try {
 					return FileSystems.newFileSystem(URI.create("jrt:/"), Map.of("java.home", jrePath.toString()));
 				} catch (IOException e) {
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
@@ -14,27 +14,29 @@
 package org.eclipse.pde.api.tools.internal.model;
 
 import java.io.IOException;
-import java.io.InputStream;
+import java.net.URI;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipFile;
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.IPath;
-import org.eclipse.core.runtime.Path;
-import org.eclipse.pde.api.tools.internal.provisional.ApiPlugin;
+import org.eclipse.jdt.internal.compiler.util.JRTUtil;
 import org.eclipse.pde.api.tools.internal.provisional.model.ApiTypeContainerVisitor;
 import org.eclipse.pde.api.tools.internal.provisional.model.IApiElement;
 import org.eclipse.pde.api.tools.internal.provisional.model.IApiTypeContainer;
 import org.eclipse.pde.api.tools.internal.provisional.model.IApiTypeRoot;
+import org.eclipse.pde.api.tools.internal.util.Signatures;
 import org.eclipse.pde.api.tools.internal.util.Util;
 
 /**
@@ -43,6 +45,13 @@ import org.eclipse.pde.api.tools.internal.util.Util;
  * @since 1.0.0
  */
 public class ArchiveApiTypeContainer extends ApiElement implements IApiTypeContainer {
+
+	/**
+	 * Keep a map of the JRT file system.
+	 *
+	 * @see #getLocation()
+	 */
+	private static final Map<String, FileSystem> JRTS = new ConcurrentHashMap<>();
 
 	/**
 	 * {@link IApiTypeRoot} implementation within an archive
@@ -58,15 +67,13 @@ public class ArchiveApiTypeContainer extends ApiElement implements IApiTypeConta
 		 * @param container archive
 		 * @param entryName zip entry name
 		 */
-		public ArchiveApiTypeRoot(ArchiveApiTypeContainer container, String entryName) {
+		public ArchiveApiTypeRoot(ArchiveApiTypeContainer container, String typeName, String entryName) {
 			super(container, entryName);
+			this.fTypeName = typeName;
 		}
 
 		@Override
 		public String getTypeName() {
-			if (fTypeName == null) {
-				fTypeName = getName().replace('/', '.').substring(0, getName().length() - Util.DOT_CLASS_SUFFIX.length());
-			}
 			return fTypeName;
 		}
 
@@ -95,46 +102,14 @@ public class ArchiveApiTypeContainer extends ApiElement implements IApiTypeConta
 				return fContents;
 			}
 			ArchiveApiTypeContainer archive = (ArchiveApiTypeContainer) getParent();
-			ZipFile zipFile;
 			try {
-				zipFile = new ZipFile(archive.fLocation);
+				Path location = archive.getLocation();
+				Path classLocation = location.resolve(getName());
+				fContents = Files.readAllBytes(classLocation);
 			} catch (IOException e) {
-				abort("Failed to open archive: " + archive.fLocation, e); //$NON-NLS-1$
-				return null;
+				abort("Failed to open class file: " + getTypeName() + " in archive: " + archive.fLocation, e); //$NON-NLS-1$ //$NON-NLS-2$
 			}
-			try {
-				ZipEntry entry = zipFile.getEntry(getName());
-				InputStream stream = null;
-				if (entry != null) {
-					try {
-						stream = zipFile.getInputStream(entry);
-					} catch (IOException e) {
-						abort("Failed to open class file: " + getTypeName() + " in archive: " + archive.fLocation, e); //$NON-NLS-1$ //$NON-NLS-2$
-						return null;
-					}
-					try {
-						fContents = stream.readAllBytes();
-						return fContents;
-					} catch (IOException ioe) {
-						abort("Unable to read class file: " + getTypeName(), ioe); //$NON-NLS-1$
-						return null;
-					} finally {
-						try {
-							stream.close();
-						} catch (IOException e) {
-							ApiPlugin.log(e);
-						}
-					}
-				}
-			} finally {
-				try {
-					zipFile.close();
-				} catch (IOException e) {
-					abort("Failed to close class file archive", e); //$NON-NLS-1$
-				}
-			}
-			abort("Class file not found: " + getTypeName() + " in archive: " + archive.fLocation, null); //$NON-NLS-1$ //$NON-NLS-2$
-			return null;
+			return fContents;
 		}
 
 		@Override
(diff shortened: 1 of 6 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
```diff
@@ -149,31 +115,46 @@ public String toString() {
...
+	 * Converts the location to a path in the applicable file system.
+	 *
+	 * @return the path corresponding to the location.
+	 * @throws IOException
+	 */
+	@SuppressWarnings("nls")
+	private Path getLocation() throws IOException {
+		if (fLocation.endsWith("jrt-fs.jar")) {
+			Path jrePath = Path.of(fLocation).getParent().getParent();
+			try (URLClassLoader loader = new URLClassLoader(new URL[] { jrePath.toUri().toURL() })) {
```
iloveeclipse, line 150: No idea what is the life cycle of this container is, and how often getLocation() is called, but isn't this change leaking file system/classloader every time (classloader will stay there as long as loaded classes are not GC'd)? Shouldn't the result be cached?

I assume you use new classloader to avoid "FS already created" error? But it would be OK to keep the FS open and re-use existing one if there is one already.

merks: I see other usage patterns similar to this, but it's better safe than sorry and create only one instance per JRE location.  That will perform better as well in any case.  I'll update the PR accordingly.  I need to fix versions anyway...  Thanks for the early feedback.  😄

merks: Note that the class loader itself is closed.  I actually wonder if I shouldn't be passing the system class loaded... 

Let's see how the tests go and I'll review that consideration when I'm back.

merks: The latest force push doesn't use a class loader at all, i.e., like how org.eclipse.jdt.internal.compiler.util.JrtFileSystem.JrtFileSystem(Jdk, String) does it.

merks:
@iloveeclipse Any more concerns? Does anyone else need to review this?

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
```diff
@@ -149,31 +122,54 @@ public String toString() {
...
+	private Path getLocation() throws IOException {
+		if (fLocation.endsWith("jrt-fs.jar")) {
+			AtomicReference<IOException> exception = new AtomicReference<>();
+			FileSystem jrtFileSystem = JRTS.computeIfAbsent(fLocation, it -> {
+				Path jrePath = Path.of(fLocation).getParent().getParent();
+				try {
+					return FileSystems.newFileSystem(URI.create("jrt:/"),
+							Collections.singletonMap("java.home", jrePath.toString()));
+				} catch (IOException e) {
+					exception.set(e);
```
iloveeclipse, line 163: If the file system already exists, you can get it in the catch block.

laeubi: Given that it is quite hard to get it right with all the different cases can't PDE reuse the code from JDT here?

merks: > If the file system already exists, you can get it in the catch block.

Each call creates a new instance.   Really, I've debugged into this, including looking at how many instances of jdt.internal.jrtfs.JrtFileSystem are created (and there are a significant number, but from other locations).  

It did also did occur to me to look at how JDT does this, which I mentioned here:

https://github.com/eclipse-pde/eclipse.pde/pull/357#discussion_r992300456

You can look at the code here:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/9562c376de84b672f4925bf2f49c10b842073f0f/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/util/JRTUtil.java#L502-L512

I did also test locally with a JRE and and JDK and the tests did all pass.

Even JDT seems to do this in multiple places, e.g., also in org.eclipse.jdt.internal.compiler.tool.JrtFileSystem.initialize().  But none of it very reusable because no one wants to over-commit on APIs...

So I believe there are no outstanding issues. 

Certainly if someone likes they may change JDT to provide something reusable and then make downstream changes at a later point in time.

iloveeclipse: The point is not that **this** code can create new file system, but **other** code (JDT) too. So the pattern is to "try create new / catch already there? just get it"

merks: No, please look at the code.
```
	JrtFileSystem(Jdk jdkHome, String release) throws IOException {
		this.jdk = jdkHome;
		this.release = release;
		JRTUtil.MODULE_TO_LOAD = System.getProperty("modules.to.load"); //$NON-NLS-1$
		HashMap<String, String> env = new HashMap<>();
		env.put("java.home", this.jdk.path); //$NON-NLS-1$
		this.fs = FileSystems.newFileSystem(JRTUtil.JRT_URI, env);
		this.modRoot = this.fs.getPath(JRTUtil.MODULES_SUBDIR);
		// Set up the root directory where modules are located
		walkJrtForModules();
	}
```

Or at this org.eclipse.jdt.internal.compiler.tool.JrtFileSystem.initialize() code:
```
	public void initialize() throws IOException {
		// initialize packages
		this.modulePathMap = new HashMap<>();
		URL jrtPath = null;

		if (this.file.exists()) {
			jrtPath = Paths.get(this.file.toPath().toString(), "lib", JRTUtil.JRT_FS_JAR).toUri().toURL(); //$NON-NLS-1$
			try (URLClassLoader loader = new URLClassLoader(new URL[] { jrtPath })) {
				HashMap<String, ?> env = new HashMap<>();
				this.jrtfs = FileSystems.newFileSystem(JRT_URI, env, loader);
				this.modules = this.jrtfs.getPath("/modules"); //$NON-NLS-1$
			}
		} else {
			return;
		}

		org.eclipse.jdt.internal.compiler.util.JRTUtil.walkModuleImage(this.file,
				new org.eclipse.jdt.internal.compiler.util.JRTUtil.JrtFileVisitor<Path>() {
			@Override
			public FileVisitResult visitModule(Path path, String name) throws IOException {
				JrtFileSystem.this.modulePathMap.put(name, path);
				return FileVisitResult.CONTINUE;
			}
		}, JRTUtil.NOTIFY_MODULES);
	}
```
I suspect you are confusing those with this pattern used for the jar/zip file system:
```
	public static FileSystem getJarFileSystem(Path path) throws IOException {
		URI uri = URI.create("jar:file:" + path.toUri().getRawPath()); //$NON-NLS-1$
		try {
			try {
				return FileSystems.getFileSystem(uri);
			} catch (FileSystemNotFoundException fne) {
				try {
					return FileSystems.newFileSystem(uri, Map.of(), ClassLoader.getSystemClassLoader());
				} catch (FileSystemAlreadyExistsException e) {
					return FileSystems.getFileSystem(uri);
				}
			}
		} catch (ProviderNotFoundException e) {
			throw new IOException("No provider for uri " + uri, e); //$NON-NLS-1$
		}
	}
```
But that's not what JDT does for the JRT file system in the **three** places (also org.eclipse.jdt.internal.compiler.apt.util.JrtFileSystem.initialize()) that it does this. Note too that unlike the zip/jar file system where there are unique URIs involved, all these places the URI  'jrt:/' and create different instances depending on the java.home value in the environment.  So they cannot use the FileSystems.getFileSystem() pattern...

So again, all these create their own new instance, none of that share that instance, and none of them catch the exception when creating it.  So I believe this is not an issue....

iloveeclipse: Well, there is lot of various uses of the API, and not all of them are correct or intended for general use in the IDE. JDT code in compiler.tool is from command line ecj compiler, which wasn't meant to be used in the multi-threaded environment of an IDE, where every one can create file system objects with any (also same) paths.

PDE code however **is** in the IDE, so it must be aware that there can be someone already created the same file system or creating it in parallel at same time.

I would recommend to look at `org.eclipse.jdt.internal.compiler.util.CtSym.init()`

```
try {
			fst = FileSystems.getFileSystem(uri);
		} catch (Exception fne) {
			// Ignore and move on
		}
		if (fst == null) {
			try {
				fst = FileSystems.newFileSystem(uri, new HashMap<>(), ClassLoader.getSystemClassLoader());
			} catch (FileSystemAlreadyExistsException e) {
				fst = FileSystems.getFileSystem(uri);
			} catch (ProviderNotFoundException e) {
				throw new IOException("Failed to create ct.sym file system for " + this.ctSymFile, e); //$NON-NLS-1$
			}
		}
```

iloveeclipse: Also please note the classloader given as third argument. It's important too.

merks: I'm not sure why you are pointing me to another another "jar:" example?  I already mentioned that **this is a different pattern** because each jar has a different URI, but the JRT file system **has only one** URI "jre:/"

If the class loader is important,  then surely org.eclipse.jdt.internal.compiler.util.JrtFileSystem.JrtFileSystem(Jdk, String) should use it to, but it doesn't.

The class loader is used only to load the provider in java.nio.file.FileSystems.newFileSystem(URI, Map<String, ?>, ClassLoader):
```
    public static FileSystem newFileSystem(URI uri, Map<String,?> env, ClassLoader loader)
        throws IOException
    {
        String scheme = uri.getScheme();

        // check installed providers
        for (FileSystemProvider provider : FileSystemProvider.installedProviders()) {
            if (scheme.equalsIgnoreCase(provider.getScheme())) {
                try {
                    return provider.newFileSystem(uri, env);
                } catch (UnsupportedOperationException uoe) {
                }
            }
        }

        // if not found, use service-provider loading facility
        if (loader != null) {
            ServiceLoader<FileSystemProvider> sl = ServiceLoader
                .load(FileSystemProvider.class, loader);
            for (FileSystemProvider provider : sl) {
                if (scheme.equalsIgnoreCase(provider.getScheme())) {
                    try {
                        return provider.newFileSystem(uri, env);
                    } catch (UnsupportedOperationException uoe) {
                    }
                }
            }
        }

        throw new ProviderNotFoundException("Provider \"" + scheme + "\" not found");
    }
```
Do we really want to load classes from the target JRE/JDK?  I'm a little doubtful of that...

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
```diff
@@ -149,31 +122,54 @@ public String toString() {
...
+	 * Converts the location to a path in the applicable file system.
+	 *
+	 * @return the path corresponding to the location.
+	 * @throws IOException
+	 */
+	@SuppressWarnings("nls")
+	private Path getLocation() throws IOException {
+		if (fLocation.endsWith("jrt-fs.jar")) {
+			AtomicReference<IOException> exception = new AtomicReference<>();
+			FileSystem jrtFileSystem = JRTS.computeIfAbsent(fLocation, it -> {
```
HannesWell, line 157: I wonder if not the String `fLocation` should be used as key here but the canonicalized/real path instead.
This should cover (the probably more edgy case) that the same target is referenced with different paths, e.g. a mixture of forward/backward slashes on Linux or Windows or the use of `.` or `..`.

If I traced the callers correctly the `ApiFileGenerationApplication` has user-defined arguments that end up here.

HannesWell (commented):
Looks good to me.
I just have remark and a little cosmetical suggestion.

And it looks like this also helps for https://github.com/eclipse-pde/eclipse.pde/issues/232.
At least all tests had passed. But maybe this was by chance.
But when I try to build with this change locally the tests still failed...

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
```diff
@@ -149,31 +122,54 @@ public String toString() {
...
+				Path jrePath = Path.of(fLocation).getParent().getParent();
+				try {
+					return FileSystems.newFileSystem(URI.create("jrt:/"),
+							Collections.singletonMap("java.home", jrePath.toString()));
```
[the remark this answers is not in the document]
merks: Yes, that's prettier.  😄

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
merks: Good idea.  Better to be conservative.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
merks: @iloveeclipse I assume that silence is implicit agreement...

iloveeclipse: > @iloveeclipse I assume that silence is implicit agreement...

I assume you have tested that & I assume the "jrt:/" must be a special case, the typical behavior I'm used to see is that JVM prevents multiple instances of file systems mapped to same url to be created. So feel free to forget my comments above :-)

HannesWell:
Holy moly, @merks it looks like you fixed the 611 failures of https://github.com/eclipse-pde/eclipse.pde/issues/232 this this. 😃

Issue #355, opened 2022-10-07
[windows] 7 errors and only 589 from 634 tests executed since I20221007-0230


iloveeclipse:
Windows build has 7 errors (I assume test classes didn't properly initialized) and only 589 from 634 tests executed in I20221007-0230.

https://download.eclipse.org/eclipse/downloads/drops4/I20221007-0230/testresults/html/org.eclipse.pde.ui.tests_ep426I-unit-win32-java11_win32.win32.x86_64_11.html

For example ClasspathResolutionTest is missing (haven't checked what else is lost).

I don't see any PDE changes, so it must be coming from platform/equinox or even tycho.

iloveeclipse:
Works again with https://download.eclipse.org/eclipse/downloads/drops4/I20221010-1840/testResults.php.
Whatever it was, it seem to be gone now.

iloveeclipse:
This **could** be related to the hang we had in `testCreationFeatureProject` which seem to open a blocking modal dialog and waits forever:

https://download.eclipse.org/eclipse/downloads/drops4/I20221011-0640/testResults.php 

```
!ENTRY org.eclipse.pde.ui.tests 1 0 2022-10-11 13:28:07.310
!MESSAGE [testCreationFeatureProject] setUp
Possible frozen test case
"main": RUNNABLE
    org.eclipse.swt.internal.win32.OS.WaitMessage(Native Method)
    org.eclipse.swt.widgets.Display.sleep(Display.java:4707)
    org.eclipse.jface.window.Window.runEventLoop(Window.java:824)
    org.eclipse.jface.window.Window.open(Window.java:799)
    org.eclipse.jface.dialogs.MessageDialog.open(MessageDialog.java:609)
    org.eclipse.jface.dialogs.MessageDialog.open(MessageDialog.java:405)
    org.eclipse.jface.dialogs.MessageDialog.openQuestion(MessageDialog.java:515)
    org.eclipse.pde.internal.ui.wizards.feature.AbstractCreateFeatureOperation.shouldOverwriteFeature(AbstractCreateFeatureOperation.java:206)
    org.eclipse.pde.internal.ui.wizards.feature.AbstractCreateFeatureOperation.createFeature(AbstractCreateFeatureOperation.java:69)
    org.eclipse.pde.internal.ui.wizards.feature.AbstractCreateFeatureOperation.execute(AbstractCreateFeatureOperation.java:58)
[...]

PR #358, opened 2022-10-12
Test fixes to uncover hidden test failures

iloveeclipse:
- don't swallow resource exceptions on teardown - they might hide fails in following tests
- don't open message or error dialogs to avoid test hangs

See https://github.com/eclipse-pde/eclipse.pde/issues/355


iloveeclipse:
Test fails are unrelated (#232 and #359).

PR #363, opened 2022-10-16
Also save HTML to Clipboard when we save RTF (#89)

mihnita:
TLDR:

A few weeks ago we added HTML support when copying to Clipboard from StyledText.
It used to be RTF only.

This PR changes code that explicitly use `TextTransfer` and `RTFTransfer` to also use `HTMLTransfer`. 

---

The original bug was filed here:
https://github.com/eclipse-platform/eclipse.platform.text/issues/101

Very likely that is the wrong place, but at the time I didn't know where the fix will end up.
Let me know if you want me to file an equivalent bug here (or close than one and open it here)

Thank you,
Mihai

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildEditor.java
@@ -147,7 +147,11 @@ public class BuildEditor extends MultiSourceEditor {
 	protected boolean hasKnownTypes() {
 		try {
 			TransferData[] types = getClipboard().getAvailableTypes();
-			Transfer[] transfers = new Transfer[] {TextTransfer.getInstance(), RTFTransfer.getInstance()};
+			Transfer[] transfers = new Transfer[] {
+					TextTransfer.getInstance(),
+					HTMLTransfer.getInstance(),
+					RTFTransfer.getInstance()
+			};
 			for (TransferData type : types) {
 				for (Transfer transfer : transfers) {
 					if (transfer.isSupportedType(type))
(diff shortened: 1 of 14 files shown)


mickaelistria:
Couldn't we avoid duplication here by directly querying possible transfers from StyledText?

mihnita:
What do you have in mind?

Add a method to `StyledText` that would do something like this?
```
Transfer[] getPossibleTransfers() {
	return new Transfer[] {
			RTFTransfer.getInstance()
			HTMLTransfer.getInstance(),
			TextTransfer.getInstance()
	};
}
```
That would have to be public to use it here. Is it common enough as a pattern to deserve being public?
(other than here) 
It would reduce the code a bit, but not much...
The double loop would still be repeated.

Or maybe add a `canPaste` method to `StyledText`?
With the same concerns: is it something worth adding "for the whole world to see"?
There are already public `copy()` / `cut()` / `paste()` method already, so maybe `boolean canPaste(Clipboard)` would make sense.

---

My initial thinking, when I've seen all this repeated code, was to add it as a public method to `PDESection`
It is the base of 10 out of 12 classes that I am touching here.
And if I make it `static` it can also be used from `MultiSourceEditor`

But checking, the method is already implemented as
```
public boolean canPaste(Clipboard clipboard) {
	return false;
}
```
So I thought there might be a good reason for things to be the way they are.
And I was afraid of side-effects, as there are 41 classes that extend `PDESection`.

mickaelistria:
> Add a method to StyledText that would do something like this?

Yes, that's what I have in mind.
The benefit is that if later we add some MarkdownTransfer or other kinds of Transfer to the StyledText, then those editors/viewers wouldn't have to be changed in order to enable it.

> Is it common enough as a pattern to deserve being public?

You tell me, you're the expert now ;)

> There are already public copy() / cut() / paste() method already, so maybe boolean canPaste(Clipboard) would make sense.

That sound indeed better.

About
```
public boolean canPaste(Clipboard clipboard) {
	return false;
}
```
I imagine this is desired for all the form-based tabs that do not build on top of a StyledText.

mihnita:
Added canPaste to `StyledText`, see https://github.com/eclipse-platform/eclipse.platform.swt/pull/437

PR #367, opened 2022-10-19
Fix: Issue 366 - NPE when exporting plug-ins via the export wizard

pzi-dsa:
The install location might be null for workspace resources. In this case, it is no longer possible to export plugins via the export wizard. Instead, those features are skipped with a warning.

Example: Synthetic features derived from Maven poms via m2e.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/exports/FeatureExportOperation.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/exports/FeatureExportOperation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2021 IBM Corporation and others.
+ * Copyright (c) 2006, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,7 @@
  *     Lars Vogel <Lars.Vogel@vogella.com> - Bug 477527
  *     Jan-Ove Weichel <janove.weichel@vogella.com> - Bug 477658
  *     Martin Karpisek <martin.karpisek@gmail.com> - Bug 525701
+ *     Patrick Ziegler - Issue 366 NPE when exporting plug-ins via the export wizard
  *******************************************************************************/
 package org.eclipse.pde.internal.core.exports;
 
@@ -856,7 +857,9 @@ public class FeatureExportOperation extends Job {
 		FeatureModelManager fmm = PDECore.getDefault().getFeatureModelManager();
 		IFeatureModel[] models = fmm.getExternalModels();
 		for (IFeatureModel model : models) {
-			map.put(model.getFeature().getId(), model.getInstallLocation());
+			if (model.getInstallLocation() != null) {
+				map.put(model.getFeature().getId(), model.getInstallLocation());
+			}
 		}
 		// remove anything that we have in the workspace models
 		models = fmm.getWorkspaceModels();


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/exports/FeatureExportOperation.java
```diff
@@ -856,7 +857,11 @@ protected String[] getPaths() {
 		FeatureModelManager fmm = PDECore.getDefault().getFeatureModelManager();
 		IFeatureModel[] models = fmm.getExternalModels();
 		for (IFeatureModel model : models) {
-			map.put(model.getFeature().getId(), model.getInstallLocation());
+			if (model.getInstallLocation() != null) {
+				map.put(model.getFeature().getId(), model.getInstallLocation());
+			} else {
+				PDECore.log(Status.warning(NLS.bind(PDECoreMessages.FeatureExportOperation_featureWithoutInstallLocationFound, model.getFeature().getId())));
```
mickaelistria, line 863: At this stage, such error message is important for user, it's not a technical message that should be in log, but something the user should be warned about.
The unsupported features should be listed and then shown in a popup after export.
Or, even better, this should happen preventively in the UI by not allowing to select those features and instead mark them with the reason why they can't be selected.

mickaelistria (changes requested):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/exports/FeatureExportOperation.java
pzi-dsa: > Or, even better, this should happen preventively in the UI by not allowing to select those features and instead mark them with the reason why they can't be selected.

I don't think that's really the problem, given that both the plug-in and feature exporter only allow you to select projects in your workspace. So this case shouldn't be possible.

To be honest, I'm not even sure why it is then necessary to load those external features. 

And if you were to export something containing this synthetic feature, you end up with an error message, regardless. So an additional popup seems redundant.
image

pzi-dsa: It looks like the change was made as part of [Bug 309884](https://bugs.eclipse.org/bugs/show_bug.cgi?id=309884). It makes sense that this feature would be required when exporting e.g. a product or the feature I've shown above. In which case it's the responsiblity of the external provider to set the install location, if they want it to work with the exporter.

The reason it shows also up for plug-ins is simply because all of those exporters use the same mechanism for creating the p2 structure and therefore it is just an unlucky side-effect. 

But if that's the case, then I agree that it doesn't make sense to log the error message. Instead, I would silently ignore those artifacts and instead rely on the pop-up that shows up when Eclipse tries to load the missing element.

mickaelistria: OK, it looks like not logging is fine. The rule is
* user needs to worry -> a popup
* some curiours developer or a contributor needs to worry -> a log message
* no-one needs to worry -> nothing.

Looks like there is an agreement it's the later case.

mickaelistria (approved):

pzi-dsa:
Is there anything holding this pull request back?

> org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testBundleRootHandling_bundleRootChangedFromNoneToDefault
> java.lang.AssertionError: expected null, but was:<plugin.a>

I've retested this test case and it works locally. It also doesn't seem to call the method I've modified, so I assume it was just a fluke?

HannesWell:
> Is there anything holding this pull request back?

Nope, there is nothing. Thanks for the heads up.
Since nobody raised concerns I'm going to merge this.

Thank you for your contribution!

> 
> > org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testBundleRootHandling_bundleRootChangedFromNoneToDefault
> > java.lang.AssertionError: expected null, but was:<plugin.a>
> 
> I've retested this test case and it works locally. It also doesn't seem to call the method I've modified, so I assume it was just a fluke?

This test unfortunately fails occasionally on the CI. So no need to worry in this case.

PR #374, opened 2022-11-03
Only add the interface to see if that could work

laeubi:
To better found out whats going on in:

- https://github.com/eclipse-pde/eclipse.pde/pull/214

### DIFF
--- a/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.15.400.qualifier
+Bundle-Version: 3.16.0.qualifier
 Bundle-Activator: org.eclipse.pde.internal.core.PDECore
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.core/plugin.xml
+++ b/ui/org.eclipse.pde.core/plugin.xml
@@ -402,7 +402,7 @@
     </extension>
     <extension
           point="org.eclipse.pde.core.source">
-          <locator class="org.eclipse.pde.internal.core.ColocationPluginSourcePathLocator" />
-          <locator class="org.eclipse.pde.internal.core.LocalMavenPluginSourcePathLocator" />
+          <locator class="org.eclipse.pde.internal.core.sourcelocators.ColocationPluginSourcePathLocator" />
+          <locator class="org.eclipse.pde.internal.core.sourcelocators.LocalMavenPluginSourcePathLocator" />
     </extension> 
 </plugin>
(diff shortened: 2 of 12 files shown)


laeubi:
First complain (as expected) is now

> The minor version should be incremented in version 3.15.400, since new APIs have been added since version 3.15.300 - The type org.eclipse.pde.core.IPluginSourcePathLocator has been added to org.eclipse.pde.core_3.15.400

so my second commit does that ...

laeubi:
Test failure is likely unrelated created https://github.com/eclipse-pde/eclipse.pde/issues/375 but API tools seem happy, so adding more code now...

HannesWell:
> Test failure is likely unrelated created #375 but API tools seem happy, so adding more code now...

It is. That test is unfortunately prone to race-conditions...

laeubi:
Some test failures but it seems no API errors...

laeubi:
And now we have the https://github.com/eclipse-pde/eclipse.pde/actions/runs/3385917236

> The re-exported type org.eclipse.pde.core.IBaseModel has been removed from org.eclipse.pde.ui_3.13.600 

but only packages marked as "internal" are modified with the last commit, and until that point API tools where happy with introducing new API...

laeubi:
It seems that just adding a new class in the internal package of pde.core triggers the api error...

HannesWell:
> It seems that just adding a new class in the internal package of pde.core triggers the api error...

Do the API tools consider that the package is marked as internal via x-friend?

Another thing that I assumed could be causing problem is the modification of the Extension-Point schema.
But reverting this locally did not make a change for me.

But in general this is all really strange...

laeubi:
> Another thing that I assumed could be causing problem is the modification of the Extension-Point schema.

As you can see in this PR all previous changes are not a problem, it is only when adding a class to the internal package

laeubi:
If I understand correctly then this code produces the message:
```

if ((visibility & visibilityModifiers) == 0) {
if ((refVisibility & visibilityModifiers) == 0) {
	// no delta
	return NO_DELTA;
}
if (Util.isAPI(refVisibility, typeDescriptor)) {
	return new Delta(deltaComponentID, IDelta.API_COMPONENT_ELEMENT_TYPE, IDelta.REMOVED, IDelta.API_TYPE, elementDescription2 != null ? elementDescription2.getRestrictions() : RestrictionModifiers.NO_RESTRICTIONS, RestrictionModifiers.NO_RESTRICTIONS, typeDescriptor.getModifiers(), typeDescriptor2.getModifiers(), typeName, typeName, new String[] {
			typeName, Util.getComponentVersionsId(component2) });
}
}
```
So it is concerned about changed visbility but I don't get what this has to do with the change...

laeubi:
Tried jsut adding an interface, raise version and implement it ... that works, so should also not be the problem here... https://github.com/eclipse-pde/eclipse.pde/pull/377

laeubi:
> Another thing that I assumed could be causing problem is the modification of the Extension-Point schema.
> But reverting this locally did not make a change for me.

It seems it is the issue here, I have now here reverted the schema and it seems API tools is happy (for now ...) stupid enough **changing** the extentsionpoint but never implement it works ... API tools only start to complain once you implement it.

HannesWell:
> > Another thing that I assumed could be causing problem is the modification of the Extension-Point schema.
> > But reverting this locally did not make a change for me.
> 
> It seems it is the issue here, I have now here reverted the schema and it seems API tools is happy (for now ...) stupid enough **changing** the extentsionpoint but never implement it works ... API tools only start to complain once you implement it.

Really strange. That really should be fixed.
But since we have now https://github.com/eclipse-pde/eclipse.pde/pull/379 working with a new Extension I think this can be closed.

PR #376, opened 2022-11-03
Add button to clean bundle pool, rewording of Reload button

kjsmita6:
- Add a "Clean" button to the target definition "Locations" section to remove any artifacts from the bundle pool which are no longer part of the target, such as if something is removed or a new version is downloaded.
- Change the "Reload" button to "Refresh" as there is another button "Reload target platform..." that does a different thing. In fact, if you do "Reload target platform..." without the target being resolved the workspace won't have a valid target platform and you will get many errors (ours go around 100k in this case). I've had to explain instructions to people like "First, press this Reload button, then once that's done reloading you press the OTHER reload button" and it's very confusing.

These changes appear in both the target definition file editor and the target platform preference page when you Edit a target.


vogella:
@HannesWell can you review?

laeubi:
@kjsmita6 thanks for your contribution, if you change UI you might want to upload some image to Github that shows how everything looks like, just copy an image (either file or clipboard) and use CTRL+V to add it to the description text.

> Add a "Clean" button to the target definition "Locations" section to remove any artifacts from the bundle pool which are no longer part of the target, such as if something is removed or a new version is downloaded.

I don't think that the target editor is a suitable place for such function:

- Actually this should be something either very seldom be used if at all by a user
- "Clean" is very unclear, why /what should I clean? Why do PDE do not clean up old stuff automatically? What are the impacts of a clean?
- It is bound to a very specific location type (the IU Location / Updatesites)

Therefore I think such function is better placed at the Preferences:
grafik


> Change the "Reload" button to "Refresh" as there is another button "Reload target platform..." that does a different thing. In fact,

I would prefer to better change the latter text, the button do not really "Refresh" anything it effectively throws away the old resolved state and starts resolving again. On the other hand, even as a long term user of PDE I have no clue what "Reload target platform..."  should actually mean, if the target platform is not current it says "Set as active target platfrom""

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
```diff
@@ -563,6 +578,35 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
...
+				}
+			} catch (CoreException e) {
+
+			}
+		}
+	}
+
+	private Job getOrCreateCleanJob() {
+		if (cleanJob == null) {
+			cleanJob = new Job(PDEUIMessages.TargetPlatformPreferencePage2_26) {
```
HannesWell, line 597: Using the `Job.create()` factory should make the code cleaner.

Furthermore you probably want to hold the `workspace` rule to prevent multiple of those jobs from running at the same time.
I assume you tried to prevent that by saving the job as field, don't you?
That is probably not the savest way. It is probably possible to create multiple instances of the surrounding class if one quickly closes and reopens the editor. Then the job of the previous editor instance would still be running.
You could also add a job family and search for existing jobs.

Furthermore just rescheduling the job on each button click is IMHO not ideal. We could consider to disable the button when the job starts and re-enable once the GC has completed.

HannesWell (commented):
Thank you for your contribution.
> 
>     * "Clean" is very unclear, why /what should I clean? Why do PDE do not clean up old stuff automatically? What are the impacts of a clean?
> 

`Clean` could indeed be confused with 'wipe-out'. What about `Collect Garbage` or `Perform GC` or `Run GC`?


>     * It is bound to a very specific location type (the IU Location / Updatesites)
> 
That's right.
But In the preferences it is hidden relatively far away (one won't stumble upon it) and the preferences for the available software sites also lists the software sites to update eclipse itself. Does it even consider the IU-locations from the targets?
However what about placing such button in the Edit... dialog of IU-locations? That is then correctly connected to the specific location type and not too far away. Altough I have to admit that it is not ideal because one then has to 'Edit' a location to collect its garbage.


> > Change the "Reload" button to "Refresh" as there is another button "Reload target platform..." that does a different thing. In fact,
> 
> I would prefer to better change the latter text, the button do not really "Refresh" anything it effectively throws away the old resolved state and starts resolving again. On the other hand, even as a long term user of PDE I have no clue what "Reload target platform..." should actually mean, if the target platform is not current it says "Set as active target platfrom""

What about naming the button either `Resolve` or even `Reresolve` or alike?

laeubi:
> But In the preferences it is hidden relatively far away (one won't stumble upon it)

in more than 10 years I about 1 or maybe to times need to really "clean" the bundle-pool, also one should not expect that people actually know what a bundle pool is. Beside that, we can't / should not add buttons that are useless for other locations, e.g. this has no meaning to directories (or will it "whipe out" all my data on disk?) or maven target or oomph targlets or ...

> Does it even consider the IU-locations from the targets?

I have no clue, at laest they end up there (but disabled) when you add things in the target dialog, but that would be a place where generic cleanup regarding cached state from updatesites is most likely to be a valid choice.

> I have to admit that it is not ideal because one then has to 'Edit' a location to collect its garbage.

If there is any "garbage" it should be collected without need for the user to press a button...

kjsmita6:
Thank you all for the comments. Replying to them all here and I'll make any code review changes separately.

@laeubi 
> If you change UI you might want to upload some image to Github that shows how everything looks like, just copy an image (either file or clipboard) and use CTRL+V to add it to the description text.

Yes, sorry I forgot to do this. It is here:
image

It appears in the corresponding location in the preferences page as well when you edit the target.

> I don't think that the target editor is a suitable place for such function:
> 
> * Actually this should be something either very seldom be used if at all by a user

Maybe it's seldom used, but the times when it's needed can make a huge difference. I was helping a coworker debug issues and it would've saved so much time if we could've done this (short of throwing out the whole folder ourselves).

> * "Clean" is very unclear, why /what should I clean? Why do PDE do not clean up old stuff automatically? What are the impacts of a clean?

* The tooltip explains it a bit by saying it removes artifacts which are no longer part of the target. See later comment for renaming to something mentioning GC.
* Why PDE doesn't clean automatically I'm not sure as I didn't create this function. It probably should but constantly checking for old artifacts seems like it could get expensive.
* From all my testing, the clean has run so fast that the progress text appears and disappears so quickly that I can't even read it.

> * It is bound to a very specific location type (the IU Location / Updatesites)
> 
> Therefore I think such function is better placed at the Preferences:

This is not for software sites, it is for the PDE target definition file editor and preference page. See the above screenshot.

> I would prefer to better change the latter text, the button do not really "Refresh" anything it effectively throws away the old resolved state and starts resolving again. On the other hand, even as a long term user of PDE I have no clue what "Reload target platform..." should actually mean, if the target platform is not current it says "Set as active target platform""
I
 actually think this is exactly what "Refresh" means. Like when you refresh a webpage it closes the webpage and reopens it essentially. Versus the "Reload target platform..." makes sense as well in this case as you are giving the workspace plugins the new target platform to use. I think the naming for this could be more clear to say this, like "Reload workspace plugins with target platform" but less wordy of course.

@HannesWell 
> Clean could indeed be confused with 'wipe-out'. What about Collect Garbage or Perform GC or Run GC?

I agree with this too, we would be clear that it isn't deleting the entire bundle pool. I like Perform GC.

> That's right. But In the preferences it is hidden relatively far away (one won't stumble upon it) and the preferences for the available software sites also lists the software sites to update eclipse itself. Does it even consider the IU-locations from the targets? However what about placing such button in the Edit... dialog of IU-locations? That is then correctly connected to the specific location type and not too far away. Altough I have to admit that it is not ideal because one then has to 'Edit' a location to collect its garbage.

See my above screenshot, this is not the case. The "Clean" function applies to the entire target and not a single location.

> What about naming the button either `Resolve` or even `Reresolve` or alike?

Actually, in my first iteration of working on this, I made it say Resolve. I changed it to Refresh as I wanted it to be clear that it is throwing out the old state and THEN resolving. I didn't do reresolve because that isn't a word ;)

> in more than 10 years I about 1 or maybe to times need to really "clean" the bundle-pool, also one should not expect that people actually know what a bundle pool is. 

I would expect plugin developers to know what a bundle pool is, which is who this change is targeting as it's about the target platform editor.

>Beside that, we can't / should not add buttons that are useless for other locations, e.g. this has no meaning to directories (or will it "whipe out" all my data on disk?) or maven target or oomph targlets or ...
> 
> > Does it even consider the IU-locations from the targets?
> 
> I have no clue, at laest they end up there (but disabled) when you add things in the target dialog, but that would be a place where generic cleanup regarding cached state from updatesites is most likely to be a valid choice.

The code uses `P2TargetUtils.garbageCollect()` which compares the bundle pool to the artifacts defined in the target platform and removes any that are in the bundle pool but not in the target.


> 
> > I have to admit that it is not ideal because one then has to 'Edit' a location to collect its garbage.
> 
> If there is any "garbage" it should be collected without need for the user to press a button...

That is a different issue then.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
kjsmita6: I will make some improvements as suggested. See one of my comments: This job has always run so quickly that I haven't even been able to read the progress view's text before it disappears.

laeubi:
> Maybe it's seldom used, but the times when it's needed can make a huge difference. I was helping a coworker debug issues and it would've saved so much time if we could've done this (short of throwing out the whole folder ourselves).

Sure but as said I think this is much to prominent for something one probably should never need (if you has need it its most probably a bug) and we have already many buttons there, adding another one what is actually for an implementation detail (that is clearing the internal bundle pool of P2) seems not valid here.

> The "Clean" function applies to the entire target and not a single location.

No it does not. A target is not required to contain only (or even any!) Update site location.

TIBCOeddie:
Jumping in to give this overall POV - the target platform editor IS an advanced editor. The user IS usually fairly well versed into PDE/bundle land. I for one would have this here (there's already some advanced concepts needed here: "Reload" button vs "Reload" link vs "Resolving" when the editor opens... all of that required quite some knowledge... Clean is usually a simple concept to "what is wrong, let's get this working again". Akin to saying why is there a workspace clean button? Because sometimes (often?) builders get things wrong, JDT/M2e... they all benefit greatly from "Clean".

laeubi (changes requested):
This is a complete different concept (Clean Projects) versus "Cleaning some cache" and is completely unrelated to the target itself (it is a property of you local installation). JDT also has useful "clean Index" function, still we don't show it on every Java Editor just because it is *sometimes* useful. If you need this function regularly, there is something broken and you should report a bug to fix this instead, either to PDE or P2 ...

TIBCOeddie:
Note I am not the OP but we work on the same product today that has had us need to clean things enough (tycho/maven fun). I see your point that clean isn't about the target platform per-se. Starting to perhaps come around to ... well, a preference page (no other place really). @laeubi where might you put it? (akin to JDT's Clean Index "if something is wrong this can help someone" feature?). I wonder if somebody working on PDE can be tagged for input/where they might put it? (if not already in this comment thread).

laeubi:
> well, a preference page (no other place really). @laeubi where might you put it?

# PDE has "Plugin Development" Preference page where such thing might fit:
grafik

# This is how it looks like for JDT:
grafik

kjsmita6:
Sorry for the long delay, I got busy with other things. I have moved the button to the preference page as requested.
image

laeubi (approved):
Looks good, but build needs some attention.

kjsmita6:
On Ubuntu it failed because the build number is wrong it seems (although I'm not familiar with that build plugin). Did I need to update it in the manifest?

HannesWell:
I can have a look tomorrow.

> On Ubuntu it failed because the build number is wrong it seems (although I'm not familiar with that build plugin). Did I need to update it in the manifest?

You probably have to increment the service version segment in the Manifest.MF (and if present also in the pom.xml) for the o.e.pde.ui Plugin by +100.

Furthermore, can you please squash your changes into one commit, with one message that covers the eventual change.
For the git history it is not of interest which iterations we made during the code-review. :)

kjsmita6:
@HannesWell I have updated the MANIFEST version from 700 to 800. I don't think I can squash now as the commits are made already. I think you could have to squash when merged into master or I would have to make a new PR?

iloveeclipse:
> I don't think I can squash now as the commits are made already

Sure you can. If you work with Eclipse / EGit, select commits in the history view, Squash / Reword und push with **force** flag to github.

iloveeclipse:
Still 4 commits :) Just select more commits for squashing.

kjsmita6:
I figured it out :) nevermind, just took a while. So many force pushes.

kjsmita6:
Hasn't been any update in a week. Is there anything needed of me?

laeubi:
Thanks for the PR and sorry for the delay!

HannesWell (commented):
Sorry, I forgot to review this due to other works. But it looks good (although it was merged already).
Thanks for your contribution.

vogella:
@kjsmita6 thanks for this feature. I used it today for the first time and with this wipe I finally made my workspace compile again! Great work @kjsmita6

kjsmita6:
@vogella thank you for this comment, it means a lot. I knew I wasn't the only one who needed this!

opcoach:
I heard about it during the Eclipse Con Europe, and I was looking for it last week because of huge workspaces after changing many times my target platform  ! Thanks a lot for this very handy feature !

PR #379, opened 2022-11-06

laeubi:
Currently a plugin can contribute source locations inside the plugin itself through the extension point

org.eclipse.pde.core.source

this has the limitation that the source must be embedded and known in advance. In some situations it would be required to provide external locations dynamically e.g. maven has the concept of source artifacts that could be queried, if a bundle contains a Bundle-Source-Location one could checkout one could have a "source-repository" and so on.

This adds the necessary parts for a new extension point

org.eclipse.pde.core.dynamicSource

and a extracts the implementation that search for Eclipse/Maven co-located source bundles.

Part of #202

### DIFF
--- a/ui/org.eclipse.pde.core/schema/dynamicSource.exsd
+++ b/ui/org.eclipse.pde.core/schema/dynamicSource.exsd
@@ -20,30 +20,7 @@
          <choice minOccurs="1" maxOccurs="unbounded">
             <element ref="locator"/>
          </choice>
-         <attribute name="point" type="string" use="required">
-            <annotation>
-               <documentation>
-                  
-               </documentation>
-            </annotation>
-         </attribute>
-         <attribute name="id" type="string">
-            <annotation>
-               <documentation>
-                  
-               </documentation>
-            </annotation>
-         </attribute>
-         <attribute name="name" type="string">
-            <annotation>
-               <documentation>
-                  
-               </documentation>
-               <appInfo>
-                  <meta.attribute translatable="true"/>
-               </appInfo>
-            </annotation>
-         </attribute>
+         <attribute name="point" type="string" use="required"/>
       </complexType>
    </element>
 
@@ -76,14 +53,14 @@
          <meta.section type="examples"/>
       </appInfo>
       <documentation>
-         The following is an example of the &lt;code&gt;source&lt;/code&gt; extension:
+         The following is an example of the &lt;code&gt;dynamicSource&lt;/code&gt; extension:
 &lt;pre&gt;
- &lt;extension point = &quot;org.eclipse.pde.core.source&quot;&gt;
-     &lt;location path=&quot;src&quot;/&gt;
+ &lt;extension point = &quot;org.eclipse.pde.core.dynamicSource&quot;&gt;
+     &lt;locator class=&quot;foo.bar.CustomSourcePathLocator&quot; /&gt;
   &lt;/extension&gt;
 &lt;/pre&gt;
 
-In the example above, the source location &lt;code&gt;src&lt;/code&gt; in the contributing plug-in has been registered.
+In the example above, the new &lt;code&gt;foo.bar.CustomSourcePathLocator&lt;/code&gt; in the contributing plug-in has been registered.
       </documentation>
    </annotation>
 
@@ -92,7 +69,7 @@ In the example above, the source location &lt;code&gt;src&lt;/code&gt; in the co
          <meta.section type="apiInfo"/>
       </appInfo>
       <documentation>
-         No Java code is required for this extension point.
+         Each contributor must provide a class that implements &lt;code&gt;org.eclipse.pde.core.IPluginSourcePathLocator&lt;/code&gt;
       </documentation>
    </annotation>
 
@@ -110,7 +87,7 @@ In the example above, the source location &lt;code&gt;src&lt;/code&gt; in the co
          <meta.section type="copyright"/>
       </appInfo>
       <documentation>
-         Copyright (c) 2004 IBM Corporation and others.
+         Copyright (c) 2022 Christoph Läubrich and others.
 &lt;br&gt;
 
 This program and the accompanying materials are made 
--- a/ui/org.eclipse.pde.core/schema/dynamicSource.exsd
+++ b/ui/org.eclipse.pde.core/schema/dynamicSource.exsd
@@ -1,0 +1,125 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="org.eclipse.pde.core" xmlns="http://www.w3.org/2001/XMLSchema">
+<annotation>
+      <appInfo>
+         <meta.schema plugin="org.eclipse.pde.core" id="dynamicSource" name="Dynamic Source Locations"/>
+      </appInfo>
+      <documentation>
+         This extension point allows PDE to find source archives for libraries at runtime.
+      </documentation>
+   </annotation>
+
+   <element name="extension">
+      <annotation>
+         <appInfo>
+            <meta.element />
+         </appInfo>
+      </annotation>
+      <complexType>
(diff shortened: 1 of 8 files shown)


laeubi:
@HannesWell do you like to review this so we get it into PDE before M3?

ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IPluginSourcePathLocator.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * A plugin source path locator is capable of locating the path of the source
+ * for a given plugin.
+ * <p>
+ * A plugin source path locator is declared as an extension
+ * (<code>org.eclipse.pde.core.source</code>).
+ * </p>
+ *
+ * @since 3.16
```
HannesWell, line 27: You also have to bump the minor version in the manifest as well.

HannesWell (commented):
Yes I'm reviewing it now.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/LocalMavenPluginSourcePathLocator.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Check if there is already a local maven source bundle available
+ *
+ */
+public class LocalMavenPluginSourcePathLocator implements IPluginSourcePathLocator {
+
+	@Override
+	public IPath locateSource(IPluginBase plugin) {
+		ISharedPluginModel model = plugin.getModel();
+		String installLocation = model.getInstallLocation();
```
HannesWell, line 32:
```suggestion
		String installLocation = plugin.getModel().getInstallLocation();
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/EclipsePluginSourcePathLocator.java
```diff
@@ -0,0 +1,47 @@
...
+		String installLocation = model.getInstallLocation();
+		if (installLocation != null) {
+			File path = new File(installLocation);
+			if (path.isFile()) {
+				File sourceFile = new File(path.getParentFile(),
+						plugin.getId() + ".source_" + plugin.getVersion() + ".jar"); //$NON-NLS-1$ //$NON-NLS-2$
+				if (sourceFile.isFile()) {
+					return new Path(sourceFile.getAbsolutePath());
+				}
+			}
```
HannesWell, line 42:
```suggestion
			// For example in a bundle pool foo_123.jar => foo.source_123.jar...
			File container = new File(installLocation).getParentFile();
			File sourceFile = new File(container, plugin.getId() + ".source_" + plugin.getVersion() + ".jar"); //$NON-NLS-1$ //$NON-NLS-2$
			if (sourceFile.isFile()) {
				return new Path(sourceFile.getAbsolutePath());
			}
```

Because checking if a file exists is a very expensive operation compared to manipulating some File/String objects I suggest to only check in the end if the expected/assumed source-jar exists. Since most plugins (in the TP) have shape jar, the check if the installation is a file is probably very often true and the last check is the relevant one anyways.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/LocalMavenPluginSourcePathLocator.java
```diff
@@ -0,0 +1,49 @@
...
+			File path = new File(installLocation);
+			if (path.isFile()) {
+				// The usual Maven covention blah-123.jar =>
+				// blah-123-sources.jar
+				String bundleFileName = path.getName();
+				String sourceFileName = bundleFileName.substring(0, bundleFileName.indexOf(".jar")) + "-sources.jar"; //$NON-NLS-1$ //$NON-NLS-2$
+				File sourceFile = new File(path.getParentFile(), sourceFileName);
+				if (sourceFile.isFile()) {
+					return new Path(sourceFile.getAbsolutePath());
+				}
```
HannesWell, line 43:
```suggestion
			// The usual Maven covention foo-123.jar => foo-123-sources.jar
			String bundleFileName = path.getName();
			String sourceFileName = bundleFileName.substring(0, bundleFileName.indexOf(".jar")) + "-sources.jar"; //$NON-NLS-1$ //$NON-NLS-2$
			File sourceFile = new File(path.getParentFile(), sourceFileName);
			if (sourceFile.isFile()) {
				return new Path(sourceFile.getAbsolutePath());
			}
```
As above it if likely faster to just check the eventually expected/assumed file.

ui/org.eclipse.pde.core/schema/dynamicSource.exsd
```diff
@@ -0,0 +1,125 @@
...
+         <attribute name="name" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+               <appInfo>
+                  <meta.attribute translatable="true"/>
+               </appInfo>
+            </annotation>
+         </attribute>
```
HannesWell, line 46:
```suggestion
         <attribute name="point" type="string" use="required"/>
```
The Schema-Editor adds the `name` and `id` attributes by default. But they seem to be unused and in a quick check it also worked without.
So can't it simply be removed?

HannesWell (changes requested):
In general this change looks good and I also tested it in conjunction with https://github.com/eclipse-m2e/m2e-core/pull/853.

Only some meta-information in the Schema-Definition have to be adjusted and I have some suggestions for code improvements.
If you want I can push them as extra commit and you can include them as desired. But I just noticed that GitHub has a batch-apply of suggestions.

Another word about Extension-Points in general:
In cases like these OSGi-DS would IMHO be much simpler to use. I wonder if there are plans to move more towards OSGi services, with the final goal to make Eclipse Extenion-Points obsolet (IIRC that's Equinox specific, isn't it). I also wonder if it would be suitable to have a compatibility layer to contribute Eclipse Extension-Point to the OSGi service registry and to read OSGi services via the Extension-Point registry. This way a smooth migration for Extension contributors and provides would be possible.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
```diff
@@ -211,7 +219,11 @@ public List<SourceLocation> getUserLocations() {
 	/**
 	 * @return array of source locations that have been added via extension point
 	 */
-	public List<SourceLocation> getExtensionLocations() {
+	public Collection<SourceLocation> getExtensionLocations() {
```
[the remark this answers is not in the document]
HannesWell: I just noticed that inlining is not possible since `AlternateSourceLocations` overrides this method. So this has to stay.

ui/org.eclipse.pde.core/schema/dynamicSource.exsd
HannesWell: @laeubi do you know if my assumption is right?
I wonder if the schema is even evaluated at runtime? Because when I delete it everything still works. It is just the PDE Extension Editor that complains and does not know how to handle the attributes/childs.

HannesWell (approved):
Looks good now 😄

The test-failures are very likely unrelated.

@laeubi please either squash my changes manually or use squash-merge when submitting this.

ui/org.eclipse.pde.core/schema/dynamicSource.exsd
laeubi: I have no clue maybe @merks know about ti, I'm from "the other side" that is using OSGi service for extends my code, I only use extension points with eclipse and even there try to avoid it :-)

laeubi:
@HannesWell thanks for your help her, finally we got it!

HannesWell:
> @HannesWell thanks for your help her, finally we got it!

Thank you for the PR!
Once the corresponding M2E part is available there is one argument less to include third-party maven deps into features. 😃

PR #380, opened 2022-11-06
Also take metatype annotations into account

laeubi:
The Metatype Service Specification defines a set of annotations to allow defining possible configuration options of a service as code.

These annotations are only retained at class level and can the used at build time to generate the proper XML definition, we should enable PDE users to use these annotations as well, so this information can then be used for example in Tycho or BND builds.

This depends on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/653


HannesWell:
> 
> do I need to wait for the next i-build for the setup to catch up?

The setup is using the I-builds repo, isn't it? So I think yes you have to wait for the next I-build so that this is available.

PR #382, opened 2022-11-07
LocalMavenPluginSourcePathLocator should not assume all plugins are jars

merks:
A plugin's install location can be a folder.
Also ".jar" may appear elsewhere in the name not just at the end.

https://github.com/eclipse-pde/eclipse.pde/pull/379

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/LocalMavenPluginSourcePathLocator.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/LocalMavenPluginSourcePathLocator.java
@@ -28,11 +28,11 @@ public class LocalMavenPluginSourcePathLocator implements IPluginSourcePathLocat
 	@Override
 	public IPath locateSource(IPluginBase plugin) {
 		String installLocation = plugin.getModel().getInstallLocation();
-		if (installLocation != null) {
+		if (installLocation != null && installLocation.endsWith(".jar")) { //$NON-NLS-1$
 			File path = new File(installLocation);
-			// The usual Maven covention foo-123.jar => foo-123-sources.jar
+			// The usual Maven convention foo-123.jar => foo-123-sources.jar
 			String bundleFileName = path.getName();
-			String sourceFileName = bundleFileName.substring(0, bundleFileName.indexOf(".jar")) + "-sources.jar"; //$NON-NLS-1$ //$NON-NLS-2$
+			String sourceFileName = bundleFileName.substring(0, bundleFileName.lastIndexOf(".jar")) + "-sources.jar"; //$NON-NLS-1$ //$NON-NLS-2$
 			File sourceFile = new File(path.getParentFile(), sourceFileName);
 			if (sourceFile.isFile()) {
 				return new Path(sourceFile.getAbsolutePath());


merks:
@laeubi  I noticed this was broken because it assumes that all plugins have .jar name.

iloveeclipse:
> @laeubi I noticed this was broken because it assumes that all plugins have .jar name.

Yep. Latest build is unusable for me because of gazillions build errors like
```
eclipse.buildId=4.26.0.I20221106-2230
java.version=17.0.4
java.vendor=Eclipse Adoptium
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data /data/4x_platform_workspace -os linux -ws gtk -arch x86_64

org.eclipse.core.jobs
Error
Mon Nov 07 09:18:15 CET 2022
An internal error occurred during: "Initializing Java Tooling".

java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 39
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4604)
	at java.base/java.lang.String.substring(String.java:2707)
	at org.eclipse.pde.internal.core.LocalMavenPluginSourcePathLocator.locateSource(LocalMavenPluginSourcePathLocator.java:35)
	at org.eclipse.pde.internal.core.SourceLocationManager.lambda$0(SourceLocationManager.java:320)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1602)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:647)
	at org.eclipse.pde.internal.core.SourceLocationManager.searchExtensionLocations(SourceLocationManager.java:321)
	at org.eclipse.pde.internal.core.SourceLocationManager.findSourcePath(SourceLocationManager.java:85)
	at org.eclipse.pde.internal.core.ClasspathUtilCore.getSourceAnnotation(ClasspathUtilCore.java:193)
	at org.eclipse.pde.internal.core.ClasspathUtilCore.getSourceAnnotation(ClasspathUtilCore.java:184)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addExtraLibrary(RequiredPluginsClasspathContainer.java:580)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addExtraClasspathEntries(RequiredPluginsClasspathContainer.java:461)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addExtraClasspathEntries(RequiredPluginsClasspathContainer.java:415)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.computePluginEntries(RequiredPluginsClasspathContainer.java:191)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.getClasspathEntries(RequiredPluginsClasspathContainer.java:112)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3277)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3435)
	at org.eclipse.jdt.internal.core.JavaProject.getResolvedClasspath(JavaProject.java:2520)
	at org.eclipse.jdt.core.JavaCore.initializeAfterLoad(JavaCore.java:4632)
	at org.eclipse.jdt.internal.ui.InitializeAfterLoadJob$RealJob.run(InitializeAfterLoadJob.java:39)
```

HannesWell (approved):
Looks good. Thank you.

In general we could consider to remove this co-located Maven source artifact locator again. Because without M2E being installed (which will provide a locator that does the same and more) the cases where this locator is useful are very rare.

iloveeclipse:
> In general we could consider to remove this co-located Maven source artifact locator again. Because without M2E being installed (which will provide a locator that does the same and more) the cases where this locator is useful are very rare.

Why was `LocalMavenPluginSourcePathLocator` added to PDE at all? PDE is build system agnostic, it doesn't care if we use Maven, Gradle etc. Shouldn't that be contributed by m2e?

laeubi:
@merks thank you for looking into it, even though the code initially was contributed by @mickaelistria and I just extracted it from `ClasspathUtilCore` to the new extension point. So probably the header should be adjusted?

See https://github.com/eclipse-pde/eclipse.pde/commit/f23c5478215ef29b09f1ffd1d2c8acc1172cd5b4

> Why was LocalMavenPluginSourcePathLocator added to PDE at all? PDE is build system agnostic, it doesn't care if we use Maven, Gradle etc. 

If I remember the discussion correctly it was argued that at laest if the file is on your disk it should be found without m2e as well and maven **repository-layout** is widly adopted (gradle uses the maven repo as well as far as I know)

> Shouldn't that be contributed by m2e?

m2e will contribute a more enhanced variant as well if this is released see
- https://github.com/eclipse-m2e/m2e-core/pull/853

mickaelistria:
> Why was LocalMavenPluginSourcePathLocator added to PDE at all? PDE is build system agnostic, it doesn't care if we use Maven, Gradle etc. Shouldn't that be contributed by m2e?

PDE is also pragmatic and meant to serve its users well. One doesn't need a full Maven support to have some minimal support for Maven conventions. Moreoever, there was/is not yet a good approach to automatically detect sources.
A few lines to better serve the #1 build tool of the Java and Eclipse ecosystem is better serving users more than being "built-tool-agnostic", with a minimal impact on performance and maintenance. Very worth it.
I personally saved a lot of time *not* having to associate sources with some jars in several PDE projects.

HannesWell:
What do you think about deleting `LocalMavenPluginSourcePathLocator` once https://github.com/eclipse-m2e/m2e-core/pull/853 is GA?
In contrast to the PDE pendant the M2E implementation is a reliable solution (if applicable) and the scenarios where the former is useful without M2E are probably seldom. Furthermore it can lead to 'works on my machine situations' that are really hard to understand.

mickaelistria:
> What do you think about deleting LocalMavenPluginSourcePathLocator once https://github.com/eclipse-m2e/m2e-core/pull/853 is GA?

I think it would be a bad idea to force initialization of m2e bundles (relatively expensive in term of RAM) if one uses a plain non-Maven plugin and what we want is just to get the source of one bundle and PDE just tries a relatively usual convention among others.
I don't see a gain in removing it; however I see it can have a negative impact on functionality (if m2e is not installed then PDE won't locate some sources which could easily be located), or on performance (forcing an entire load of m2e just to check a file exists).

If this is the issue, then we can just rename it to not mention Maven but mention the convention Maven uses and that is also usable and used without it, and rename the class to ColocatedSourcesSuffixPathLocator ?

>  the scenarios where the former is useful without M2E are probably seldom.

I don't think those cases are so seldom or so unlikely or so stupid to be rejected. One can define a .target file without m2e referencing a local directory which contains bundles.jar and bundles-sources.jar. The bundle-sources.jar, the initially Maven convention, has also got mainstream in many software that doesn't use OSGi but ship Sources: the *-sources.jar are available besides the *.jar.

To be honest, I don't get what's the actual issue LocalMavenPluginSourcePathLocator that is worth the churn.

merks:
@mickaelistria  I tend to agree.  Moreover, I think the argument for keeping it would be even stronger if the implementation supported `<bundle-name>_<version>.jar` being mapped to `<bundle-name>.source_<version>.jar`, i.e., using the convention for conventionally-built source bundle.  Or maybe is that already otherwise supported via a different mechanism?

laeubi:
There is also 
https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/EclipsePluginSourcePathLocator.java

and since it is extensible feel free to implement another one if this still do not fit :-)

The main idea for making this extensible was/is that we can have different schemes either by default or by external plugins whatever suffice-

PR #395, opened 2022-11-09
Use varargs in parent test classes

akurtakov:
Remove useless methods.


akurtakov:
jenkins build succeeded

PR #396, opened 2022-11-09
Test if forcing a ApiBuilderTest.beforeClass() call fixes the tests

merks:


merks:
This is just a test to see if this makes the tests run properly...

merks:
This doesn't help fix the 611 failures.

PR #397, opened 2022-11-09
dummy commit

vik-chand:
test

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/AbstractIllegalTypeReference.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/AbstractIllegalTypeReference.java
@@ -41,7 +41,7 @@ import org.eclipse.pde.api.tools.internal.util.Signatures;
  * Base implementation of a problem detector for type references
  *
  * @since 1.1
- * @noextend This class is not intended to be sub-classed by clients.
+ * @noextend This class is not intended to be sub-classed by clients. test
  */
 public abstract class AbstractIllegalTypeReference extends AbstractProblemDetector {
 


laeubi:
@vik-chand as you can see even the GH suceeds here, but Jenkins failed. Sometimes its the other way round or both fail ...

vik-chand:
https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-397/1/ failed with 611 API tool failures whereas the next one had other issue.

The equivalent tests in integration build page is fairly stable here https://download.eclipse.org/eclipse/downloads/drops4/I20221109-0600/testresults/html/org.eclipse.pde.api.tools.tests_ep426I-unit-macM1-java17_macosx.cocoa.aarch64_17.html#ApiToolsPluginTestSuite

For 611 failure issue, there is some dependency/config/releng issue in Tycho Surefire/Maven. I am not an expert in that area. I see lot of people has already tried fixing it.  In the worst case scenario, we can disable it in ONLY in this workflow ( so that these tests run on Integration build page and issues are detected there).  Till couple of years back or little more, I remember that only pde.ui tests used to run with gerrit. I know it is annoying but not a blocker. If there is a way to recreate this API tool problem in eclipse IDE, I will be more than happy to investigate/fix it.

vik-chand:
Closing this PR since this was a dummy PR

laeubi:
@mickaelistria has done some investigations to reproduce it in a local build, maybe he can help here.

And yes its kind of random, what makes it even harder, so I think it is more a timing issue than a problem with the configuration itself.

vik-chand:
> @mickaelistria has done some investigations to reproduce it in a local build, maybe he can help here.
> 
> And yes its kind of random, what makes it even harder, so I think it is more a timing issue than a problem with the configuration itself.

Yes, I saw https://github.com/eclipse-pde/eclipse.pde/pull/354. A lot of time has been spent on this. Perhaps a way to disable these 611 tests in this workflow ( the test should still run on Integration build page) may be a better option. If someone can provide that patch with a comment linking to [bug](https://github.com/eclipse-pde/eclipse.pde/issues/232) , there would work too.

Issue #345, opened 2022-09-29
Intermittent test failures in API tools


vik-chand:
org.eclipse.pde.api.tools.builder.tests.compatibility.ProjectTypeContainerTests	testPackageNames	Error	Component 'API Tools Tests Plug-in A' in the baseline 'workspace' is disposed

org.eclipse.core.runtime.CoreException: Component 'API Tools Tests Plug-in A' in the baseline 'workspace' is disposed
at org.eclipse.pde.api.tools.internal.model.BundleComponent.baselineDisposed(BundleComponent.java:1271)
at org.eclipse.pde.api.tools.internal.model.BundleComponent.getBundleDescription(BundleComponent.java:992)
at org.eclipse.pde.api.tools.internal.model.BundleComponent.createApiDescription(BundleComponent.java:390)
at org.eclipse.pde.api.tools.internal.model.Component.getApiDescription(Component.java:106)
at org.eclipse.pde.api.tools.builder.tests.ApiBuilderTest.exportApiComponent(ApiBuilderTest.java:375)
at org.eclipse.pde.api.tools.builder.tests.compatibility.CompatibilityTest.setUp(CompatibilityTest.java:161)
at junit.framework.TestCase.runBare(TestCase.java:140)
at junit.framework.TestResult$1.protect(TestResult.java:122) 

These are the failing test cases with similar stack

ProjectTypeContainerTests::testPackageNames
BundleVersionTests::test005
BundleMergeSplitTests::test007
BundleMergeSplitTests::test003

jukzi:
duplicates https://github.com/eclipse-pde/eclipse.pde/issues/553

Issue #261, opened 2022-08-02
org.eclipse.pde.api.tools.builder.tests.compatibility.BundleMergeSplitTests fails


laeubi:
The org.eclipse.pde.api.tools.builder.tests.compatibility.BundleMergeSplitTests fails some times, it should either be fixed, disabled or explain why this particular test fails:

```
org.eclipse.core.runtime.CoreException: Component 'Test Plug-in' in the baseline 'workspace' is disposed
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.baselineDisposed(BundleComponent.java:1271)
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.getBundleDescription(BundleComponent.java:992)
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.createApiDescription(BundleComponent.java:390)
	at org.eclipse.pde.api.tools.internal.model.Component.getApiDescription(Component.java:106)
	at org.eclipse.pde.api.tools.builder.tests.ApiBuilderTest.exportApiComponent(ApiBuilderTest.java:375)
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleMergeSplitTests.setupTest(BundleMergeSplitTests.java:343)
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleMergeSplitTests.test005(BundleMergeSplitTests.java:178)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at junit.framework.TestCase.runTest(TestCase.java:177)
```

vik-chand:
Also here https://download.eclipse.org/eclipse/downloads/drops4/I20221109-1850/testresults/html/org.eclipse.pde.api.tools.tests_ep426I-unit-win32-java11_win32.win32.x86_64_11.html

vik-chand:
I will look at this in 4.27 as this also occurs in nighty build. I will try to recreate on my local setup and see if I can do something about it.

vik-chand:
Go from PDE
[...]

PR #408, opened 2022-11-17
Fix #369 [Spy] StackOverflow in the ContextSpy when selecting the Osg…

### LINKED ISSUE #369: [Spy] StackOverflow in the ContextSpy when selecting the OsgiContext
In a new install of the Eclipse for RCP and RAP developers 2022-09, open the context spy, select the root context OSGiContext in the ContextSpy, and get : 

java.lang.StackOverflowError
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getCalculatedComputations(EclipseContext.java:734)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.pushComputation(EclipseContext.java:743)
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:67)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:265)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:236)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:606)
	at org.eclipse.e4.ui.internal.workbench.ExtensionTrackeContextFunction.compute(ExtensionTrackeContextFunction.java:43)
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:70)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:265)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:236)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:606)
	at org.eclipse.e4.ui.internal.workbench.ExtensionTrackeContextFunction.compute(ExtensionTrackeContextFunction.java:43)
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:70)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:265)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:236)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.getLocal(EclipseContext.java:606)
..... 

It seems to be a problem with the context fonction called recursively...  

No problem when selecting other contexts...

opcoach:
Catch also errors when getting the values for a context to be displayed in the context spy.

### DIFF
--- a/ui/org.eclipse.pde.spy.context/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.spy.context/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.spy.context;singleton:=true
-Bundle-Version: 1.0.300.qualifier
+Bundle-Version: 1.0.400.qualifier
 Bundle-Vendor: %provider-name
 Automatic-Module-Name: org.eclipse.pde.spy.context
 Bundle-RequiredExecutionEnvironment: JavaSE-11
--- a/ui/org.eclipse.pde.spy.context/src/org/eclipse/pde/internal/spy/context/ContextDataProvider.java
+++ b/ui/org.eclipse.pde.spy.context/src/org/eclipse/pde/internal/spy/context/ContextDataProvider.java
@@ -121,7 +121,7 @@ public class ContextDataProvider extends ColumnLabelProvider implements ITreeCon
 			for (String key : selectedContext.localContextFunction().keySet())
 				try {
 					cfValues.put(key, selectedContext.get(key));
-				} catch (Exception e) {
+				} catch (Throwable e) {
 					cfValues.put(key, NO_VALUE_COULD_BE_COMPUTED + " (Exception : " + e.getClass().getName() + ")"); //$NON-NLS-1$ //$NON-NLS-2$
 				}
 			result.addAll(cfValues.entrySet());


vik-chand:
We are in RC1 phase. Do we want to put this in 4.27? or is it critical enough for 4.26RC2

laeubi:
I'm not sure if catching `Throwable` is really a good idea see:

https://stackoverflow.com/a/6083271/9503281

especially as it could be any kind of "bad situation" especially Errors like OOM ...

opcoach:
@vik-chand : the context spy does not display the OSGi context content because one context function generates a Stack Overflow... This fix is general and would have been useful to detect earlier this problem (which will be fixed in this bug https://github.com/eclipse-platform/eclipse.platform.ui/issues/454).  For me it should be fixed for the next 4.26 final release of December (or at least the platform issue 454).   Another question : why the tests are not running ? 

@laeubi : sure it is better to catch each exception one by one and react differently for each, instead of Throwable. But we are in a low level component that don't know what can happen in some context function computations. Context spy behavior must, in this case, warn the user that a context function is not working, and display the exception in red. For me this is the good way to do in this level of component. If you have another idea or point of view we can discuss of course.

opcoach:
@vik-chand question : I have seen emails where people were complaining about test failures during CI ... because it was broken. Does it mean I can merge this PR now anyway ?

vik-chand:
Wait for RC1 announcement. Then we get approvals for RC2 and then we merge. The local 611 api tools errors on PR-HEAD can be ignored. And there are few intermittent tests as well like
 - https://github.com/eclipse-pde/eclipse.pde/issues/261
 - https://github.com/eclipse-pde/eclipse.pde/issues/300
 - https://github.com/eclipse-pde/eclipse.pde/issues/345

vik-chand:
API tool tests should be run only if you made change in *pde.api*tools* plugin.  They test the API tools. However API tool testing happens for all plugins ( that should not be removed). Even if we can make a change which does this, we can escape this problem since 90% of changes in PDE are not in API tools.

opcoach:
Ok ! You can merge it when it will be possible !

vik-chand:
+1 for RC2, since master is open now, committing this early so that we have more time for testing and verification.

vik-chand (commented):
LGTM

PR #409, opened 2022-11-18
Lookup source bundles in the current target platform

laeubi:
Currently a "source bundle" is required to have a special header so PDE can discover it, but actually certain location types (e.g. P2 Sites and maven) have other ways to identify/discover the source bundle. Even the Targetplatform model already has the concept of a location that knows about the source for a bundle.

This adds a sourcepath locator that is capable of discovering such native source bundles from the target platform location.


laeubi:
Incooperated here: https://github.com/eclipse-pde/eclipse.pde/pull/425

PR #410, opened 2022-11-18
Add complexity attribute for source locators

laeubi:
This enables dynamic source providers to state how complex there computation is so the using code can decide to query faster ones first, or even decide if showing some progress or scheduling jobs is suitable.

@vik-chand I'd like to get approval for this to be included in RC2 because of the following reasons:

1. The extension point was only added in this release (and relative late) so there are no consumers we can break or where there are issues
2. this intentionally do not contain any code changes (what could be done in 2023-03 release) but possible implements should be forced to define this property from day-1 so we have the information available.

### DIFF
--- a/ui/org.eclipse.pde.core/plugin.xml
+++ b/ui/org.eclipse.pde.core/plugin.xml
@@ -403,7 +403,11 @@
     </extension>
         <extension
           point="org.eclipse.pde.core.dynamicSource">
-          <locator class="org.eclipse.pde.internal.core.EclipsePluginSourcePathLocator" />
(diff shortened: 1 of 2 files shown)


HannesWell (approved):
Looks good and I agree that submitting this soon will be beneficatial.

vik-chand:
+1 for RC2, since master is open now, committing this early so that we have more time for testing and verification.

laeubi:
Thanks @vik-chand and @HannesWell for review/approval!

vik-chand:
The code change is picked up in I20221119-0600. Please test/verify on this build.

PR #422, opened 2022-12-06
Deprecate ApiFileGenerationApplication

vogella:
Eclipse platform migrated to Tycho for generating the API baseline. Yon also also add the generation to your Tycho build. Also platform is not using this application anymore it is now deprecated and may be deleted in the future-

 </plugin>
			<plugin>
	            <groupId>org.eclipse.tycho</groupId>
				<artifactId>tycho-apitools-plugin</artifactId>
	            <version>${tycho.version}</version>
				<executions>
					<execution>
						<id>generate</id>
						<goals>
							<goal>generate</goal>
						</goals>
						<configuration>
                  			<skip>${skipAPIDescription}</skip>
                  			<projectName>${project.artifactId}_${qualifiedVersion}</projectName>
laeubi marked this conversation as resolved.
						</configuration>
					</execution>
				</executions>
			</plugin>

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiFileGenerationApplication.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiFileGenerationApplication.java
@@ -17,10 +17,17 @@ import org.eclipse.equinox.app.IApplication;
 import org.eclipse.equinox.app.IApplicationContext;
 import org.eclipse.pde.api.tools.internal.provisional.ApiPlugin;
 
+/**
+ * Prefer using the tycho-apitools-plugin to generate the api description file
+ *
+ */
+@Deprecated
 public class ApiFileGenerationApplication implements IApplication {
 
+	private static String DEPRECATION_WARNING = "DEPRECATED, PLEASE MOVE TO THE tycho-apitools-plugin TO GENERATE THE API DESCRIPTION FILE"; //$NON-NLS-1$
 	@Override
 	public Object start(IApplicationContext context) throws Exception {
+		ApiPlugin.logErrorMessage(DEPRECATION_WARNING);
 		APIFileGenerator generator = new APIFileGenerator();
 		String[] args = (String[]) context.getArguments().get(IApplicationContext.APPLICATION_ARGS);
 		generator.projectName = find("projectName", args); //$NON-NLS-1$


vogella:
Did we also replace ApiAnalysisApplication with Tycho or only the API generation part?

mickaelistria:
Could this application log a message? An application is usually not consumed via API but using the launcher with the applicationId. Some consumers would still not be informed if we just deprecate the API.

mickaelistria:
> Did we also replace ApiAnalysisApplication with Tycho or only the API generation part?

Not yet, only API generation has a good and complete replacement at the moment.

vogella:
> Could this application log a message? An application is usually not consumed via API but using the launcher with the applicationId. Some consumers would still not be informed if we just deprecate the API.

I added an error output message, which should be visible to any user.

mickaelistria (approved):
I approve that.
In general, I approve moving as much of Ant support out of PDE. If someone wants to ensure Ant build support, it would better fit in a separate project (I remember it used to be Eclipse Athena providing that, but since it dies long ago, there is apparently not so much interest).

laeubi:
I just wanted to note that this is actually not **ant task** the application was just previously **called by ant** (but you could have done this with other things as well).

laeubi:
> Did we also replace ApiAnalysisApplication with Tycho or only the API generation part?

This is work in progress see
- https://github.com/eclipse-tycho/tycho/pull/1328

vogella:
> > Did we also replace ApiAnalysisApplication with Tycho or only the API generation part?
> 
> This is work in progress see
> 
> * [Add API-analysis Mojo for Tycho eclipse-tycho/tycho#1328](https://github.com/eclipse-tycho/tycho/pull/1328)

Great. Once this is done, we should review which of the Ant Tasks from the org.eclipse.pde.api.tools.internal.tasks package are not used anymore by platform and deprecated these.

mickaelistria:
Thanks for the correction @laeubi ; I was mistaken with the other conversation. I'm then a bit skeptical that removing applications is a good idea. The benefit of applications is that they can easily be integrated with *any* tool: bash, Gradle, Tycho, Ant... without further integration effort; this seems important as we don't want to lock consumers of API Tools to 1 particular build tool or 1 particular process.
But as it's just a deprecation, it's all right and it's actually a good test. We may get attention from other consumers and their feedback will help to decide how to continue support of the API application or not.

HannesWell (commented):
Looks good to me.
And while it is right that applications give great flexibility, my guess is that there are not many, If no usages at all without Tycho. I don't think there are many people that build Eclipse Plugins without Tycho and therefore I think this applies to this app top.
But if I'm wrong the deprecation notice will hopefully encourage other users to let ist know.

PR #425, opened 2022-12-14
Lookup source bundles in the current target platform

laeubi:
Currently a "source bundle" is required to have a special header so PDE can discover it, but actually certain location types (e.g. P2 Sites and maven) have other ways to identify/discover the source bundle. Even the Targetplatform model already has the concept of a location that knows about the source for a bundle.

This adds a sourcepath locator that is capable of discovering such native source bundles from the target platform location and implements an ordered lookup that first queries lower complexity locators.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BundleManifestSourceLocationManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BundleManifestSourceLocationManager.java
@@ -13,20 +13,26 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.core;
 
-import java.util.ArrayList;
+import java.io.File;
+import java.net.URI;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.equinox.frameworkadmin.BundleInfo;
 import org.eclipse.osgi.util.ManifestElement;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.pde.core.plugin.IPluginBase;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.core.target.TargetBundle;
 import org.eclipse.pde.internal.core.plugin.PluginBase;
 import org.osgi.framework.BundleException;
 import org.osgi.framework.Constants;
@@ -45,7 +51,9 @@ public class BundleManifestSourceLocationManager {
 	/**
 	 * Maps SourceLocationKeys (plugin name and version) to IPluginModelBase objects representing source bundles
 	 */
-	private Map<SourceLocationKey, IPluginModelBase> fPluginToSourceBundle = new LinkedHashMap<>(0);
+	private Map<SourceLocationKey, IPluginModelBase> fPluginToSourceBundle = Map.of();
+
+	private Map<SourceLocationKey, TargetBundle> fPluginToTargetBundle = Map.of();
 
 	/**
 	 * Returns a source location that provides source for a specific plugin (specified by name and version)
@@ -54,13 +62,45 @@ public class BundleManifestSourceLocationManager {
 	 * @return a source location or <code>null</code> if no location exists for this plugin
 	 */
 	public SourceLocation getSourceLocation(String pluginName, Version pluginVersion) {
-		IPluginModelBase plugin = fPluginToSourceBundle.get(new SourceLocationKey(pluginName, pluginVersion));
-		if (plugin != null) {
-			SourceLocation location = new SourceLocation(new Path(plugin.getInstallLocation()));
-			location.setUserDefined(false);
-			return location;
+		return locationOf(new SourceLocationKey(pluginName, pluginVersion)).orElse(null);
+	}
+
+	private Optional<SourceLocation> locationOf(SourceLocationKey key) {
+		return pluginModel(key).or(() -> targetBundle(key));
+	}
+
+	private Optional<SourceLocation> targetBundle(SourceLocationKey key) {
+		TargetBundle targetBundle = fPluginToTargetBundle.get(key);
+		if (targetBundle != null) {
+			BundleInfo bundleInfo = targetBundle.getBundleInfo();
+			if (bundleInfo != null) {
+				URI location = bundleInfo.getLocation();
+				if (location != null) {
+					try {
+						File file = new File(location);
+						SourceLocation sourceLocation = new SourceLocation(new Path(file.getAbsolutePath()));
+						sourceLocation.setUserDefined(false);
+						return Optional.of(sourceLocation);
+					} catch (RuntimeException e) {
+						// cannot be used then...
+					}
+				}
+			}
 		}
-		return null;
+		return Optional.empty();
+	}
+
+	private Optional<SourceLocation> pluginModel(SourceLocationKey key) {
+		IPluginModelBase plugin = fPluginToSourceBundle.get(key);
+		if (plugin != null) {
+			String path = plugin.getInstallLocation();
+			if (path != null) {
+				SourceLocation location = new SourceLocation(new Path(path));
+				location.setUserDefined(false);
+				return Optional.of(location);
+			}
+		}
+		return Optional.empty();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


opcoach:
Great ! Do you mean that plugins will be searched now in this order : first in the maven .m2 repository ( as all projects build populate it and we often have everything inside), and then in the bundle pools of the workspace and then somewhere else ? That would be a really nice upgrade !!

laeubi:
1. Check if there is is source bundle next to the executable one
2. Check if there is a source bundle with maven naming convention next to the executable one
3. Check if there is a Target Bundle (in the current active target platform) that is a source bundle and is the target for the executable bundle
4. check if there are more extensions (e.g. m2e will provide one too find it from the configured remote repositories)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
```diff
@@ -446,6 +463,27 @@ protected BundleManifestSourceLocationManager initializeBundleManifestLocations(
 
 	private static final class SourceExtensions {
 		final Collection<SourceLocation> locations = new LinkedHashSet<>();
-		final List<IPluginSourcePathLocator> locators = new ArrayList<>();
+		final List<OrderedPluginSourcePathLocator> locators = new ArrayList<>();
```
HannesWell, line 466: Can't you just make this a `List<Map.Entry<IPluginSourcePathLocator, LocatorComplexity>>` and pass a comparator to `Collections.sort()` like `Comparator.comparing(Entry::getValue)`?
That would save the extra inner static class.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformPluginSourcePathLocator.java
```diff
@@ -0,0 +1,75 @@
...
+ * Investigates the current active target if a matching source bundle can be
+ * found.
+ */
+public class TargetPlatformPluginSourcePathLocator implements IPluginSourcePathLocator {
+
+	@Override
+	public IPath locateSource(IPluginBase plugin) {
+		ITargetPlatformService platformService = TargetPlatformService.getDefault();
+		try {
+			TargetBundle[] bundles = platformService.getWorkspaceTargetDefinition().getBundles();
```
HannesWell, line 38: Can't you build a mapping from plugin to source bundle once and re-use it until the TP is re-loaded? IIRC one can listen to this event.
Then this operation would be very cheap after the cache has been build.

But at the moment I'm not 100% certain what the key of such mapping would be (since IPluginBase is not a BundleInfo or TargetBundle), but probably the bundle file path or similar would be a good start.

HannesWell (changes requested):
In general I like the direction, but have a few suggestions for enhancements.

Since it was also discussed, I would like to add some words (although they don't apply to this change directly):  In general I think it is good to have some from of flag for source bundles, may it be maintained within the Bundle-Manifest (or somewhere else) or in the P2-metadata. Just anything that clearly states this is the source of that bundle.
Just relying on naming conventions is IMHO too weak and just leads to similar 'issues' like with 'test'-fragments.

laeubi:
> Since it was also discussed, I would like to add some words (although they don't apply to this change directly): In general I think it is good to have some from of flag for source bundles, may it be maintained within the Bundle-Manifest (or somewhere else) or in the P2-metadata. Just anything that clearly states this is the source of that bundle.
> Just relying on naming conventions is IMHO too weak and just leads to similar 'issues' like with 'test'-fragments.

This works for years in maven to have a convention based on the base name and having a suffix so I have no strong feelings here, at least it is nothing the core PDE has to worry about how it is decided if better meta-data is available. The concept of "Source-Bundles" is just broken because requiring that they are "bundles" that do not contain anything executable is really strange, java even ships the sources as a Zip-File and do not require them to be a jar file, sources can even be a plain folder.

Nerveless the goal is not to change the whole world but to restore a function that was clearly the intend of the Target Platform API designer, because currently the information about source is only used in the UI to display a different icon...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
laeubi: I think we do not safe much here and are prepared to further enhancements e.g. I think the next would be to have some kind of priority to sort items on the same complexity level, e.g. as you said naming conventions might be only a weak assumption as you said and we probably want to prefer other metrics first, also this eases debugging.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformPluginSourcePathLocator.java
laeubi: This is something one might maybe more implement in the Targetplatform itself because having such state in each component using it and especially with extensions that do not have a lifecycle is often cumbersome, for the moment I think it does not matter much and can be improved later, if one is certain work could even be parallelized here :-)

I even first did a Map-like approach here https://github.com/eclipse-pde/eclipse.pde/pull/409 but in the end it didn't worked out so well and was hard to understand (and didn't solved the caching issue anyways).

laeubi:
I now changed this so PDE first looks in the target (what should be the best source of source bundles anyways) and then queries extensions afterwards.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformPluginSourcePathLocator.java
HannesWell: For small targets it probably does not matter, but for large there could be a noticeable impact. But I'm neither sure about that nor do I have experience in this exact area and it just an guess and actually one would have to make some performance measurements. But if it isn't too complicated it should be addressed.

Couldn't we add a method `getBundles(String symbolicName)` that returns all bundles with that name and the TargetDefinition maintains the mapping from name -> TargetBundles internally?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
```diff
@@ -441,11 +461,43 @@ private static SourceExtensions processExtensions() {
...
+
+		@Override
+		public int compareTo(OrderedPluginSourcePathLocator o) {
+			return complexity.compareTo(o.complexity);
+		}
+
+	}
+
+	private enum LocatorComplexity {
+		low, medium, high, unkown;
```
HannesWell, line 501: Is the concept of complexity still relevant with your latest change?
If not I suggest to remove it again.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BundleManifestSourceLocationManager.java
```diff
@@ -45,7 +51,9 @@ public class BundleManifestSourceLocationManager {
 	/**
 	 * Maps SourceLocationKeys (plugin name and version) to IPluginModelBase objects representing source bundles
 	 */
-	private Map<SourceLocationKey, IPluginModelBase> fPluginToSourceBundle = new LinkedHashMap<>(0);
+	private Map<SourceLocationKey, IPluginModelBase> fPluginToSourceBundle = Map.of();
+
+	private Map<SourceLocationKey, TargetBundle> fPluginToTargetBundle = Map.of();
```
HannesWell, line 56: One thing that could be improved to map SourceLocationKeys directly to the SourceLocation. That would save the computation in each look-up and would probably simplify `getSourceLocations()`.

But I'm fine with doing that later. :)

HannesWell (approved):
Much better 😃 Looks good to me, although I could look at the code and did test it.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
laeubi: Yes for example m2e will most probably have a mediom and/or high one implementation.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BundleManifestSourceLocationManager.java
laeubi: I alligned this with how pde aready works, and even if it should be not required, it seems this is for cases where the location might has changed but I can't find a reall use-case for that.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationManager.java
HannesWell: Understand 👍🏽
I think it would then be good to split the commit into two. One that introduces the complexity and one that adds support for 'native target-sources'.
You probably also want to update the commit message of this that commit.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BundleManifestSourceLocationManager.java
HannesWell: Yes, that would be different and then should also be changed for the existing code.
It was more a general remark for a potential improvement that could be done later.

I would assume that the location can change at all without re-loading the target?

PR #428, opened 2022-12-27
Make the "Cleanup Manifest" action more accessible/visible to the user

laeubi:
PDE has a wizzard to cleanup a manifest, but this is currently quite hidden. One can select from the context menu Plugin-Tools>Organize Manifests, or one can open the manifest editor and under "Exporting" to open the wizard. Especially the second one indicates tha this is more a task done before exporting the plugin (what most of the time happens in a CI build instead from the IDE) and is not accessible when one of the other tabs are open.

This change adds a new icon at the toolbar next to the run icons so it is visible on every tab so people hopefully use tthis feature more.


HannesWell (approved):
Looks good. 👍🏽

But I have looked at it at GH and have not tested it.

Issue #306, opened 2022-08-23
Validation button on product shows more issues than validation on runtime


vogella:
Validation issues reported with the validation button on the product:

![image](https://user-images.githubusercontent.com/139910/186101307-7402eb78-380d-40c9-a284-c94fd727f539.png)

Validation issues reported with the validation button on the launch configuration:

![image](https://user-images.githubusercontent.com/139910/186101464-63003a59-26fd-4914-bb65-d282503f2175.png)

@HannesWell do we have two different validations? The one from the launch configuration displays correct results.

vogella:
> Have you tried a recent I-build?

This was with yesterdays I-build. 

>If not I can have a look at it next week when back from vacation.

Thanks

HannesWell:
> > If not I can have a look at it next week when back from vacation.
> 
> Thanks

I have to admit that haven't been on vacation for that long and simply forgot this. 😅
But it turned that this, just like #343 are the root cause for #437 and should all be fixed with #438.

PR #434, opened 2023-01-19
Do not explicitly include commons-io in template

akurtakov:
Nothing requires it directly and Maven central and Orbit had different symbolic names so better let the tooling resolve whatever it needs(if at all).

### DIFF
--- a/ui/org.eclipse.pde.ui.templates/templates_3.5/E4Application/$pluginId$.product
+++ b/ui/org.eclipse.pde.ui.templates/templates_3.5/E4Application/$pluginId$.product
@@ -25,7 +25,6 @@
       <plugin id="org.apache.batik.css"/>
       <plugin id="org.apache.batik.i18n"/>
       <plugin id="org.apache.batik.util"/>
-      <plugin id="org.apache.commons.commons-io"/>
       <plugin id="org.apache.commons.jxpath"/>
       <plugin id="org.apache.commons.logging"/>
       <plugin id="org.apache.felix.scr"/>


gireeshpunathil:
I am new to this project, so pardon me if this is irrelevant. Should the `version` be bumped in line 4 with this change? or that version is meant for something else?

`<product name="$pluginId$" id="$pluginId$.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="1.0.0.qualifier" useFeatures="false" includeLaunchers="true">`

akurtakov:
> I am new to this project, so pardon me if this is irrelevant. Should the `version` be bumped in line 4 with this change? or that version is meant for something else?
> 
> `<product name="$pluginId$" id="$pluginId$.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="1.0.0.qualifier" useFeatures="false" includeLaunchers="true">`

This file is a *template* used by PDE wizards to create new application so it makes sense to start new product from version 1.0.

akurtakov:
@HannesWell Can you help with why commons.io is not auto resolved automatically in not listed in the feature?

HannesWell:
> @HannesWell Can you help with why commons.io is not auto resolved automatically in not listed in the feature?

I can have a look at this today's or tomorrows evening.

HannesWell:
After a quick look at the Test I belive the reason for the failure is that automatic inclusion of deps is not enabled by default for Plugin-ased Products.

Setting the flag probably fixes this and #437, but I can check that when on my computer.

HannesWell:
> After a quick look at the Test I belive the reason for the failure is that automatic inclusion of deps is not enabled by default for Plugin-ased Products.
> 
> Setting the flag probably fixes this and #437, but I can check that when on my computer.

After a more in depth analysis it turned out that it is actually #306/#343.
The automatic inclusion of requirements is turned on by default for all types of Products. It is only turned of the plugin-based launch configs (for backward-compatibility reasons).
Created #438 to fix the former and then this should succeed, at least it passed in my IDE then.

akurtakov:
Jenkins build succeeds without issues.

PR #435, opened 2023-01-22
Miscellaneous clean-ups and minor code improvements.

HannesWell:
Does anybody want to review this?

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
@@ -19,7 +19,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Dictionary;
-import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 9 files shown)


akurtakov:
Haven't looked in details but overall I like where this is going so if you don't get replies just merge if/when you're confident.

HannesWell:
Looks like the API-tools test errors are back... I already thought they luckily just vanished since the previous build of this PR was all green. Have seen the failures in other PRs as well.

Does anybody know what changed in the meantime?

akurtakov:
I have no idea. They used to be 611 so at least some amount of them no longer fail but why is beyond me.

merks:
From past debugging, the problem was caused when the builder kicked in but for tests the builder should be disabled.

Maybe @iloveeclipse remembers the details better than me...

merks:
If something doesn't properly turn off the parallel builds then depending on how quickly the build kicks in,  it could cause random failures. In any case, it's these that disable the parallel build:

org.eclipse.pde.api.tools.builder.tests.ApiBuilderTest.beforeClass()
org.eclipse.pde.api.tools.tests.AbstractApiTest.beforeClass()

Maybe it needs to be done in more places?  Or maybe the order in which tests runs is random and sometimes some tests run before these two? Just a theory; this did bite me when I was fixing an API tools bug...

Issue #437, opened 2023-01-25
validateProduct test fails since I20230124-0530


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20230124-0530/testresults/html/org.eclipse.pde.ui.templates.tests_ep427I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

Seeing references to missing `jdom` I assume this is a regression from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/815

```
Generated product fails validation: [Missing Constraint: Import-Package: javax.servlet; version="0.0.0", Missing Constraint: Import-Package: javax.servlet.http; version="0.0.0", Missing Constraint: Import-Package: javax.servlet.jsp; version="0.0.0", Missing Constraint: Import-Package: org.apache.commons.beanutils; version="0.0.0", Missing Constraint: Import-Package: org.jdom; version="0.0.0", Missing Constraint: Import-Package: org.jdom.input; version="0.0.0"]

java.lang.AssertionError: Generated product fails validation:
[Missing Constraint: Import-Package: javax.servlet; version="0.0.0", Missing Constraint: Import-Package: javax.servlet.http; version="0.0.0", Missing Constraint: Import-Package: javax.servlet.jsp; version="0.0.0", Missing Constraint: Import-Package: org.apache.commons.beanutils; version="0.0.0", Missing Constraint: Import-Package: org.jdom; version="0.0.0", Missing Constraint: Import-Package: org.jdom.input; version="0.0.0"]
at org.junit.Assert.fail(Assert.java:89)
at org.eclipse.pde.ui.templates.tests.TestPDETemplates.validateProduct(TestPDETemplates.java:193)
```

@mickaelistria : please check.

akurtakov:
I'm pretty sure that the root cause is the same as the one preventing https://github.com/eclipse-pde/eclipse.pde/pull/434 to be merged.

iloveeclipse:
@akurtakov : I have no idea about things mentioned in #434.
The question for me: is this something that **must** be fixed because it is a production issue or is this just unfortunate test setup / data that can be fixed?

akurtakov:
It's a bug in PDE resolving IMO that wasn't covered by a test and this change exposes it.

mickaelistria:
The root cause may then be https://github.com/eclipse-pde/eclipse.pde/issues/429

merks:
@mickaelistria 

[...]

Issue #343, opened 2022-09-26
org.osgi.service.prefs reported as missing for RCP applications


vogella:
If I create a new RCP application using the org.eclipse.e4.rcp feature and its dependencies, PDE reports org.osgi.service.prefs reported as missing

![image](https://user-images.githubusercontent.com/139910/192289122-91b0698a-4b9c-42a0-bd2d-f49f5677fe5c.png)

@HannesWell is this really missing or just reported incorrectly?

HannesWell:
> @HannesWell is this really missing or just reported incorrectly?

I assume the automatic inclusion of dependencies is enabled in the product and in the launch config?
Is it present in the TargetPlatform?
At least version 1.1.2 (which fits the bundle version range) is present in the current SimRel repo:
https://download.eclipse.org/releases/2022-09/202209141001/plugins/

If you can provide a minimal reproducer with target-platform and product file I can debug the issue.

PR #438, opened 2023-01-25
[Editor] Consider complete Set of launched Bundles in Product-validation

HannesWell:
The Validation provided in the Product-Editor only considered the Set of Plug-ins/Bundles directly contained in the product and did not consider transitive dependencies included in launches of the product. With this change this is fixed so that the Validation in the Product-Editor sees the same set of Bundles like the validation when the product is really being launched.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/306
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/343
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/437
Required for https://github.com/eclipse-pde/eclipse.pde/pull/434

### DIFF
--- a/ui/org.eclipse.pde.ui.templates.tests/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.ui.templates.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for PDE templates
 Bundle-SymbolicName: org.eclipse.pde.ui.templates.tests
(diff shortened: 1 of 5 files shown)


HannesWell:
Looks like the API-tools test errors are back... I already thought they luckily just vanished (e.g. https://github.com/eclipse-pde/eclipse.pde/pull/435 was full green).

akurtakov:
So what do we do here?

HannesWell:
> So what do we do here?

Since I'm relativly sure that the Test-Failures are unrelated (seen the same in other PRs) I think we should merge this now so that we can proceed with other PRs. 

I'll have a Look at the failures at the weekend.

Issue #433, opened 2023-01-18
Run Configuration doesn't add multiple versions of a bundle


mickaelistria:
Currently, m2e uses gson 2.10.1 and 2.9.1; both are included in the Run Configuration, either with "Include all available plugins" or with explicitly selecting the included plugins, selecting both versions.
The clicking "Validate" shows that the older version seems ignored and thus missing; running the configuration (starting an IDE) shows in the Installation Details of the application under test that the older gson is not included. As a result, it seems currently not possible to properly debug m2e as a whole.
I've seen other cases like this for other projects.
This is a blocker for several development workflows.

laeubi:
I think you need to use feature for this (I have created a feature based product for such purpose that you can turn into a run config).

mickaelistria:
> I think you need to use feature for this

That can be a workaround, but I don't want to use features for this on the long term. Having mulitple versions of a bundle in the launch config used to work.

HannesWell:
Actually it is possible to add multiple versions of a Plugin/Bundle or a Feature to a Run-Configuration and that is tested extensively in [PluginBasedLaunchTest](https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/PluginBasedLaunchTest.java) and friends.

> The clicking "Validate" shows that the older version seems ignored and thus missing; running the configuration (starting an IDE) shows in the Installation Details of the application under test that the older gson is not included.

What you are experiencing is probably a result of the resolution failure in the TP, similar like https://github.com/eclipse-pde/eclipse.pde/issues/429:

![grafik](https://user-images.githubusercontent.com/44067969/213286320-06a12e03-36db-4954-81b7-5b4c1e732577.png)



If you add the required older gson bundle version to the provided Run-config for m2e (as extra Bundle, since that config is based on features) you can simply continue after the validation error message and everything works in the launched eclipse and you can even verify in the About dialog that both gson versions are present.

Just like https://github.com/eclipse-pde/eclipse.pde/issues/429 I can have a look for this (hopefully at the weekend).

mickaelistria:
> If you add the required older gson bundle version to the provided Run-config for m2e (as extra Bundle, since that config is based on features) you can simply continue after the validation error message and everything works in the launched eclipse and you can even verify in the About dialog that both gson versions are present.

That's not what I see with I20230113-1800
Target Platform:
[...]

PR #441, opened 2023-01-26
Honor multiple versions of same bundle in launch config

mickaelistria:
Do not consider only bundle id when generating the runtime metadata, resolving to actually configured bundle as soon as possible.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/433

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
@@ -16,7 +16,6 @@ package org.eclipse.pde.internal.launching.launcher;
 import java.io.*;
 import java.net.URL;
 import java.util.*;
-import java.util.Map.Entry;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.variables.IStringVariableManager;
 import org.eclipse.core.variables.VariablesPlugin;
@@ -121,8 +120,10 @@ public class LaunchConfigurationHelper {
 			}
 			// if target's config.ini has the osgi.bundles header, then parse and compute the proper osgi.bundles value
 			String bundleList = properties.getProperty(PROP_OSGI_BUNDLES);
-			if (bundleList != null)
-				properties.setProperty(PROP_OSGI_BUNDLES, computeOSGiBundles(TargetPlatformHelper.stripPathInformation(bundleList), bundles, bundlesWithStartLevels));
+			if (bundleList != null) {
+				boolean autoStart = configuration.getAttribute(IPDELauncherConstants.DEFAULT_AUTO_START, false);
+				properties.setProperty(PROP_OSGI_BUNDLES, computeOSGiBundles(TargetPlatformHelper.stripPathInformation(bundleList), bundles, bundlesWithStartLevels, autoStart));
+			}
 		} else {
 			String templateLoc = configuration.getAttribute(IPDELauncherConstants.CONFIG_TEMPLATE_LOCATION, (String) null);
 			if (templateLoc != null) {
@@ -134,8 +135,9 @@ public class LaunchConfigurationHelper {
 			}
 		}
 		// whether we create a new config.ini or read from one as a template, we should add the required properties - bug 161265
+		boolean autostart = configuration.getAttribute(IPDELauncherConstants.DEFAULT_AUTO_START, false);
 		if (properties != null) {
-			addRequiredProperties(properties, productID, bundles, bundlesWithStartLevels);
+			addRequiredProperties(properties, productID, bundles, bundlesWithStartLevels, autostart);
 		} else {
 			properties = new Properties();
 		}
@@ -145,7 +147,6 @@ public class LaunchConfigurationHelper {
 		String osgiBundles = properties.getProperty(PROP_OSGI_BUNDLES);
 		int start = configuration.getAttribute(IPDELauncherConstants.DEFAULT_START_LEVEL, 4);
 		properties.put("osgi.bundles.defaultStartLevel", Integer.toString(start)); //$NON-NLS-1$
-		boolean autostart = configuration.getAttribute(IPDELauncherConstants.DEFAULT_AUTO_START, false);
 
 		// Special processing for launching with p2 (simple configurator)
 		if (osgiBundles != null && osgiBundles.contains(IPDEBuildConstants.BUNDLE_SIMPLE_CONFIGURATOR) && bundles.containsKey(IPDEBuildConstants.BUNDLE_SIMPLE_CONFIGURATOR)) {
@@ -192,7 +193,7 @@ public class LaunchConfigurationHelper {
 		return properties;
 	}
 
-	private static void addRequiredProperties(Properties properties, String productID, Map<String, List<IPluginModelBase>> bundles, Map<IPluginModelBase, String> bundlesWithStartLevels) {
+	private static void addRequiredProperties(Properties properties, String productID, Map<String, List<IPluginModelBase>> bundles, Map<IPluginModelBase, String> bundlesWithStartLevels, boolean autoStart) {
 		if (!properties.containsKey("osgi.install.area")) //$NON-NLS-1$
 			properties.setProperty("osgi.install.area", "file:" + TargetPlatform.getLocation()); //$NON-NLS-1$ //$NON-NLS-2$
 		if (!properties.containsKey("osgi.configuration.cascaded")) //$NON-NLS-1$
@@ -205,7 +206,7 @@ public class LaunchConfigurationHelper {
 		if (properties.containsKey("osgi.splashPath")) //$NON-NLS-1$
 			resolveLocationPath(properties.getProperty("osgi.splashPath"), properties, bundles); //$NON-NLS-1$
 		if (!properties.containsKey(PROP_OSGI_BUNDLES))
-			properties.setProperty(PROP_OSGI_BUNDLES, computeOSGiBundles(TargetPlatform.getBundleList(), bundles, bundlesWithStartLevels));
+			properties.setProperty(PROP_OSGI_BUNDLES, computeOSGiBundles(TargetPlatform.getBundleList(), bundles, bundlesWithStartLevels, autoStart));
 		if (!properties.containsKey("osgi.bundles.defaultStartLevel")) //$NON-NLS-1$
 			properties.setProperty("osgi.bundles.defaultStartLevel", "4"); //$NON-NLS-1$ //$NON-NLS-2$
 	}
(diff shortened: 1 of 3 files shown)


iloveeclipse:
Mickael, please note: this should be done only for non singleton bundles. I don't think I see the change respects that.
I mean this attribute: `Bundle-SymbolicName: org.eclipse.e4.ui.workbench;singleton:=true`

mickaelistria:
> Mickael, please note: this should be done only for non singleton bundles. I don't think I see the change respects that.

I disagree. If a user says "I want those 2 versions in the runtime", then both should be in, independently of the bundle specificity (eg dependencies, singleton...). Then it's up to the runtime to pick the version that makes most sense and honor the singleton aspect.
If we filter too much in PDE, then we may even have PDE explicitly removing 1 bundle for some reasons that PDE has implemented, but actually build a case that would be different from what Equinox would chose. ie PDE would pick singletonBundle 2.0.0 only when the same launch containing both singletonBundle 1.0.0 and 2.0.0 would make Equinox favor 1.0.0 for some reason.
So overall, PDE is not the right place where to honor that, it has to honor user request and the launched application will take care of honoring OSGi rules.

laeubi:
Installing more than one singleton bundle/fragment is an error and rejected by OSGi ("Equinox"), so it makes sense for the Management Agent ("PDE") to only include one.

But this is more a failure in an earlier stage, the Run Configuration page already should provide an error (probably only at validation) and then ask the user if `A` or `B` should be included.

mickaelistria:
> Installing more than one singleton bundle/fragment is an error and rejected by OSGi ("Equinox")

That's not true, example launching my IDE with the `-console` flag:
```
g! ss org.eclipse.lsp4e
"Framework is launched."


id	State       Bundle
307	STARTING    org.eclipse.lsp4e_0.16.0.202301162058
308	STARTING    org.eclipse.lsp4e.debug_0.15.1.202301162058
g! install /home/mistria/Downloads/eclipseide-rust-1.2.4-linux.gtk.x86_64/eclipse/plugins/org.eclipse.lsp4e_0.13.12.202206011407.jar
Bundle ID: 526
```
What is forbidden by OSGi is not installing, but starting multiple singleton bundles:
```
g! start 526
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.lsp4e [526]
  Another singleton bundle selected: osgi.identity; type="osgi.bundle"; version:Version="0.16.0.202301162058"; osgi.identity="org.eclipse.lsp4e"; singleton:="true"
```

Having validation early to mention that to user is fine, but if users decide to not care about validation warning, the application should be provisioned as decided by the user, even if this includes 2 versions of the singleton.

laeubi:
> What is forbidden by OSGi is not installing, but starting multiple singleton bundles

The error says it can't _resolve_ the bundle, so previous it is just in _installed_ state (what is kind of pointless in general as you can actually don't do much with such a bundle anyways), so for a run config (where one want a resolvable state), it should at least bring up the validation dialog as it does for other validation errors.

mickaelistria:
>  it should at least bring up the validation dialog as it does for other validation errors.

I agree with that, but this is not really in the scope of that change; and it may even actually already be working

mickaelistria:
Test failure is
```
java.lang.AssertionError: expected null, but was:<plugin.a>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotNull(Assert.java:756)
	at org.junit.Assert.assertNull(Assert.java:738)
	at org.junit.Assert.assertNull(Assert.java:748)
	at org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testDelete_projectWithChangedBundleRoot(WorkspaceModelManagerTest.java:170)
```
I've tried to reproduce the failure locally, quite unsuccessfully. Is there something particular about this test that I should look at? It seems not so related to other changes.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
```diff
@@ -229,40 +233,28 @@ private static String computeOSGiBundles(String bundleList, Map<String, List<IPl
...
-				if (buffer.length() > 0)
-					buffer.append(',');
-				buffer.append(id);
-				if (index != -1 && index < token.length() - 1)
-					buffer.append(token.substring(index));
-				initialBundleSet.add(id);
+			IPluginModelBase model = getLatestModel(id, bundles);
+			if (model != null) {
+				String startInfo = index + 1 < token.length() ? token.substring(index + 1) : ""; //$NON-NLS-1$
+				allBundles.put(model, startInfo);
```
HannesWell, line 245:
```suggestion
				allBundles.putIfAbsent(model, startInfo);
```
Otherwise the values specified in the launch config are always overwritten by the values in the given `initialBundleList`.
But I think that this is actually not sufficient because existing entries often have `default:default` as value. So the value should be merged with an existing value where only `default` for the start-level/autostart property is overwritten and a non-default value is left unchanged.
Map.merge with a corresponding remapping function could do that job.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
```diff
@@ -229,40 +233,28 @@ private static String computeOSGiBundles(String bundleList, Map<String, List<IPl
...
-				if (buffer.length() > 0)
-					buffer.append(',');
-
-				String slinfo = entry.getValue();
-				buffer.append(id);
-				buffer.append('@');
-				buffer.append(slinfo);
+			if (!allBundles.containsKey(model) && //
+					!IPDEBuildConstants.BUNDLE_OSGI.equals(id)) { // write out all bundles in osgi.bundles - bug 170772
+				allBundles.put(model, entry.getValue());
```
HannesWell, line 253: Wouldn't that be simpler?
```suggestion
			if (!IPDEBuildConstants.BUNDLE_OSGI.equals(id)) { // write out all bundles in osgi.bundles - bug 170772
				allBundles.putIfAbsent(model, entry.getValue());
			}
```

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
```diff
@@ -229,40 +233,28 @@ private static String computeOSGiBundles(String bundleList, Map<String, List<IPl
...
 			}
 		}
 
-		return buffer.toString();
+		return allBundles.entrySet().stream().map(entry -> getBundleURL(entry.getKey(), true) + getStartData(entry.getValue(), defaultAuto)).collect(Collectors.joining(",")); //$NON-NLS-1$
```
HannesWell, line 257: For better readability I would instead of this one-liner would rather go with
```suggestion
		return allBundles.entrySet().stream().map(entry -> {
			IPluginModelBase bundle = entry.getKey();
			String startInfo = entry.getValue();
			return getBundleURL(bundle, true) + getStartData(startInfo, defaultAuto);
		}).collect(Collectors.joining(",")); //$NON-NLS-1$
```
or
```suggestion
		StringJoiner osgiBundles = new StringJoiner(","); //$NON-NLS-1$
		allBundles.forEach((bundle, startInfo) -> {
			String bundleURL = getBundleURL(bundle, true);
			String startData = getStartData(startInfo, defaultAuto);
			osgiBundles.add(bundleURL + startData);
		});
```

HannesWell (changes requested):
Change looks good in general and fixes the issues.

But the logic to merge pre-specified bundles and those specified in the launch config should be enhanced (see comments below).

HannesWell:
> > it should at least bring up the validation dialog as it does for other validation errors.
> 
> I agree with that, but this is not really in the scope of that change; and it may even actually already be working

Agree as well and it is indeed already working. Just made a1 and a2 singletons and then the launch validation complains:

grafik


> Test failure is
> 
> ```
> java.lang.AssertionError: expected null, but was:<plugin.a>
> 	at org.junit.Assert.fail(Assert.java:89)
> 	at org.junit.Assert.failNotNull(Assert.java:756)
> 	at org.junit.Assert.assertNull(Assert.java:738)
> 	at org.junit.Assert.assertNull(Assert.java:748)
> 	at org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testDelete_projectWithChangedBundleRoot(WorkspaceModelManagerTest.java:170)
> ```
> 
> I've tried to reproduce the failure locally, quite unsuccessfully. Is there something particular about this test that I should look at? It seems not so related to other changes.

Unfortunately this pops up regularly in the CI, but like you I wasn't able to reproduce and fix that locally. :/

mickaelistria:
I updated the PR according to your feedback @HannesWell .

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java
```diff
@@ -229,40 +231,29 @@ private static String computeOSGiBundles(String bundleList, Map<String, List<IPl
...
-					buffer.append(',');
-
-				String slinfo = entry.getValue();
-				buffer.append(id);
-				buffer.append('@');
-				buffer.append(slinfo);
+			IPluginModelBase model = getLatestModel(id, bundles);
+			if (model != null) {
+				// when present, startInfo from intialBundlesList overrides the one from the config
+				String startInfo = index + 1 < token.length() ? token.substring(index + 1) : allBundles.get(model);
```
HannesWell, line 245: I think the start-info from the intialBundlesList should only overwrite an exsting start-level or autostart value if that value is empty or `"default"`. A value specified in the launch config should take precedence over a default value.

mickaelistria: The current (before this patch) behavior seems to be that if the bundle comes from initialBundlesList, then other startInfo settings are skipped ( https://github.com/eclipse-pde/eclipse.pde/blob/3f9f33643122a8109460e88362593d67fde42ed5/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationHelper.java#L254 excludes processing of bundles coming from initialBundlesList ); so basically it means that initialBundlesList takes higher priority over other settings, ie it can override them.
So I believe this change keeps the current state. Or am I missing 1 case?

HannesWell: Indeed. I didn't see another case. Then lets keep the behavior as it is in this PR (which is also simpler).
Nevertheless I think it would be better to prefer the value from the launch config. But this can be done in a follow up.

HannesWell (approved):

PR #442, opened 2023-01-26
Minor cleanups

akurtakov:


akurtakov:
It's interesting whether tests will still fail if there is change in api-tools.

akurtakov:
Same unrelated failures. Merging.

PR #443, opened 2023-01-26
Move pde.api.tools.annotations to Java-11

HannesWell:
Just noticed that the `pde.api.tools.annotations` plugin still requires Java-1.8. I noticed some error messages in the PDE-build in that regard and therefore wonder if that is related to the new test failures in the api.tools tests.
Lets see what the build tells.


HannesWell:
Does anybody know if there a specifc reason why this was kept at Java 1.8?
Although that this did not fix the failures in the API tests I think that would be a reasonable change unless something important speaks against it.

laeubi:
I think the question is more, are there any advantages? Any new language features to use (I assume not?), because it might be then that such an annotation no longer be used on Java 8 projects anymore (not sure how java handles this when they have only source retention), but that might be a reason for 1.8

Also I think moving a project "just" to a higher EE without using anything from the higher EE is a bit useless, because the point of such requirement is to declare the *lowest* requirement for a bundle so having a higher one has no advantage if it is not used.

akurtakov:
I looked at the log as @HannesWell pointed https://ci.eclipse.org/pde/job/eclipse.pde/job/master/191/consoleFull and there is:
```
13:54:56  [INFO] --- tycho-compiler-plugin:4.0.0-SNAPSHOT:compile (default-compile) @ org.eclipse.pde.api.tools.annotations ---
13:54:56  [INFO] Compiling 5 source files to /home/jenkins/agent/workspace/eclipse.pde_master/apitools/org.eclipse.pde.api.tools.annotations/target/classes
13:54:56  Failed to init Classpath for jar file /opt/tools/java/oracle/jdk-8/1.8.0_202/jre/lib/sunrsasign.jar
13:54:56  java.nio.file.NoSuchFileException: /opt/tools/java/oracle/jdk-8/1.8.0_202/jre/lib/sunrsasign.jar
13:54:56  	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
13:54:56  	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
13:54:56  	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
13:54:56  	at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
13:54:56  	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:148)
13:54:56  	at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
13:54:56  	at java.base/java.nio.file.Files.readAttributes(Files.java:1851)
13:54:56  	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1264)
13:54:56  	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:709)
13:54:56  	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:243)
13:54:56  	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:172)
13:54:56  	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:186)
13:54:56  	at org.eclipse.jdt.internal.compiler.batch.ClasspathJar.initialize(ClasspathJar.java:204)
13:54:56  	at org.eclipse.jdt.internal.compiler.batch.FileSystem.<init>(FileSystem.java:235)
13:54:56  	at org.eclipse.jdt.internal.compiler.batch.Main.getLibraryAccess(Main.java:3498)
13:54:56  	at org.eclipse.jdt.internal.compiler.batch.Main.performCompilation(Main.java:4744)
13:54:56  	at org.eclipse.jdt.internal.compiler.batch.Main.compile(Main.java:1802)
13:54:56  	at org.eclipse.tycho.compiler.jdt.JDTCompiler.compileInProcess(JDTCompiler.java:385)
13:54:56  	at org.eclipse.tycho.compiler.jdt.JDTCompiler.performCompile(JDTCompiler.java:118)
13:54:56  	at copied.org.apache.maven.plugin.AbstractCompilerMojo.execute(AbstractCompilerMojo.java:414)
13:54:56  	at org.eclipse.tycho.compiler.AbstractOsgiCompilerMojo.doCompile(AbstractOsgiCompilerMojo.java:393)
13:54:56  	at org.eclipse.tycho.compiler.AbstractOsgiCompilerMojo.execute(AbstractOsgiCompilerMojo.java:374)
13:54:56  	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
13:54:56  	at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute(MojoExecutor.java:301)
13:54:56  	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:211)
13:54:56  	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:165)
13:54:56  	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:157)
13:54:56  	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:121)
13:54:56  	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:81)
13:54:56  	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:56)
13:54:56  	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:127)
13:54:56  	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:294)
13:54:56  	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:192)
13:54:56  	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:105)
13:54:56  	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:960)
13:54:56  	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:293)
13:54:56  	at org.apache.maven.cli.MavenCli.main(MavenCli.java:196)
13:54:56  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
13:54:56  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
13:54:56  	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
13:54:56  	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
13:54:56  	at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:282)
13:54:56  	at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:225)
13:54:56  	at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode(Launcher.java:406)
13:54:56  	at org.codehaus.plexus.classworlds.launcher.Launcher.main(Launcher.java:347)
```
No idea what and how important this is (maybe even misleading output) but as its 1.8 specific and doesn't happen with 11+ it is simpler to move ahead.

laeubi:
@akurtakov this is just an info output that @iloveeclipse added to latest ECJ so nothing to worry about.

akurtakov:
From releng POV every extra stacktrace in the already huge log troubles me.

laeubi:
> From releng POV every extra stacktrace in the already huge log troubles me.

The you should report this to JDT, this is not limited to Java 8 usage....

akurtakov:
@iloveeclipse so is there any hope for this misleading stacktrace complain to be fixed in JDT, aka not printed?

iloveeclipse:
> @iloveeclipse so is there any hope for this misleading stacktrace complain to be fixed in JDT, aka not printed?

Sure. Create a bug please.

iloveeclipse:
> it might be then that such an annotation no longer be used on Java 8 projects anymore

This *is* the reason, yes. But I assume the solution for those who still use Java 8 & these annotations is to keep on using them by fetching them from the old platform.

So I personally think that is OK to bump the version to at least Java 11.

HannesWell:
Mainly this was a guess/uneducated attempt, if this would help to fix the api-tools test failure.

Since these annotations are put on the project classpath and as Christoph pointed out that we don't use newer language feature I would not mind to keep them at java-8 as long as it does not harm the build of PDE (which it probably doesn't).

akurtakov:
Let's make a decision on this one and be done here as it seems clear what happens if we push it so it's a choice we have to make.. 
I say go for Java 11 .

laeubi:
> Mainly this was a guess/uneducated attempt, if this would help to fix the api-tools test failure.

I must confess I'm a bit reserved about such "guess" changes, actually this change does not fix any problem (neither the API tools nor the unwanted exception) nor does it improves anything (we can't use any newer features or provide better code) and the result will run on Java 11 or any later platform so probably just for uniform bundle shape but then we better drop the `JDK=BREE` thing altogether and simply raise all BREEs without exception.

akurtakov:
I haven't had the time luxury to deep investigate issues in years thus a lot of the things happen to be "reduce matrix to investigate"  (which could be classified as "uneducated guess") and it happens to be the only thing allowing me to still throw things out.

akurtakov:
Seriously this discussion doesn't help with this particular issue and should happen elsewhere. Can we please limit to +/-1 on this PR for the next 24 h and depending on result either merge or abandon .
Here is my +1.

laeubi:
I vote -1 then at the current state...

iloveeclipse:
Usually I would vote with -1 on a change that seem to be unneeded.

However I see following here:
- the annotations were used (mostly) for Eclipse based code. Such code should stay at target platform it uses already, and not use annotations from "later" platform version, so probability of regression is probably low and the fix should be simple.
- mixing JLS 8 and 9+ projects in a platform build in long term just complicates maintenance, alone because everyone who stumbles over that code wonders "why it is like it is".

So I'm with Alex and let move that to Java 11 too.

merks:
Note the investigations here indicate a timing problem causing the tests to fail:

https://github.com/eclipse-pde/eclipse.pde/pull/446

Another timing problem is already fixed (or so it appears):

https://github.com/eclipse-pde/eclipse.pde/commit/442d717f4ca655a976384867a9b8d5b5c745eee0

So I don't think changing this will help with anything.

As such, I vote -1.

HannesWell:
I just checked in depth how the api.tools annotations are used and they are added as additional bundles to a Plug-ins classpath from the current TP via an entry in the `build.properties` like the following (see also https://github.com/eclipse-pde/eclipse.pde/issues/447 to simplify that):
```
additional.bundles = org.eclipse.pde.api.tools.annotations
```
Have to admit that I didn't knew that before, especially since the platform still mainly uses the javadoc-tags (which I think we should change and the good news is a converter for that exists already).
Anyways because the annotation bundle is obtained from the TP and one will probably have a hard time creating Plugins that target a recent Eclipse version and Java below 11 (except for o.e.osgi and EMF almost everything require Java-11 now AFAIK) I think there is no benefit in keeping the BREE for the annotation bundle lower. Those that use Java<11 simply have to use older Eclipse versions.
Therefore, although it probably solves no problem for the sake of simplicity I vote +1.

akurtakov:
So there are 3 "+1"s and 2 "-1"s thus merging.

PR #444, opened 2023-01-27
Fix NPE in TargetPlatformPreferencePage

HannesWell:
The null-case was forgotten in 2f33eea55a4af1fb858a4ce2ac98feb6cd74f723

CC @iloveeclipsesince since you usually discover such bugs quite quickly. :)


iloveeclipse:
> CC @iloveeclipsesince since you usually discover such bugs quite quickly. :)

I can only discover bugs in things I use. Or did I miss that in some review?

HannesWell:
> > CC @iloveeclipsesince since you usually discover such bugs quite quickly. :)
> 
> I can only discover bugs in things I use. Or did I miss that in some review?

No no you didn't miss anything. I encountered that NPE immediately when the target-platform preference in a debugged Eclipse I assumed that it's a bug that is probably encountered quickly and because you are often the first one that discovers such bugs I wanted to let you know so you don't have extra work to create an issue or investigate. :)

PR #445, opened 2023-01-27
Ensure that ApiBuilderTest disables RUN_API_ANALYSIS_AS_JOB

merks:
Also ensure it initializes the ApiTestingEnvironment's target platform.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiBuilderTest.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiBuilderTest.java
@@ -72,7 +72,6 @@ import org.eclipse.pde.ui.tests.util.FreezeMonitor;
 import org.eclipse.ui.dialogs.IOverwriteQuery;
 import org.eclipse.ui.wizards.datatransfer.FileSystemStructureProvider;
 import org.eclipse.ui.wizards.datatransfer.ImportOperation;
-import org.junit.BeforeClass;
 import org.osgi.service.prefs.BackingStoreException;
 
 import junit.framework.Test;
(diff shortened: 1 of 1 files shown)


merks:
@HannesWell @iloveeclipse 

Does this change look okay?  Are the GitHub Actions failures "normal"?

HannesWell:
> @HannesWell @iloveeclipse
> 
> Does this change look okay? Are the GitHub Actions failures "normal"?

The GH workflow failure is not normal and they look like those currently on the master.
Nevertheless this change is very reasonable, thanks for your investigation and for Christoph's hint in https://github.com/eclipse-pde/eclipse.pde/pull/354#issuecomment-1406165679!
I have seen Jenkins pass in the past days sometimes, somay this was just lucky. But nevertheless I think this is an improvement and should be merged.

In general those tests should be migrated to Junit-4 or 5 (when/if that works in the Releng Build), but at least for me that would be a task for later.

iloveeclipse:
> Does this change look okay? Are the GitHub Actions failures "normal"?

Yep, Rein damit!

HannesWell (approved):

akurtakov:
As time consuming as it is moving tests to JUnit 4.x or even 5.x is best step for long term maintenance. Luckily it could be done step by step most of the time. I've been doing that when I look at some test for quite some time but help is more than welcome.

merks:
Yes, only time will tell if the tests pass on Jenkins by luck or for real.

@laeubi  Thanks again for sharing your insight!

@akurtakov  The test extends org.eclipse.jdt.core.tests.builder.BuilderTests so that has to move first..

laeubi:
> The test extends org.eclipse.jdt.core.tests.builder.BuilderTests so that has to move first..

and this extends [TestCase](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/junit/extension/TestCase.java)
and this extend [PerformanceTestCase](https://github.com/eclipse-platform/eclipse.platform.releng/blob/master/bundles/org.eclipse.test.performance/src/org/eclipse/test/performance/PerformanceTestCase.java)
that finally extends (junit3)TestCase :-)

akurtakov:
So many suboptimal and plain wrong things. To start with PerformanceTestCase should never make it to not performance tests to start with.

laeubi:
What offered today by `PerformanceMeter` should today be a JUnit Rule maybe even something build in:
https://junit.org/junit4/javadoc/4.12/org/junit/rules/Stopwatch.html

akurtakov:
I've opened https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/828 for the long term discussion about these perf tests as they have been plain useless for the last couple fo years.

PR #446, opened 2023-01-27
Add print statements to try to figure out what's wrong with the tests.

merks:
This is not intended to ever be committed!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 pipeline {
 	options {
-		timeout(time: 40, unit: 'MINUTES')
+		timeout(time: 240, unit: 'MINUTES')
 		buildDiscarder(logRotator(numToKeepStr:'5'))
 		disableConcurrentBuilds(abortPrevious: true)
 		timestamps()
--- a/apitools/org.eclipse.pde.api.tools.tests/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools.tests/META-INF/MANIFEST.MF
@@ -20,7 +20,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.jdt.core.tests.builder;bundle-version="[3.8.0,4.0.0)",
  org.eclipse.jdt.core.tests.compiler;bundle-version="[3.8.0,4.0.0)",
  org.eclipse.ant.core,
- org.eclipse.pde.ui.tests
+ org.eclipse.pde.ui.tests,
+ org.eclipse.pde.api.tools.annotations
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Export-Package: org.eclipse.pde.api.tools.anttasks.tests,
  org.eclipse.pde.api.tools.apiusescan.tests,
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiBuilderTest.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiBuilderTest.java
@@ -1020,7 +1020,8 @@ public abstract class ApiBuilderTest extends BuilderTests {
 			env = new ApiTestingEnvironment();
 			env.openEmptyWorkspace();
 			env.setAutoBuilding(false);
-			ApiTestingEnvironment.setTargetPlatform();
+		} else {
+			env.waitForAutoBuild();
 		}
 		setBuilderOptions();
 		super.setUp();
@@ -1034,6 +1035,7 @@ public abstract class ApiBuilderTest extends BuilderTests {
 		this.debugRequestor.clearResult();
 		super.tearDown();
 		FreezeMonitor.done();
+		env.waitForAutoBuild();
 	}
 
 	/**
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiTestingEnvironment.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/ApiTestingEnvironment.java
@@ -27,10 +27,13 @@ import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.IJavaModelMarker;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
@@ -45,6 +48,7 @@ import org.eclipse.pde.api.tools.tests.util.ProjectUtils;
 import org.eclipse.pde.core.target.ITargetDefinition;
 import org.eclipse.pde.core.target.ITargetPlatformService;
 import org.eclipse.pde.internal.core.PDECore;
+import org.eclipse.pde.internal.core.PluginModelManager;
 import org.eclipse.pde.internal.core.natures.PDE;
 import org.eclipse.pde.ui.tests.util.TargetPlatformUtil;
 
@@ -748,4 +752,53 @@ public class ApiTestingEnvironment extends TestingEnvironment {
 			TargetPlatformUtil.setRunningPlatformSubSetAsTarget("Current bundles target platform", null); //$NON-NLS-1$
 		}
 	}
+
+	@Override
+	public void cleanBuild() {
+		super.cleanBuild();
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void cleanBuild(String projectName) {
+		super.cleanBuild(projectName);
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void fullBuild() {
+		super.fullBuild();
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void fullBuild(String projectName) {
+		super.fullBuild(projectName);
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void incrementalBuild() {
+		super.incrementalBuild();
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void incrementalBuild(IPath projectPath) {
+		super.incrementalBuild(projectPath);
+		waitForAutoBuild();
+	}
+
+	@Override
+	public void waitForAutoBuild() {
+		super.waitForAutoBuild();
+		try {
+			Job.getJobManager().join(PluginModelManager.class, new NullProgressMonitor());
+			Job.getJobManager().join(ResourcesPlugin.FAMILY_MANUAL_BUILD, new NullProgressMonitor());
+		} catch (OperationCanceledException | InterruptedException e) {
+			e.printStackTrace();
+		}
+		super.waitForAutoBuild();
+	}
+
 }
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/annotations/AnnotationTest.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/annotations/AnnotationTest.java
@@ -259,6 +259,8 @@ public abstract class AnnotationTest extends ApiBuilderTest {
 			// populate the workspace with initial plug-ins/projects
 			createExistingProjects("tagprojects", true, true, false); //$NON-NLS-1$
 		}
+
+		getEnv().waitForAutoBuild();
 	}
 
 	@Override
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/compatibility/BundleMergeSplitTests.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/compatibility/BundleMergeSplitTests.java
@@ -103,7 +103,7 @@ public class BundleMergeSplitTests extends ApiBuilderTest {
 		this.baseline = null;
 		IApiBaseline wsbaseline = manager.getWorkspaceBaseline();
 		if (wsbaseline != null) {
-			wsbaseline.dispose();
+			// wsbaseline.dispose();
 		}
 		for (IProject project : getEnv().getWorkspace().getRoot().getProjects()) {
 			getEnv().removeProject(project.getFullPath());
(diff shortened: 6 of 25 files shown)


iloveeclipse:
> This is not intended to ever be committed!

So may be you can convert this to a "Draft" PR?

merks:
I was going to ask how to convert to a draft, but Google answered that for me.  😛

merks:
@HannesWell @akurtakov 

FYI, there is some type of timing problem where activities on background threads cause failures exactly like we see in the jobs:

image

This only happens when I set breakpoints and stop on them for a while.  Further investigation is required...

merks:
This is the "bad thing" that causes the problem:

image

Even when auto build is off, the following from org.eclipse.core.internal.events.AutoBuildJob.build(boolean) still kicks in:

image

And org.eclipse.pde.api.tools.internal.WorkspaceDeltaProcessor.resourceChanged(IResourceChangeEvent) line 238 is doing cleanAndDisposeWorkspaceBaseline of the project.  There are various "wait for auto build"s already so I need to figure out where best one is needed...

merks:
I see this schedule yet another build job:

image

That leads to the preceding state.  But org.eclipse.pde.internal.core.PluginModelManager.UpdateClasspathsJob has no family so there isn't a good way to wait for it.

laeubi:
> That leads to the preceding state. But org.eclipse.pde.internal.core.PluginModelManager.UpdateClasspathsJob has no family so there isn't a good way to wait for it.

Maybe all ApiTool Jobs should get a JobFamily, then it would even possible to give them a nice icon.

merks:
There is quite a bit of such family testing in PDE:

image

Note that UpdateClasspathsJob has nothing to do with API analysis; it's just a delayed action to prevent UI lockup (looking at the history).

At this point I'm just focused on getting the tests to properly/consistently, not on making PDE better...

HannesWell:
> Even when auto build is off, the following from org.eclipse.core.internal.events.AutoBuildJob.build(boolean) still kicks in:
> [...]
> And org.eclipse.pde.api.tools.internal.WorkspaceDeltaProcessor.resourceChanged(IResourceChangeEvent) line 238 is doing cleanAndDisposeWorkspaceBaseline of the project. There are various "wait for auto build"s already so I need to figure out where best one is needed...

Yes that is really a problem that also hit me multiple times when writing tests and I already spent some time trying to really turn-off auto-builds in PDE-tests but without success.

Maybe we should enhance the platform-core so that disabled autobuilds means really nothing starts to build without explicit calls in the code?

iloveeclipse:
> Maybe we should enhance the platform-core so that disabled autobuilds means really nothing starts to build without explicit calls in the code?

The platform does **not** start a build, it just sends an event. The **clients** can do whatever they want or not do anything on this event, which is what PDE should do in this case.

HannesWell:
> > Maybe we should enhance the platform-core so that disabled autobuilds means really nothing starts to build without explicit calls in the code?
> 
> The platform does **not** start a build, it just sends an event. The **clients** can do whatever they want or not do anything on this event, which is what PDE should do in this case.

Maybe the `org.eclipse.pde.api.tools.internal.WorkspaceDeltaProcessor` should then ignore events issued by incremental builds if auto-building is off (if it is possible to detect that reliably)?

iloveeclipse:
> ignore events issued by incremental builds if auto-building is off (if it is possible to detect that reliably)?

This is what I've meant, I just haven't checked who and where in PDE triggers what exactly on disabled autobuild. And yes, one can reliable know if autobuild is disabled or not.

> incremental builds

There are no incremental builds if autobuild is disabled, at least nothing that is **automatically** triggered by platform. All builds that happen by disabled autobuild can / should be triggered by clients only.

merks:
> > ignore events issued by incremental builds if auto-building is off (if it is possible to detect that reliably)?
> 
> This is what I've meant, I just haven't checked who and where in PDE triggers what exactly on disabled autobuild. And yes, one can reliable know if autobuild is disabled or not.
> 
> > incremental builds
> 
> There are no incremental builds if autobuild is disabled, at least nothing that is **automatically** triggered by platform. All builds that happen by disabled autobuild can / should be triggered by clients only.

@iloveeclipse 

Unfortunately this is not true 😭  as I explained here:

https://github.com/eclipse-pde/eclipse.pde/pull/446#issuecomment-1407318641

The auto builds are disabled but  org.eclipse.core.internal.events.AutoBuildJob.AutoBuildOffJob.run(IProgressMonitor) send PRE_BUILD and POST_BUILD events nevertheless.

@HannesWell 

The current "sledgehammer" changes do ignore PRE_BUILD events that do arrive when auto build is disabled:

image

But, apparently  that too was not enough so there is more going on and now with waits and the end up setup and at the end of teardown the builds take forever such that they get aborted even after I increased the timeout:

https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-446/

But the log no longer has `junit.framework.AssertionFailedError: Missing API component` errors...

I don't know if it's worth increasing the timeout further just to see if the build finishes eventually with fewer failures....

iloveeclipse:
> Unfortunately this is not true  as I explained here:

Once again, this is not a **build**, it is a resource chwnge event.

merks:
Yes it’s a PRE_BUILD event exactly like one gets before a build when auto build is enabled and it cleans up (build?) state such that no API component is found right after that information has been computed, depending on timing.

Unfortunately the tests fail massively for me locally on windows so pretty much a mine field of problems where the behavior changes when you set break points.

merks:
I'm investigating all changes that are happening on threads other than the main thread.  DelayedSnapshotJob does disposes, but I don't think that affects the workspace baseline.  Also I see org.eclipse.pde.api.tools.internal.util.Util.BuildJob occasionally running.

@HannesWell 

Is there anything that argues against requiring the org.eclipse.pde.api.tools.annotations bundle in the org.eclipse.pde.api.tools.tests bundle?  I've been running this test with minimal requirements and this one is missing, causing all kinds of failures that took a long time to track down to this as the cause.  The pom.xml has these:
```
				<configuration>
					<dependency-resolution>
						<extraRequirements>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.osgi.compatibility.state</id>
								<versionRange>0.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-feature</type>
								<id>org.eclipse.jdt</id>
								<versionRange>0.0.0</versionRange>
							</requirement>
							<requirement>
								<type>eclipse-plugin</type>
								<id>org.eclipse.pde.api.tools.annotations</id>
								<versionRange>0.0.0</versionRange>
							</requirement>
						</extraRequirements>
					</dependency-resolution>
				</configuration>
```
Which makes me wonder what else might be missing in my IDE launch....

HannesWell:
> @HannesWell
> 
> Is there anything that argues against requiring the org.eclipse.pde.api.tools.annotations bundle in the org.eclipse.pde.api.tools.tests bundle? I've been running this test with minimal requirements and this one is missing, causing all kinds of failures that took a long time to track down to this as the cause. The pom.xml has these:

No objection from me. In general tests are IMHO far less of a concern since we can change them at any time. And at the moment, at least for this PR I would do any (sledgehammer) change that at least helps to identify the root problem. Once we have that we can still think of a better/simpler/more suitable solution.

> Which makes me wonder what else might be missing in my IDE launch....

The 'Plugin-test' environment created in Tycho and the one created by 'Run as JUnit Plugin Test' unfortunately often too different which makes it harder to reproduce errors locally, if are caused by the environment.
What I sometimes do is to create a launch config using 'Run as JUnit Plugin Test', then abort the run and then modify the config in the dialog to be based on selected bundles (instead of all available once), then select the test-bundle, the SWT fragment for your platform and `org.eclipse.osgi.compatibility.state` and whatever else is specified as `extraRequirements` in the pom. Finally you have to set the checkbox to include missing requirements automatically (maybe check and uncheck it multiple times, because that box is flaky, didn't have the time to fix it yet...).
That should result in a test environment closer to the one running in the build.

In general I think we should enhance the JUnit Plug-in tests and @laeubi suggested a new JUnit OSGi Test a while ago that also includes that but it did not yet happen (and it's another discussion).

laeubi:
You can also run an individual test with

 `mvn clean verify -Pbuild-individual-bundles -Dtest=<fqn of test>#testMethod`

or in (remote) debug mode with

 `mvnDebug clean verify -Pbuild-individual-bundles -Dtest=<fqn of test>#testMethod`

or right click on the `pom.xml` > `Run As` > `Maven Build ...` and then enter as "Goals" `clean verify -Pbuild-individual-bundles -Dtest=<fqn of test>#testMethod`` from within the IDE

merks:
Some of the tests do getWorkspaceBaseline().dispose() but sometimes it's not properly rebuilt by a subsequent test which I found here:

image

merks:
There are fewer failure, and they are different ones...  Progress?  😱

merks:
The good news is I can reproduce exactly the 112 errors locally...

merks:
So this problem is caused by the fact that the enableBaselineOptions is set differently for different tests:

image

And the tests that fail don't actually do a build (for whatever reason, perhaps caused by other hacks I did) and therefore they see the a problem generated during an earlier test that is unexpected if that test runs first.  If I force a build, or run that test all by itself, the problem goes away.

But now I have so many horrible hacks it's hard to know what's fixing what and the performance has degraded horribly.

These tests with so much leakage between them, along with tools changing important state in background jobs,  is really just a minefield of problems happening or waiting to happen.  😭

laeubi:
This is just an idea but the whole purpose for "disabling autobuild" and "Not run in job" was to prevent background jobs from running, but this has obviously failed.

Wont it be then much more better to simply throw away all this attempts, accept the async nature and change the method that sets the baseline to accept a `BaselineListener` the is notified when the baseline is set? The listener can then be passed to the job or use a job listener or whatever and can signal to the caller when it is done and then the test can just wait for that?

Or even better one can use CompletableFuture as a return value, then code can do something like:

```
CompletableFuture<Baseline> baselineFuture = ....;
baselineFuture.thenAcept(baseline-> ... verify what ever ...);
baselineFuture.join(30, TimeUnit.Seconds);
```

merks:
The async stuff is just one of many problems and it has not been the problem blocking me yesterday and today.  Generally waiting for  (all) the jobs has dealt with the async stuff now.  I'm really not sure the extent to which "we" want to add APIs to the API tools just to be used by tests.  Ideally tests just test what's there and ideally the tests would not be so poorly implemented.  The current problem is that a build doesn't even run (sometimes) because all the projects are already in the workspace from a previous test.

What would be ideal (for me) is if someone else did all the work to deal with these horrible tests.  👅

merks:
That big hammer seems to have done the trick:

image

But the previous build ran for 4 hours and aborted:

image

I had it happen once locally that it seemed to wait forever for the jobs.  I'll kick off another build just to see what happens...

iloveeclipse:
> But the previous build ran for 4 hours and aborted:

Most likely because of `JavaModelManager.getIndexManager().waitForIndex(enableIndex, null);` 
The code in question seem to be buggy, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/41

Note, JDT / JDT debug / PDE tests have "dump on freeze" code running, that prints thread dumps on timeout.
However this was not the case here:

https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-446/11/console

```
13:02:48  org.eclipse.pde.genericeditor.extension.tests.Bug531602FormattingTests.testCommentsAndWhitespacePersists  Time elapsed: 0.017 s
13:02:48  
13:02:48  Results:
13:02:48  
13:02:48  Failures: 
13:02:48    UpdateUnitVersionsCommandTests.testVersionSort:55->confirmVersionUpdates:92 ID: org.eclipse.fake.7 has the incorrect version. expected:<[1.0.0.v2]> but was:<[0.0.0]>
13:02:48  
13:02:48  Tests run: 29, Failures: 1, Errors: 0, Skipped: 0
13:02:48  
Cancelling nested steps due to timeout
16:03:29  Sending interrupt signal to process
16:03:30  [INFO] 
16:03:30  [INFO] -----------< org.eclipse.pde:org.eclipse.pde.unittest.junit >-----------
```

Why? Probably some suite misses to setup `org.eclipse.pde.ui.tests.util.FreezeMonitor`?

merks:
Yes, waiting in a loop for the index jobs was exactly what happened locally!

iloveeclipse:
> Probably some suite misses to setup `org.eclipse.pde.ui.tests.util.FreezeMonitor`?

If you seen this locally, can you please check why `org.eclipse.pde.ui.tests.util.FreezeMonitor` was not executed? probably some test suite misses that?

HannesWell:
Great work Ed! Thanks a lot.
A approximately doubled build time is not so nice but at least we have some green check-marks now and can have a look if we can find finer grained solutions. And AFAICT all changes that lead to the current situation are only don in tests aren't they?

What I find interesting is that locally respectively executed as JUnit Plugin Test in the IDE those tests always pass (at least on my computer). And as far as I know it was the same for the others. So unless one puts break-points at the right places like you did Ed, it looks like local builds are always lucky.

> 
> What would be ideal (for me) is if someone else did all the work to deal with these horrible tests. 👅

I think those tests should be reworked to be more stable and to use at least Junit-4 (maybe with 5 in mind) as said before. I can attempt that later (if nobody else did it until then), but currently have other things with higher priority for me.

> > But the previous build ran for 4 hours and aborted:
> 
> Most likely because of `JavaModelManager.getIndexManager().waitForIndex(enableIndex, null);` The code in question seem to be buggy, see [eclipse-jdt/eclipse.jdt.core#41](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/41)
> 
> Note, JDT / JDT debug / PDE tests have "dump on freeze" code running, that prints thread dumps on timeout. However this was not the case here:
> 
> https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-446/11/console


IIRC those issues appeared with the API-tools tests a while ago and there is already a Bugzilla issue for that.

laeubi:
@merks I think technically you are the export for this tests now, from a code POV it seems some of your improvements might already be helpful and could be merged as a separate change (e.g. assigning of a group to the jobs) and probably improve things a bit? This would also ensure that this PR is getting smaller in its context and archived improvements don't get lost later on.

merks:
Yes, I am currently reworking the changes so that they are reduced to the essentials and will have less impact on the build time; I'm finding other inconsistencies as I am doing so.  E.g., there are places the workspace description is modified, but those modification has no effect because the modified description needs to be set back into the workspace.

None of the changes to the underlying framework at this point are needed or useful for the actual code base; I'd like to keep those to a minimum...

So please have patience will I continue to work on this...

akurtakov:
@merks Please submit PRs for these inconsistencies as otherwise next one that looks into test failures will bang his head in the same wall.

merks:
I know we'd all like to get this out of the way sooner rather than later, but the performance is really brutal.  I was trying to simplify the changes today but the tests take forever to run, so I went for a walk and when I got back my machine has overheated and shutdown.  Not conducive to quite progress.

One horrible aspect of these tests is that the setup is expensive and it is repeated for each test case while in the end the setup could happen once (more like @beforeClass and @afterClass).  Often the teardown does not remove projects, so they leak from one test class to the next.

So please have patience; I will not have time for this tomorrow...

laeubi:
I think @akurtakov is not meaning to solve this as a whole, but if you find an inconsistency / isolated thing to change (e.g. as simple as setting a job group, waiting for an auto-group, ...), you should just submit this as a standalone PR, so even if those don't solve the overall story, they are already included.

This then might also show what kind of change leads to longer runs.

merks:
This is superseded by 

https://github.com/eclipse-pde/eclipse.pde/pull/456

which built in 15 minutes without errors and minimizes the number of changes.

PR #448, opened 2023-01-28
Use new static factory method to create UIJobs

HannesWell:
The factory was introduce in https://github.com/eclipse-platform/eclipse.platform.ui/pull/387.


akurtakov:
Some version bumps are needed.

HannesWell:
> Some version bumps are needed.

Done.
Besides that there are only the known 349 test failures, so this is ready for submission.

PR #449, opened 2023-01-28
Simplify implementations of IApiTypeContainer.getPackageNames()

HannesWell:
and ensure the names are always unique and sorted.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/TypeScope.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/TypeScope.java
@@ -13,7 +13,6 @@
  *******************************************************************************/
 package org.eclipse.pde.api.tools.internal.builder;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
```diff
@@ -238,7 +238,7 @@ public String[] getPackageNames() throws CoreException {
 		init();
 		synchronized (this) {
 			if (fPackageNames == null) {
-				fPackageNames = fPackages.keySet().toArray(String[]::new);
```
HannesWell, line 241: @merks in b973ca8f012ff6de044b90485f38685bc3fef8b6 you changed this from sorted to not sorted. Was this intentional?

merks: It's initialized by `fPackages = new TreeMap<>();`.

HannesWell: May bad. Thanks for the clarification! Checked the other implementations as well, but that's the only one where sorting is not necessary.

HannesWell:
Jenkins succeeds and on GH there are only the usual 349 failures in the API-tools tests, thus ready for submission.

PR #450, opened 2023-01-28
Unify Maven build arguments into maven.config and remove unnecessary

HannesWell:
Add a .mvn/maven.config that contains the maven arguments shared in the Jenkins-CI, GitHub workflow and for a local build. This also enables the 'build-individual-bundles' profile for all of the mentioned scenarios, so that one does not have to specify it manually anymore.

The profile 'bree-libs' is never definied in any pom.xml and therefore removed as CLI argument.
The property 'project.build.sourceEncoding=UTF-8' is already defined in the 'eclipse-platform-parent' pom and is therefore unnecessary.

In the Jenkins build additionally define the
properties'maven.test.error.ignore=true' and
'maven.test.failure.ignore=true' in order to only mark the build as unstable if there are test-failures/errors so that it is simpler to distinguish from other failures in the build.

In Jenkins not specifying '-Dmaven.repo.local=$WORKSPACE/.m2/repository' should the PDE-build use the default Maven-cache of the Jenkins container that is persisted so that dependencies are only downloaded once. Or did I understand this wrong or is there another reason to not use that default maven-cache?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,5 @@ Snap.*
 # maven
 /*/*/target/
 .DS_Store
-.polyglot.META-INF
-.polyglot.feature.xml
+.polyglot.*
+pom.tycho
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,13 +36,10 @@ jobs:
         mvn
         --batch-mode
         --global-toolchains ${{ github.workspace }}/.github/toolchains.xml
-        -Pbuild-individual-bundles
         -Pbree-libs
         -Papi-check
         -Dmaven.test.failure.ignore=true
-        -Dcompare-version-with-baselines.skip=false
-        -Dproject.build.sourceEncoding=UTF-8
-        -DtrimStackTrace=false
+        -Dmaven.test.error.ignore=true
         clean verify
     - name: Upload Test Results
       uses: actions/upload-artifact@v3
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,4 @@
+-Pbuild-individual-bundles
+-Dcompare-version-with-baselines.skip=false
+-DtrimStackTrace=false
+--fail-at-end
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,7 @@ pipeline {
 		timestamps()
 	}
 	agent {
-		label "centos-latest"
+		label 'centos-latest'
 	}
 	tools {
 		maven 'apache-maven-latest'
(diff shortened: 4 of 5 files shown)


akurtakov:
Bree-libs is defined at https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/33b1aadedfb1f5058cd4a8922d1b44ac52c8e157/eclipse-platform-parent/pom.xml#L701

HannesWell:
> Bree-libs is defined at https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/33b1aadedfb1f5058cd4a8922d1b44ac52c8e157/eclipse-platform-parent/pom.xml#L701

Thanks. Restored that in Jenkinsfile and GH-workflow. Since not everybody has tool-chains set up, it should not be added to the maven.config to not make local-builds more difficult.

laeubi:
> In Jenkins not specifying '-Dmaven.repo.local=$WORKSPACE/.m2/repository' should the PDE-build use the default Maven-cache of the Jenkins container

It is not safe to share maven caches across different jobs (even though the won't survive anyways as we are using a fresh docker container on each build)

HannesWell:
> > In Jenkins not specifying '-Dmaven.repo.local=$WORKSPACE/.m2/repository' should the PDE-build use the default Maven-cache of the Jenkins container
> 
> It is not safe to share maven caches across different jobs (even though the won't survive anyways as we are using a fresh docker container on each build)

As far as I understood it, but I can definitely be wrong, is that the local .m2 cache in the containers is backed by some EF Maven-cache and is therefore faster than a new/empty cache in each container (like it is now on the master).
And I see some mounting happen in the k8 templates that look like that. But I'm not very familiar with kubernetes and therefore could be wrong?

Why would be using the default provided by the infra team be unsafe? Shouldn't they then change the default to something save?

laeubi:
I can only tell why this is a usual Jenkins configuration for maven builds, but configuration might vary.
If it is made sure that the repo is always only used by one build it is save, otherwise not as Maven do not offer any inter-process synchronization.

HannesWell:
> I can only tell why this is a usual Jenkins configuration for maven builds, but configuration might vary. If it is made sure that the repo is always only used by one build it is save, otherwise not as Maven do not offer any inter-process synchronization.

Understand. So since there is uncertainty in this regard I just restored the `-Dmaven.repo.local=$WORKSPACE/.m2/repository` argument to complete this PR.
If we are later sure that it can be removed (or it is even better to remove it), then we can do that in a later PR.

Nevertheless the build of PDE is still improved. Especially since local builds don't have to specify `-Pbuild-individual-bundles` anymore. Do you others think that is a interesting thing to have for other Eclipsee TLPs (Equinox, Platform, JDT)?

.gitignore
```diff
@@ -10,7 +10,7 @@ Snap.*
 /.project
 
 # maven
-/*/*/target/
+target/
```
merks, line 13: Isn’t this a concern?

https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target

merks (commented):
Not sure ignoring all target folders works.

akurtakov:
> Nevertheless the build of PDE is still improved. Especially since local builds don't have to specify `-Pbuild-individual-bundles` anymore. Do you others think that is a interesting thing to have for other Eclipsee TLPs (Equinox, Platform, JDT)?

In my book - everything that creates less surprises for developers is more than welcome and -Pbuild-individual-bundles is probably the leading hurdle for newcomers. Thus please do it.

.gitignore
HannesWell: Uff. It is indeed a problem. Good catch! Thanks. That would have caused trouble.
Reverted this.

HannesWell:
> > Nevertheless the build of PDE is still improved. Especially since local builds don't have to specify `-Pbuild-individual-bundles` anymore. Do you others think that is a interesting thing to have for other Eclipsee TLPs (Equinox, Platform, JDT)?
> 
> In my book - everything that creates less surprises for developers is more than welcome and -Pbuild-individual-bundles is probably the leading hurdle for newcomers. Thus please do it.

OK, great. I wrote in on my list as well. At the moment I have other things with higher priority on list, which I would like to complete for the spring release (which always comes faster than expected) additionally I'm moving this week and threfore busy with that. But if I find time later I will continue with other projects unless somebody else already did it.

akurtakov:
Let's merge this one.

PR #454, opened 2023-02-02
Standardize name of the compiler in different property sheets

alshamams:
The name of the compiler under the plug-in development property page is appearing differently in different contexts. Since we are already under the plugin development tab, just the name `Compiler` should suffice.

Eclipse preference property page:
image

Quick-fix property page:
image

### DIFF
--- a/ui/org.eclipse.pde.ui/plugin.properties
+++ b/ui/org.eclipse.pde.ui/plugin.properties
@@ -143,7 +143,7 @@ differentiate between binary and source projects.
 
 PropertyPage.pluginDevelopment = Plug-in Development
 PropertyPage.selfHosting = Runtime Classpath
-PropertyPage.compilers = Compiler
+PropertyPage.compilers = Compilers
 Plugin.dependencies.container = Plug-in Dependencies
 
 Plugin.WorkingSet = Plug-ins and Fragments
--- a/ui/org.eclipse.pde.ui/plugin.properties
+++ b/ui/org.eclipse.pde.ui/plugin.properties
@@ -143,7 +143,7 @@ differentiate between binary and source projects.
 
 PropertyPage.pluginDevelopment = Plug-in Development
 PropertyPage.selfHosting = Runtime Classpath
-PropertyPage.compilers = Plug-in Manifest Compiler
+PropertyPage.compilers = Compiler
 Plugin.dependencies.container = Plug-in Dependencies
 
 Plugin.WorkingSet = Plug-ins and Fragments


akurtakov:
Please sign ECA .

merks:
This is just an observation, but it is kind of odd that the plug-in compiler can specify project-specific settings but not the schema compiler.  Also annoying is that the feature compiler cannot have project-specific settings (although I have discovered that you can edit those manually):

image

I could imagine, if anyone cared enough, that schema compiler and feature compiler project-specific preferences (properties) would be made available...

alshamams:
> I could imagine, if anyone cared enough, that schema compiler and feature compiler project-specific preferences (properties) would be made available...

@merks Thanks, I will investigate and see if I can enable those two as well (I noticed the missing compilers but I wasn't sure if it was an issue or just the design).

alshamams:
> @akurtakov Please sign ECA .

@akurtakov I have signed the ECA just now. Thank you.

vik-chand:
There is still a difference - Compiler Vs Compilers in the 2 scenarios. I will suggest to keep the same for both.

vik-chand:
> This is just an observation, but it is kind of odd that the plug-in compiler can specify project-specific settings but not the schema compiler. Also annoying is that the feature compiler cannot have project-specific settings (although I have discovered that you can edit those manually):
> 
> image
> 
> I could imagine, if anyone cared enough, that schema compiler and feature compiler project-specific preferences (properties) would be made available...


I guess manually they can be changed ( and they would be reflected ) - assuming user knows that error is 0, warning 1 etc
	/**
	 * schema preferences
	 */
	public static final String S_CREATE_DOCS = "compilers.s.create-docs"; //$NON-NLS-1$
	public static final String S_DOC_FOLDER = "compilers.s.doc-folder"; //$NON-NLS-1$
	public static final String S_OPEN_TAGS = "compilers.s.open-tags"; //$NON-NLS-1$

	/**
	 * feature preferences
	 */
	public static final String F_UNRESOLVED_PLUGINS = "compilers.f.unresolved-plugins"; //$NON-NLS-1$
	public static final String F_UNRESOLVED_FEATURES = "compilers.f.unresolved-features"; //$NON-NLS-1$

	/**

vik-chand:
> > I could imagine, if anyone cared enough, that schema compiler and feature compiler project-specific preferences (properties) would be made available...
> 
> @merks Thanks, I will investigate and see if I can enable those two as well (I noticed the missing compilers but I wasn't sure if it was an issue or just the design).

I will recommend opening a new issue for this. This again can be a good candidate for new contributors

alshamams:
> There is still a difference - Compiler Vs Compilers in the 2 scenarios. I will suggest to keep the same for both.

@vik-chand Have made the changes. Please review.

merks:
@akurtakov I think we could merge this too.  It's just changes to names in a properties file...

akurtakov:
It's just a string change so go for it.

PR #455, opened 2023-02-02
Enable Api analysis in the new plugins

vik-chand:


akurtakov:
Note that there are more bundles changed in this PR so probably their version bumps are hidden as they are not built due to their deps being broken (aka existing commits).

PR #457, opened 2023-02-03
synchronise both the views with the template selector

gireeshpunathil:
The plug-in template selector has two components - template list, and the help text (FormBrowser). While the first widget follows the toggle button properly, the formbrowser does not. Fix that by toggling between enabled and disabled states following the template selector button.

Before the change: when the template selector is off, the template list is disabled, but the form browser is not.

image

After the change: both the template list as well as the form browser is disabled.
image

image


vogella:
Thanks

PR #463, opened 2023-02-09
Fix PluginRegistryTests

akurtakov:
Instead of using hardcoded version or the bundle version (which should match the one in OSGi container) fetch the version from OSGi and use that version for quering the plugin registry.


akurtakov:
Jenkins build succeeds.

PR #468, opened 2023-02-13
Do not require org.eclipse.osgi.services

laeubi:
Requiring org.eclipse.osgi.services bundle is bad as it actually is only meant to export some osgi packages that can came from everywhere, therefore a bundle should instead import the package of the desired service and not bound to one specific provider (that is planned to be deprecated).

@vik-chand @akurtakov as only test code is affected, maybe something we can still merge as "configuration change" to get down dependency count on  `org.eclipse.osgi.services`?


akurtakov:
As per https://www.eclipse.org/eclipse/development/plans/freeze_plan_4_27.php "Ongoing changes to documentation, tests or examples do not require approval. "  . Aka merging today or even for RC1/2 is fine as long as it doesn't destabilize tests.

laeubi:
Then lets try it out... it could be reverted if really causing any problem quite easy. This is the last (direct) reference I found so far in platform code :+1:

PR #474, opened 2023-02-13
BundleComponent: Always extract into a unique temporary directory

sratz:


merks:
@akurtakov 

Would it be okay to merge this today? The changes look good and it adds a new test to verify that the changes are in fact good....

merks:
FYI, the Windows host seems to have a general infrastructure (Tycho?) problem that also turned up here:

https://github.com/eclipse-pde/eclipse.pde/pull/454

laeubi (approved):
Looks good to me!

akurtakov:
Looks good to me and unless project lead says otherwise I think it's fine to be merged for M3.

laeubi:
@sratz you probably want to squash it into one commit and use the descriptive commit message for this PR as well.

merks:
@HannesWell Is busy completing his move...  Maybe he'll have time for a quick peek...

sratz:
> @sratz you probably want to squash it into one commit and use the descriptive commit message for this PR as well.

I intentionally separated the first cleanup commit from the second fix commit, so I suppose rebase+commit would be the cleanest option.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/BundleComponent.java
```diff
@@ -698,30 +696,21 @@ protected IApiTypeContainer createApiTypeContainer(String path) throws CoreExcep
...
-										FileManager.getManager().recordTempFileRoot(parent.getCanonicalPath());
-									} else {
-										FileManager.getManager().recordTempFileRoot(file.getCanonicalPath());
-									}
+								// also create a temp directory with unique name first,
+								// then extract the file inside that directory.
+								// Without that additional parent directory, this would clash, e.g.
+								// due to pre-existing files in the temp dir or when comparing with
+								// a baseline during API analysis: one file would simply overwrite
+								// the other one, never finding any issues.
```
HannesWell, line 710:
```suggestion
								// Create a uniquely named temp-directory and extract the file inside it.
								// Without that additional parent directory, this could clash, e.g.
								// due to pre-existing files in the temp dir or when comparing with
								// a baseline during API analysis: one file would simply overwrite
								// the other one and thus a difference cannot be found.
```

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/util/Util.java
```diff
@@ -2035,7 +2035,22 @@ public static File getEEDescriptionFile() {
 	 * @since 1.1
 	 */
 	public static File createTempFile(String prefix, String suffix) throws IOException {
-		File file = File.createTempFile(prefix, suffix);
+		File file = Files.createTempFile(prefix, suffix).toFile();
```
HannesWell, line 2038: In order to avoid code duplication I suggest to extract the shared code from the `createTempFile()` and createTempDirectory() method (in the GH UI I can only suggest changes for changed lines, not existing lines. Therefore you probably have to apply the suggestions manually).

Besides that javadoc parameters without any description are worthless and IMHO better removed (or a description is added, but I think that's not necessary here).

All of that (plus the removal of the  since-tag should be done in the first clean-up commit).


```suggestion
	/**
	 * Creates a new file in the users' <code>temp</code> directory
	 *
	 * @return a new temp file
	 * @throws IOException
	 * @since 1.1
	 */
	public static File createTempFile(String prefix, String suffix) throws IOException {
		return recordTempFile(Files.createTempFile(prefix, suffix).toFile());
	}

	/**
	 * Creates a new directory in the users' <code>temp</code> directory
	 *
	 * @return a new temp directory
	 * @throws IOException
	 */
	public static File createTempDirectory(String prefix) throws IOException {
		return recordTempFile(Files.createTempDirectory(prefix).toFile());
	}

	private static File recordTempFile(File file) throws IOException {
		file.deleteOnExit(); 
		FileManager.getManager().recordTempFileRoot(file.getCanonicalPath());
		return file;
	}
```

HannesWell (approved):
The change in general looks good and makes sense with the description given in the Bug-report.
I just have two minor suggestions regarding code-style. I think it would be good to apply them, but it is not critical.

>> @sratz you probably want to squash it into one commit and use the descriptive commit message for this PR as well.
>
> I intentionally separated the first cleanup commit from the second fix commit, so I suppose rebase+commit would be the cleanest option.

That's reasonable 👍🏽

HannesWell (approved):
LGTM.
Thanks @sratz for this PR!

@akurtakov do you want to submit this now or wait until M3 completed?
If we want to have this in 2023-03 I think it is better now than later.

merks:
He alsready approved above:

akurtakov:
And it's in!

merks:
This is evidence of really great teamwork from everyone involved!  🏆

akurtakov:
Unfortunately the new test failed in  I-builds see https://download.eclipse.org/eclipse/downloads/drops4/I20230214-1800/testresults/html/org.eclipse.pde.api.tools.tests_ep427I-unit-cen64-gtk3-java19_linux.gtk.x86_64_19.html :
```
Cannot invoke "org.eclipse.pde.api.tools.internal.provisional.model.IApiBaseline.getApiComponent(String)" because "baseline2" is null

java.lang.NullPointerException: Cannot invoke "org.eclipse.pde.api.tools.internal.provisional.model.IApiBaseline.getApiComponent(String)" because "baseline2" is null
at org.eclipse.pde.api.tools.model.tests.ApiBaselineTests.testIdenticalNestedJarFileNamesAreHandledCorrectly(ApiBaselineTests.java:232)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
at java.base/java.lang.reflect.Method.invoke(Method.java:578)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12) 
```

PR #478, opened 2023-02-16
Spurious API tool error may occur if a new bundle added ( with same package as existing bundle)

vik-chand:
package as existing bundle)


laeubi:
@vik-chand is this something we should merge now new stream is open?

vik-chand:
> @vik-chand is this something we should merge now new stream is open?

thanks, I will merge it tomorrow after updating version

vik-chand:
There are 359 test failures. I am looking at it.

iloveeclipse:
@vik-chand : would you please rebase your branch on master instead of merging master into your branch?

@laeubi : we should add here same git checks like in the platform.

vik-chand:
There were some issues in rebase - Created https://github.com/eclipse-pde/eclipse.pde/pull/520

I was able to recreate the test failures locally

Issue #472, opened 2023-02-13
Move spies to dedicated Feature?


HannesWell:
### Discussed in https://github.com/eclipse-pde/eclipse.pde/discussions/415

<div type='discussions-op-text'>

<sup>Originally posted by **HannesWell** November 28, 2022</sup>
As part of their migration to PDE the E4 spies have been added to the `org.eclipse.pde` feature:
https://github.com/eclipse-pde/eclipse.pde/blob/fc645e8682da8e853f21451e9833cabd9918337e/features/org.eclipse.pde-feature/feature.xml#L162-L216

Consequently everybody that uses the `org.eclipse.pde` feature now gets them installed if it is wanted or not.
Since the spies are not a core functionality of PDE I think it would be good to move them to a dedicated feature (for example `org.eclipse.pde.spies`) so that one can choose to install them in a product or not.

@vogella and @opcoach, you were the main drivers behind the migration, what do you think?</div>

laeubi:
Why is `org.eclipse.tools.layout.spy` part of PDE? The description says " Plug-in Menu Spy" but I have no clue what a "menu plugin" should be?

Beside that it seems reasonable to having a "spy" feature .

HannesWell:
> Why is `org.eclipse.tools.layout.spy` part of PDE? The description says " Plug-in Menu Spy" but I have no clue what a "menu plugin" should be?
> 

No idea. This happened before my time.
Maybe @vogella knows this?

PR #479, opened 2023-02-16
Move PDE spies to dedicated new Feature 'o.e.pde.spies'

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/472

When this is merged we have to make sure that the new feature ends up in the Eclipse-SDK repo.
Can someone tell me where to do this?
The SimRel contribution has to be adjusted too in https://git.eclipse.org/c/simrel/org.eclipse.simrel.build.git/tree/ep.aggrcon

Alternatively we could simply add it to the org.eclipse.sdk feature. Then it should end in both of the mentioned repos automatically. But I'm not sure if this is the right place.

@akurtakov should we consider this for RC1? Or better wait for 4.28?

### DIFF
--- a/features/org.eclipse.pde.spies-feature/feature.xml
+++ b/features/org.eclipse.pde.spies-feature/feature.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature
+      id="org.eclipse.pde.spies"
+      label="%featureName"
+      version="1.0.0.qualifier"
+      provider-name="%providerName"
+      license-feature="org.eclipse.license"
+      license-feature-version="0.0.0">
+
+   <description>
+      %description
+   </description>
+
+   <copyright>
+      %copyright
+   </copyright>
+
+   <license url="%licenseURL">
+      %license
+   </license>
+
+   <plugin
+         id="org.eclipse.tools.layout.spy"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.core"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.model"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.css"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.preferences"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.context"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.bundle"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.pde.spy.event"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+</feature>
--- a/features/org.eclipse.pde-feature/feature.xml
+++ b/features/org.eclipse.pde-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde"
       label="%featureName"
-      version="3.14.1600.qualifier"
+      version="3.15.0.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
@@ -24,7 +24,7 @@
       <import plugin="org.osgi.annotation.versioning"/>
       <import plugin="org.osgi.annotation.bundle"/>
       <import plugin="org.osgi.service.component.annotations"/>
-      <import plugin="org.osgi.service.metatype.annotations" />
+      <import plugin="org.osgi.service.metatype.annotations"/>
    </requires>
 
    <plugin
@@ -154,62 +154,6 @@
 
    <plugin
          id="org.eclipse.pde.genericeditor.extension"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.tools.layout.spy"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.core"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.model"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.css"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.preferences"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.context"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.bundle"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.eclipse.pde.spy.event"
          download-size="0"
          install-size="0"
          version="0.0.0"
--- a/features/org.eclipse.pde.spies-feature/.project
+++ b/features/org.eclipse.pde.spies-feature/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.pde.spies-feature</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.pde.FeatureBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.pde.FeatureNature</nature>
+	</natures>
+</projectDescription>
--- a/features/org.eclipse.pde.spies-feature/.settings/org.eclipse.core.resources.prefs
+++ b/features/org.eclipse.pde.spies-feature/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/features/org.eclipse.pde.spies-feature/.settings/org.eclipse.core.runtime.prefs
+++ b/features/org.eclipse.pde.spies-feature/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
--- a/features/org.eclipse.pde.spies-feature/build.properties
+++ b/features/org.eclipse.pde.spies-feature/build.properties
@@ -1,0 +1,15 @@
+###############################################################################
+# Copyright (c) 2023, 2023 Hannes Wellmann and others.
+#
+# This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License 2.0
+# which accompanies this distribution, and is available at
+# https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#     Hannes Wellmann - initial API and implementation
+###############################################################################
+bin.includes = feature.xml,\
+               feature.properties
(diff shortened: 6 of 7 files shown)


akurtakov:
Let's wait for 4.28. I would rather have lesschurn in RC phases .

features/org.eclipse.pde.spies-feature/feature.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature
+      id="org.eclipse.pde.spies"
+      label="%featureName"
+      version="1.0.0.qualifier"
```
HannesWell, line 5: Alternatively the feature could be named `o.e.pde.spy` and its version could be aligned with the `o.e.pde` feature whose current version is `3.15.0`.

HannesWell:
> Let's wait for 4.28. I would rather have lesschurn in RC phases .

Acknowledge.

HannesWell:
> When this is merged we have to make sure that the new feature ends up in the Eclipse-SDK repo.
> Can someone tell me where to do this?
> The SimRel contribution has to be adjusted too in https://git.eclipse.org/c/simrel/org.eclipse.simrel.build.git/tree/ep.aggrcon

@akurtakov can you tell me where the eclipse-sdk repo is assembled so that we can add the new feature to it?

Furthermore are there more remarks with respect to naming and versioning the new feature?

HannesWell:
> > Can someone tell me where to do this?
> 
> I guess you mean this: https://github.com/eclipse-platform/eclipse.platform.releng/blob/master/features/org.eclipse.sdk/feature.xml

Thanks for the link, but actually I wanted to keep the spies out of the sdk feature, since I don't consider them as such an integral part. I think what I was looking for was
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml

On the other hand, it looks like the o.e.sdk feature gets more other features included so maybe the spies should be added there as well. Do you know if EPP uses the sdk directly or only the included sub-features?

CC @vogella, @opcoach

merks:
The DSL and Modeling EPP packages include org.eclipse.sdk.

opcoach:
The spies are using a lot of different concepts... 

For the model spy it uses modeling, for event spy it is bound to the jdk (because it was possible to click on a class to display the code, but I think we could simplify it as it is broken), the preference spy directly uses eclipse.core,  css spy uses specific css plugins, ...

The goal of the spies is to help the developers to make E4 plugins, this is logical to let the spies in the PDE. And they are not delivered in the final application like the PDE.

HannesWell:
> The goal of the spies is to help the developers to make E4 plugins, this is logical to let the spies in the PDE. And they are not delivered in the final application like the PDE.

So what would be your suggestion then?
My suggestion is to include the pde.spies feature in the eclipse/updates repo as well as in the Eclipse SimRel release repo so that everybody can consume it if desired and to add it to the following EPPs:
- Eclipse IDE for Eclipse Committers
- Eclipse IDE for Java and DSL Developers
- Eclipse Modeling Tools

Basically the packages that include the o.e.sdk plus the committer package.

But it looks like more 'tools' features where added to the o.e.sdk recently so perhaps it would be useful after all to add pde.spies there as well?

opcoach:
The spies must be provided automatically when we develop plugins. Once you have the PDE, you must have the spies without any additional install. Before they were migrated in the PDE they stood in e4.tools and  that was the mess to download them. 
It must be packed with the PDE and in all EPP containing PDE.

merks:
@opcoach Are you suggesting that PDE's feature should include the new spies feature?

opcoach:
Yes !  Like the plugin/manifest editor, spies are needed to develop E4 plugins.

HannesWell:
> @opcoach Are you suggesting that PDE's feature should include the new spies feature?

That would contradict the purpose of this PR.

> Yes ! Like the plugin/manifest editor, spies are needed to develop E4 plugins.

I agree that they are often very useful, but not in every case they are wanted, for example in our use-case of PDE.
So my suggestion as a compromise would be to add them to the o.e.sdk and to those EPP that don't include the SDK but PDE directly.

opcoach:
Here is the use case for the spies :  
When you download one of these available packages : 
- Eclipse IDE for committers
- Eclipse IDE for RCP and RAP developers
- Eclipse Modeling tools

you MUST have the spies directly available without any additional download (as they are delivered with the PDE). 

For all the other packages, the spies could be added manually, if needed, using the eclipse update site containing your feature. For instance, if you develop some pure C++ code, you will not develop plugins and you don't need them. But, if you are using the  "IDE for C/C++ developers" and you need to add manually the PDE feature because you want to develop both C++ code and plugins, you must get the spies included in the PDE. 

Just to remind ... before moving the spies in the PDE, they were managed in another project (e4 tools), and needed to be installed manually. They were delivered on the market place. It took about 6 years to finally accept to have the spies delivered and managed in the PDE.  

So the spy feature should be included in the PDE feature. If you want to provide also outside of the PDE why not, but without the PDE it is not really useful. 

A question : what do you mean by : 'but not in every case there are wanted for example in our use-case of PDE'... ?  When you use the PDE, you develop plugin right ?  So you need editors to develop, and the spies to control the runtime... even for those who still develop E3 legacy plugins, as the spies reflect the E4 runtime content... 

@HannesWell, I apologize if I didn't get well all the purpose of this PR, but I want to be sure that we are not going back in the spy management so let's discuss about it.

HannesWell:
> @HannesWell, I apologize if I didn't get well all the purpose of this PR, but I want to be sure that we are not going back in the spy management so let's discuss about it.

Yes, lets discuss it. It would have been a bit better if we had the discussion before the action, which was the reason I created a 'discussion' in PDE and pinged you and Lars, but unfortunately the only response was one up-vote. Anyway, better now than never. :)


> A question : what do you mean by : 'but not in every case there are wanted for example in our use-case of PDE'... ? When you use the PDE, you develop plugin right ? So you need editors to develop, and the spies to control the runtime... even for those who still develop E3 legacy plugins, as the spies reflect the E4 runtime content...

In our product Eclipse based product users can create Projects that are customized PDE Plugin projects. Therefore we use PDE under the hood, but don't provide all possibilities of PDE to the users and therefore don't need the spies there.
At the moment we just include the PDE feature in our product since that used to contain what we need, but since the last release also the spies are included.
Instead of including the pde-feature I could simply list all desired PDE plugins in another feature of our product, but I assumed that others might be in a similar situation like we are and don't expect the spies to be in their application now. And until now nobody objected. :)

Nevertheless I fully agree that for the EPPs it would be reasonable to include the spies, which is why I suggested to add them to the org.eclipse.sdk and where necessary in the package directly.
In your use case where one installs PDEs from the SimRel repo one would then indeed have to select the spies explicitly.
Personally I find the spies very useful while developing Plugins and Products, but I would not consider them a must have.

In general I find it a bit unfortunate that the spies can only inspect the application they are running in and that therefore one has to explicitly include them into a product/application when starting one from a Dev Eclipse. While at development time spies are helpful I don't think that one always wants to ship them with the product (regardless of if it contains PDE or not).

laeubi:
I can only add how we did it with the m2e-pde feature in the past, it was not included in m2e core feature but added to the relevant EPPs, so if an EPP contains PDE you can just add the spy feature as well, that would be the most flexible approach.

merks:
Do we agree it would be good to include the spies in the Eclipse SDK product?

Looking at these EPP product definitions, I would expect three of them to want to include the spies:

image

But I expect the JEE product would **not** want to include the spies, which I think backs @HannesWell reasoning.  One does not use PDE only to develop Eclipse UI plugins...

@jonahgraham I think it's a simple matter to include the spies feature in the products that want them.  Shall I do that?

laeubi:
>  One does not use PDE only to develop Eclipse UI plugins...

Exactly, that's why I already suggested to more separate those, e.g. some more belong to E4/E3 tooling wile other more belong to some kind of OSGi Tooling where others are specific to Equinox ...

But as A first step I think move them from the pde feature to an own one and add it to the corresponding EPP would keep status-quo and will allow peopel to have more flexible configurations.

merks:
They are there own feature now.  That's the point of these already-merged changes.  😁

laeubi:
> They are there own feature now. That's the point of these already-merged changes. grin

I know, but there where concerns, so I think the only thing needed (after this change) would be to add the feature to all EPPs that currently include PDE + the SDK... so nothing would change, and then one can decide / improve on this later on.

merks:
Yes, that was my conclusion and my suggestion; I offered to do the EPP changes if there is agreement from the project lead. I'm not sure who needs to agree to the SDK changes...

HannesWell:
I would be fine with either way, only include the spies in the EPP packages that use the PDE or o.e.sdk feature, where it is reasonable, or to include into the o.e.sdk feature.
Given that the o.e.sdk feature also contains for example the new ByteCode view I think we could justify to have there spies as well and the o.e.sdk is probably not directly included in third-party products any way.
I would just use the SDK Feature in a TP and select the contained Features for a Product.

Anyway, even if added to the SDK Feature the Spies still have to be added to those EPPs that include PDE directly (If suitable).

HannesWell:
> Yes, I will add them to EPP if @jonahgraham agrees and I think it would be good to add them to the SDK because it doesn't hurt and will be helpful to many...

Thank you. 👍🏽 
I can provide a PR to add them to the SDK this evening, or do you want to do that?
Once they are in the SDK, this can be reverted https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/951

opcoach:
Ok !

 I think everything is clear now and I do agree with your conclusions. 😊. Thank you @HannesWell for your explanation. 

I am really sorry I didn't join the discussion sooner ... Actually I receive every day dozens of GitHub notifications, including those where I am only mentioned.. it is not possible to read everything and I clean them sometimes ... I will try to be better on this and find a better setting... 😊

jonahgraham:
It is ok by me to include in those EPP packages and I encourage you to make such a pr for EPP. 

The contents of the individual packages are decided on by the package maintainers, not the EPP project lead. So please email eppdev or cc the package maintainers in the gerrit so they can have their input.

HannesWell:
> @HannesWell Sorry there was too much to do around the 2023-03 release today for me to do the SDK changes...

No problem, that's not too hard. :)
Just created https://github.com/eclipse-platform/eclipse.platform.releng/pull/212.



> Ok !
> 
> I think everything is clear now and I do agree with your conclusions. 😊. Thank you @HannesWell for your explanation.

You're welcome and I'm glad we got a consensus on this topic. 😊

> 
> I am really sorry I didn't join the discussion sooner ... Actually I receive every day dozens of GitHub notifications, including those where I am only mentioned.. it is not possible to read everything and I clean them sometimes ... I will try to be better on this and find a better setting... 😊

I know that problem and sometimes also loose some pings. I'll just try it more often the next time. 😄

laeubi:
Just if you don't know, you can click the (Un)watch arrow, then choose custom and then select for what events you want to get notified. If someone mention you you will get _always_ notified.

HannesWell:
The new pde.spies feature is in the o.e.sdk. @merks @jonahgraham do you want to add them to the corresponding EPPs or should I create Gerrits (EPP is still on Gerrit, isn't it?)?

merks:
There are a couple of steps that are needed first.  EPP builds from 2023-06 staging, so first that feature needs to be contributed SimRel and because it's new, that can't/won't happen until M1.  We should ensure that we categorize the new feature like we did for the 3 new JDT views.   Then we can modify EPP to use those new features.  All of this is straightforward and I will can help drive this as soon as the platform contributes M1. 

I suppose we could contribute an earlier 4.28-I-build to SimRel to get the process started earlier.  I can't see that this would cause any harm.  I'd want to contribute a specific I-Build.  I don't think any I-build ever get deleted.  Is that correct?

HannesWell:
> We should ensure that we categorize the new feature like we did for the 3 new JDT views. Then we can modify EPP to use those new features. All of this is straightforward and I will can help drive this as soon as the platform contributes M1.

Yes that would be good. Could you point me to where this is done?
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml seems not to be the place since there are no categories.



> I suppose we could contribute an earlier 4.28-I-build to SimRel to get the process started earlier. I can't see that this would cause any harm. I'd want to contribute a specific I-Build. I don't think any I-build ever get deleted. Is that correct?

That would be great. AFAIK they are deleted after/with the release, but not before.

merks:
For the JDT views it was done like this:

https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/200683

It's really not safe to do this type of change textually.  You really must set up a development environment with the appropriate tools installed and use the structured editor for this purpose:

https://ci.eclipse.org/simrel/

And yes, my sense was these just accumulate for the entire release cycle and poof so we could point at one we believe to be relatively stable...

jonahgraham:
> The new pde.spies feature is in the o.e.sdk. @merks @jonahgraham do you want to add them to the corresponding EPPs or should I create Gerrits (EPP is still on Gerrit, isn't it?)?

EPP is still on gerrit - but it [was supposed to change to github today](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2439). Please create gerrit/github PR and I will help by adding the package maintainers to the CC. I don't think it will be controversial, but the contents of the individual packages are up to the maintainers of those packages.

HannesWell:
@merks categorized the new pde.spies feature and contributed it to SimRel, https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1012#issuecomment-1498681260.
Furthermore he also already made the change to include them into the Eclipse committers, modelling and RCP EPPs:
https://github.com/eclipse-packaging/packages/pull/4

Thank you Ed for that (I also had this on my list, but was lazy and wanted to wait for M1 to pass)!

PR #486, opened 2023-02-23
Fix #427 : look for spy parts in all application model and from any window

### LINKED ISSUE #427: Spy handler does not work in new window
**AS IS:**

1) Open any spy, e.g. Window -> Spies -> Preference Spy
2) `E4 Spies window` window (literally) opens and shows "Preference Spy"
3) Open new Eclipse window via Window -> New Window
4) Open any spy, e.g. Window -> Spies -> Preference Spy like before **but from new window**
5) Nothing happens. No  `E4 Spies window` window will be shown.
6) Go back to first window and open same spy there.  `E4 Spies window` window opens and shows as many **multiple** spies views as we tried to open them in the second window
image

**TO BE:**

Spies should work from any window and should not duplicate already existing spy views.

I see the code that opens spies is in `org.eclipse.pde.spy.core.SpyHandler` and in the debugger it seem to find "right" spy window every time, but it seem to fail to find already created spy view and from here on things seem to go wrong. I'm not e4 expert, but looks like this could be something e4 related, may be not PDE specific.

opcoach:
Actually the possible existing spyPart was searched using the **partService.findPart** method, but this method is looking for parts in the current window and not in all the application.  

So if another window is created using Window -> 'New window' the part was not found, and the spy window was not opened. 

On the other hand, if the spy is moved outside the spy window (in a new window or in the application), and closed, it was also not found. 

So I changed the way to find the spy part using **modelService.findElements**. If it is found, it is simply reactivated and bring to top whatever its current location.  If it is not found, it is automatically located in the spy window. 

_Remark_ :  when the spy window is closed with some spies opened inside, it is reopened with only one (the others are not rendered but are still in the partStack). Actually, when the window is closed it seems to use setToBeRendered instead of setVisible on the content.. may be, the spy window could only call setVisible on close. If we want this behavior, open another issue.


opcoach:
A question : what is a problem for the windows build ?

akurtakov:
> A question : what is a problem for the windows build ?

https://github.com/eclipse-tycho/tycho/issues/2149

opcoach:
I merged my branch with master, but there is still the same problem ...

image

opcoach:
And code freeze started today ? ... arg ...

opcoach:
I will deliver it again with a clean rebase... the merge is messy

iloveeclipse:
@opcoach : did sou manage to recreate the PR? 

https://github.com/eclipse-pde/eclipse.pde/issues/427

Issue #362, opened 2022-10-13
Add an action to convert a bundle import into a package import


laeubi:
Currently one can right click on a required bundle and select "Compute Dependency Extends" and then PDE shows all classes that use a package from this dependency.

It would be useful to use that function to provide a new action "Convert to Import-Package" that do the following:

1. Computes all this "Dependency Extends" and filter unique package names
2. Removes the required bundle
3. add all unique packages to package imports.

laeubi:
Yes that's a bit sad but:

- one could use that as an initial way to see what packages might need versioning and improve
- use it for other cases (not all software is platform)

To make this work really good, also annotation of API interfaces with `@ProviderType` / `@Consumer` would be good (or PDE should understand `@noimplement`/ `@noextend` for this regard... so its a bit like a puzzle and we need to find all the missing pieces.

laeubi:
I have now created:
- https://github.com/eclipse-pde/eclipse.pde/issues/491

so we get an easy to use tool to add version exports to bundles...

HannesWell:
> The following comment is **not a critique** of the new feature, it's just an observation. In particular it's relevant to note that the platform pretty much does not version any of its exported packages, so converting a bundle requirement with a version range on one of the Platform's bundles to an package import without version ranges is a loss of information...

That is right, but I think it should not hinder us to add such feature in PDE. IMHO just because something is possible one has to do it everywhere. But since this is a step in the right direction we should do it. Of course it would be nice if all platform packages were already versioned correctly. But we probably have to do that afterwards then (but before we use this new tool in platform).

PR #488, opened 2023-02-24
Add an action to convert a bundle import into a package import

laeubi:
Currently there is no convient way to convert a bundle from using Require-Bundle to Import-Package (what is recommended to use).

This adds a new option in the OrganizeManifest action to do this conversion automatically in the follwoing way:

- Automatic Managed Dependencies are set to using Import-Package
- Current Require-Bundles are converted to Automatic Managed Dependencies
- the package imports are recalculated and updated

Fix https://github.com/eclipse-pde/eclipse.pde/issues/362


laeubi:
## Select the Organize Manifest Action
(Alternatively select some projects and choose `Plugin Tools` > `Organize Manifests...`)
grafik

## Choose the "Recompute Import-Package ..." cleanup
grafik

## The Manifest will be transformed and the result can be reviewed
grafik

PR #489, opened 2023-02-24
Fix #427 : look for spy parts in all application model and from any window

### LINKED ISSUE #427: Spy handler does not work in new window

opcoach:
Recreate pull request after a rebase instead of merge...


laeubi:
@opcoach anything that needs to be done here or should we merge it?

laeubi:
@opcoach @vogella please either finish or discard this PR if it of no interest anymore.

laeubi:
Since no feedback from authors/reviewers I'm closing this now.

Issue #210, opened 2022-07-14
WorkspacePluginModelManager: Comparison method violates its general contract!


laeubi:
I got the following strange exception while editing a component:

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.TimSort.mergeLo(TimSort.java:781)
	at java.base/java.util.TimSort.mergeAt(TimSort.java:518)
	at java.base/java.util.TimSort.mergeForceCollapse(TimSort.java:461)
	at java.base/java.util.TimSort.sort(TimSort.java:254)
	at java.base/java.util.Arrays.sort(Arrays.java:1515)
	at java.base/java.util.ArrayList.sort(ArrayList.java:1749)
	at java.base/java.util.Collections.sort(Collections.java:179)
	at org.eclipse.osgi.internal.module.VersionHashMap.reorder(VersionHashMap.java:87)
	at org.eclipse.osgi.internal.module.ResolverImpl.resolve(ResolverImpl.java:472)
	at org.eclipse.osgi.internal.resolver.StateImpl.resolve(StateImpl.java:602)
	at org.eclipse.osgi.internal.resolver.StateImpl.resolve(StateImpl.java:662)
	at org.eclipse.pde.internal.core.MinimalState.internalResolveState(MinimalState.java:244)
	at org.eclipse.pde.internal.core.MinimalState.resolveState(MinimalState.java:221)
	at org.eclipse.pde.internal.core.PluginModelManager.modelsChangedSynchronized(PluginModelManager.java:298)
	at org.eclipse.pde.internal.core.PluginModelManager.modelsChanged(PluginModelManager.java:244)
	at org.eclipse.pde.internal.core.AbstractModelManager.fireModelProviderEvent(AbstractModelManager.java:37)
	at org.eclipse.pde.internal.core.WorkspaceModelManager.createAndFireEvent(WorkspaceModelManager.java:343)
	at org.eclipse.pde.internal.core.WorkspacePluginModelManager.createAndFireEvent(WorkspacePluginModelManager.java:552)
	at org.eclipse.pde.internal.core.WorkspaceModelManager.processModelChanges(WorkspaceModelManager.java:322)
	at org.eclipse.pde.internal.core.WorkspaceModelManager.processModelChanges(WorkspaceModelManager.java:280)
	at org.eclipse.pde.internal.core.WorkspacePluginModelManager.processModelChanges(WorkspacePluginModelManager.java:541)
[...]

PR #490, opened 2023-02-24
Fix ordering problem in PDEState

laeubi:
Currently the comparator in PDEState vialoates the transitive contract of compare in the follwoing situation:

- There are two bundles of different versions with the same bsn and they both export the same package
- There is another bundle with a different bsn and it also export the package

This leads to the follwoing problem:
The two bundles with the same BSN are always compared with their version, but the other bundle is compared against the two ones using the bundle id.

That leads to the situation that if these three bundles are A, B, C that A > B and B > C but A <= C

This changes the compare function to always compare the version of the supplier, what gives a predictable order of A, B, C and only if BSN is equal AND version is equal finally the bundle id is used to impose a total order.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/210

@tjwatson please review, it seems I can't add you as a reviewer (not committer of PDE?)
@merks I included your change for the path as well
@vik-chand I think we should include this in this release if possible, because this throws a random exception depending on system state, so there are only two cases:

1. It does not happen in that case this is manly a no-op
2. It does happen and in that case the update failed

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
@@ -79,80 +79,12 @@ public class PDEState extends MinimalState {
 		if (resolve) {
 			final String systemBSN = getSystemBundle();
 			fState.getResolver().setSelectionPolicy(new Comparator<BaseDescription>() {
+
 				@Override
-				public int compare(BaseDescription bd1, BaseDescription bd2) {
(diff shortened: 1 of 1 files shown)


vik-chand:
Unless this is a regression, I think we can wait for early 4.28 because we are already past RC1. Also the use case described looks like a corner case. Subject to review by Tom, we can put this in early 4.28 so that we have enough time for testing 
@tjwatson  What do you think?

merks:
FYI, It sounds reasonable to me.

tjwatson:
I would hold off until next release.

PR #495, opened 2023-03-04
toggle framework selection based on project type

gireeshpunathil:
In the new pulg-in project wizard (first page),
when the plug-in project is not Java project
it does not make sense to enable the target
framework (OSGi versus Eclipse) to be select-able,
as those types are relevant only for Java projects.

Make the framework selection dependent on project type.

before:
image

after:
image


laeubi:
@gireeshpunathil a bundle (or especially fragment) can contain only resources without a problem, so why do you think it is only relevant for java projects?

gireeshpunathil:
> @gireeshpunathil a bundle (or especially fragment) can contain only resources without a problem, so why do you think it is only relevant for java projects?

@laeubi - so when I created a project using the wizard using Eclipse framework and OSGi framework (with the Java project type disabled), I see no difference in the generated code, hence I came to this conclusion. When there are only resources in bundles / fragments, does the selection of OSGi / Eclipse make any difference anywhere in the project?

laeubi:
The difference is only the settings, so if you choose "an OSGi framework" and then "standard" this should set a preference so e.g. plugin.xml is not shown in the manifest editor.

If you choose "Eclipse" and then always choose "next" you can create a bundle containing sample help content.

gireeshpunathil:
thanks @laeubi for the clarifications! so this change does not stand. closing.

PR #497, opened 2023-03-07
Update classpath settings: overhaul to fix multiple issues

haubi:
When computing the classpath entry for a library found in the PDE plugin, the result has export=true and no source attachment. If the already existing classpath entry differs in whatever irrelevant parameter, the computed classpath entry is added as well, so Update classpath settings fails with "Build path contains duplicate entry".

Instead, use the library path only to check for the already existing classpath entry.


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -236,7 +236,7 @@ protected static IBuild getBuild(IProject project) throws CoreException {
 		return (buildModel != null) ? buildModel.getBuild() : null;
 	}
 
-	private static void addLibraryEntry(IProject project, IPluginLibrary library, IPath sourceAttachment, IClasspathAttribute[] attrs, ArrayList<IClasspathEntry> result) throws JavaModelException {
+	private static void addLibraryEntry(IProject project, IPluginLibrary library, IPath sourceAttachment, IClasspathAttribute[] attrs, HashSet<IPath> paths, ArrayList<IClasspathEntry> result) throws JavaModelException {
```
HannesWell, line 239:
```suggestion
	private static void addLibraryEntry(IProject project, IPluginLibrary library, IPath sourceAttachment, IClasspathAttribute[] attrs, Set<IPath> paths, ArrayList<IClasspathEntry> result) throws JavaModelException {
```

HannesWell (commented):
Thank you for this contribution. In general it sounds reasonable, but maybe you can add a minimal reproducer to demonstrate the issue, ideally as as a test-case?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
haubi: done.

laeubi:
You need to rebase your change and maybe increase the version of `org.eclipse.pde.core`

haubi:
One more thought: Is the new test found/executed by some magic, or should it be added to some (which?) testsuite?

haubi:
@laeubi Sorry for a probably dumb question, but I'm quite new to Eclipse development and a bit lost:
Since 7ec2f759360c683c7440dff1dd641c1b590352f4 I have problems in my local Eclipse workspace to resolve biz.aQute.bndlib, and my only idea so far is to run Help->Perform Setup Tasks..., but that one fails to find aQute.bnd.osgi 5.5.0.
What am I missing?

laeubi:
This actually should be the way, can you try to Help > Check For Updates and after that execute the setup tasks again?

haubi:
I see - ok, never mind, I'll continue based on an older commit until then.

HannesWell (commented):
> One more thought: Is the new test found/executed by some magic, or should it be added to some (which?) testsuite?

Yes, you have to add your test to `org.eclipse.pde.ui.tests.AllPDEMinimalTests`.

I have two little style remarks below.

HannesWell:
> Sorry. This transition week when 4.27 is almost but not quite released but master is moving beyond to 4.28 is always problematic and isn’t properly resolved until next Wednesday morning when the product and project catalogs are updated.
> 
> I can hack together a local workaround, but not until tomorrow morning.

Wouldn't it be possible to create the 2023-06 Repository List already that points to `https://download.eclipse.org/eclipse/updates/4.28-I-builds/` and use that in the setup?
Currently I my PDE workspace has many errors and I guess it is because of that. :/

@haubi that's the reason I was not yet able to try out your change. Sorry for that delay.

Regarding the test case. What do you think about adding a property to the classpath entry that would not be added by default and then assert in the test that it is not overridden?

merks:
There are sort of three setup problems:   

- The URI from which to install; should be 4.28-Ibuild; not super important if it's wrong, i.e., currently 4.27-IBuild.  
- The URI from which to set the API baseline;  final URI currently unknown, but could use 4.27-Ibuild for the time being; if it's wrong you won't get errors about needing to increment versions relative to the baseline.
- The URI from which to resolve the modular target platform; should be 4.28; if it's wrong and there are new things added to the target and are needed by bundles, resolution will fail.

These are all problems for Platform, PDE, Equinox, and JDT setups, so best solved centrally in the project catalog, but such changes also affect all projects...  Thinking of the least disruptive workaround...

merks:
Here's the simple workaround.   Copy the following text to the clipboard:
```
<xmi:XMI xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:setup="http://www.eclipse.org/oomph/setup/1.0">
  <setup:RedirectionTask
      sourceURL="https://download.eclipse.org/eclipse/updates/4.27-I-builds"
      targetURL="https://download.eclipse.org/eclipse/updates/4.28-I-builds"/>
  <setup:RedirectionTask
      sourceURL="https://download.eclipse.org/eclipse/updates/4.26/R-4.26-202211231800"
      targetURL="https://download.eclipse.org/eclipse/updates/4.27-I-builds"/>
</xmi:XMI>
```

Use Navigate -> Open Setup -> Installation.  Paste onto the root Installation object:

image

Save, and then do Perform Setup Tasks... so that the installation is updated to 4.28, the API baseline to 4.27, and the target platform for 4.28.

If there continue to be API baseline errors (as was the case for me), use the following workaround to "refresh" the baseline in a way that API tools is aware that it needs to do a new build:

image

After the release happens next week, those redirections won't matter anymore before different URIs will be used automatically, including the proper location of the final 4.27 release build, which is currently not present.

haubi:
@merks This has worked out, thanks!

@HannesWell Good idea, 've added checks for some attributes, as well as for computing the classpath from scratch.
Beyond that: Newly added libraries get the 'exported=true' property set in classpath.
To me, this feels like "bad practice": Is this by intention?

haubi:
To get rid of the exported=true by default for binary libs, let me propose two more commits in https://github.com/haubi/eclipse.pde/compare/no-dup-cp-for-lib...new-libs-private-in-cp, which are based on this PR.
Does this make sense? Even as part of this PR?

HannesWell (commented):
I finally had the time to try it out and the change is fine.

While looking at the change in more depth I noticed a few more places that could be polished, but hey are not critical. Please see the comments below.

With the comments below applied this change looks good to me would be ready for submission.

haubi:
Thanks for the improvements, highly appreciated!
Will create another PR regarding the default choice for external once this PR is done.

Since the messy code around sourceAttachment being null contributes to the original duplication problem, I prefer adding another test case and fix as part of this PR.
After all, I don't have any better idea about the use case behind that code other than adding or modifying the source attachment.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathupdater/ClasspathUpdaterTest.java
```diff
@@ -50,6 +57,32 @@ public void testUpdateClasspath_CreateFromScratch() throws Exception {
...
+		IPluginModelBase model = PluginRegistry.findModel(project.getProject());
+
+		IClasspathEntry[] cp = ClasspathComputer.getClasspath(project, model, sourceMap, false, false);
+		jProject.setRawClasspath(cp, null);
+
+		assertClasspathAttribute("JavaSE-17", IClasspathAttribute.MODULE, true, Boolean::valueOf);
+		assertClasspathAttribute("library.jar", IClasspathAttribute.TEST, true, Boolean::valueOf);
+		assertClasspathProperty("library.jar", "exported=true", (e) -> e.isExported());
+		assertClasspathProperty("library.jar", "source",
+				(e) -> "A.jar".equals(nullOr(e.getSourceAttachmentPath(), IPath::lastSegment)));
```
haubi, line 81: Somehow I failed to properly use java Optional here, hence the 'nullOr'. Any better idea from Java experts?

haubi:
Well... one never has enough test cases:
One more situation that could have been in mind behind the original implementation, but also had the duplication problem:
Update an existing classpath entry to find a newly added source attachment, but without overriding that one's location (and without overriding any other attribute, but that eventually was forgotten in the original idea).

haubi:
Again I've no idea why CI build fails, otherwise I've nothing more to add right now.

merks:
It looks like some baseline checking problem:

> 11:40:57  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.pde.spy.core: Baseline and reactor have the same fully qualified version, but different content
> 11:40:57  [ERROR] different
> 11:40:57  [ERROR]    META-INF/ECLIPSE_.RSA: different
> 11:40:57  [ERROR]    META-INF/ECLIPSE_.SF: different
> 11:40:57  [ERROR] -> [Help 2]

haubi:
Well... unfortunately, I'm absolutely clueless about who is supposed to do what about this error.

HannesWell:
> Well... unfortunately, I'm absolutely clueless about who is supposed to do what about this error.

I suspect that the qualifier of the spies Plug-ins has to be bumped because they were not included in the I-builds repo for a few builds, since I have moved them out of the PDE feature and was to slow to add them to the I-builds repo. And therefore not the version from the baseline has been used to replace the artifact and they now have new signatures. A simple qualifier bump is probably sufficient.
I can do that this evening unless somebody else does it earlier. :)

HannesWell:
Created https://github.com/eclipse-pde/eclipse.pde/pull/506 to fix the baseline problem.

haubi:
Never ending story: Changing the _order_ of an existing classpath IMO is a bad idea as well. WDYT?

iloveeclipse:
> Changing the _order_ of an existing classpath IMO is a bad idea as well. WDYT?

For sure. Very bad, because can cause very subtle runtime problems not visible at compile time.

haubi:
For completion, I thought about extending tests to verify that even 'variable'-kind classpath entries are preserved, because they aren't computed based on some PDE level metadata.
To keep it simple, my idea was to use the ECLIPSE_HOME variable, but this seems to fail in the CI build.
Any idea what else I should use here? Must admit that we don't use them in the real world application here.
And: More ideas about what else to test (and how)?

HannesWell:
Sorry I lost track about this PR and after a quick look saw that it has grown a lot.
I'll try to review it as soon as possible, but can't make any promise.

haubi:
> For completion, I thought about extending tests to verify that even 'variable'-kind classpath entries are preserved, because they aren't computed based on some PDE level metadata.

Nah, the original implementation actually does _not_ preserve 'variable'-kind classpath entries... about to revert that one.

haubi:
Since this became an overhaul, if you prefer I won't mind to sqash, maybe to one test and one fix commit.

HannesWell:
> Since this became an overhaul, if you prefer I won't mind to sqash, maybe to one test and one fix commit.

Yes please just squash it, even one commit would be fine for me.

HannesWell:
I started to review your PR and have to admit it is hard the review the mixture of code refactoring and actual behavior changes and fixes.
Would you mind to move the refactoring of the `ClasspathComputer` code that does not change the behavior in a first commit and the actual behavioral changes/fixes in a second commit together with the addition of tests, so that the changes in `ClasspathComputer` are minimal in the second commit? If the refactoring in the first commit is mainly for preparation of the second commit and not valuable alone, that's fine.
Sorry for the back and fourth and late reply!

HannesWell:
@haubi I just moved pde.core and pde.ui to Java-17 in https://github.com/eclipse-pde/eclipse.pde/pull/518 (and plan to move all other PDE Plugins as well soon), so you may want to consider to use features available in Java-17 too.
For example instead of creating instances of `ClasspathComputer` it may be useful to use one or two record classes that contain the data previously passed via numerous method parameters. But I'll leave the decision to you to use records or not, it is just a suggestion.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -57,86 +58,125 @@ public class ClasspathComputer {
...
+		Stream<IClasspathEntry> collecting = Stream.concat(reloaded.stream(), Stream.of(original));
+		// in original order, collect unique entries only, and
+		// prevent them from being collected as new entries below
+		Map<IPath, IClasspathEntry> collectingByKey = collecting
+				.collect(Collectors.toMap(e -> keyOfPath(e.getPath()), e -> e, (first, dupe) -> first));
+		result.addAll(Stream.of(original).map(e -> collectingByKey.remove(keyOfPath(e.getPath())))
+				.filter(e -> e != null).collect(Collectors.toList()));
+		// append new entries in the order they were loaded
+		result.addAll(reloaded.stream().filter(e -> collectingByKey.remove(keyOfPath(e.getPath())) != null)
+				.collect(Collectors.toList()));
```
HannesWell, line 137:
```suggestion
		Arrays.stream(original).map(e -> collectingByKey.remove(keyOfPath(e.getPath())))
				.filter(e -> e != null).forEachOrdered(result::add);
		// append new entries in the order they were loaded
		reloaded.stream().filter(e -> collectingByKey.remove(keyOfPath(e.getPath())) != null)
				.forEachOrdered(result::add);
```
This avoids the creation of intermediate List-objects.

haubi:
> I started to review your PR and have to admit it is hard the review the mixture of code refactoring and actual behavior changes and fixes. Would you mind to move the refactoring of the `ClasspathComputer` code that does not change the behavior in a first commit and the actual behavioral changes/fixes in a second commit together with the addition of tests, so that the changes in `ClasspathComputer` are minimal in the second commit? If the refactoring in the first commit is mainly for preparation of the second commit and not valuable alone, that's fine. Sorry for the back and fourth and late reply!

Wel... before squashing, in the former 8 commits I've tried to implement that one-by-one approach.
However, for the ordering problem I found myself unable to fix without that refactoring, which rendered the smaller fixes obsolete anyway. The alternative would have been to use the 'create from scratch' variant and add an independent merge strategy, much like I did anyway [here in TEA](https://git.eclipse.org/r/plugins/gitiles/tea/tea/+/350459fb8a6b9e4cf26e9f6e66d575a180b5caeb/org.eclipse.tea.library.build/src/org/eclipse/tea/library/build/util/ClasspathUpdater.java#98) for use with existing Eclipse releases.
And for the "behavior change": Actually, it's about fixing bugs only, since even the recent "don't ignore VARIABLE entries" change was identified as bug in discussion #532.
After all, isn't the idea of having tests to ensure proper behavior even across refactoring?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
haubi: done

haubi:
> @haubi I just moved pde.core and pde.ui to Java-17 in #518 (and plan to move all other PDE Plugins as well soon), so you may want to consider to use features available in Java-17 too. For example instead of creating instances of `ClasspathComputer` it may be useful to use one or two record classes that contain the data previously passed via numerous method parameters. But I'll leave the decision to you to use records or not, it is just a suggestion.

Well, I haven't used any Java 17 feature before. Heck, I won't call myself an experienced Java programmer yet.
So unless there's an issue with the class object, I prefer to stop refactoring here.

haubi:
Is org.eclipse.pde.api.tools.tests.ApiToolsPluginTestSuite failing because of my changes?

haubi:
For the MacOS verification build, ClasspathUpdaterTest and others fail: Something seems wrong with the Java setup there.
For the Windows verification build, PluginBasedLaunchTest fails: Exceptions tell "URI is not hierarchical".

I don't think the ClasspathUpdaterTest should be agnostic to the running JRE version, should it be?

HannesWell:
> Wel... before squashing, in the former 8 commits I've tried to implement that one-by-one approach.
> However, for the ordering problem I found myself unable to fix without that refactoring, which rendered the smaller fixes obsolete anyway.

This was one fix per commit, wasn't it? For me and probably others it would be simpler to follow if it is, first refactoring only (with same behavior/bugs as before), then bug-fixing in a small commit. But if that is too much work, I'll review it as it is.

> After all, isn't the idea of having tests to ensure proper behavior even across refactoring?

If they already pass, then yes :)

> Is org.eclipse.pde.api.tools.tests.ApiToolsPluginTestSuite failing because of my changes?

Likely not. We activated Windows and Mac workflows recently and they don't have all failures fixed yet.
At the moment you should only care about Linux and Jenkins. A goog indicator is often the master branch. What fails there can also fail in your PR.

> For the MacOS verification build, ClasspathUpdaterTest and others fail: Something seems wrong with the Java setup there. For the Windows verification build, PluginBasedLaunchTest fails: Exceptions tell "URI is not hierarchical".
> 
> I don't think the ClasspathUpdaterTest should be agnostic to the running JRE version, should it be?

Yes, I'm currently working on this in https://github.com/eclipse-pde/eclipse.pde/pull/541.

haubi:
> > Wel... before squashing, in the former 8 commits I've tried to implement that one-by-one approach.
> > However, for the ordering problem I found myself unable to fix without that refactoring, which rendered the smaller fixes obsolete anyway.
> 
> This was one fix per commit, wasn't it? For me and probably others it would be simpler to follow if it is, first refactoring only (with same behavior/bugs as before), then bug-fixing in a small commit. But if that is too much work, I'll review it as it is.

Exactly: My intention indeed was to have one commit adding one failing test, the next commit to fix the bug being tested for.
But for the sorting problem, fixing without refactoring was not reasonable at all. And refactoring while keeping the bugs alive was not reasonable as well. So I debugged through the original (buggy) code to gain the remaining ideas behind, and put them into the tests. Then I've done the refactoring to actually implement those ideas and get the tests working.

The only thing that might be useful now, if you prefer, is to split the single commit once: add all tests first, do the refactoring second.

haubi:
Any progress possible here now?

HannesWell (commented):
> Any progress possible here now?

Sorry for the delay and thanks for your patience.
I didn't have the time for a exhaustive in-depth review, but the overall direction looks good and many details as well. For now I only have a few minor remarks and questions.
For the next two weeks I will be on vacation, but when back I will have a complete review and testing as soon as possible, so that we hopefully will have this change in RC1 of the next release-cycle. At the moment we are in the freeze period for 2023-06 anyways and the master will re-open in about 7~10 days.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -58,86 +59,125 @@ public class ClasspathComputer {
...
+		List<IClasspathEntry> result = new ArrayList<>();
+		// also collect original entries not explicitly reloaded
+		Stream<IClasspathEntry> collecting = Stream.concat(reloaded.stream(), Stream.of(original));
+		// in original order, collect unique entries only, and
+		// prevent them from being collected as new entries below
+		Map<IPath, IClasspathEntry> collectingByKey = collecting
+				.collect(Collectors.toMap(e -> keyOfPath(e.getPath()), e -> e, (first, dupe) -> first));
+		Arrays.stream(original).map(e -> collectingByKey.remove(keyOfPath(e.getPath()))).filter(e -> e != null)
+				.forEachOrdered(result::add);
+		// append new entries in the order they were loaded
```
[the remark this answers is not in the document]
haubi: I'd prefer to use "reload" like anywhere else, rather than "add" or even "load".

haubi:
Since I'm about to be offline until September, please feel free to reword/fixup/squash/adopt/etc. as you see fit.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -58,86 +59,125 @@ public class ClasspathComputer {
...
+			Map<String, IPath> sourceLibraryMap, boolean clear, boolean overrideCompliance) throws CoreException {
+		ClasspathComputer cpc = new ClasspathComputer(project, model, sourceLibraryMap, clear, overrideCompliance);
+		return cpc.compute();
+	}
+
+	private IClasspathEntry[] compute() throws CoreException {
 
 		// add JRE and set compliance options
 		String ee = getExecutionEnvironment(model.getBundleDescription());
-		result.add(createEntryUsingPreviousEntry(javaProject, ee, PDECore.JRE_CONTAINER_PATH));
```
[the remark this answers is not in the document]
jukzi: internal classes ("internal" in the package name) are not API and can be removed if not used anymore

haubi:
Any chance to get this in (or further reviewed) any time soon?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/AllPDEMinimalTests.java
```diff
@@ -49,7 +50,8 @@
 	AllPDERuntimeTests.class,
 	// ExportBundleTests.class,
 	AllLauncherTests.class, AllLogViewTests.class, ProjectCreationTests.class, BundleRootTests.class,
-	PluginRegistryTestsMinimal.class, ClasspathResolverTest.class, PDESchemaHelperTest.class,
+	PluginRegistryTestsMinimal.class, ClasspathResolverTest.class, ClasspathUpdaterTest.class,
+		PDESchemaHelperTest.class,
```
HannesWell, line 54: This will not be necessary anymore once https://github.com/eclipse-pde/eclipse.pde/pull/745 is merged.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -65,79 +72,86 @@ public static void setClasspath(IProject project, IPluginModelBase model) throws
 
 	public static IClasspathEntry[] getClasspath(IProject project, IPluginModelBase model, Map<String, IPath> sourceLibraryMap, boolean clear, boolean overrideCompliance) throws CoreException {
 		IJavaProject javaProject = JavaCore.create(project);
-		ArrayList<IClasspathEntry> result = new ArrayList<>();
-		IBuild build = getBuild(project);
+		List<IClasspathEntry> originalClasspath = clear ? List.of() : Arrays.asList(javaProject.getRawClasspath());
```
HannesWell, line 75: Shouldn't we call `updateTestAttribute` or something similar on retained original source-entries if the plugin is a test-plugin?
Maybe in `collectInOriginalOrder()`?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -202,28 +216,32 @@ private static IClasspathAttribute[] getClasspathAttributes(IProject project, IP
...
 
-	private static void addSourceFolder(IBuildEntry buildEntry, IProject project, HashSet<IPath> paths,
-			ArrayList<IClasspathEntry> result, boolean isTestPlugin) throws CoreException {
+	private static void addSourceFolders(IBuildEntry buildEntry, ClasspathConfiguration context)
+			throws CoreException {
 		String[] folders = buildEntry.getTokens();
+		IProject project = context.javaProject.getProject();
 		for (String folder : folders) {
 			IPath path = project.getFullPath().append(folder);
-			if (paths.add(path)) {
```
HannesWell, line 210: I assume duplicated tokens are now handled in the very end implicitly since` collectInOriginalOrder()` only uses the first entry with a path?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -237,45 +255,63 @@ protected static IBuild getBuild(IProject project) throws CoreException {
...
-					result.add(oldEntry);
-					return;
-				}
+			// If we have the same binary root but new or different source,
+			// we should recreate the entry, preserving unrelated attributes.
+			if (sourceAttachment == null) {
+				// no override: stick to existing one, if any
+				sourceAttachment = oldEntry.getSourceAttachmentPath();
 			}
+			if (sourceAttachment != null && sourceAttachment.equals(oldEntry.getSourceAttachmentPath())) {
```
HannesWell, line 278: Can't we keep the original code here?
```suggestion
			if ((sourceAttachment == null && oldEntry.getSourceAttachmentPath() != null) || (sourceAttachment != null && sourceAttachment.equals(oldEntry.getSourceAttachmentPath()))) {
```

HannesWell (commented):
Thanks for your patience, I finally had the time to review your change in more detail and in general it looks good to me.

I pushed two (temporary) follow-up commits to stream-line the changes a bit.
The main point is to use a record instead of adding class fields to `ClasspathComputer` and
pass arguments directly where suitable and not too many/far. This reduces the boiler-plate due to the fields. Furthermore I reworked `collectInOriginalOrder()` a bit to make it IMHO a bit simpler, but it the logic should be retained.
What do you think about that?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
haubi: Exactly!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
haubi: If I remember correctly, I have had issues around this original code because it was ignorant against the (new==null && old==null) situation, but this really was because the subsequent code wasn't preserving some attributes.
So indeed, combined with the other fixes, the original code here makes sense again.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
haubi: Isn't the TEST attribute preserved together with any other attribute we preserve anyway?

HannesWell: They are (probably) preserved, but IIRC the old implementation also added them if the project had a Test-Name, but the source entries are not marked as test (e.g. for an old Project).

HannesWell: I had a quick look and pushed another suggestion. Simply updating all src-entries in the end would ensure it is in all cases of old/retained and new entries and whose test-attribute aligns with the project having a test-name or not. And it would also avoid the need to pass the flag if the project has a test-name or not down all the methods.
What do you think?

haubi:
Based on your recent suggestion, now I _set_ the test attribute if advised by the project name matching the 'test_plugin_pattern', but don't _clear_ if not matching the pattern, in order to retain existing test attributes, and added unit tests accordingly.
WDYT?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -108,6 +109,7 @@ private static IClasspathEntry[] collectInOriginalOrder(List<IClasspathEntry> or
 		});
 		// using the order of reloading, append new entries (in the map still)
 		result.addAll(resultingReloadedByPath.values());
+		result.replaceAll(e -> updateTestAttribute(isTestPlugin, e));
```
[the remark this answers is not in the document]
HannesWell: > If I understand correctly, this would remove each TEST attribute from CPE_SOURCE entries when the plugin name isn't matching the pattern, no?

That's correct.

> Shouldn't we just set the TEST attribute when indicated by the pattern, but leave alone when not indicated?

That is something that can be discussed.


> And why does updateTestAttribute() touch CPE_SOURCE entries only, but not CPE_LIBRARY (or similar, if any?) entries referring to bundle-local content?

This probably used to be like this since PDE does not (yet) have the concept of test-dependencies, like for example Maven. IIRC the `TEST` flag was mainly introduced for the Maven use-case where you can have dependencies that should only be available for test-src-folders. For now in PDE the test-src-folders only prevent usage of test-src in non-test-sources.

HannesWell:
> Based on your recent suggestion, now I _set_ the test attribute if advised by the project name matching the 'test_plugin_pattern', but don't _clear_ if not matching the pattern, in order to retain existing test attributes, and added unit tests accordingly. WDYT?

Yes that seems reasonable. It is not always the case that test-plugins/fragments follow the naming convention. In that case one probably does not want to have the TEST-attribute removed.
I assume that it is more likely that one has a 'test'-plugin that does not follow the naming convention than a Plugin that matches accidentially the test-plugin name pattern but is actually not a 'test'-plugin. Until the beginning of this week we even had a test-plugin in PDE that didn't follow the naming convention.

To make a long story short, I'm fine with this.
Please squash the FIXUPs into the two commits (one for the test, one for change) or even into one commit (the test commit and the logic-change commit are disjunct so that would be fine) and then I think this is ready for submission.

HannesWell (approved):
@haubi thanks for this contribution, this is really good work.
And thanks for your patience. 

This change will be integrated in tonight's I-build and consequently will also be part of the Eclipse 2023-12 milestone 1, just in case you want to use it early.

PR #504, opened 2023-03-12
Simplify comparison-implementations in PDEState

HannesWell:


HannesWell:
@laeubi you did some changes in that area recently, do you want to review this PR?

laeubi:
The crucial part is the comparator, so if you made sure that nothing functional changes but only style changes it should be fine because otherwise you will see errors at runtime about unstable sorting.

HannesWell:
> The crucial part is the comparator, 

I know. :)

> so if you made sure that nothing functional changes but only style changes it should be fine 

I tried to do that, but I was hoping for your review to check that. :)

> because otherwise you will see errors at runtime about unstable sorting.

I expected that, can you tell how good the test-coverage in this area is?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
```diff
@@ -106,85 +105,61 @@ public int compare(BaseDescription o1, BaseDescription o2) {
...
-			}
-		}
-		// as a last resort, compare the bundle id...
-		long id1 = s1.getBundleId();
-		long id2 = s2.getBundleId();
-		return (id1 < id2) ? -1 : ((id1 == id2) ? 0 : 1);
+	private Comparator<BaseDescription> createHigherLocalVersionFirstComparator(String systemBSN) {
+		return systemBundlesFirst(systemBSN) //
+				.thenComparing(BaseDescription::getVersion, HIGHER_VERSION) //
+				.thenComparing(BaseDescription::getSupplier,
```
laeubi, line 111: Seems  like now the supplier is (again) first compared, what should not be the case (see comment below)

laeubi (changes requested):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
HannesWell: Just like before the suppliers are first compared by their version, then if they are in the workspace and last by their bundle-id.
But for me it looks like this is currently the case.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
```diff
@@ -106,85 +100,62 @@ public int compare(BaseDescription o1, BaseDescription o2) {
...
+			.thenComparingInt(Version::getMicro) //
+			.thenComparing(PDEState::hasGenericQualifier) // false<true
+			.thenComparing(Version::getQualifier) //
+			.reversed(); //
+
+	private static final Comparator<BundleDescription> IN_WORKSPACE_FIRST = (s1, s2) -> {
+		String n1 = s1.getName();
+		String n2 = s2.getName();
+		// From previous comparison we know that both versions are fully equal
+		if (n1 != null && n1.equals(n2) && !hasGenericQualifier(s1.getVersion())) {
```
HannesWell, line 136: Here is the only difference that I know of. Previously the version of the first BaseDescription was checked if it has a qualifier, but now it uses the version of the supplier for that check. I believe that makes more sense because I have never seen for example an exported package having a qualifier.
Furthermore it would make the comparator more complex.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
laeubi: Okay I was confused here maybe instead of using it inline extract the code below into a field like the other ones,

HannesWell: OK, good. Did that.
Then the body of the method `createHigherLocalVersionFirstComparator()` is simple enough that it can be inlined into the only caller.

vogella:
@laeubi do you still require changes? If yes, what are these?

HannesWell:
> @laeubi do you still require changes? If yes, what are these?

Thanks for the heads up.
@laeubi if you don't object I will merge this tomorrow.

PR #509, opened 2023-03-16
Add support for loading the baseline from a file that list bundles

laeubi:
Currently the ApiAnalysisApplication only accepts atarget that then is resolved and afterwards transformed in a list of files.

This change add support for passing a list of bundles directly to prevent this kind of indirection and save workload if the caller already knows the list of bundles (e.g. Tycho).

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java
@@ -17,6 +17,8 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov would you mind triggering an I-build? Because of how API Tools works today one can only reliable verify/test this with a new ibuild repository.

akurtakov:
Could you trigger one youself ? https://ci.eclipse.org/releng/job/Builds/job/I-build-4.28 . Let me know if you can't so I give you access .

laeubi:
I triggered it know, just not sure if "everyone" can/should trigger an build, but I interpret your answer as you have no concerns :-)

akurtakov:
We should trust each other to the sensible thing - wait a bit if build is close to finish or kill a just started build and etc. but it's a matter of common sense from the committer. Personally I believe that if someone is trusted to push code to the repo that one should be trusted to deal with builds too :smile:

Issue #507, opened 2023-03-15
NPE after loading target platform using Eclipse 4.27


mfvalenta:
After loading our target platform in Eclipse 4.27 RC2, the following NPE happens repeatedly and my java project will not build, making Eclipse 4.27 unusable. The cause appears to be https://github.com/eclipse-pde/eclipse.pde/pull/425.

!ENTRY org.eclipse.ui.ide 2 1 2023-03-15 14:32:42.071
!MESSAGE Problems occurred while trying to save the state of the workbench.
!SUBENTRY 1 org.eclipse.core.resources 2 566 2023-03-15 14:32:42.072
!MESSAGE Problems occurred during save.
!STACK 0
java.lang.NullPointerException: Cannot invoke "java.lang.String.hashCode()" because "this.fBundleName" is null
	at org.eclipse.pde.internal.core.SourceLocationKey.hashCode(SourceLocationKey.java:56)
	at java.base/java.util.HashMap.hash(HashMap.java:338)
	at java.base/java.util.HashMap.put(HashMap.java:610)
	at org.eclipse.pde.internal.core.BundleManifestSourceLocationManager.setTargetBundles(BundleManifestSourceLocationManager.java:282)
	at org.eclipse.pde.internal.core.SourceLocationManager.initializeBundleManifestLocations(SourceLocationManager.java:469)
	at org.eclipse.pde.internal.core.SourceLocationManager.getBundleManifestLocator(SourceLocationManager.java:255)
	at org.eclipse.pde.internal.core.SourceLocationManager.getBundleManifestLocation(SourceLocationManager.java:247)
	at org.eclipse.pde.internal.core.SourceLocationManager.searchBundleManifestLocations(SourceLocationManager.java:340)
	at org.eclipse.pde.internal.core.SourceLocationManager.findSourcePath(SourceLocationManager.java:86)
	at org.eclipse.pde.internal.core.ClasspathUtilCore.getSourceAnnotation(ClasspathUtilCore.java:193)
	at org.eclipse.pde.internal.core.PDEClasspathContainer.addExternalPlugin(PDEClasspathContainer.java:90)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addPlugin(RequiredPluginsClasspathContainer.java:352)
	at org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addDependency(RequiredPluginsClasspathContainer.java:297)
[...]

PR #516, opened 2023-03-18
Fix hasCode / equals to scope will null values in SourceLocationKey

laeubi:
Currently the SourceLocationKey will produce a NPE if the bundle symbolic name is null but the constructor never ensures that it is not null.

This replace the hand written hascode / equals with eclipse generated Java 1.7+ variant that can savly accept nulls.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/507

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationKey.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationKey.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.core;
 
+import java.util.Objects;
 import org.osgi.framework.Version;
 
 /**
@@ -36,25 +37,21 @@ public class SourceLocationKey {
 
 	@Override
 	public boolean equals(Object obj) {
-		if (obj instanceof SourceLocationKey) {
-			SourceLocationKey key = (SourceLocationKey) obj;
-			if (fVersion != null && key.fVersion != null) {
-				return fBundleName.equals(((SourceLocationKey) obj).fBundleName) && fVersion.equals(((SourceLocationKey) obj).fVersion);
-			} else if (fVersion == null && key.fVersion == null) {
-				return fBundleName.equals(((SourceLocationKey) obj).fBundleName);
-			}
+		if (this == obj) {
+			return true;
 		}
-		return false;
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+		SourceLocationKey other = (SourceLocationKey) obj;
+		return Objects.equals(fBundleName, other.fBundleName) && Objects.equals(fVersion, other.fVersion);
 	}
 
 	@Override
 	public int hashCode() {
-		if (fVersion == null) {
-			return fBundleName.hashCode();
-		}
-		int result = 1;
-		result = 31 * result + fBundleName.hashCode();
-		result = 31 * result + fVersion.hashCode();
-		return result;
+		return Objects.hash(fBundleName, fVersion);
 	}
 }


laeubi:
@mfvalenta please check if that fixes the issue for you!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationKey.java
```diff
@@ -36,25 +37,21 @@ public SourceLocationKey(String bundleName) {
...
+		if (this == obj) {
+			return true;
 		}
-		return false;
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
```
HannesWell, line 48:
```suggestion
		if (!(obj instanceof SourceLocationKey)) {
			return false;
		}
```

HannesWell (approved):
LGTM. From the description that should fix the issue.

Just one little style suggestion

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocationKey.java
laeubi: You should then suggest this to JDT as its all auto-generated by eclipse ;-)

merks: I'm trying to imagine in which case line 44 would ever return false?  If the getClass() are == and we are in a SourceLocationKey instance then the obj must be a SourceLocationKey instance as well. Or?

laeubi: Not sure what line 44 you mean (original or from @HannesWell ) but usual pattern generated by eclipse that fulfills the strict equals / hasCode contract of the JDK are:

1. Check if both objects are **the same**
2. Check if the other object is null
3. Check if both objects are of exact same type
4. Compare the actual values

merks: I mean this line seems to me not reachable:

image

laeubi: equals accept an Object, so obj can be for example a String or anything else not of instance `SourceLocationKey`

HannesWell: > 3\. Check if both objects are of exact **same type**

You can configure the code-generator of Eclipse to use `instanceof` instead comparing class identity.

And as far as I know it is not mandated that two objects need to have the exact same type to be considered equals. For example all classes that implement List are considered equal if both objects implement list and they contain the same sequence of elements. And IIRC this is even stated in the java-doc of List.equals().

The object class only has very basic requirements:
```
Indicates whether some other object is "equal to" this one. 

The equals method implements an equivalence relationon non-null object references: 
•It is reflexive: for any non-null reference value x, x.equals(x) should return true. 
•It is symmetric: for any non-null reference values x and y, x.equals(y)should return true if and only if y.equals(x) returns true. 
•It is transitive: for any non-null reference values x, y, and z, if x.equals(y) returns true and y.equals(z) returns true, then x.equals(z) should return true. 
•It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return trueor consistently return false, provided noinformation used in equals comparisons on theobjects is modified. 
•For any non-null reference value x, x.equals(null) should return false. 

An equivalence relation partitions the elements it operates oninto equivalence classes; all the members of anequivalence class are equal to each other. Members of anequivalence class are substitutable for each other, at leastfor some purposes.
```

So I think it is the choice of the designer if objects need to have the exact same type or not, as long as the general rules of `Object.equals` are met.

merks: Silly me.  I didn't take note that you were replacing the "class == class" test.  😝  Never mind.

laeubi: >  For example the objects of all classes that implement List are considered equal if both objects implement list and they contain the same sequence of elements. And IIRC this is even stated in the java-doc of List.equals().

It is not "even stated" it is because the List API defines a different (weaker) contract.

> So I think it is the choice of the designer if objects need to have the exact same type or not, as long as the general rules of Object.equals are met.

If you don't check this, then the rules are not meet. Just take the simple example that A extends B and A checks for instance of A and B checks for instanceof B in the equals method, then a.equals(b) mit return true but b.equals(a) return false!

This can lead to very subtile and hard to track errors, we already see in PDE and platform because equals/hashCode are implemented not carefully.

So my point here still stands, that either you need to define it in the API (what is not the case here), the class must be final (so no subclasses are possible) or beeing very strict in that regard.

HannesWell: > If you don't check this, then the rules are not meet. Just take the simple example that A extends B and A checks for instance of A and B checks for instanceof B in the equals method, then a.equals(b) mit return true but b.equals(a) return false!

That's right. But as you said one should always take care when implementing hashCode/equals. :)
Such cases are difficult anyway, because when you are looking at an instance of class B from an A perspective it looks equals, but it is isn't.

But in this case there are no subclasses and the class is internal, so it would not be an issue (until somebody introduces a subclass).

But completly different suggestion:
Since Platform moves to Java-17 PDE can move as well and this class looks like a record.
So why not convert it into a record and we have this problem solved and much less code? :)

laeubi: Felle fre to do yo I just want to fix the NPE for now ;-)

HannesWell: Done in https://github.com/eclipse-pde/eclipse.pde/pull/518.

PR #517, opened 2023-03-18
Ensure that the encoded profile ID does not exceed the max file length

merks:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng/issues/215


merks:
I took a very conservative approach here such that mostly profiles will continue to have the same ID as before, even for quite long profile names.   Only those profile IDs that would very likely exceed the max file length, and hence didn't work at all before, will have a more strictly shortened ID...

merks:
I'm doubtful the test failures are caused by these changes...

laeubi (approved):
Looks good!

HannesWell:
> I'm doubtful the test failures are caused by these changes...

But as far as I can tell the API-tools tests are pretty stable since you fixed them.

merks:
@HannesWell  So these failures probably are specific to these changes?  The PDE UI tests work very poorly on Windows so generally it's hard to run tests on Windows with any kind of confidence...

HannesWell:
> @HannesWell So these failures probably are specific to these changes? The PDE UI tests work very poorly on Windows so generally it's hard to run tests on Windows with any kind of confidence...

I would have been surprised if it is related and looking at the exact same failures in https://github.com/eclipse-pde/eclipse.pde/pull/516 I think we have again some permanent API-Tools failures. 😞
So the good news is this change is fine.

> The PDE UI tests work very poorly on Windows so generally it's hard to run tests on Windows with any kind of confidence...

Unfortunately yes. I have on my TODO list to modernize those tests and to make them more robust, but not with no-1 priority.

PR #518, opened 2023-03-18
Move o.e.pde.core and o.e.pde.ui to Java-17 and use records

HannesWell:
Steps to migrate:
- Set JRE System Library to Java-17, via the Configure-Build-Path Editor
- Set the BREE to Java-17
- Perform minor version bump


vogella:
Using Records in Eclipse source code is awesome! Thanks for working on this @HannesWell.

Are the test failures related or just the typical random API tool errors?

merks:
These same tests were failing for me to in another PR over the weekend.  Maybe I can find a bit of time to look at them; hopefully they fail the same way locally...

merks:
I did a rebase to see if all the tests pass now.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SourceLocation.java
```diff
@@ -15,47 +15,10 @@
...
-	}
-
-	@Override
-	public String toString() {
-		return path.toOSString();
-	}
-
-	@Override
-	public boolean equals(Object obj) {
-		if (obj instanceof SourceLocation) {
```
merks, line 49: The replaced class ignored userDefined for equality.  I’m not sure that matters though.

HannesWell: Right. 
While implementing this change it seemed that this value is not used. 
So I better remove it, which solves this too.

HannesWell: Done.
But without the `userDefined` field this class actually just wraps an `IPath`. So it can just be replaced by `IPath`.

vogella:
Very nice, I suggest to merge this

vogella:
Not only windows tests are failing https://github.com/eclipse-pde/eclipse.pde/issues/523

HannesWell:
The build used to be green.
But since one or two days more don't matter for this change I would like to fix at least two of the three platforms.
I use a Windows computer so I can investigate the failures there. The not-hierarchical URI error will probably not be too hard to fix.

HannesWell:
I didn't have the time yet to look into the Windows/Mac test yet, but they are also failing on the master branch.
Because the linux build and the Jenkins build succeeded I think this should no longer be hold back. Therefore I'm submitting this now.

HannesWell:
Since (almost?) all other PDE Plugins depend on the two that are already moved to Java-17 I think it makes sense to move all others as well, to ensure that they are all moved the same way.
I plan to do that in the next days.

HannesWell:
> Since (almost?) all other PDE Plugins depend on the two that are already moved to Java-17 I think it makes sense to move all others as well, to ensure that they are all moved the same way. I plan to do that in the next days.

Done in https://github.com/eclipse-pde/eclipse.pde/pull/547.

PR #519, opened 2023-03-20
Fix API tests that fail because they need to support testing Java 17

merks:


HannesWell:
Great! Thank you Ed.
If the tests are reworked ins one day I think the Java Version should be conputed dynamically based on the highest BREE.

laeubi:
Is this something specific to PDE or does other repositories also need updates? I'm asking this because I see API Application failures that mention Java 17 on Windows + Mac (but works on Linux).

merks:
Yes, this was specific this test suite's setup.

Issue #125, opened 2022-05-26
testDelete_projectWithChangedBundleRoot fails intermittently


vik-chand:
```
expected null, but was:<plugin.a>

java.lang.AssertionError: expected null, but was:<plugin.a>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotNull(Assert.java:756)
at org.junit.Assert.assertNull(Assert.java:738)
at org.junit.Assert.assertNull(Assert.java:748)
at org.eclipse.pde.core.tests.internal.WorkspaceModelManagerTest.testDelete_projectWithChangedBundleRoot(WorkspaceModelManagerTest.java:170)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:271)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner
```

It fails on I build page and sometimes on tests while pr validation too.

PR #520, opened 2023-03-20
Spurious API tool error may occur if a new bundle added ( with same

vik-chand:
package as existing bundle)


vik-chand:
Sole test failure is because of https://github.com/eclipse-pde/eclipse.pde/issues/125

vik-chand:
On retrigger of the tests, all tests pass.

Issue #523, opened 2023-03-21
Several test failures under windows because not correctly closed files


laeubi:
```
2023-03-21T12:13:29.7080967Z Contains: Could not delete: D:\a\eclipse.pde\eclipse.pde\ui\org.eclipse.pde.ui.tests\target\work\data\test.headerformatting\META-INF\MANIFEST.MF.
2023-03-21T12:13:29.7081688Z java.nio.file.FileSystemException: D:\a\eclipse.pde\eclipse.pde\ui\org.eclipse.pde.ui.tests\target\work\data\test.headerformatting\META-INF\MANIFEST.MF: The process cannot access the file because it is being used by another process
2023-03-21T12:13:29.7082382Z 	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
2023-03-21T12:13:29.7082786Z 	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
2023-03-21T12:13:29.7083176Z 	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
2023-03-21T12:13:29.7083584Z 	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
2023-03-21T12:13:29.7083999Z 	at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
2023-03-21T12:13:29.7084365Z 	at java.base/java.nio.file.Files.deleteIfExists(Files.java:1191)
2023-03-21T12:13:29.7084719Z 	at org.eclipse.core.internal.filesystem.local.LocalFile.internalDelete(LocalFile.java:222)
2023-03-21T12:13:29.7085144Z 	at org.eclipse.core.internal.filesystem.local.LocalFile.internalDelete(LocalFile.java:250)
2023-03-21T12:13:29.7085535Z 	at org.eclipse.core.internal.filesystem.local.LocalFile.delete(LocalFile.java:139)
2023-03-21T12:13:29.7085912Z 	at org.eclipse.core.internal.localstore.DeleteVisitor.delete(DeleteVisitor.java:66)
2023-03-21T12:13:29.7086292Z 	at org.eclipse.core.internal.localstore.DeleteVisitor.visit(DeleteVisitor.java:156)
[...]

PR #524, opened 2023-03-21
Close streams opened by ProjectCreationTests.testHeaderFormatting()

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/issues/523


laeubi:
It is at least undesired but expected in the sense that no one seem to ever has tested running that on mac... I'll try to find out whats wrong there but have no access to real mac computer to test/debug stuff so its might take some time to find out whats wrong.

iloveeclipse:
OK, jenkins succeeded, I plan to merge despite other checks failing on Mac/Windows, assuming they are unrelated.

laeubi:
> "That" is what exactly?

The API tools Application / api-tools-mojo

> Is there a trick to see concrete test fails?

Once the build is done, it should post a test result here with links that leads to test-failures

> OK, jenkins succeeded, I plan to merge despite other checks failing on Mac/Windows, assuming they are unrelated.

Well Jenkins only runs on Linux, so it is quite expected that windows related failures do not show up there :-)

> If it is not running on Mac, should / can it be disabled on Mac?

If I can't find a solution, this might be the ultimate last resort, but for now the aim is more to fix that and make the failures visible.

PR #528, opened 2023-03-22
Remove compute button and sync checkbox from the feature editor

lppedd:
Two points:

1. `IFeature` has a `computeImports` method, which I didn't touch because that - even if unused in this project - would be a breaking API change.
2. The help content doesn't explain what the _Compute_ button or _Recompute when..._ checkbox do, but the image example is outdated.

Closes: eclipse-pde#37

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/RequiresSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/RequiresSection.java
@@ -18,16 +18,11 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.ui.editor.feature;
 
-import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
-
 import java.util.ArrayList;
 import java.util.Iterator;
 
 import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.Platform;
-import org.eclipse.core.runtime.Plugin;
 import org.eclipse.core.runtime.Status;
-import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.GroupMarker;
 import org.eclipse.jface.action.IMenuManager;
@@ -57,7 +52,6 @@ import org.eclipse.pde.internal.core.ifeature.IFeature;
 import org.eclipse.pde.internal.core.ifeature.IFeatureImport;
 import org.eclipse.pde.internal.core.ifeature.IFeatureModel;
 import org.eclipse.pde.internal.core.ifeature.IFeaturePlugin;
-import org.eclipse.pde.internal.ui.IPDEUIConstants;
 import org.eclipse.pde.internal.ui.PDEPlugin;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
 import org.eclipse.pde.internal.ui.dialogs.FeatureSelectionDialog;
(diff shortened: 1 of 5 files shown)


lppedd:
@HannesWell FYA

lppedd:
Not sure about the failing tests, they seem unrelated to the scope of this change, which is fair since I've also removed UI elements.

HannesWell:
Thanks @lppedd, this looks good on the first sight, I'll review it in detail in the next days.

> Not sure about the failing tests, they seem unrelated to the scope of this change, which is fair since I've only removed UI elements.

That's probably right. We recently added workflows on more platforms and still have to fix some tests, so you don't have to worry about that.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/RequiresSection.java
```diff
@@ -89,7 +84,7 @@ public Object[] getElements(Object parent) {
 	}
 
 	public RequiresSection(FeatureDependenciesPage page, Composite parent) {
-		super(page, parent, Section.DESCRIPTION, new String[] {PDEUIMessages.FeatureEditor_RequiresSection_plugin, PDEUIMessages.FeatureEditor_RequiresSection_feature, PDEUIMessages.FeatureEditor_RequiresSection_remove, PDEUIMessages.FeatureEditor_RequiresSection_compute});
+		super(page, parent, Section.DESCRIPTION, new String[] {PDEUIMessages.FeatureEditor_RequiresSection_plugin, PDEUIMessages.FeatureEditor_RequiresSection_feature, PDEUIMessages.FeatureEditor_RequiresSection_remove});
 		getSection().setText(PDEUIMessages.FeatureEditor_RequiresSection_title);
 		getSection().setDescription(PDEUIMessages.FeatureEditor_RequiresSection_desc);
```
HannesWell, line 89: I think the value of the property `FeatureEditor_RequiresSection_desc` should be rephrased to something like:
```
List Plug-ins and Features that will need to be present before installing this feature.
```

HannesWell (changes requested):
> Two points:
> 
>     1. `IFeature` has a `computeImports` method, which I didn't touch because that - even if unused in this project - would be a breaking API change.

`IFeature` is not a public API, it is just an internal Interface so you can just delete the `computeImports` method and its implementation in the `Feature` class. You can then also delete all methods that are only called by the implementation. But you will probably find that out quickly but simply deleting the implementation. Most called methods are private so you should get an error/warning if they are not called anymore.

> 
>     2. The help content doesn't explain what the _Compute_ button or _Recompute when..._ checkbox do, but the image example is outdated.

Unfortunatly I cannot find the help for that part of the Editor quickly (🙄), but the doc for PDE lives in https://github.com/eclipse-platform/eclipse.platform.common/tree/master/bundles/org.eclipse.pde.doc.user.

If you want to update that as well, it would be great. :)

HannesWell:
> @HannesWell @lppedd would be great to finish this today so it can go into M1

The M1 freeze period already started yesterday at 12pm.
@lppedd nevertheless it would great to have this for M2. :)

lppedd:
@HannesWell @laeubi I'm busy as of now but this will be done before the end of the week.

vogella:
> @HannesWell @laeubi I'm busy as of now but this will be done before the end of the week.

Thanks, end of week is approaching, would be nice to have this in.

HannesWell (approved):
@lppedd I have applied the suggested changes and pushed them to your branch.

Since this overall not a big change I plan to merge this as soon as the build succeeds so that we can have it just in time for the 2023-06 release.

HannesWell:
@lppedd thanks for this PR.

PR #508, opened 2023-03-16
autocomplete for build.properties

gireeshpunathil:
A rudimentary implementation of the auto-complete feature for the build.properties file editor for features, plugins and fragments. Searches and auto-fills valid property names.

Pending items would be to handle corner cases (if any), provide a help text, provide potential suggestions for the values. But I prefer to handle one item at a time - I am still learning things!

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/503

PR #526, opened 2023-03-22
Api tools clean up

HannesWell:
Apply "Convert to enhanced for' loops" clean-up to API-tools and subsequently in another commit improve the code manually where potential for improvements was noticed.

PR #529, opened 2023-03-23
autocomplete for build.properties: follow-up

gireeshpunathil:
a few review comments are included in this follow-up PR:
 - removed inheriting from ICompletionListener
 - hence removed completion listener registration (ICompletionListener.addCompletionListener)
 - usage of array creation shortcut
 - better string initialization
 - localization of fContentAssistantProcessor variable

Refs: https://github.com/eclipse-pde/eclipse.pde/pull/508

/cc @HannesWell

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildSourceViewerConfiguration.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildSourceViewerConfiguration.java
@@ -43,7 +43,6 @@ public class BuildSourceViewerConfiguration extends ChangeAwareSourceViewerConfi
 	private BasePDEScanner fPropertyValueScanner;
 
 	private ContentAssistant fContentAssistant;
-	private BuildPropertiesContentAssistProcessor fContentAssistantProcessor;
 
 	private abstract class AbstractJavaScanner extends BasePDEScanner {
 
@@ -241,6 +240,7 @@ public class BuildSourceViewerConfiguration extends ChangeAwareSourceViewerConfi
 
 	@Override
 	public IContentAssistant getContentAssistant(ISourceViewer sourceViewer) {
+		BuildPropertiesContentAssistProcessor fContentAssistantProcessor;
 		if (fSourcePage != null && fSourcePage.isEditable()) {
 			if (fContentAssistant == null) {
 				// Initialize in SWT thread before using in background thread:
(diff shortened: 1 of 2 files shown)


vogella:
@HannesWell please check if you review comments have been correctly applied. Thanks @gireeshpunathil 
 for the fast follow-up.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildSourceViewerConfiguration.java
```diff
@@ -241,6 +240,7 @@ public boolean affectsColorPresentation(PropertyChangeEvent event) {
 
 	@Override
 	public IContentAssistant getContentAssistant(ISourceViewer sourceViewer) {
+		BuildPropertiesContentAssistProcessor fContentAssistantProcessor;
```
HannesWell, line 243: This can be moved down to the only location where the variable is assigned.

HannesWell (approved):
Looks good. Thanks you. :)

Just one minimal suggestion below.
Personally I like to have the Sonar-Lint Plugin installed for static code analysis. Although it is sometimes pedantic, in my opinion its suggestions often lead to better to understand code (and sometimes even prevents bugs). So I can just recommend to install it.


@vogella I somehow thought that there would be a quickfix to merge split variable declaration and initialization, but I couldn't found it. Do you know if there is one?

In the API-Tools I saw it very often, that variables are declared in the beginning of a method but initialized much later and for example re-used over multiple loop.
IMHO this is bad style since it makes the scope of variables larger than necessary, which makes it harder to understand and has the potential to introduce bugs (see https://github.com/eclipse-pde/eclipse.pde/pull/526 for an potential example ).

vogella:
I think you were looking for override assignemt with Move declaration if necessary.

image

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/build/BuildSourceViewerConfiguration.java
gireeshpunathil: @HannesWell - thanks, brought it to the lowest possible point, and converted to a `var style declaration` as per the `quick-fix` suggestion that appeared in the context.

gireeshpunathil:
thanks @HannesWell and @vogella for the pointers and suggestions.

I installed sonar-lint, but not sure what effect it made - there wasn't any pro-active suggestion by the plug-in at the declaration site. when I checked the rules,  I see only these:

image

however, when I brought down the declaration to the initialisation site, the editor suggested a quick-fix to use `var` to take advantage of the type inference at the initialisation context. Not sure if this is coming from the plug-in or the JDT core itself.

I also notice that the preferences showed up by the plug-in is fundamentally different from the screen-shot shared by @vogella . Are we talking about two different things here? Am I missing something?

I am still learning!

HannesWell (approved):
Looks good :)
Thank you.

HannesWell:
> I think you were looking for override assignemt with Move declaration if necessary.

Thank you! that is what I was looking for. But it looks like that it does not work for Variables used in loops.


>I also notice that the preferences showed up by the plug-in is fundamentally different from the screen-shot shared by @vogella . Are we talking about two different things here? Am I missing something?

Yes sorry that was confusing. What I asked Lars for is a build in Source Clean up action provided by JDT. Sonar Lint on the other hand is a static code analyzer tool that shows you blue markers in your code where ever it finds issues.
For example in the APIUsePatternTab:

grafik

Some suggestions are very useful, some are not suitable in the cases here. So while it definitely makes sense to look at classes with many linter issues, it is not necessary the goal to get completely rid of them. One always has to decide if the suggesting from Sonar-Lint is suitable the location.

PR #530, opened 2023-03-23
Delete the temporary file after being parsed

laeubi:
Currently one gets a lots of lingering files named:

```
/tmp/eed17021414477589663537.ee
/tmp/eed15599516348791254164.ee
/tmp/eed10981864096532456217.ee
/tmp/eed9998009589164187122.ee
/tmp/eed2679993409650199583.ee
```

in the temp directory of the system even though after the file is read it is never used anywhere.


HannesWell (approved):
Looks sane.

Actually the `FileManager` of the API-Tools should delete this file eventually.
However it might be the the case that it is not working in Tycho or is too late in cleaning up.

In general I think that there should be a way provided by JDT to create a ExecutionEnvironmentDescription from a Map of properties, or even better from a given VM(Installation) path and a selected EE.

I'll create a PR in the next days.

laeubi:
Yes they are deleted on shutdown of the application, but still the accumulate a lot (not sure if they are created multiple times?!) and as they are not needed afterwards I think there is no need to wait until that point.

Issue #480, opened 2023-02-16
[mac][win] URI is not hierarchical / Intermittent failure in 1 platform (testTwoVersionsOfSameBundleBundlesInfo,testTwoVersionsOfSameBundleConfigIni)


vik-chand:
org.eclipse.pde.ui.tests.launcher.PluginBasedLaunchTest | testTwoVersionsOfSameBundleBundlesInfo | Error | URI is not hierarchical                                                                           java.lang.IllegalArgumentException: URI is not hierarchical                                                  	at java.base/java.io.File.<init>(File.java:420)                                                  	at  org.eclipse.pde.ui.tests.launcher.PluginBasedLaunchTest.testTwoVersionsOfSameBundleBundlesInfo(PluginBasedLaunchTest.java:883)                                                  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)                                                  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)                                                  	at  java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                                  	at java.base/java.lang.reflect.Method.invoke(Method.java:568)                                                  	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)                                                  	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)                                                  	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)                                                  	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)                                                  	at

laeubi:
This can also be seen here:
https://github.com/eclipse-pde/eclipse.pde/runs/12359056671

Issue #494, opened 2023-03-03
testProducts_productAddedToDeriveFolderNestedInSrcFolder fails intermittently


vik-chand:
r | Cannot invoke "java.util.Collection.size()" because "products" is null                                                                           java.lang.NullPointerException: Cannot invoke "java.util.Collection.size()" because "products" is null                                                  	at  org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.assertSingleProductWithId(WorkspaceProductModelManagerTest.java:151)                                                  	at  org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.testProducts_productAddedToDeriveFolderNestedInSrcFolder(WorkspaceProductModelManagerTest.java:101)                                                  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)                                                  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)                                                  	at  java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                                  	at java.base/java.lang.reflect.Method.invoke(Method.java:568)                                                  	at org.junit.runners.mode
-- | --

Issue #375, opened 2022-11-03
Random failing org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.testProducts_productAddedToDerivedSourceFolder


laeubi:
The test [org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.testProducts_productAddedToDerivedSourceFolder](https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-374/2/testReport/junit/org.eclipse.pde.core.tests.internal/WorkspaceProductModelManagerTest/testProducts_productAddedToDerivedSourceFolder/ )randomly fails with:

```
java.lang.NullPointerException: Cannot invoke "java.util.Collection.size()" because "products" is null
	at org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.assertSingleProductWithId(WorkspaceProductModelManagerTest.java:151)
	at org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.testProducts_productAddedToDerivedSourceFolder(WorkspaceProductModelManagerTest.java:87)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:271)
```

Issue #359, opened 2022-10-12
WorkspaceProductModelManagerTest fails on Jenkins


iloveeclipse:
Test failed with build on unrelated changes in #358.

See https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-358/1/testReport/junit/org.eclipse.pde.core.tests.internal/WorkspaceProductModelManagerTest/testProducts_productAddedToDeriveFolderNestedInSrcFolder/

```
java.lang.NullPointerException: Cannot invoke "java.util.Collection.size()" because "products" is null
	at org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.assertSingleProductWithId(WorkspaceProductModelManagerTest.java:151)
	at org.eclipse.pde.core.tests.internal.WorkspaceProductModelManagerTest.testProducts_productAddedToDeriveFolderNestedInSrcFolder(WorkspaceProductModelManagerTest.java:101)
```

PR #541, opened 2023-03-29
Fix 'URI is not hierarchical' failure in PluginBasedLaunchTest

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/480.

In PluginBasedLaunchTest the test cases
- testTwoVersionsOfSameBundleBundlesInfo
- testTwoVersionsOfSameBundleConfigIni

failed on Windows (and probably similar on Mac) because the URL/URI value of the '-configuration' program argument is not correctly build. On windows it is something similiar to 'file:C:/path1/path2/' while it should be 'file:///C:/path1/path2/'.
This makes the test case resilient against this kind of malformed URL/URIs.

It is great that we now build on all platforms, which should prevent such failures in the future.

This also reminds me about https://github.com/eclipse-platform/eclipse.platform/issues/35.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/PluginBasedLaunchTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/PluginBasedLaunchTest.java
@@ -19,20 +19,23 @@ import static org.eclipse.pde.ui.tests.launcher.FeatureBasedLaunchTest.concat;
 import static org.eclipse.pde.ui.tests.launcher.FeatureBasedLaunchTest.toDefaultStartData;
 import static org.eclipse.pde.ui.tests.util.TargetPlatformUtil.bundle;
 import static org.eclipse.pde.ui.tests.util.TargetPlatformUtil.resolution;
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 import static org.osgi.framework.Constants.REQUIRE_BUNDLE;
 import static org.osgi.framework.Constants.RESOLUTION_OPTIONAL;
 
 import java.io.File;
-import java.io.FileInputStream;
-import java.net.URI;
+import java.io.InputStream;
+import java.net.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.debug.core.*;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.target.NameVersionDescriptor;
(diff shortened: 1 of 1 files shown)


HannesWell:
Let's see if this also fixes the intermittent test failures in `Workspace(Product)ModelManagerTest`.
This would be a simple lucky punch that fixes multiple open issues. Locally on my computer the test succeeded like before.

HannesWell:
> Let's see if this also fixes the intermittent test failures in `Workspace(Product)ModelManagerTest`. This would be a simple lucky punch that fixes multiple open issues. Locally on my computer the test succeeded like before.

There was now at least one successful pass. In case the change in this PR is really a fix this would fix the following issues:

- https://github.com/eclipse-pde/eclipse.pde/issues/21
- https://github.com/eclipse-pde/eclipse.pde/issues/125
- https://github.com/eclipse-pde/eclipse.pde/issues/255
- https://github.com/eclipse-pde/eclipse.pde/issues/300
- https://github.com/eclipse-pde/eclipse.pde/issues/375
- https://github.com/eclipse-pde/eclipse.pde/issues/359
- https://github.com/eclipse-pde/eclipse.pde/issues/494

laeubi:
I don't see any of the "not hirachical" anymore so this this seems to work an we can merge this?!

HannesWell:
> I don't see any of the "not hirachical" anymore so this this seems to work an we can merge this?!

I also added some more commits to investigate the test-failure in that class in the mac workflow that have to be adjusted. So this PR is not read for merge now. I will complete it this evening.

HannesWell:
I now moved the part that aims to fix the WorkspaceManagerTests into https://github.com/eclipse-pde/eclipse.pde/pull/544 and focus in this PR on fixing `PluginBasedLaunchTest`.

HannesWell:
Looks like there is an issue on Mac to detect the available VMInstalls. Running the folllowing snippted:
```
System.out.println("Install-types");
for (IVMInstallType installType : JavaRuntime.getVMInstallTypes()) {
	System.out.println("Name: " + installType.getName());
	System.out.println("id: " + installType.getId());
	for (IVMInstall vmInstall : installType.getVMInstalls()) {
		System.out.println("VM loc: " + vmInstall.getInstallLocation());
	}
}
```
Produces the following console output on Mac:
```
Install-types
Name: Standard VM
id: org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType
Name: Execution Environment Description
id: org.eclipse.jdt.launching.EEVMType
Name: Standard 1.1.x VM
id: org.eclipse.jdt.launching.Standard11xVMType
```
While for example on Linux it is:
```
Install-types
Name: Execution Environment Description
id: org.eclipse.jdt.launching.EEVMType
Name: Standard VM
id: org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType
VM loc: /usr/lib/jvm/temurin-17-jdk-amd64
Name: Standard 1.1.x VM
id: org.eclipse.jdt.launching.Standard11xVMType
```

And I'm relatively sure that this is the reason for the failure of the `PluginBasedLaunchTest` on Mac.
Of course it could be worked around somehow, but the problem seem to be fundamental to me.
@iloveeclipse can you tell if something has changed in JDT in this regard in the past? Or is this a set up mistake in the mac-runners?
Can anybody with access to a Mac computer debug that test case and tell if there is any issue?

HannesWell:
For now I'll leave the mac-failures out of this PR and just stick with fixing the originally intended test-case. That's already something.
The mac-failures can be fixed in another PR if somebody can tell how.

laeubi:
> Looks like there is an issue on Mac to detect the available VMInstalls.

see the following PR for background and mitigation:
- https://github.com/eclipse-platform/eclipse.platform/pull/390
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/211

PR #545, opened 2023-03-31
Use p2.inf to pull in depedency to org.eclipse.jdt.launching.macosx

laeubi:
jdt.launcher requires special treatment to find jvms on MacOS that is handeled with an additional bundle and the pde.ui test depend on the launching framework.

This adds a p2.inf that pulls in this additional depenedency using a p2.inf for the test framework resoloution.

Lets see if this helps for PDE as well @HannesWell


laeubi:
Only three tests left ...

HannesWell:
Great, thanks!
So the problems in https://github.com/eclipse-pde/eclipse.pde/pull/541#issuecomment-1491044930 are solved. Nice.

PR #546, opened 2023-03-31
Adjust pde.build to pack200 removal in p2

akurtakov:
Now that p2 no longer deals with pack200, pde.build should stop doing so too.
Tracked in https://github.com/eclipse-equinox/p2/issues/238 .


akurtakov:
IIRC pde.build tests are not executed with maven so testing only happens with I-build thus if verification is successful I'll merge and see whether all are fixed in next I-build.

akurtakov:
Jenkins build and test is successful and the number of tests is exactly the same like the last master build. Merging.

PR #547, opened 2023-03-31
Bump BREE for bundles with transitive Java-17 dep

akurtakov:
Declare this dependency directly to allow actually using Java 17 features in these bundles.


HannesWell:
Thank you!

PR #548, opened 2023-03-31
Remove manual tycho-surefire dependency

akurtakov:
Bundle o.e.pde.genericeditor.extension.tests no longer requires o.e.osgi.compatibility.state bundles or it was even copy-paste from somewhere.
This reduces possible differences between I-build tests and Maven testing.


akurtakov:
Tests number is exactly like last master build.

PR #550, opened 2023-04-01
Support automatic manifest generation in new project wizard

laeubi:
Currently one needs a manual setup steps to create a project that uses automatic manifest generation.

This adds a new option in the new plugin-project wizard to enable the automatic manifest generation.

## How it works

Create a new Plugin project and select the standard OSGi framework type and select "Generate OSGi metadata automatically"

grafik

Review and maybe adjust the usual settings like name and version

grafik

Finish the wizard to get the project created
grafik


HannesWell:
Great 👍🏽

Eventually there should be a N&N entry for the whole manifest-generation story.

laeubi:
> Eventually there should be a N&N entry for the whole manifest-generation story.

I created individual N&N already and will try to write down a larger story once all the pieces are there.

PR #552, opened 2023-04-01
Use p2.inf to pull in depedency to org.eclipse.jdt.launching.macosx

laeubi:
jdt.launcher requires special treatment to find jvms on MacOS that is handeled with an additional bundle and the pde.smartimport test depend on the launching framework.

This adds a p2.inf that pulls in this additional depenedency using a p2.inf for the test framework resoloution.


laeubi:
One more successful mac test ...

PR #560, opened 2023-04-04
Synchronize values changed in the Overview with the instructions file

laeubi:
Currently one can edit the instructions file and after safe the manifest is updated (and thus also the editor), but the other way does not work, if one for example edit the name in the overview page the changes are just overwritten on the next compile.

This installs listeners and implements a very basic model for the instructions so changes are mirrored to the instructions file as it already happens for the MANIFEST.MF source tab.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyAnalysisSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyAnalysisSection.java
@@ -40,8 +40,13 @@ public class DependencyAnalysisSection extends PDESection {
 	}
 
 	private String getFormText() {
-		boolean editable = getPage().getModel().isEditable();
-		if (getPage().getModel() instanceof IPluginModel) {
+		IBaseModel model = getPage().getModel();
+		if (model == null) {
+			return ""; // we don't know the type so we can't give //$NON-NLS-1$
+						// a suitable hint (at the momment)
+		}
+		boolean editable = model.isEditable();
+		if (model instanceof IPluginModel) {
 			if (editable)
 				return PDEUIMessages.DependencyAnalysisSection_plugin_editable;
 			return PDEUIMessages.DependencyAnalysisSection_plugin_notEditable;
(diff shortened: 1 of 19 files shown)


laeubi:
@vik-chand @HannesWell it seems I missed the freeze yesterday, nerveless I wanted this to be merged as soon as possible, would you as PL approve this?

Even though it might bee seen as a new feature, it also fixes some issues one currently see with using that feature as currently sometimes the manifest editor disappears and the manifest is not regenerated when things run "too fast" after each other and users might be confused when there changed values are not taken into account.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ManifestEditor.java
```diff
@@ -356,9 +350,19 @@ public void editorContextAdded(InputContext context) {
 		}
 	}
 
+	@Override
+	public void doSave(IProgressMonitor monitor) {
+		// TODO Auto-generated method stub
+		super.doSave(monitor);
+	}
+
```
HannesWell, line 358: Is this still necessary? It looks obsolete.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyAnalysisSection.java
```diff
@@ -40,8 +40,9 @@ public DependencyAnalysisSection(PDEFormPage page, Composite parent, int style)
 	}
 
 	private String getFormText() {
-		boolean editable = getPage().getModel().isEditable();
-		if (getPage().getModel() instanceof IPluginModel) {
+		IBaseModel model = getPage().getModel();
+		boolean editable = model != null && model.isEditable();
+		if (model instanceof IPluginModel) {
```
HannesWell, line 45: In case the model is null this method returns the text for fragments. I don't think that is intended.
If null is a valid case now that should be be handled somehow. My first suggestion would be to use the Plugin text case case the model is null and the 'type' is therefore unknown.
```suggestion
		if (model instanceof IPluginModel || model == null) {
```

HannesWell (approved):
I only checked the existing code that is not only related to the new BND processing to ensure existing functionality is not broken and that looks good.
Since you just added the BND processing recently there is nothing to break that used to work in a previous release/milestone.
Therefore and since you said it contains fixes I'm fine with merging this today.

Only two remarks below. They are not critical but should be addressed eventually.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ManifestEditor.java
laeubi: Removed

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyAnalysisSection.java
laeubi: It could happen that the model is (not yet) there, in such a case one can not give a suitable hint and I return the empty string for now.

HannesWell (approved):
Looks good. Thank you. 👍🏽

Issue #534, opened 2023-03-27
arbitrary behavior for export package dialog


gireeshpunathil:
steps to reproduce:

 - create a plugin project
 - open design editor
 - select `runtime` and click add
 - toggle the `Show non-java packages` a couple of times
 - repeat this with selecting and deselecting items in the list
 - watch the behavior of `Add` button

I see arbitrary behaviors for the `Add` button, w.r.t being enabled / disabled.

vik-chand:
Nice catch! That's a bug. I can recreate it too

gireeshpunathil:
relevant code: https://github.com/eclipse-pde/eclipse.pde/blob/91ab8626d037671ea7d5a7009db6378d3eeae46d/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java#L383-L434

I am unable to locate the code pertinent to the button and its selection change callback, so not sure if this is an in-built behaviour of the `ConditionalListSelectionDialog` class or not.

PR #567, opened 2023-04-06
make export dialog add button follow list selection

gireeshpunathil:
There are potential race conditions that are happening between the dialog creation with empty list and the thread that populates the list and fires selection change events.

An easy hack is to make sure the button is enabled whenever there are items in the list. This always
works, as the default behavior of list makes sure
that there is at least one item selected in a
non-empty list.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/534


vik-chand (changes requested):
Perhaps a null check on fElements is a good idea. I see null check for fElements in some other parts in this file.

vik-chand:
Also update
Copyright (c) 2005, 2015 

to 

Copyright (c) 2005, 2023

gireeshpunathil:
@vik-chand -fixed both, thanks.

vik-chand (approved):

Issue #307, opened 2022-08-23
Remove "Run API analysis parallel to the build job" option from UI preferences


vogella:
IIRC @iloveeclipse planned to remvoe the "Run API analysis parallel to the build job" option from the preferences once this is stable.

Andrey, do you have time to remove it next release?

![image](https://user-images.githubusercontent.com/139910/186105793-9e9d1c42-6007-4231-98ae-33ae4b918aee.png)

vik-chand:
@iloveeclipse  Can you try this for 4.26RC1 or 4.27M1?

iloveeclipse:
had no time, sorry. 
@vogella : can you propose a PR **to remove UI** for setting this preference?
The preference itself is still in use for tests, they need (for simplicity) synchronous execution.

PR #568, opened 2023-04-07
remove RunAPIAnalysisBuilderAsJob option from perferences UI

gireeshpunathil:
Remove all references to the UI preference for running API analysis in parallel to the builder, as a scheduled job, While retaining the underlying boolean ICoreConstants.RUN_API_ANALYSIS_AS_JOB so that the tests that depend on it are unaffected, though it is false always.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/307

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PreferenceInitializer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PreferenceInitializer.java
@@ -75,7 +75,6 @@ public class PreferenceInitializer extends AbstractPreferenceInitializer {
 		PDEPreferencesManager corePrefs = PDECore.getDefault().getPreferencesManager();
 		corePrefs.setDefault(ICoreConstants.WORKSPACE_PLUGINS_OVERRIDE_TARGET, true);
 		corePrefs.setDefault(ICoreConstants.DISABLE_API_ANALYSIS_BUILDER, false);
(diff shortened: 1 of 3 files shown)


gireeshpunathil:
If I understand the issue report correctly, we just remove the option from the UI, but do not affect the underlying setting (set to false always) so that the tests run smooth. Is my understanding correct @iloveeclipse @vogella ?

iloveeclipse:
Yes

vik-chand:
@iloveeclipse  Do we want to keep this option or do we want to remove it?  I am not very sure ( as it can be very helpful). What is your opinion?

iloveeclipse:
> @iloveeclipse Do we want to keep this option or do we want to remove it? I am not very sure ( as it can be very helpful). What is your opinion?

Since we switched to run as job I've never needed the preference option **in the UI as user**.
The preference option itself is still needed by the test, but if I read it right, this change doesn't touch that.

stephan-herrmann:
This change causes 
```
Warning: NLS missing message: MainPreferencePage_RunAPIAnalysisBuilderAsJob in: org.eclipse.pde.internal.ui.pderesources
```

vik-chand:
> This change causes
> 
> ```
> Warning: NLS missing message: MainPreferencePage_RunAPIAnalysisBuilderAsJob in: org.eclipse.pde.internal.ui.pderesources
> ```

@gireeshpunathil Can you please look at this?

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/pull/619

Issue #566, opened 2023-04-06
standard OSGi plugin creation settings mutates the creation wizard


gireeshpunathil:
steps to recreate:
1. project -> plugin project
2. select OSGi standard type
3. finish
4. create another plugin project
5. this time select Eclipse as the target type
6. the `next` button is now disabled, disallowing to create plugins from templates
7. clean / change the workspace to restore the original behaviour

probably related : https://github.com/eclipse-pde/eclipse.pde/pull/550 

/cc @laeubi

laeubi:
I have looked into this and the problem seem when "Generate OSGi Metadata automatically" is checked, so a workaround is to first uncheck that option and then switch to "Eclipse" project type, will now look into it why it is considered even when not selected...

vik-chand:
Let's fix this in 4.28M2

PR #569, opened 2023-04-07
Automatic flag in TemplateListSelectionPage must check if OSGi is active

laeubi:
The "automatic" flag is actually bound to the enablement of an OSGi Framework project type, so we need to disable it when OSGI is not selected.

@HannesWell @vik-chand this should be easy enough to be included in current M1?

Fix https://github.com/eclipse-pde/eclipse.pde/issues/566


gireeshpunathil:
I applied this patch and tested locally, everything looks good!

HannesWell (approved):
Looks reasonable.

> @HannesWell @vik-chand this should be easy enough to be included in current M1?

But we have already missed M1 and I believe it is not a blocker (you desribed an acceptable workaround) to require a respin.

> I applied this patch and tested locally, everything looks good!

Thanks for checking!

PR #570, opened 2023-04-09
migrate to Java 17 switch semantics

gireeshpunathil:
replace legacy switch block with java 17 semantics in `eclipse.pde.internal.ui` package, wherever relevant.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/filter/SizeFilter.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/filter/SizeFilter.java
@@ -36,31 +36,21 @@ public class SizeFilter implements IFilter {
 
 	@Override
 	public boolean accept(final ImageElement element) {
-		boolean accept = true;
+		
 
-		switch (mWidthType) {
-			case TYPE_EXACT :
-				accept &= (element.getImageData().width == mWidth);
-				break;
-			case TYPE_BIGGER_EQUALS :
-				accept &= (element.getImageData().width >= mWidth);
-				break;
-			case TYPE_SMALLER_EQUALS :
-				accept &= (element.getImageData().width <= mWidth);
-				break;
-		}
+		boolean accept = true & switch (mWidthType) {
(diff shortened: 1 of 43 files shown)


mickaelistria:
Thank you!

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/templates/PreprocessorParser.java
```diff
@@ -224,20 +214,13 @@ private void parse() throws Exception {
...
-				case T_NEQ :
-					opcode = OP_NEQ;
-					break;
-			}
+			opcode = switch (token) {
+				case T_AND -> opcode = OP_AND;
+				case T_OR -> OP_OR;
+				case T_EQ -> OP_EQ;
+				case T_NEQ -> OP_NEQ;
+				default -> opcode;
```
HannesWell, line 222: the default could be zero and the swtich expression could be assigned to the opcode variable during initialization already.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/filter/SizeFilter.java
```diff
@@ -36,31 +36,21 @@ public SizeFilter(final int width, final int widthType, final int height, final
...
-			case TYPE_BIGGER_EQUALS :
-				accept &= (element.getImageData().height >= mHeight);
-				break;
-			case TYPE_SMALLER_EQUALS :
-				accept &= (element.getImageData().height <= mHeight);
-				break;
-		}
+		
+
+		boolean accept = true & switch (mWidthType) {
```
HannesWell, line 41: The leading `true` is not necessary here.

And the parenthesis in the following cases can probably be deleted too.

HannesWell (commented):
Thank you @gireeshpunathil for this clean up, this is much appreciated.

Sorry for the late reply. I just have to minor remarks below, in case you want to address them in a follow up.

vogella:
Very nice to see code cleanup activities in PDE. 

@gireeshpunathil I hope to see more of these in the future!

Issue #563, opened 2023-04-05
quick-fix for extraneous empty line(s) in manifest editor


gireeshpunathil:
-leave an empty line in MANIFEST.MF and save
 - error is produced on: `Extraneous empty line causes the file to be invalid`
 - quick fix menu is enabled, but upon clicking it says `no fixes available for Extraneous empty line causes the file to be invalid`

It can be bettered via either:
 - disable the quick fix when one is not available
 - provide a quick fix to remove extraneous lines (on the lines of how the quick fix works when an extraneous line is missing at the end of the file)

iloveeclipse:
@gireeshpunathil , @vik-chand : could you please look into that soon, I hate workspace with compilation errors.

iloveeclipse:
I assume the code has to ignore trailing empty lines (e.g. check if any of the following lines contain non-whitespace characters)

vik-chand:
Thanks @iloveeclipse I have reverted this commit. Gireesh will create a new PR after addressing this issue.

PR #573, opened 2023-04-12
refix: quick-fix for extraneous empty line(s) in manifest editor

gireeshpunathil:
Extraneous lines are erroneous to the manifest compiler, but it does not recommend / provide a resolution. Add a quick fix to remove extraneous line. takes care of the empty valid empty lines at the end of manifests.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/563
Refs: https://github.com/eclipse-pde/eclipse.pde/pull/564
Refs: https://github.com/eclipse-pde/eclipse.pde/pull/572

/cc @iloveeclipse @vik-chand  - please review. The issue with the previous one was that I was checking whether the pertinent line is the last one or not (in which case we ignore it, as the last line can be empty as per the spec), but that logic did not cater to multiple last lines. 

The new version addresses this by getting the rest of the content in the file and trim it to see if those are empty. I tested it and looks like it works.

Also, I converted this as a warning, and removed the return statement, so that this does not block the build.


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -138,11 +138,21 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
 					return;
 				}
 				// parse
+				String rest = document.get().substring(lineInfo.getOffset() + lineInfo.getLength()).trim();
```
iloveeclipse, line 141: Please don't do that. This does eager reading almost all of the the manifest on every line!

iloveeclipse (changes requested):

HannesWell:
Since this change seems to be non trivial could you please add some tests for it?
Maybe one for the simple case and one for the one that caused problemlos in the first attempt?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -138,11 +138,21 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
...
 				if (line.length() == 0) {
 					// Empty Line
 					if (l == 0) {
 						report(PDECoreMessages.BundleErrorReporter_noMainSection, 1, CompilerFlags.ERROR, PDEMarkerFactory.CAT_FATAL);
 						return;
+
+					} else if (rest != "") { //$NON-NLS-1$
+						VirtualMarker marker = report(PDECoreMessages.BundleErrorReporter_noNameHeader, lineNumber,
```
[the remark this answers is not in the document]
gireeshpunathil: thanks @iloveeclipse for the proposal. I had tried a similar approach earlier, but abandoned. My plan was to identify multiple continues empty lines and mark them together. The fact that the markers do not allow aggregations in their attributes (only accepts strings and integers as values), it was tricky to implement that.

However, your suggestion to `remember` empty lines and reporting at the end individually makes sense to me, I will follow that; thanks!

vik-chand:
> > Since this change seems to be non trivial could you please add some tests for it?
> > Maybe one for the simple case and one for the one that caused problemlos in the first attempt?
> 
> I was looking for prior arts in manifest compiler test cases. I came across 2 types and both pose challenges to my scenario, so I thought I will ask here:
> 
>     1. tests in https://github.com/eclipse-pde/eclipse.pde/tree/c835eb17aa9248cffad5beb47d367424c3e7f571/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/model/bundle: these validate the correctness of manifest format to great extent, however, none of those exercise the manifest parser.
> 
>     2. https://github.com/eclipse-pde/eclipse.pde/blob/c835eb17aa9248cffad5beb47d367424c3e7f571/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/ee/ExecutionEnvironmentTests.java : performs full-blown build of a project, and requires the manifests and other artifacts in the disk (I guess).
> 
> 
> Are there better options that does these?
> 
>     * able to create a manifest in memory like string buffers
> 
>     * able to compile it and collect the errors and warnings

I will propose to do this (PDE compiler errors/warning)  via a separate issue. API tools do have test case for API tool scenarios ( however there too , we don't have tests for quickfixes). In JDT, we have tests for JDT compiler errors and warnings. It  will be good to have a set of test case for PDE compiler problems as well( not only extra lines but all of the other ones) ( we could look at both JDT as well as API tool test scenarios to get started with this work).

vik-chand:
@gireeshpunathil  Can you please update this patch wrt master?
@iloveeclipse  Can you please review post the update from Gireesh?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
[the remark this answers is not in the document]
gireeshpunathil: @iloveeclipse  - pushed a new change based on the review comments, please have a look. The approach is to identify empty lines in each iteration and push the locations to a list instead of marking inline, and then mark those only when a non-empty line is visited. this solves two concerns: no pressure is incurred in the iteration, legitimate empty lines at the end are automatically ignored.

gireeshpunathil:
@vik-chand @HannesWell @iloveeclipse - pushed a change to improve the parsing, pls have a look. I have tested locally and everything looks ok; will continue to look for ways to add test scenarios in subsequent PR(s).

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -139,17 +141,32 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
...
+						emptyLines.add(Integer.valueOf(l));
+						continue;
 					}
 					/* flush last line */
 					if (header != null) {
 						fHeaders.put(header.getName().toLowerCase(), header);
 						header = null;
 					}
 					break; /* done processing main attributes */
+				} else {
```
iloveeclipse, line 160: This `else` is unneeded and shows whole block as aa huge warning in Eclipse:
image

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -139,17 +141,32 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
...
+					for (Integer index : emptyLines) {
+						VirtualMarker marker = report(PDECoreMessages.BundleErrorReporter_noNameHeader, index + 1,
+								CompilerFlags.WARNING, PDEMarkerFactory.M_EXTRANEOUS_EMPTY_LINES,
+								PDEMarkerFactory.CAT_OTHER);
```
iloveeclipse, line 164: `PDEMarkerFactory.CAT_OTHER` is intentional? It is not specified in the error reported above.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -139,17 +141,32 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
...
 					/* flush last line */
 					if (header != null) {
 						fHeaders.put(header.getName().toLowerCase(), header);
 						header = null;
 					}
 					break; /* done processing main attributes */
+				} else {
+					for (Integer index : emptyLines) {
+						VirtualMarker marker = report(PDECoreMessages.BundleErrorReporter_noNameHeader, index + 1,
+								CompilerFlags.WARNING, PDEMarkerFactory.M_EXTRANEOUS_EMPTY_LINES,
```
iloveeclipse, line 163: Why do we use `CompilerFlags.WARNING` here if we are sure and report user that the manifest is invalid?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -139,17 +141,32 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
...
 				// parse
 				if (line.length() == 0) {
-					// Empty Line
+					// Empty Line in the beginning
 					if (l == 0) {
 						report(PDECoreMessages.BundleErrorReporter_noMainSection, 1, CompilerFlags.ERROR, PDEMarkerFactory.CAT_FATAL);
 						return;
+
+						// empty line elsewhere
+					} else if (l != document.getNumberOfLines() - 1) {
```
iloveeclipse, line 150: Can we please extract `document.getNumberOfLines()` call **before** the for loop? It looks like a non trivial getter call.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ExtraneousLinesResolution.java
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	public ExtraneousLinesResolution(int type, IMarker marker) {
+		super(type, marker);
+	}
+
+	/**
+	 * Resolves the problem by extracting the empty line and removing it.
+	 */
+	@Override
+	protected void createChange(BundleModel model) {
```
iloveeclipse, line 45: Please check how **multiple** empty lines are handled via quick fix.
After applying quickfix the manifest is corrupted, some **proper** lines are removed:

image
After 
image

vik-chand:
@gireeshpunathil  For multi-fix, we may need to override public IMarker[] findOtherMarkers(IMarker[] markers) {
See AddAutomaticModuleResolution or ConfigureProblemSeverityForPDECompilerResolution as an example

gireeshpunathil:
ok - multi-fix is new to me.

when I debugged, what I could see is: 
 - the parser builds a list of markers, with a set of line numbers.
 - the resolver is called for each marker
 - a resolution here means deleting a line and updating the doc
 - this means subsequent markers are invalid, as the liner numbers referred by those are now different in the doc

any suggestions?

laeubi:
> the resolver is called for each marker

If I remember correctly, you can override a method to get all markers at once somehow to handle them as one batch.

vik-chand:
For multi-fix of adding since tag ( in API Tools), after adding each since tag, the line number does change. It will be worth seeing what is done there to overcome this issue.

gireeshpunathil:
did not find another resolver that does a similar thing.

I tried a hack by persisting the state across `createChange` calls, but soon realized that it wont work: the document in question is then subjected for race condition. The document modification in one thread fires an event in another, which in turn causes re-parsing of the manifest and potential generation of new markers.

so one reasonable approach I see is to aggregate all the warnings in one marker, and resolve it atomically. Might look little odd from the ui perspective, but I think it is the right thing to do from a use case point of view too: no one would want to fix one extraneous line and leave another untouched, do they?

WDYT?

iloveeclipse:
Would work, sure. Just create a marker pointing to very first line. In theory one could simply avoid storing *any* line info (except first) and simply go over file and remove every empty line.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: fixed it, changed the logic to avoid eager reading.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: fixed it, the redundant else block is taken out.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: we initially started with `CompilerFlags.ERROR`, but then downplayed it to warning due to the regression it caused. Is it ok to see its stability for sometime before we elevate it to an error?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: fixed.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: the other categories are so specific and do not match with this context. If and when we decide to elevate this into an error, we could use `PDEMarkerFactory.CAT_FATAL` .WDYT?

gireeshpunathil:
> Would work, sure. Just create a marker pointing to very first line. In theory one could simply avoid storing any line info (except first) and simply go over file and remove every empty line.

@iloveeclipse - pushed a change to this effect. please have a look!

gireeshpunathil:
> Please don't create multiple commits for this feature and squash previous commits with the next commit that would handle requested changes in one commit.

sure, will follow that in future. (in the previous project I worked, we leave the review commits separate for sometime, so that the reviewers can `visually` see the delta changes; and squash into one once things are ready to merge. I am still learning new practices everyday!)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
iloveeclipse: > then downplayed it to warning due to the regression it caused

Well, if we **know** the manifest is invalid, why do we want to report only a **warning**?

Regarding regression: we should only make sure the code does the right thing :-)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
iloveeclipse: Sure. I'm not an OSGI expert, but if you know that manifest containing an empty line is invalid, it is surely a fatal issue?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: @vik-chand - pls share your thoughts here. Error or Warning? I don't have a strong opinion at this point, so would go by the consensus between you two!

vik-chand: Lets keep to whatever was before ( ie Error)

gireeshpunathil: thanks @vik-chand and @iloveeclipse . pushed (and squashed) a change to that effect, pls have a look. thanks!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: fixed this too - using `PDEMarkerFactory.CAT_FATAL` now.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ExtraneousLinesResolution.java
```diff
@@ -0,0 +1,72 @@
...
+		System.out.println(this);
+		IDocument doc = model.getDocument();
+		int line = 0;
+		try {
+			String emptyLines = (String) marker.getAttribute("emptyLine"); //$NON-NLS-1$
+			StringTokenizer st = new StringTokenizer(emptyLines, ","); //$NON-NLS-1$
+			int adjust = 0;
+			while (st.hasMoreElements()) {
+				line = Integer.parseInt(st.nextToken()) - adjust++;
```
iloveeclipse, line 54: If we would save lines in the opposite order, we could avoid "adjust" calculation here because we would walk from the end of the file to begin.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -193,6 +206,25 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
 				}
 
 			}
+			if (emptyLines.size() > 0) {
+				// build a marker with all the line info aggregated into it,
+				// with the first empty line as the anchor line.
+				// could have been easy if the marker accepts maps, but it
+				// doesn't at the moment. TODO: enhance marker to process maps.
+				StringBuilder str = new StringBuilder();
```
iloveeclipse, line 214: You could use StringJoiner like
```
StringJoiner str = new StringJoiner(","); //$NON-NLS-1$
for (Integer i : emptyLines) {
	str.add(Integer.toString(i));
}
```
and so don't need to add separators / remove last one

iloveeclipse (dismissed):
Should be mostly OK now, just few smaller nitpicks.

gireeshpunathil:
@iloveeclipse - I have been on vacation for almost a month, will address the review comments sometime this week.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -193,6 +206,25 @@ protected void parseManifest(IDocument document, IProgressMonitor monitor) {
 				}
 
 			}
+			if (emptyLines.size() > 0) {
+				// build a marker with all the line info aggregated into it,
+				// with the first empty line as the anchor line.
+				// could have been easy if the marker accepts maps, but it
+				// doesn't at the moment. TODO: enhance marker to process maps.
```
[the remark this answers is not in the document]
gireeshpunathil: removed the TODO, to avoid confusion.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ExtraneousLinesResolution.java
gireeshpunathil: thanks for the suggestion, applied that logic - here, and the matching logic with the order reversal at the marker side.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ExtraneousLinesResolution.java
```diff
@@ -0,0 +1,72 @@
...
+				IRegion l = doc.getLineInformation(line);
+				doc.replace(l.getOffset(), l.getLength() + 1, ""); //$NON-NLS-1$
+			}
+		} catch (BadLocationException | CoreException e) {
```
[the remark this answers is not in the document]
gireeshpunathil: addressed, logged the error. also added `NumberFormatException` to the troop.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
gireeshpunathil: thanks, I used this class to compose the marker string.

gireeshpunathil:
@iloveeclipse - apologies for the delay, was stuck with split priorities. I have addressed all the comments, pls have a look. thanks!

gireeshpunathil:
I also found a strange thing while testing this. If I add empty lines between `Bundle-SymbolicName: el;singleton:=true` and `Bundle-Version: 1.0.0.qualifier` lines in the manifest, then it produces lot of errors like this:

image


I tested with master and can see it is a pre-existing scenario. Looks like the first fatal error caused skipping of the rest of the parsing, including the `require-bundle` validation. Will investigate outside of this PR. Just mentioning here to avoid confusion.

gireeshpunathil:
@jukzi - did you add / modify any commits to this PR? I do see some actions above but do not see any files changed, so just wondering what it could have been! thanks.

@iloveeclipse - when you get some time, pls review this PR, or else let me know if you still have `request changes`.

jukzi:
i only rebased

gireeshpunathil:
I don't know what is the usual practices in this situation for this project; this PR is waiting for more than a month for review or cancellation of previously set X mark.

@iloveeclipse @HannesWell @vik-chand - pls advise.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -2353,6 +2353,8 @@ NoLineTerminationResolutionCreate_description=Adds a line break at the end of th
 NoLineTerminationResolutionCreate_label=Add a line break after the header
 NoLineTerminationResolutionRemove_description=Removes all whitespace characters from the last line of the manifest
 NoLineTerminationResolutionRemove_label=Remove excess whitespace from the end of the manifest
+ExtraneousLineResolutionRemove_label=Remove extraneous lines
```
jukzi, line 2356: Would be more clear to add "empty" before "lines".  That's even more explanatory then "extraneous"

jukzi (dismissed):
Please add a Juni Test that shows removing multiple lines works as expected.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ExtraneousLinesResolution.java
gireeshpunathil: yes, this is solved now.

vik-chand:
> Please add a Juni Test that shows removing multiple lines works as expected.

None of the quickfixes in Manifest file has JUnit. It will be a good idea to create a separate issue to create Junit test for quickfixes.

jukzi:
> None of the quickfixes in Manifest file has JUnit.

@vik-chand then please you finish review.

vik-chand:
> > None of the quickfixes in Manifest file has JUnit.
> 
> @vik-chand then please you finish review.

Sure, I will review this

vik-chand:
My review comments - 

1) For multiple empty lines, we just have 1 error marker. If we look at any other error marker and its behaviour, we generally have 1  error marker per problem. It is not obvious that the 2nd empty line is also an error in itself.

2) Multi-fix removes all lines not marked with error which again is confusing and I haven't seen such behavior in any pde or jdt or other quick fix. We should strive to be consistent. However trying multifix would require a preference value (which may or may not be present in the UI). Already we have too many preference options. Also it will require to deal with the line change number.

Based on this, either
A) We should not do multi-fix with this PR ( and make another issue out of it). I don't anticipate lot of work for that multifix issue.
B) We should align the multi-fix in this PR itself which is consistent with other multi-fixes in PDE and JDT.

 JUnit for quickfixes of PDE ( on similar lines of JDT UI - I believe JDT UI has one) is a good idea. In any case, we should also create an issue for this so as to not lose this work item.

gireeshpunathil:
thanks @vik-chand for the review. IMO, the problem is with the current design of quick-fix resolution framework in the empty lines context. we could very well attach one marker per error. But when it comes to the resolution site, applying of the fix poses challenge. When one fix is applied, it renders the rest of the markers invalid - due to the fact that the line numbers stored in those markers are changed due to the first resolution. Depending on relative position of the marker that is being resolved, the other markers may contain valid or invalid data.

I can go back to single-fix (from where we started originally), but that again may look a weak feature from end user perspective? let me dig a little more to see if there is a way to synchronize multiple markers w.r.t resolutions. If it proves to be impossible, I am happy to fall back to single fix. Ideas welcome!

gireeshpunathil:
ok, I was doing a little digging around how to develop multi-fix in a fool-proof manner, and looked at how similar quick-fix implementations are doing.

 - RemoveUnknownExecEnvironments: shows only one error in the first fault line, but resolution applies to all applicable lines.
 - RemoveRequireBundleResolution: shows errors in all the relevant lines, but multi-fix leads to resolution to only the first item.
 - RemoveImportPackageResolution: shows errors in all the relevant lines, quick-fix initially shows only on line, but upon selecting the resolution other lines appear in the view, but upon selecting multi-fix, only the first one gets removed.

I stopped looking further after these 3. In short, the multi-fix is already inconsistent, and looks like we went to and fro a couple of times in this PR in search of ideal behaviour.

I can invest more time on this, but would like to know if it makes sense to land this as is, given the prior arts. I can then spend time on implementing a marker aggregator function in the `AbstractManifestMarkerResolution` class that will help adjusting all the affected markers in-line with a single change. That will help me addressing all the above issues one and for all. WDYT?

@vik-chand @HannesWell

vik-chand:
> * RemoveImportPackageResolution

You can open a bug for RemoveRequireBundleResolution & RemoveImportPackageResolution

So with this PR we are mimicking this behavior ( RemoveUnknownExecEnvironments). I think in the quickfix if you make make it  explicit( via text) that all extraneous lines will be deleted.

What do you think @HannesWell  ?

HannesWell:
I haven't follow the discussion closely, but in general I agree that it would be good if the quick-fix removes all extraneous empty lines on the first run. And I also think that it should be consistent and other quick-fixes in a similar situation should behave the same. IMHO it would be worth to invest some time into this. :)

vik-chand (approved):
Also please open an issue for junits for quickfixes

HannesWell (changes requested):
Just tried out this change and I get errors when there are entries outside the main section (which is valid):

```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: Host
Bundle-SymbolicName: host;singleton:=true
Bundle-RequiredExecutionEnvironment: JavaSE-11

Name: com/ibm/icu/text/TitlecaseTransliterator.class
SHA-256-Digest: bGXIJlHHRm7FZikT6z4B1DZw87oPajgDwhHlAojb5a0=
```
You can see these extra entries for example in all jar-signed artifacts in the TP. So I think an error in this case is a false-positive.

gireeshpunathil:
@HannesWell - pushed a change to fix this issue, can you please have a look? thanks!

laeubi:
@HannesWell anything left here?

HannesWell:
> @HannesWell anything left here?

I'll have to check it again in the next days, have to familiaries myself again with this change.

laeubi:
> > @HannesWell anything left here?
> 
> I'll have to check it again in the next days, have to familiaries myself again with this change.

Any updates here?

PR #575, opened 2023-04-13
migrate to Java 17 switch semantics: extras

gireeshpunathil:
replace legacy switch block with java 17 semantics:
1. short-circuit initialization in the switch expression
2. default case match with the original initialization
3. reduce complex boolean expression to an elegant form

Refs: https://github.com/eclipse-pde/eclipse.pde/pull/570#pullrequestreview-1379419914

/cc @HannesWell

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/templates/PreprocessorParser.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/templates/PreprocessorParser.java
@@ -213,7 +213,7 @@ public class PreprocessorParser {
 			}
 
 			int opcode = switch (token) {
-				case T_AND -> opcode = OP_AND;
+				case T_AND -> OP_AND;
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/templates/PreprocessorParser.java
```diff
@@ -212,14 +212,12 @@ private void parse() throws Exception {
 				continue;
 			}
 
-			int opcode = 0;
-
-			opcode = switch (token) {
+			int opcode = switch (token) {
 				case T_AND -> opcode = OP_AND;
```
merks, line 216: I think this assignment is not intentional.

gireeshpunathil: @merks - agree, looks like accidentally retained in the previous change. fixed it now.

Issue #581, opened 2023-04-18
Improve PDE onboarding icon for the black theme


vogella:
The onboarding icon for the resource perspective is beautiful. Thanks @BeckerWdf 

![image](https://user-images.githubusercontent.com/139910/232732868-29241dfc-f6d6-40f1-b07c-c68a8049131a.png)

The light background in the PDE onboarding icon does not look good in the dark theme.

![image](https://user-images.githubusercontent.com/139910/232733023-4aa80b86-16df-4cdb-8011-3b52176361f2.png)

(same for the JDT perspective BTW)

@BeckerWdf I think you were planning to add transparency or to remove the background but I was unable to find the issue / PR for it.

PR #582, opened 2023-04-18
Make Onboarding Icon More High Fidelity

BeckerWdf:
Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/581


BeckerWdf:
@vogella: Why can't I add you as reviewer on PRs in PDE?

BeckerWdf:
I made the icon completely new in a more high fidelity version:

Screenshot 2023-04-18 at 14 44 42
Screenshot 2023-04-18 at 14 44 27

vogella:
Thanks @BeckerWdf this is much nicer!

> @vogella: Why can't I add you as reviewer on PRs in PDE?

Don't know

vogella (commented):
LGTM

merks:
I think only committers can assigned reviewers. I think also only a committer can close an issue other than their own.  I'm not sure how those rules are made. I suppose that's to prevent arbitrary people from doing potentially disruptive things, i.e., for our safety and comfort...

laeubi:
Yes, only committers can assign reviewers. So one solution would be to nominate @BeckerWdf as a PDE committer ;-)

vogella:
> Yes, only committers can assign reviewers. So one solution would be to nominate @BeckerWdf as a PDE committer ;-)

Please go ahead and nominate @BeckerWdf as PDE committer. This icon work is worth the official recognition.

laeubi:
> I made the icon completely new in a more high fidelity version:

I'm a bit late to the party but wanted to remark that the PDE icon is representing a "wall plug" stick into a "power outlet" so maybe we should have symbolize this with some lines or join it like in the original PDE icon:

grafik

https://github.com/eclipse-platform/eclipse.platform.images/blob/master/org.eclipse.images/eclipse-svg/org.eclipse.pde.ui/icons/view16/plugins.svg

that has some shining/3d effect maybe this effect can be used to not make the icon look too "flat", if I wouldn't know the original it now looks a bit more than a cloud or some bubbles crashing into an invisible wall :-)

BeckerWdf:
>  the PDE icon is representing a "wall plug" stick into a "power outlet" 

I know and it was my intention so symbolise this but in a flat look style. If I did not succeed that's a pitty.

> that has some shining/3d effect maybe this effect can be used to not make the icon look too "flat", 

As the onboarding icon should be more like a water mark I did remove shining/3d effect by intention

>if I wouldn't know the original it now looks a bit more than a cloud or some bubbles crashing into an invisible wall :-)

That's bad. Can you elaborate a bit more and this so that I am able to improve the icon. What makes it cloudy and where is the invisible wall?

laeubi:
I'll try to provide a screenshot, but try to describe in more details as text an hope this already clarifies some things:

- if you want it to look like a watermark I think it should be a solid black shape with a transparency of e.g 60% so it adapts good to different background (e.g. on my system the background color is grey not white) and looks like it is transparently "stamped" over the background.
- because there is a space between the the two parts, it does not looks "connected" and for a real plug I would expect to see the plug contacts, so it looks like there is something (invisible) that separates the two things.
- if you look at the large image above, you see that the right part is a bit wider and more sharp (not rounded) because usually that part needs some "grip" to plug it in/out and is one continuous "thing"
- The left part on the other hand has a more flat and consists of three parts the one the plug fits in, then it has a (one pixel) "ring" and then the cable so currently it looks a bit too rounded and not as something that fits in.

BeckerWdf:
> the large image above

Do you mean my grey icon or your colorful icon?

laeubi:
The color icon, I just made a screenshot of a zoomed variant so one can better see the details.

BeckerWdf:
@laeubi: How do you like this once

Screenshot 2023-05-11 at 09 21 12
Screenshot 2023-05-11 at 09 20 58

BeckerWdf:
> That looks better, I just think the plug should really not be rounded and the cable should get a slighly different colour !

The plug is the right part?

> 
> You also might want to try it out with the standard GTK background color what is neither white nor black but some gray tone grafik

Can you send me the RGB values of that background

BeckerWdf:
Thanks for the feedback. I did what you proposed. Better now?

With GTK background:
onboarding_plugins_gtk

Mac dark and light:
Screenshot 2023-05-11 at 10 11 30

Screenshot 2023-05-11 at 10 11 46

laeubi:
Yes this looks now better (for me, so probably other want to give feedback too), what do you think about the outline idea?

grafik

BeckerWdf:
I don't like the idea of the outline. With that we will loose the "whatermark"-like style

laeubi:
As said it currently looks not like a "watermark" either... also googling for "watermark" gives a wide range of icons classified as watermarks, so I don't thing there is "the watermark" but usually one would think about one that is half-transparent over some background where content shines through  (not the case here) or something that has some kind of embossed into a thick-paper (also do not really apply here) ... so if thats really desired it should have at laest a transparency of maybe 70% ...

laeubi:
The lines feel to thick here (I think at laest half the current width), and the one marked red should not be there and instead one added in green

grafik

laeubi:
This looks good except that I think the line is still to thick ....

vogella:
Looks fine to me too, but I also like the "old" icon

laeubi:
It looks too thick for this example as well :-)
This makes the image look really "fat" and heavy...

BeckerWdf:
Any other opinions?

BeckerWdf:
> > Any other opinions?
> 
> To which of many proposals ? :-)

To the last one. To the thickness of the strokes. To ....

BeckerWdf:
> [#582 (comment)](https://github.com/eclipse-pde/eclipse.pde/pull/582#issuecomment-1543598485) is OK, but it would be easier to answer if we could have before / after diff.

See: https://github.com/eclipse-pde/eclipse.pde/pull/597

PR #590, opened 2023-04-29
Merge 'unit-tests' GH workflow into 'ci' workflow

HannesWell:
This reduces the number of files and therefore simplifies the build infrastructure.


.github/workflows/ci.yml
```diff
@@ -18,4 +18,7 @@ jobs:
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/checkMergeCommits.yml@master
   build:
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/mavenBuild.yml@master
-
+  public-test-results:
+    if: ${{ always() }}
```
HannesWell, line 22: I wonder if the test publication can/should run always or only if the `build` job succeeded?
If the later is true, then this if is obsolete.

HannesWell:
I wonder if it would be possible to merge the entire [publishTestResults](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/publishTestResults.yml) into the [mavenBuild](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/mavenBuild.yml) workflow?
To me it looks like they should always work together, so if technically possible, it should be just one workflow call.

HannesWell:
> I wonder if it would be possible to merge the entire [publishTestResults](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/publishTestResults.yml) into the [mavenBuild](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/mavenBuild.yml) workflow? To me it looks like they should always work together, so if technically possible, it should be just one workflow call.

It looks like this is not possible for permission/security reasons:
https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches

And I assume threrfore we cannot even merge the invocations of the called workflows -> Closing.

.github/workflows/ci.yml
laeubi: No it must run always because the build can fail with a failing test and we want to cover this case..

PR #603, opened 2023-05-26
Align e4 template for toolbar naming with menu contribution

vogella:
Currently the PDE wizard show a mix of 3.x and e4 namings, that is inconsistent and should be align.
Also make it clear that this work on any platform using 4.x API.


HannesWell (approved):
LGTM

PR #604, opened 2023-06-06
use try-with-resource

jukzi:
to close streams.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiElement.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiElement.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2009 IBM Corporation and others.
+ * Copyright (c) 2008, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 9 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiElement.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2009 IBM Corporation and others.
+ * Copyright (c) 2008, 2023 IBM Corporation and others.
```
vogella, line 2: FYI - updating this year is optional; I personally find it a waste of effort. If you like to do this, that is fine but in case you also find it non-value add, you can skip this step.

vogella:
@iloveeclipse freeze period is already over or is the calender broken again? @HannesWell Maybe we could move the usage  of the freeze check to use something in our repo so that we have better access to it? Like a json file defining the dates.

laeubi:
The calendar is here: https://calendar.google.com/calendar/embed?src=prfk26fdmpru1mptlb06p0jh4s%40group.calendar.google.com&ctz=Europe%2FParis according to that we are in a Quiet period.
If required I think it makes much sense to ask the calendar maintainers to give access to project leads and share a link that can be used to import the calendar e.g. in Thunderbird, and alike...

HannesWell:
> Maybe we could move the usage of the freeze check to use something in our repo so that we have better access to it? Like a json file defining the dates.

I would prefer to have a single source of truth and if the calender is needed for other things, I would just keep it.
If necessary I agree that increasing the number of people with write access would be better.

Btw. the workflow prints a link to the console to access the calender in the Browser.

jukzi:
only unrelated random fail https://github.com/eclipse-pde/eclipse.pde/issues/261 on win
mind to merge?

akurtakov:
Random failing tests are with us for too long and these improvements should help to reduce "randomness" in the long term thus please merge.

PR #606, opened 2023-06-06
BundlePluginBase: fix wrong loop

jukzi:
Codereview: Loop would throw IndexOutOfBoundsException

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bundle/BundlePluginBase.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bundle/BundlePluginBase.java
@@ -183,9 +183,7 @@ public class BundlePluginBase extends PlatformObject implements IBundlePluginBas
 		if (index >= 0) {
 			// copy up to the removed library
 			StringBuilder buffer = new StringBuilder();
-			for (int i = 0; i > index; i++) {
-				buffer.append(value.charAt(i));
-			}
+			buffer.append(value, 0, index);
 			int after = index + name.length();
 			// delete (skip) comma
 			if (after < value.length()) {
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bundle/BundlePluginBase.java
```diff
@@ -183,7 +183,7 @@ private void removeLibrary(IPluginLibrary library, IManifestHeader header) {
 		if (index >= 0) {
 			// copy up to the removed library
 			StringBuilder buffer = new StringBuilder();
-			for (int i = 0; i > index; i++) {
+			for (int i = 0; i < index; i++) {
 				buffer.append(value.charAt(i));
 			}
```
HannesWell, line 188: What about replacing the entire loop just by the following?
```suggestion
			buffer.append(value, 0, index);
```
This would make the code simpler and maybe even faster.

The same could be done for the last loop that copies the remaining part.
It could just be `buffer.append(value, after, value.length());`.

Besides that I wonder if this code works correctly if the the library to remove is the last element. It looks like a trailing comma will then remain.
Or if the library to remove is the only element. Then the headers value will be empty, which is also an error.

HannesWell (commented):
Thanks!
This fix looks good, but I think we could take the opportunity and make it even better.
Do you want to address the comments below too?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bundle/BundlePluginBase.java
jukzi: ok, lets use StringBuilder.append(CharSequence, start, end) also in the loop below. I don't feel confident to propose a fix for the other odds.

HannesWell:
> I'll look into the other discovered issues in a follow up.

Just for the record, after looking at the code again, the affected code path seems to be used only in some very edgy cases I as not able to re-produce. For now, I will not take any further action.
If somebody else is affected by this, a fix is of course welcome.

PR #608, opened 2023-06-06
fix org.eclipse.pde.api.tools.internal.comparator.Delta.hashCode()

jukzi:
Did use identity hashcode.
see https://rules.sonarsource.com/java/RSPEC-2116


HannesWell (approved):
Yes this hashCode ist wrong.
Thanks for the fix!
Havn't checked where this is used in a Set/as Map-key, but if it is, this is a hard to find bug.

PR #610, opened 2023-06-07
Add missing "static" modifier for constants

jukzi:
To reduce memory per instance

### DIFF
--- a/ui/org.eclipse.pde.ui.templates/templates_3.1/helloOSGiService/java/DictionaryImpl.java
+++ b/ui/org.eclipse.pde.ui.templates/templates_3.1/helloOSGiService/java/DictionaryImpl.java
@@ -4,19 +4,19 @@ import java.util.Set;
 
 public class DictionaryImpl implements Dictionary {
 
-	private final Set<String> fWords = Set.of("$word1$", "$word2$", "$word3$");
-	private final String fLanguage = "$language$";
+	private static final Set<String> WORDS = Set.of("$word1$", "$word2$", "$word3$");
+	private static final String LANGUAGE = "$language$";
 
 	public String getLanguage() {
-		return fLanguage;
+		return LANGUAGE;
 	}
 
 	public boolean check(String word) {
-		return fWords.contains(word);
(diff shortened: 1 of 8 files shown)


ui/org.eclipse.pde.ui.templates/templates_3.1/helloOSGiService/java/DictionaryImpl.java
```diff
@@ -5,7 +5,7 @@
 public class DictionaryImpl implements Dictionary {
 
 	private final Set<String> fWords = Set.of("$word1$", "$word2$", "$word3$");
```
HannesWell, line 7: Why not make this static too?

ui/org.eclipse.pde.ui.templates/templates_3.1/helloOSGiServiceComponent/java/DictionaryImpl.java
```diff
@@ -8,7 +8,7 @@
 public class DictionaryImpl implements Dictionary {
 
 	private final Set<String> fWords = Set.of("$word1$", "$word2$", "$word3$");
```
HannesWell, line 10: Static too?

HannesWell (commented):
Looks good, thanks, but it looks like there is room for more.

ui/org.eclipse.pde.ui.templates/templates_3.1/helloOSGiServiceComponent/java/DictionaryImpl.java
jukzi: done

merks:
If they're static, then the names fWords and fLanguage don't follow the Java naming conventions for static final constants...

HannesWell:
But this still needs qualifier bumps in 

- org.eclipse.pde.ds.ui
- org.eclipse.pde.ui.templates

jukzi:
thanks @HannesWell  - aka "early bird" - for all the review an version bumps.

Issue #613, opened 2023-06-10
Error marker being reported on incorrect line number during manifest validation


alshamams:
In the MANIFEST.MF file, the error reporting in some cases is displayed on line 1 by default. Such a situation arises due to the difference in the validation process during :
1) when manifest is first loaded, and
2) when it is inline edited.
In the first case, the validation takes place in the OSGI layers without obtaining the erroneous line number, as is the case with bundle description validation, thereby displaying it on line 1.
In the second case, the line number is also taken into account and the error marker is displayed accordingly.
The former approach can lead to uncertainty with respect to the exact location of the error.

PR #614, opened 2023-06-10
Fix line number in error reporting during manifest validation:

alshamams:
The plug-in manifest is validated in two phases: i) when it is first loaded, ii) when the manifest is inline edited. In the former case, the validation is carried out in OSGI layers, and the exception caught in the pde modules with no bearing on the faulty line number, in case of issues.

This commit tries to parse the exception message in a best effort basis, extracts the header that is subjected for the fault and obtains the line number of the header. It falls back on line 1, in case of parse error.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/613


vik-chand (changes requested):
1) The variable names like t1,t2,hdr can be improved
2) StringTokenizer is legacy, split should be preferred
3) Also a short comment on what you are trying to achieve

alshamams:
In some cases manifest validation takes place in the OSGI layers without obtaining the erroneous line number, as is the case with bundle description validation, thereby displaying it on line 1. This may lead to confusion regarding line number.
This commit tries to parse the exception message, extracts the header from the error message and obtains the line number of the header. It falls back on line 1, in case of parse error.

vik-chand (changes requested):
The declaration at line 149 could be moved down. Update license year if required. 
split could be renamed as splitArray. Also one liner comment on what you are doing.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
```diff
@@ -144,7 +144,18 @@ protected void validate(IProgressMonitor monitor) {
...
-					report(e.getMessage(), 1, CompilerFlags.ERROR, PDEMarkerFactory.CAT_FATAL);
+					// Extract header from the error message and obtain the line
+					// number of the header
+					String msg = e.getMessage();
+					String[] splitArray = msg.split(":"); //$NON-NLS-1$
+					String firstString = splitArray[0];
+					String[] splitToken = firstString.split(" "); //$NON-NLS-1$
+					String lastString = splitToken[splitToken.length - 1];
+					IHeader header = null;
+					header = getHeader(lastString);
```
gireeshpunathil, line 155:
```suggestion
					IHeader header = getHeader(lastString);
```

gireeshpunathil (approved):

vik-chand (approved):

vik-chand:
Please rebase wrt latest.

Issue #624, opened 2023-06-18
back-porting Use the actual system-packages from the JVM for Product/Site export #284 to 4.23


gireeshpunathil:
Context: recently jetty bundles were upgraded in response to security findings. When this was backported to 4.23, launch / export operations (anything that involved PDE) started to fail with reason similar to: `Unsatisfied import package java.io_0.0.0`

steps to simulate the issue:

 - create a simple plugin project - from a template.
 - open the manifest, and add `Import-Package: java.io` to it
 - save it and export it (as deployable plugins)

an error as shown below appears, and the export fails.

![image](https://github.com/eclipse-pde/eclipse.pde/assets/6447530/e12ef787-4c3d-4534-b194-2fd70765a81a)

I have performed `git bisect` to determine that [this commit](https://github.com/eclipse-pde/eclipse.pde/commit/ea34245635594625ba7a6dbbd84fe35046009b4e) will fix the issue.
I see [Validation error when launching child eclipse](https://github.com/eclipse-pde/eclipse.pde/issues/194) and [Fix system package launch validation](https://github.com/eclipse-pde/eclipse.pde/pull/221) from last year that have gone into 4.23, but [Use the actual system-packages from the JVM for Product/Site export](https://github.com/eclipse-pde/eclipse.pde/pull/284) does not look like. Unfortunately, I am unable to back-port this, as I see the pde-build repos have moved, and I see only pre-built jars and classes for the source, and I don't understand how to meaningfully back-port this.

any suggestions are appreciated!

/cc @HannesWell @vik-chand

laeubi:
The "Export Plugins" has always had different issues and it seems no one really maintains this., so I would recommend to use Tycho instead to build your bundles / plugins.

gireeshpunathil:
thanks @laeubi - is this supported by the IDE? I mean, are there UI driven wizards to perform tyco builds? or this has to be done manually in the command line?

laeubi:
You can use m2eclipse or you can add a shortcut to the externals tools menu.

HannesWell:
> You can use m2eclipse or you can add a shortcut to the externals tools menu.

While I agree with Christoph that building Plugins, Features, Products, etc. is done best with Tycho+Maven, using Tycho also requires setting up a corresponding Maven+Tycho build (i.e. at least a root `pom.xml`). If you have done that you can use m2e as described by Christoph to run the build.

[...]

PR #625, opened 2023-06-18
back-port Use the actual system-packages from the JVM for export

gireeshpunathil:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/194 for PDE-build

Refs: https://github.com/eclipse-pde/eclipse.pde/issues/624

Refs: https://github.com/eclipse-pde/eclipse.pde.build/pull/16

(commit of this PR should be paired together with the commit in the referenced PR in eclipse.pde.build repo, as they constitute a single commit logically, but got split up physically due to the the fact that they were in two repos then)

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java
@@ -410,13 +410,7 @@ public class TargetPlatformHelper {
 		if (vm == null || !JavaRuntime.isModularJava(vm)) {
 			return null;
 		}
-
-		String release = null;
-		Map<String, String> complianceOptions = environment.getComplianceOptions();
-		if (complianceOptions != null) {
-			release = complianceOptions.get(JavaCore.COMPILER_COMPLIANCE);
-		}
-
+		String release = environment.getProfileProperties().getProperty(JavaCore.COMPILER_COMPLIANCE);
 		try {
 			Collection<String> packages = new TreeSet<>();
 			File jrtPath = new File(vm.getInstallLocation(), "lib/" + JRTUtil.JRT_FS_JAR); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


gireeshpunathil:
@HannesWell - I split the commit into two and backported one part here, and the other in the pde build repo. I am struggling to build the combination locally (even after cloning both the repos in the same project, I get a lot of errors, trying to get some meaning of it), but the backport should work in theory, as the merge was clean.

Please have a look!

HannesWell:
> @HannesWell - I split the commit into two and backported one part here, and the other in the pde build repo. I am struggling to build the combination locally (even after cloning both the repos in the same project, I get a lot of errors, trying to get some meaning of it), but the backport should work in theory, as the merge was clean.

I have to admit that I don't fully recall how pde.ui (which used to be the name for this repo) and pde.build are build together, but I believe that pde.build used to be independent and pde.ui depended on pde.build, so you should be able to build the latter standalone. A quick look at https://github.com/eclipse-pde/eclipse.pde.build/blob/R4_23_maintenance/org.eclipse.pde.build/pom.xml shows that on the maintenance branch the parent has a SNAPSHOT-version (which is likely not available anymore after the release) and also has a unexpected group- and artifactId.
I recommend you look up how you build the maintenance branches at your build system and replicate the maven commands locally or work it out with @vik-chand. :)

That being said and looking at this change, I remember that the change now in this PR is only a small (probably not noticeable) performance optimization but not a functional change. You can see that by looking in the implementation of the old code.
Therefore I think it is not necessary to back-port it.

gireeshpunathil:
@HannesWell - thanks for all the tips, I am now able to validate the back-port: by building both on the command line, and through eclipse.

 - building in the command line was very tricky: it involved replacing the occurrences of `4.23.0-SNAPSHOT` to `4.29.0-SNAPSHOT` in all the pom.xml files
 - building in eclipse was easy: source both the reps, import `o.e.pde.core` and `o.e.pde.build`, downgrading the `org.eclipse.jdt.launching` bundle version to match what was available in `4.23` timeframe.
 - I am also able to validate that the reported issue is fixed without the change in this PR.

so going to close this PR in lieu of https://github.com/eclipse-pde/eclipse.pde.build/pull/16. let me know if we are on the same page; and also pls review that PR.

thanks once again! lots of learning indeed!

HannesWell:
You're welcome and I'm glad you were successful finding the tricky bits. 👍

Issue #628, opened 2023-06-19
Provisioning agent not found on shutdown


laeubi:
I sometimes get the following error on shutdown of eclipse
```
org.eclipse.core.runtime.CoreException: Provisioning agent not found
	at org.eclipse.pde.internal.core.target.P2TargetUtils.getAgent(P2TargetUtils.java:381)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.getProfileRegistry(P2TargetUtils.java:966)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.cleanOrphanedTargetDefinitionProfiles(P2TargetUtils.java:220)
	at org.eclipse.pde.internal.core.PDECore$1.saving(PDECore.java:329)
	at org.eclipse.core.internal.resources.SaveManager.executeLifecycle(SaveManager.java:393)
	at org.eclipse.core.internal.resources.SaveManager$1.run(SaveManager.java:209)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.resources.SaveManager.broadcastLifecycle(SaveManager.java:212)
	at org.eclipse.core.internal.resources.SaveManager.save(SaveManager.java:1166)
	at org.eclipse.core.internal.resources.SaveManager.save(SaveManager.java:1145)
	at org.eclipse.core.internal.resources.DelayedSnapshotJob.run(DelayedSnapshotJob.java:51)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
!SUBENTRY 2 org.eclipse.pde.core 4 0 2023-06-19 17:05:24.466
!MESSAGE Provisioning agent not found
```
I assume this can happen if the agent is shutdown already, this should not result in an error visible to the user.

PR #629, opened 2023-06-19
Do not fail if cleaning orphan targets fails on shutdown

laeubi:
Cleaning is not an operation that should fail the saving and show an error to the user therefore we only write a warning to the log.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/628

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
@@ -325,8 +325,12 @@ public class PDECore extends Plugin implements DebugOptionsListener {
 		IWorkspace workspace = ResourcesPlugin.getWorkspace();
 		workspace.addSaveParticipant(PLUGIN_ID, new ISaveParticipant() {
 			@Override
-			public void saving(ISaveContext saveContext) throws CoreException {
-				P2TargetUtils.cleanOrphanedTargetDefinitionProfiles();
+			public void saving(ISaveContext saveContext) {
+				try {
+					P2TargetUtils.cleanOrphanedTargetDefinitionProfiles();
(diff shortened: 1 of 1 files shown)


HannesWell (approved):
Looks reasonable to me.
I also noticed this message a few time recently using the I-build, I wonder why this now pops up.

laeubi:
> I also noticed this message a few time recently using the I-build, I wonder why this now pops up.

I have not analyzed this in detail but as it happens on shutdown I assume the service is already gone at this point in time when the save action is triggered, we should probably have an eye on it but until now I can't see any problems beside this message, and we should until then not bother the user, I'll check if maybe something is missing from https://github.com/eclipse-equinox/p2/pull/287 but I would assume that then some test would fail...

laeubi:
I think the error message is also misleading here, actually it seems PDE can't find the `IProvisioningAgentProvider` ...

I can see the service is there bot no one seems using it:

```
{org.eclipse.equinox.p2.core.IProvisioningAgentProvider}={service.id=74, service.bundleid=2196, service.scope=bundle, service.ranking=10000, osgi.ds.satisfying.condition.target=(osgi.condition.id=true), component.name=org.eclipse.oomph.p2.core.provisioning.agent, component.id=38}
  "Registered by bundle:" org.eclipse.oomph.p2.core_1.25.0.v20230509-0732 [2196]
  "No bundles using service."
```
... and ...

```
{org.eclipse.equinox.p2.core.IProvisioningAgentProvider}={service.id=179, service.bundleid=2525, service.scope=bundle, osgi.ds.satisfying.condition.target=(osgi.condition.id=true), component.name=org.eclipse.equinox.p2.di.agentProvider, component.id=108}
  "Registered by bundle:" org.eclipse.equinox.p2.core_2.10.100.v20230617-1417 [2525]
  "No bundles using service."
```

Issue #555, opened 2023-04-01
[Mac] testImport no-configuration (org.eclipse.ui.tests.smartimport.PlainJavaProjectTest) fails


HannesWell:
In the Mac workflow the PlainJavaProjectTest fails sometimes with the following error.
```
No matching widget found with Matcher matching when all matchers match: [Matcher matching widget with the same type as or type extending class org.eclipse.swt.widgets.Button, Widget matcher matching widgets with style: 8, Matcher matching widgets with text that without mnenomic matches: is "OK"]
class org.eclipse.swt.widgets.Shell[shell: 'Event Details'] with text 'Event Details'
 class org.eclipse.swt.widgets.Composite[shell: 'Event Details']
  class org.eclipse.swt.widgets.Composite[shell: 'Event Details']
   class org.eclipse.swt.custom.SashForm[shell: 'Event Details']
    class org.eclipse.swt.widgets.Composite[shell: 'Event Details']
     class org.eclipse.swt.widgets.Composite[shell: 'Event Details']
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with text 'Plug-in:'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'Plug-in:' with text 'org.eclipse.urischeme'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'org.eclipse.urischeme' with text 'Severity:'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'Severity:' with text ''
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'Severity:' with text 'Error'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'Error' with text 'Date:'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label 'Date:' with text '4/1/23, 15:02:01.840'
      class org.eclipse.swt.widgets.Label[shell: 'Event Details'] with label '4/1/23, 15:02:01.840' with text 'Message:'
      class org.eclipse.swt.widgets.Text[shell: 'Event Details'] with label 'Message:' with text 'file:/Users/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.ui.tests.smartimport/target/work/Contents/Info.plist'
     class org.eclipse.swt.widgets.Composite[shell: 'Event Details'] with label 'Message:'
[...]

Issue #553, opened 2023-04-01
org.eclipse.pde.api.tools.builder.tests.compatibility.BundleVersionTests "the baseline 'workspace' is disposed"


laeubi:
Sometimes the test fails with

```
Component 'exportedBundle Plug-in' in the baseline 'workspace' is disposed
org.eclipse.core.runtime.CoreException: Component 'exportedBundle Plug-in' in the baseline 'workspace' is disposed
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.baselineDisposed(BundleComponent.java:1259)
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.getBundleDescription(BundleComponent.java:980)
	at org.eclipse.pde.api.tools.internal.model.BundleComponent.createApiDescription(BundleComponent.java:388)
	at org.eclipse.pde.api.tools.internal.model.Component.getApiDescription(Component.java:106)
	at org.eclipse.pde.api.tools.builder.tests.ApiBuilderTest.exportApiComponent(ApiBuilderTest.java:364)
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleVersionTests.setupTest(BundleVersionTests.java:162)
	at org.eclipse.pde.api.tools.builder.tests.compatibility.BundleVersionTests.test001(BundleVersionTests.java:192)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.runTest(TestCase.java:983)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)
[...]

PR #632, opened 2023-06-23
Factory for SAXParser parsing XML without DOCTYPE

jukzi:
External entity resolution is not supported by PDE (see PDECoreMessages.XMLErrorReporter_ExternalEntityResolution) but still the SAXParser did follow external links where DefaultHandler.resolveEntity was not overwritten.
At many places PDE already overwrote DefaultHandler.resolveEntity to prevent external resolution. With the new configuration that method is not even called anymore.

This change offers and uses a configuration that
* reports an Exception if .xml contains DOCTYPE or
* does just ignore external references (as a fall back if the exception would show up to cause trouble).

Also the caching of used parsers in possibly other threads is removed because the SAXParser is not guaranteed to be thread-safe. Only the factory is reused, because that is effectively final after creation. Reusing SAXParser is not a big help nowadays - see XmlParserFactoryTest.main(String[]) for performance test. In my measurement successive parser creations takes only ~ 0.06 ms.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/XmlParserFactoryTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/XmlParserFactoryTest.java
@@ -1,0 +1,203 @@
+/*******************************************************************************
+ *  Copyright (c) 2023 Joerg Kubitz and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.pde.core.tests.internal;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.net.SocketException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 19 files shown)


jukzi:
random fails:
BundleVersionTests https://github.com/eclipse-pde/eclipse.pde/issues/553
PlainJavaProjectTest https://github.com/eclipse-pde/eclipse.pde/issues/555

jukzi:
random failing ApiDescriptionProcessorTests https://github.com/eclipse-pde/eclipse.pde/issues/647

jukzi:
ignoring random failing tests

HannesWell:
Nice job on this and the other PRs regarding XML-parser creation.

Since the `XmlTransformerFactory` and `XmlDocumentBuilderFactory` class only contain one not too big method, I think it would be good to move them to the `XmlParserFactory` introduced in this change. The method names are already very expressive so I don't think it is necessary to have them in a dedicated class.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/XmlParserFactoryTest.java
```diff
@@ -0,0 +1,203 @@
...
+				public void run() {
+					try (Socket socket = serverSocket.accept()) {
+						try (BufferedReader in = new BufferedReader(new InputStreamReader(socket.getInputStream()))) {
+							String firstLine = in.readLine();
+							try (OutputStream outputStream = socket.getOutputStream()) {
+								outputStream.write("HTTP/1.1 200 OK\r\n".getBytes(StandardCharsets.UTF_8));
+							}
+							assertTrue(firstLine, firstLine.startsWith("GET"));
+							assertFalse("Server received secret: " + firstLine, firstLine.contains("secret")); // var3
+							assertFalse("Server was contacted", true);
```
HannesWell, line 137: Looks like this is always throwing an exception.
Can you tell why the test passes nevertheless?

jukzi: One test expects an exception (but does not come so far because of another exception) while another test makes sure the code is not called at all.

jukzi:
> Since the `XmlTransformerFactory` and `XmlDocumentBuilderFactory` class only contain one not too big method, I think it would be good to move them to the `XmlParserFactory` introduced in this change. The method names are already very expressive so I don't think it is necessary to have them in a dedicated class.

https://github.com/eclipse-pde/eclipse.pde/pull/667

PR #634, opened 2023-06-23
DocumentBuilderFactory: Parse xml without external entity resolution

jukzi:
External entity resolution is not supported by PDE (see PDECoreMessages.XMLErrorReporter_ExternalEntityResolution). PDE should not contact or even inject external sources into Documents.


jukzi:
ignoring random failing PlainJavaProjectTest https://github.com/eclipse-pde/eclipse.pde/issues/555

PR #635, opened 2023-06-23
PDEState: unconditionally close stream

jukzi:
For eager return on manifestStream == null jarFile was not closed.

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -318,45 +318,29 @@ public class PDEState implements IPDEBuildConstants, IBuildPropertiesConstants {
 
 	//Return a dictionary representing a manifest. The data may result from plugin.xml conversion  
 	private Dictionary<String, String> basicLoadManifest(File bundleLocation) {
-		InputStream manifestStream = null;
-		ZipFile jarFile = null;
 		try {
 			if ("jar".equalsIgnoreCase(IPath.fromOSString(bundleLocation.getName()).getFileExtension()) && bundleLocation.isFile()) { //$NON-NLS-1$
-				jarFile = new ZipFile(bundleLocation, ZipFile.OPEN_READ);
-				ZipEntry manifestEntry = jarFile.getEntry(JarFile.MANIFEST_NAME);
-				if (manifestEntry != null) {
-					manifestStream = jarFile.getInputStream(manifestEntry);
+				try (ZipFile jarFile = new ZipFile(bundleLocation, ZipFile.OPEN_READ)) {
+					ZipEntry manifestEntry = jarFile.getEntry(JarFile.MANIFEST_NAME);
+					if (manifestEntry != null) {
+						try (InputStream manifestStream = jarFile.getInputStream(manifestEntry)) {
+							if (manifestStream != null) {
+								return new Hashtable<>(ManifestElement.parseBundleManifest(manifestStream, null));
+							}
+						}
+					}
 				}
 			} else {
-				manifestStream = new BufferedInputStream(new FileInputStream(new File(bundleLocation, JarFile.MANIFEST_NAME)));
+				try (InputStream manifestStream = new FileInputStream(new File(bundleLocation, JarFile.MANIFEST_NAME))) {
+					return new Hashtable<>(ManifestElement.parseBundleManifest(manifestStream, null));
+				}
 			}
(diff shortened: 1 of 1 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -317,45 +317,32 @@ private String getQualifierPropery(String bundleLocation) {
...
+							if (manifestStream != null) {
+								result.putAll(ManifestElement.parseBundleManifest(manifestStream, null));
+								return result;
+							}
+						}
+					}
 				}
 			} else {
-				manifestStream = new BufferedInputStream(new FileInputStream(new File(bundleLocation, JarFile.MANIFEST_NAME)));
+				try (InputStream manifestStream = new BufferedInputStream(new FileInputStream(new File(bundleLocation, JarFile.MANIFEST_NAME)))) {
```
HannesWell, line 335: ManifestElement.parseBundleManifest() already wraps the stream into a BufferedStream. No need to wrap it twice.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -317,45 +317,32 @@ private String getQualifierPropery(String bundleLocation) {
...
 		try {
 			if ("jar".equalsIgnoreCase(IPath.fromOSString(bundleLocation.getName()).getFileExtension()) && bundleLocation.isFile()) { //$NON-NLS-1$
-				jarFile = new ZipFile(bundleLocation, ZipFile.OPEN_READ);
-				ZipEntry manifestEntry = jarFile.getEntry(JarFile.MANIFEST_NAME);
-				if (manifestEntry != null) {
-					manifestStream = jarFile.getInputStream(manifestEntry);
+				try (ZipFile jarFile = new ZipFile(bundleLocation, ZipFile.OPEN_READ)) {
+					ZipEntry manifestEntry = jarFile.getEntry(JarFile.MANIFEST_NAME);
+					if (manifestEntry != null) {
+						try (InputStream manifestStream = jarFile.getInputStream(manifestEntry)) {
```
HannesWell, line 326: "Closing this ZIP file will close all of the input streams previously returned by invocations of the `getInputStream`".
So I think we can skip the nested try in this case. :)

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -317,45 +317,32 @@ private String getQualifierPropery(String bundleLocation) {
 
 	//Return a dictionary representing a manifest. The data may result from plugin.xml conversion  
 	private Dictionary<String, String> basicLoadManifest(File bundleLocation) {
```
HannesWell, line 319: I think it would simplify the method if the processing of the input stream would be extraced into a separat method and then called from both locations.
Together with the other two suggestions the code would then be:
```
	private Dictionary<String, String> basicLoadManifest(File bundleLocation) {
		try {
			if ("jar".equalsIgnoreCase(IPath.fromOSString(bundleLocation.getName()).getFileExtension()) && bundleLocation.isFile()) { //$NON-NLS-1$
				try (ZipFile jarFile = new ZipFile(bundleLocation, ZipFile.OPEN_READ)) {
					ZipEntry manifestEntry = jarFile.getEntry(JarFile.MANIFEST_NAME);
					if (manifestEntry != null) {
						return readManifestEntries(jarFile.getInputStream(manifestEntry));
					}
				}
			} else {
				try (InputStream manifestStream = new FileInputStream(new File(bundleLocation, JarFile.MANIFEST_NAME))) {
					return readManifestEntries(manifestStream);
				}
			}
		} catch (IOException | BundleException e) {
			//ignore
		}

		//It is not a manifest, but a plugin or a fragment
		return null;
	}

	private Dictionary<String, String> readManifestEntries(InputStream manifestStream) throws IOException, BundleException {
		Hashtable<String, String> result = new Hashtable<>();
		result.putAll(ManifestElement.parseBundleManifest(manifestStream, null));
		return result;
	}
```

HannesWell (commented):
Looks good in general.
A few suggestion for further improvements below.

In a follow up, one could get rid of Hashtable/Dictionary here. It is far too often used in PDE in places were it is not necessary.

jukzi:
ignoring random fail PlainJavaProjectTest https://github.com/eclipse-pde/eclipse.pde/issues/555

Issue #554, opened 2023-04-01
[Win] Unknown test (org.eclipse.pde.ui.tests.launcher.FeatureBasedLaunchTest) failed 


HannesWell:
In the Windows workflow the FeatureBasedLaunchTest recently started to fail with the following error sometimes.
```
An internal error occurred during: "Load Target Platform".
java.lang.AssertionError: An internal error occurred during: "Load Target Platform".
	at org.eclipse.pde.ui.tests.util.TargetPlatformUtil.loadAndSetTarget(TargetPlatformUtil.java:112)
	at org.eclipse.pde.ui.tests.util.TargetPlatformUtil.createAndSetTarget(TargetPlatformUtil.java:148)
	at org.eclipse.pde.ui.tests.util.TargetPlatformUtil.setRunningPlatformSubSetAsTarget(TargetPlatformUtil.java:91)
	at org.eclipse.pde.ui.tests.util.TargetPlatformUtil.setRunningPlatformAsTarget(TargetPlatformUtil.java:80)
	at org.eclipse.pde.ui.tests.launcher.AbstractLaunchTest.setupTargetPlatform(AbstractLaunchTest.java:62)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.RunBefores.invokeMethod(RunBefores.java:33)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
	at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:271)
	at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:271)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
[...]

PR #636, opened 2023-06-23
BndProjectManager: close Processor

jukzi:
Processor is a Closeable that was not closed.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/BndProjectManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/BndProjectManager.java
@@ -91,13 +91,14 @@ public class BndProjectManager {
 	}
 
 	static synchronized Workspace getWorkspace() throws Exception {
-		Processor run = new Processor();
-		run.setProperty(Constants.STANDALONE, TRUE);
-		IPath path = PDECore.getDefault().getStateLocation().append(Project.BNDCNF);
 		if (workspace == null) {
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
Don't close the processor it is passed to the workspace and managed by this!

jukzi:
As far as i can see the workspace only calls .getProperty() / getProperties() on it.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/BndProjectManager.java
```diff
@@ -91,15 +91,16 @@ private static void setupProject(Project bnd, IProject project) throws CoreExcep
...
 
 	static synchronized Workspace getWorkspace() throws Exception {
-		Processor run = new Processor();
-		run.setProperty(Constants.STANDALONE, TRUE);
-		IPath path = PDECore.getDefault().getStateLocation().append(Project.BNDCNF);
-		if (workspace == null) {
-			workspace = Workspace.createStandaloneWorkspace(run, path.toFile().toURI());
-			workspace.addBasicPlugin(TargetRepository.getTargetRepository());
-			workspace.refresh();
+		try (Processor run = new Processor()) {
```
laeubi, line 94: can you move the creation of the processor inside the `if (workspace == null) `

laeubi (approved):
Looks good

jukzi:
ignoring random fail FeatureBasedLaunchTest https://github.com/eclipse-pde/eclipse.pde/issues/554

PR #639, opened 2023-06-23
fix BuildStamps.incBuildStamp(IProject)

jukzi:
postIncrement an reassignment is a NOOP. Stamp was never increased!


gireeshpunathil (approved):

iloveeclipse:
This bug exists since ever, unbelievable :-( 

What's more interesting, the code that uses this feature can probably be just removed, since we haven't seen any unexpected problems with that. 

I wonder more if that would cause the ancient PDE code that uses this broken timestamps to cause regressions now, since no one ever seen this code properly answering.

Note: I'm not against the merge.

jukzi:
Let's try what happens with code corrected.

HannesWell:
This is indeed a very nice catch.

> Let's try what happens with code corrected.

I traced the callers and assume we get a few more API-builds that did not happen before.

PR #640, opened 2023-06-23
fix TarFile.TarInputStream.read() to return unsigned byte

jukzi:
https://rules.sonarsource.com/java/RSPEC-4517


HannesWell (approved):
LGTM

PR #641, opened 2023-06-23
fix FeatureGenerator.addLauncher()

jukzi:
Even though Entry.equals(String) was defined String.equals(Entry) == false. It's pure implementation detail of the Set created in FeatureGenerator.createSet(String[]) if s.equals(e) or e.equals(s) is called and also Entry.hashCode() does not match String.hashCode();


jukzi:
I don't know how to test this well, because for me, locally, PDEBuildTestSuite shows so much errors anyways
image

PR #642, opened 2023-06-24
Support auto-start configuration of transitively included bundles in Products

HannesWell:


laeubi:
This is very useful, until now one has to edit the product config in source to archive this.

HannesWell:
> This is very useful, until now one has to edit the product config in source to archive this.

Yes, this bother me as well (at the moment in M2E), which is why I added this. :)

merks:
Indeed this seems a useful feature for the launcher, but I'm slightly confused by the "product config" comment because products do need to configure this themselves properly in order to function as a product.  Probably I misunderstood something...

HannesWell:
> Indeed this seems a useful feature for the launcher, but I'm slightly confused by the "product config" comment because products do need to configure this themselves properly in order to function as a product. Probably I misunderstood something...

This PR essentially does two things:
1. It allows users to select Plugins for auto-start level configuration in the `Configuration` Section of the Product-Editor that are only transitively included.
In the following screenshot that is for example `slf4j-api` (not that this specific bundle would be necessary), if your product e.g. only consists of the e4.rcp feature.
grafik
2. It supports that the auto-start levels configured for only transitively included bundles are written into the a launch-configuration created/updated when you launch a Feature-based Product from the Product editor.

Now writing about it, I noticed that the second point is probably not working for Plugin-based Products... I'll check that.

merks:
Very cool indeed!  You'll want to write a new and noteworthy about it.  😁

HannesWell:
> Very cool indeed! You'll want to write a new and noteworthy about it. 😁

So many news to write 😄 

> Now writing about it, I noticed that the second point is probably not working for Plugin-based Products... I'll check that.

This is now fixed. While working on this I noticed one reason why users had to list dependencies in their Plugin-Based products was that such plugins with auto-start configuration were not added to a launch-config when they were not listed. This is now also changed (I think for the better).

PR #643, opened 2023-06-24
Set recommended start-levels only for (transitively) included bundles

HannesWell:
In the configuration section of the Product editor only add a recommended auto-start configuration for a bundle, if it is (transitively) included into the product. Before this change a recommended configuration was always added, even if the bundle was not part of a launched product.


HannesWell:
Should be mentioned in the N&N together with https://github.com/eclipse-pde/eclipse.pde/pull/642.

PR #646, opened 2023-06-26
Use new ILog.of(...) method instead of Platform.getLog(...)

laeubi:


laeubi:
Failures reported here:
- https://github.com/eclipse-pde/eclipse.pde/issues/648
- https://github.com/eclipse-pde/eclipse.pde/issues/647

PR #658, opened 2023-07-06
Remove 'Eclipse-LazyStart' headers in favor of 'Bundle-ActivationPolicy'

HannesWell:
Replace the use of the legacy and Eclipse-specific MANIFEST.MF header 'Eclipse-LazyStart: true' by the OSGi standard compliant header 'Bundle-ActivationPolicy: lazy' (which is already used in most cases.

@tjwatson do you know any reason to keep the old 'Eclipse-LazyStart: true' header?

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
@@ -4,7 +4,6 @@ Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.api.tools.ui; singleton:=true
 Bundle-Version: 1.3.100.qualifier
 Bundle-Localization: plugin
-Eclipse-LazyStart: true
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.27.0,4.0.0)",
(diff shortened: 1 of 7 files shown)


tjwatson:
For PDE, no.  For runtime, it is cheap to keep allowing it.

laeubi:
@HannesWell what do you think about adding a warning marker to the manifest for this header?

akurtakov:
There already is such a warning. Maybe it's not enabled by default but I remember seeing it and thus doing such conversion (IIRC there was even a quickfix).

HannesWell:
> For PDE, no. For runtime, it is cheap to keep allowing it.

Actually I was referring to if there is a reason to keep it in the Manifest. Since support for that is already there and I see no problem with that code a.t.m. and the header is still in use (at least not fully removed from the SDK code-base), I think we should keep it to help users migrating.



> @HannesWell what do you think about adding a warning marker to the manifest for this header?

I think there should be one. :)

> There already is such a warning. Maybe it's not enabled by default but I remember seeing it and thus doing such conversion (IIRC there was even a quickfix).

There is indeed already one, but it is not shown, if a `Bundle-ActivationPolicy` header is also present in the Manifest:
https://github.com/eclipse-pde/eclipse.pde/blob/f0eeea56119fafadb9dcf36a4a50459f394f9965/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java#L1594

There is also a warning for the predecessor of `Eclipse-LazyStart` called `Eclipse-AutoStart`, but it is also not shown if `Bundle-Bundle-ActivationPolicy` is present:
https://github.com/eclipse-pde/eclipse.pde/blob/f0eeea56119fafadb9dcf36a4a50459f394f9965/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java#L1576

I think it would be better to always show that warning, regardless of `Bundle-ActivationPolicy` is present or not. If one then migrates to `Bundle-ActivationPolicy` and didn't notice it is present already one gets a warning about duplicated headers.

PR #660, opened 2023-07-10
Replace use-cases of internal PDE build Utils by standard Java methods

HannesWell:
Simplifies the code and also makes it a little bit faster.

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ProductGenerator.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ProductGenerator.java
@@ -20,6 +20,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -177,7 +178,7 @@ public class ProductGenerator extends AbstractScriptGenerator {
 		if (initialInf.exists()) {
 			//copy over initial contents
 			try {
-				StringBuffer inf = Utils.readFile(initialInf);
+				String inf = Files.readString(initialInf.toPath(), StandardCharsets.ISO_8859_1);
 				buffer.append(inf);
 				buffer.append('\n');
 			} catch (IOException e) {
@@ -189,7 +190,7 @@ public class ProductGenerator extends AbstractScriptGenerator {
 
 		try {
 			Files.createDirectories(Path.of(root));
-			Files.writeString(Path.of(root, "p2.inf"), buffer.toString()); //$NON-NLS-1$
+			Files.writeString(Path.of(root, "p2.inf"), buffer.toString(), StandardCharsets.ISO_8859_1); //$NON-NLS-1$
 		} catch (IOException e) {
 			return false;
 		}
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
@@ -827,52 +827,6 @@ public final class Utils implements IPDEBuildConstants, IBuildPropertiesConstant
 	}
 
 	/**
-	 * 
-	 * @param buf
-	 * @param start
-	 * @param target
-	 * @return int
-	 */
-	static public int scan(StringBuffer buf, int start, String target) {
-		return scan(buf, start, new String[] {target});
-	}
-
-	/**
-	 * 
-	 * @param buf
-	 * @param start
-	 * @param targets
-	 * @return int
-	 */
-	static public int scan(StringBuffer buf, int start, String[] targets) {
-		for (int i = start; i < buf.length(); i++) {
-			for (String target : targets) {
-				if (i < buf.length() - target.length()) {
-					String match = buf.substring(i, i + target.length());
-					if (target.equals(match))
-						return i;
-				}
-			}
-		}
-		return -1;
-	}
-
-	/**
-	 * Return a buffer containing the contents of the file at the specified location.
-	 * 
-	 * @param target the file
-	 * @return StringBuffer
-	 * @throws IOException
-	 */
-	static public StringBuffer readFile(File target) throws IOException {
-		return new StringBuffer(new String(Files.readAllBytes(target.toPath())));
-	}
-
-	static public StringBuffer readFile(InputStream stream) throws IOException {
-		return new StringBuffer(new String(stream.readAllBytes()));
-	}
-
-	/**
 	 * Custom build scripts should have their version number matching the
 	 * version number defined by the feature/plugin/fragment descriptor.
 	 * This is a best effort job so do not worry if the expected tags were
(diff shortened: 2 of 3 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
```diff
@@ -884,18 +839,18 @@ static public StringBuffer readFile(InputStream stream) throws IOException {
 	 *
 	 */
 	public static void updateVersion(File buildFile, String propertyName, String version) throws IOException {
-		StringBuffer buffer = readFile(buildFile);
-		int pos = scan(buffer, 0, propertyName);
+		StringBuilder buffer = new StringBuilder(Files.readString(buildFile.toPath(), Charset.defaultCharset()));
```
laeubi, line 842: Is a stringbuilder actually neded here?

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ProductGenerator.java
```diff
@@ -176,7 +177,7 @@ public boolean generateP2Info() throws CoreException {
 		if (initialInf.exists()) {
 			//copy over initial contents
 			try {
-				StringBuffer inf = Utils.readFile(initialInf);
+				String inf = Files.readString(initialInf.toPath(), Charset.defaultCharset());
```
jukzi, line 180: When explicitly using a charset it should be thought of whether that charset is really the right one to use. using no charset at least gives a static code analysis warning. In this case my best guess is that it should be a fixed, ASCII compatible charset. or even only 7 bit Ascii if it's not agreed to use UTF-8.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
jukzi: note that JDKs also offers an two valued indexOf for String.
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#indexOf(java.lang.String,int)

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/builder/SourceGenerator.java
```diff
@@ -422,27 +426,27 @@ private void replaceXMLAttribute(String location, String tag, String attr, Strin
 		if (!featureFile.exists())
 			return;
 
-		StringBuffer buffer = null;
+		StringBuilder buffer = null;
 		try {
-			buffer = Utils.readFile(featureFile);
+			buffer = new StringBuilder(Files.readString(featureFile.toPath(), Charset.defaultCharset()));
```
jukzi, line 431: It's bad to have several places to define the charset to be used, when its supposed to use same charset.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
HannesWell: > Is a stringbuilder actually neded here?

Thinking about it, not really. At least not until the very end.
Does anybody know a more elegant way to replace a substring based on indices?

Btw. I also traced who is writing the file and seems to the AntScript class, which uses UTF-8, so we can just the no-argument overload.
https://github.com/eclipse-pde/eclipse.pde/blob/c8754be8f6e23e9c5aebf6bf11b59515e98d262d/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ant/AntScript.java#L50C56-L50C78

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ProductGenerator.java
HannesWell: OK. I'll try to find out what charset is used to write those files.

But for me, passing no charset does not produce a warning for me, at least with Sonar-Lint.

HannesWell: Looks like this file is only written with `Files.writeString()` below in this method.
That being said, the file is also loaded as Properties file, so it should be `ISO_8859_1` encoded.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/builder/SourceGenerator.java
HannesWell: This is a `feature.xml` file and can have any encoding, depending on what a user sets up. So we should find that out and then use it when writing the feature file below.
Querying the workspace for the file's charset requires an additional dependency to the resources-plugin. I cannot tell if this is desired. Does anybody have a better idea?

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/Utils.java
jukzi: great.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/builder/SourceGenerator.java
jukzi: The best way to change XML is javax.xml.transform.TransformerFactory

jukzi: otherwise just UTF_8 -  i have never seen any non-UTF_8 xml

jukzi: damn https://bugs.eclipse.org/bugs/show_bug.cgi?id=98069 - some people write xml with other encoding

HannesWell: > The best way to change XML is javax.xml.transform.TransformerFactory

I'll try that.

> damn https://bugs.eclipse.org/bugs/show_bug.cgi?id=98069 - some people write xml with other encoding

Too bad, just reading/writing it using UTF-8, would have been very convenient.

jukzi:
lgtm

HannesWell:
Thanks.
I think I also found a solution for the encoding issue, but I have to verify/test that it really reads the encoding declared in the XML file as expected.

HannesWell:
> I think I also found a solution for the encoding issue, but I have to verify/test that it really reads the encoding declared in the XML file as expected.

Just checked it and it really reads the encoding declared in the XML header. And even better, Eclipse warns a user if one declares in the project preferences for the file another encoding than in the file header.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/builder/SourceGenerator.java
```diff
@@ -463,20 +472,29 @@ private void replaceXMLAttribute(String location, String tag, String attr, Strin
...
 			} catch (IOException e) {
 				//ignore
 			}
 		}
 	}
 
+	private static Charset getFileEncoding(Path file) throws IOException, XMLStreamException {
+		try (InputStream is = Files.newInputStream(file)) {
+			XMLInputFactory factory = XMLInputFactory.newFactory();
+			factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
```
HannesWell, line 492: @jukzi did you recall if PDE features need doctype resolution at all?

jukzi:
Not at all.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/builder/SourceGenerator.java
HannesWell: > On the other hand, since this parser is only used to read the encoding, the parser could not come into the situation to attempt to resolve them at all.

Just checked it and not supporting DTD at all is not a problem even if present, as expected.

PR #667, opened 2023-07-12
Refactor XML handling

jukzi:
class PDEXmlProcessorFactory combines classes:
 XmlDocumentBuilderFactory,
 XmlParserFactory,
 XmlTransformerFactory

[performance] reuse single DocumentBuilderFactory instance see PDEXmlProcessorFactoryTest.main(String[]) - 100x faster


ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,344 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+import org.xml.sax.SAXParseException;
+import org.xml.sax.helpers.DefaultHandler;
+
+public class PDEXmlProcessorFactoryTest {
+
+	final List<Path> tmpFiles = new ArrayList<>();
```
HannesWell, line 59:
```suggestion
	@Rule
	TemporaryFolder tempFolder = new TemporaryFolder();
```
Instead of cleaning up files manually a `org.junit.rules.TemporaryFolder` can do that for us.

In the code below you can then create new files in the temp-folder for example with `tempFolder.newFile("test.txt")`.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,344 @@
...
+					+ "<!ENTITY % var2 \"<!ENTITY var3 SYSTEM 'http://localhost:" + localPort + "/?%var1;'>\">\n" //
+					+ "%var2;\n";
+			Files.writeString(tempDtd, dtdContent);
+			StringBuilder sb = new StringBuilder();
+			sb.append("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n");
+			URI dtdURi = tempDtd.toUri();
+			String dtdUrl = dtdURi.toURL().toString();
+			sb.append("<!DOCTYPE var1 SYSTEM \"" + dtdUrl + "\">\n");
+			sb.append("<Body>&var3;&var4;</Body>");
+			String xml = sb.toString();
```
HannesWell, line 240: Text-Block + String.format again:
```suggestion
			String xml = String.format("""
					<?xml version="1.0" encoding="UTF-8"?>
					<!DOCTYPE var1 SYSTEM "%s">
					<Body>&var3;&var4;</Body>""", tempDtd.toUri().toURL());
```

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,344 @@
...
+		try {
+			Path tempSecret = Files.createTempFile("test", ".txt");
+			tmpFiles.add(tempSecret);
+			Files.writeString(tempSecret, "secret");
+			Path tempDtd = Files.createTempFile("test", ".dtd");
+			tmpFiles.add(tempDtd);
+			String dtdContent = "<!ENTITY % var1 SYSTEM \"" + tempSecret.toUri().toURL() + "\">\n" //
+					+ "<!ENTITY var4 SYSTEM \"" + tempSecret.toUri().toURL() + "\">\n" //
+					+ "<!ENTITY % var2 \"<!ENTITY var3 SYSTEM 'http://localhost:" + localPort + "/?%var1;'>\">\n" //
+					+ "%var2;\n";
```
HannesWell, line 232:
```suggestion
			String dtdContent = String.format("""
					<!ENTITY % var1 SYSTEM "%s">
					<!ENTITY var4 SYSTEM "%s">
					<!ENTITY % var2 "<!ENTITY var3 SYSTEM 'http://localhost:%s/?%var1;'>">
					%var2;
					""", tempSecret.toUri().toURL(), tempSecret.toUri().toURL(), localPort);
```
A text-block and String.format would make this simpler to read. Once String-Templates are available in a future Java version and we can use that version, it can be simply replaced to use a String-template.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,344 @@
...
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	private InputStream createNormalXml(int localPort) {
+		StringBuilder sb = new StringBuilder();
+		sb.append("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n");
+		sb.append("<Body>hello</Body>");
+		String xml = sb.toString();
```
HannesWell, line 251:
```suggestion
		String xml = """
				<?xml version="1.0" encoding="UTF-8"?>
				<Body>hello</Body>""";
```

HannesWell (commented):
Looks much better 👍🏽
I have a few remarks below how the code could be further improved.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDEXmlProcessorFactory.java
```diff
@@ -0,0 +1,166 @@
...
+	private static final SAXParserFactory SAX_FACTORY_ERROR_ON_DOCTYPE_NS = createSAXFactoryWithErrorOnDOCTYPE(true);
+	private static final SAXParserFactory SAX_FACTORY_IGNORING_DOCTYPE = createSAXFactoryIgnoringDOCTYPE();
+
+	/**
+	 * Creates TransformerFactory which throws TransformerException when
+	 * detecting external entities.
+	 *
+	 * @return javax.xml.transform.TransformerFactory
+	 */
+	public static TransformerFactory createTransformerFactoryWithErrorOnDOCTYPE() {
```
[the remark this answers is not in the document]
jukzi: I don't mind that code saving much: it reduces flexibility to configure transformer. (Also taking into account to copy&paste this class to other projects) A bigger (performance) improvement would be to not need to recreate factory. But as far as i have seen it's not used frequently anyway.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDEXmlProcessorFactory.java
```diff
@@ -0,0 +1,166 @@
...
+		SAXParserFactory f = SAXParserFactory.newInstance();
+		try {
+			// force org.xml.sax.SAXParseException for any DOCTYPE:
+			f.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true); //$NON-NLS-1$
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+		return f;
+	}
+
```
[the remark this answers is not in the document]
jukzi: I would like to keep it as a fall back if we discover a case where an exception is too harsh.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
jukzi: interesting. i will consider that next time. for now i am happy with the non-magic solution.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,344 @@
...
+			String message = e.getMessage();
+			assertTrue(message, message.contains("DOCTYPE"));
+		}
+	}
+
+	@Test
+	public void testDocumentBuilderFactoryWithoutExternalEntity() throws Exception {
+		DocumentBuilderFactory documentBuilderFactory = PDEXmlProcessorFactory
+				.createDocumentBuilderFactoryWithErrorOnDOCTYPE();
+		testParseXmlWithExternalEntity(documentBuilderFactory.newDocumentBuilder(), this::createNormalXml);
```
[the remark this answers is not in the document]
jukzi: we'll need createDocumentBuilderFactoryWithErrorOnDOCTYPE on other projects

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
jukzi: it's not obvious to me as a reader, that the "%"  does not need to be escaped and harder to understand which parameter lands where. I prefer the old-school thing.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
jukzi: i am not familiar to read String.format

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
jukzi: Ok readable :-)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDEXmlProcessorFactory.java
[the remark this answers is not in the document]
HannesWell: For which one?
We should not keep internal methods for other projects, because if it is not used it can happen that one day in the future somebody comes along and cleans up this class and then that other project suddenly breaks. Everything that is not API can be removed at any time. Furthermore even a method that is not called has maintanance cost, because somebody reading this class has to understand what it is doing and then wonders who's calling it and why no code in PDE does that. In that regard no code is the best code.


And at least for the Eclipse SDK (Equinox, Platform, JDT and PDE), PDE is at the end of the dependency hierarchy, so other projects are outside the SDK? Or, since you mentioned copy-pasting, do you want to add this class to other Eclipse SDK projects as it is?
Then I suggest to add it to one at the top of the hierarchy, probably equinox or eclipse(-runtime), as an internal class, add a comment that states that this is intended to be used in the entire Eclipse-SDK (Equinox, Platform, JDT and PDE, but not more) so that one knows where to search callers.
This way the code can be re-used, but we don't have to make it an API.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
HannesWell: The next time can be now. :)
I added another commits that applies it, feel free to squash it yours. This commit also uses more detailed assertions.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
HannesWell: OK. But then at least the same style (with regards to StringBuilder vs. String concatenation) should be used.
In my extra commit I changed it to use String concatenation.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
HannesWell: ACK.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
```diff
@@ -0,0 +1,336 @@
...
+					try (Socket socket = serverSocket.accept()) {
+						try (BufferedReader in = new BufferedReader(new InputStreamReader(socket.getInputStream()))) {
+							String firstLine = in.readLine();
+							System.out.println(socket.getInetAddress() + ": " + firstLine);
+							try (OutputStream outputStream = socket.getOutputStream()) {
+								outputStream.write("HTTP/1.1 200 OK\r\n".getBytes(StandardCharsets.UTF_8));
+							}
+							assertThat(firstLine, startsWith("GET"));
+							assertThat(firstLine, not(containsString("secret")));
+							fail("Server was contacted");
```
HannesWell, line 266: Have you checked if AssertionErrors thrown by this method are considered by JUnit for a test result, i.e. can they let the test fail? My first assumption would be that they don't.
So if they are relevant for test results, they probably should be collected and for example thrown again in the Server's close method.
Alternatively, the Server could collect only the first line it has read and provide it to the test method and the assertions could be done there. This way errors would not have to be catched, collected and re-thrown. 
If the assertions are actually meaning-less, they should be removed.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDEXmlProcessorFactory.java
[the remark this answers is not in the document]
jukzi: I struggled to find a place that both needs this methods itself (which would make sure it is not deleted because unused and not API) and is available in all other modules as well. I would highly appreciate help to find a good place, Otherwise i would copy & paste the (small) solution to other projects and keep only the test at one module.
Part of it would be needed for example at 
equinox ExtensionRegistry or org.eclipse.core.internal.resources but in at least 8 distinct git repositories. I think it's very bad to share no-API across repositories.
@jonahgraham any idea?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/PDEXmlProcessorFactoryTest.java
jukzi: that commit failed to run the test - @Rule has to be public

HannesWell: Yes. Thanks, for fixing.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDEXmlProcessorFactory.java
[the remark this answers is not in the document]
HannesWell: We could also register a provider of such factories as (declarative) OSGi Service and can then query the service from any bundle in the runtime. Registering it in something like equinox.common. maybe even cascade it.

But I'm fine doing that in a later step and for now submit this as it is.

HannesWell (approved):

PR #673, opened 2023-07-20
Ignore class files in eclipse.pde

fedejeanne:
Add `**/*.class` to `.gitignore` so that (among others) all class files that are built under `eclipse.pde/build/org.eclipse.pde.build.tests/resources/**` are ignored and the _Git staging_ view in Eclipse doesn't look like this

image


.gitignore
```diff
@@ -8,6 +8,7 @@ heapdump.*
 core.*
 Snap.*
 /.project
+**/*.class
```
HannesWell, line 11: Just `*.class` should be sufficient.

fedejeanne: You're right, thanks. I corrected it in 7dd071979509feaf45c850aa914ebbc6c9b0113f

HannesWell (approved):
Looks good.
Thanks.

PR #674, opened 2023-07-21
Show a progress indicator when opening a PDE run configuration.

fedejeanne:
The progress indicator uses `IProgressMonitor.UNKNOWN` as the total amount of work because the (sometimes) long running operation is deep in the call stack: it goes all the way down to the `TracingOptionsManager` in `org.eclipse.pde.core` (dependency issues).

The operation can **not** be canceled.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/679

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TracingOptionsManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TracingOptionsManager.java
@@ -24,16 +24,16 @@ import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.plugin.PluginRegistry;
 
@@ -44,8 +44,8 @@ public class TracingOptionsManager {
 		super();
 	}
 
-	public Map<String, String> getTemplateTable(String pluginId) {
-		Map<String, String> tracingTemplate = getTracingTemplate();
+	public Map<String, String> getTemplateTable(String pluginId, IProgressMonitor monitor) {
+		Map<String, String> tracingTemplate = getTracingTemplate(monitor);
 		Map<String, String> defaults = new HashMap<>();
 		tracingTemplate.forEach((key, value) -> {
 			if (belongsTo(key, pluginId)) {
@@ -60,9 +60,9 @@ public class TracingOptionsManager {
 		return pluginId.equalsIgnoreCase(firstSegment);
 	}
 
-	public Map<String, String> getTracingOptions(Map<String, String> storedOptions) {
+	public Map<String, String> getTracingOptions(Map<String, String> storedOptions, IProgressMonitor monitor) {
 		// Start with the fresh template from plugins
-		Map<String, String> defaults = getTracingTemplateCopy();
+		Map<String, String> defaults = getTracingTemplateCopy(monitor);
 		if (storedOptions != null) {
 			// Load stored values, but only for existing keys
 			storedOptions.forEach((key, value) -> {
@@ -74,21 +74,29 @@ public class TracingOptionsManager {
 		return defaults;
 	}
 
-	public Map<String, String> getTracingTemplateCopy() {
-		return new HashMap<>(getTracingTemplate());
+	public Map<String, String> getTracingTemplateCopy(IProgressMonitor monitor) {
+		return new HashMap<>(getTracingTemplate(monitor));
 	}
 
-	private synchronized Map<String, String> getTracingTemplate() {
-		if (template == null) {
-			Map<String, String> temp = new HashMap<>();
-			IPluginModelBase[] models = PluginRegistry.getAllModels();
-			Arrays.stream(models).map(TracingOptionsManager::getOptions).filter(Objects::nonNull).forEach(p -> {
-				@SuppressWarnings({ "rawtypes", "unchecked" })
-				Map<String, String> entries = (Map) p;
-				temp.putAll(entries); // All entries are of String/String
-			});
-			template = temp;
+	private synchronized Map<String, String> getTracingTemplate(IProgressMonitor monitor) {
+		if (template != null) {
+			return template;
 		}
+
+		Map<String, String> temp = new HashMap<>();
+		IPluginModelBase[] models = PluginRegistry.getAllModels();
+		SubMonitor subMonitor = SubMonitor.convert(monitor, models.length);
+
+		for (IPluginModelBase model : models) {
+			subMonitor.split(1);
+			Properties options = TracingOptionsManager.getOptions(model);
+			if (options != null) {
+				@SuppressWarnings({ "rawtypes", "unchecked" })
+				Map<String, String> entries = (Map) options;
+				temp.putAll(entries); // All entries are of String/String
+			}
+		}
+		template = temp;
 		return template;
 	}
 
@@ -129,7 +137,7 @@ public class TracingOptionsManager {
 	}
 
 	public void save(Path file, Map<String, String> map, Set<String> selected) {
-		Map<String, String> properties = getTracingOptions(map);
+		Map<String, String> properties = getTracingOptions(map, null);
 		properties.keySet().removeIf(key -> {
 			IPath path = IPath.fromOSString(key);
 			return path.segmentCount() < 1 || !selected.contains(path.segment(0));
@@ -138,7 +146,7 @@ public class TracingOptionsManager {
 	}
 
 	public void save(Path file, Map<String, String> map) {
-		saveOptions(file, getTracingOptions(map));
+		saveOptions(file, getTracingOptions(map, null));
 	}
 
 	private static Properties getOptions(IPluginModelBase model) {
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -1093,6 +1093,7 @@ public class PDEUIMessages extends NLS {
 	public static String TracingTab_AttributeLabel_TracingChecked;
 	public static String TracingTab_AttributeLabel_TracingNone;
 
+	public static String TracingBlock_initializing_tracing_options;
 	public static String TracingBlock_restore_default;
 	public static String TracingBlock_restore_default_selected;
 
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
@@ -18,16 +18,21 @@ package org.eclipse.pde.internal.ui.launcher;
 
 import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.BiFunction;
 
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.jface.dialogs.IDialogSettings;
+import org.eclipse.jface.operation.IRunnableContext;
 import org.eclipse.jface.viewers.ArrayContentProvider;
 import org.eclipse.jface.viewers.CheckboxTableViewer;
 import org.eclipse.jface.viewers.IStructuredSelection;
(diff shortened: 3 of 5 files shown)


HannesWell:
With this change now always the progress-indicator pops when opening the launch-configuration Editor, even if it is just for a very short time.
A pop-up that vanishes within the fraction of a second so that the user cannot read its content is IMHO disturbing.
Wouldn't it be possible to show the pop-up only after a certain amount of time or estimate better if the operation is long-runnning?

fedejeanne:
> Wouldn't it be possible to show the pop-up only after a certain amount of time or estimate better if the operation is long-runnning?

I made some changes and shifted all the heavy lifting to the **activation** of the _Tracing_ tab, so I managed to get rid of the dialog at the beginning. 

The problem with the flickering still remains, but it now happens while changing some of the configurations in the _Tracing_ tab and the flickering happens in the lower part of the _Run configurations..._ dialog (in the progress bar). The good thing is that fixing that is easier than fixing the dialog, here's a draft PR to do that --> https://github.com/eclipse-platform/eclipse.platform/pull/591

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
```diff
@@ -61,6 +66,59 @@
 
 public class TracingBlock {
 
+	private class TracingCache {
```
HannesWell, line 69: I don't think maintaining an own cache is not necessary since the TracingOptionsManager already caches the template internally and since `TracingOptionsManager.getTracingTemplate()` seem to be the expensive method (it reads the tracing options from all plugins, which is relatively heavy IO), a second cache seem unnecessary.

HannesWell (commented):
I looked into this and as written below an additional cache seems not to be necessary.

But shifting the first call to the TracingManager seems reasonable.
I suggest we add a method to `TracingOptionsManager` like `ensureInitialized()` that basically calls getTracingTemplate() and passes a `ProgressMonitor` to it so that we can at least see progress. It shouldn't be canclable because otherwise the state would become inconsistent or more complex to manage.

I'm currently working on a greater clean up of many of the launcher classes, which would also make the TracingOptionsManager a bit simpler. Maybe you want to await that before adding the suggested `ensureInitialized()` method, in order to avoid conflicts.

fedejeanne:
> I'm currently working on a greater clean up of many of the launcher classes, which would also make the TracingOptionsManager a bit simpler. Maybe you want to await that before adding the suggested ensureInitialized() method, in order to avoid conflicts.

👌 
Please let me know when your PR is ready/merged and I can change this one.

HannesWell:
> Please let me know when your PR is ready/merged and I can change this one.

Since https://github.com/eclipse-pde/eclipse.pde/pull/755 is now submitted, we can continue on this one. :)

HannesWell:
> I suggest we add a method to `TracingOptionsManager` like `ensureInitialized()` that basically calls getTracingTemplate() and passes a `ProgressMonitor` to it so that we can at least see progress. It shouldn't be canclable because otherwise the state would become inconsistent or more complex to manage.

Thinking about it again, a `TracingOptionsManager.ensureInitialized(IProgressMonitor)` method could be canceled but then all results computed until then should be discarded.

To achieve this it is probably sufficient to replace `TracingOptionsManager.getTracingTemplate()` by 
```
	public void ensureInitialied(IProgressMonitor monitor) {
		getTracingTemplate(monitor);
	}

	private synchronized Map<String, String> getTracingTemplate(IProgressMonitor monitor) {
		if (template == null) {
			Map<String, String> temp = new HashMap<>();
			IPluginModelBase[] models = PluginRegistry.getAllModels();
			SubMonitor subMonitor = SubMonitor.convert(monitor, models.length);
			Arrays.stream(models).map(TracingOptionsManager::getOptions).filter(Objects::nonNull).forEach(p -> {
				subMonitor.split(1);
				@SuppressWarnings({ "rawtypes", "unchecked" })
				Map<String, String> entries = (Map) p;
				temp.putAll(entries); // All entries are of String/String
			});
			template = temp;
		}
		return template;
	}
```
and pass just null as monitor in all other callers of `getTracingTemplate()`.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
fedejeanne: I removed the cache but I left the internal class (it's now called `TracingOptionsManagerDelegate`). The reason for this is that the class provides a good encapsulation and helps me channel all calls to the "expensive methods" in `TracingOptionsManager`.

fedejeanne:
I just pushed some changes and took your advice about using the `ProgressMonitor` inside the method `TracingOptionsManager::getTracingTemplate`, thank you for that! I made some adjustments though:
* I didn't add any method to make sure the template is initialized ( _e.g._ `ensureInitialized`) because I had no need for it
* I converted the stream into a `for`-loop to properly report the progress since one needs to call `monitor.split(1)` for every model and not for every _filtered_ (non-null) model. 
* I inverted the `if`-logic in the method to reduce the [cognitive complexity (nesting level)](https://docs.sonarsource.com/sonarqube/latest/user-guide/metric-definitions/).

## Important: do **NOT** merge yet
This PR is blocked by https://github.com/eclipse-platform/eclipse.platform/pull/766

fedejeanne:
Test failure already documented here: https://github.com/eclipse-pde/eclipse.pde/issues/820

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TracingOptionsManager.java
```diff
@@ -74,21 +74,29 @@ public Map<String, String> getTracingOptions(Map<String, String> storedOptions)
...
+			Properties options = TracingOptionsManager.getOptions(model);
+			if (options != null) {
 				@SuppressWarnings({ "rawtypes", "unchecked" })
-				Map<String, String> entries = (Map) p;
+				Map<String, String> entries = (Map) options;
 				temp.putAll(entries); // All entries are of String/String
-			});
-			template = temp;
+			}
+			subMonitor.split(1);
```
HannesWell, line 97: Please move `subMonitor.split(1)` to the beginning of the loop.
`SubMonitor.split()` cleans up an previously created other children and therefore 'completes/calls `done()` for them.
Consequently calling split at the end of the loop has the effect that the first progress reporting for the loop happens after the second model has been processed because only then a pending submonitor-child exists.
This means a slight delay of the first feedbeck.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
```diff
@@ -58,6 +64,62 @@
...
+	/**
+	 * This class encapsulates all calls to the tracing options manager and runs them using a
+	 */
+	private static final class TracingOptionsManagerDelegate {
+
+		private IRunnableContext fRunnableContext;
+
+		public void setRunnableContext(IRunnableContext runnableContext) {
+			fRunnableContext = runnableContext;
+		}
```
HannesWell, line 76: If you prefer to have those methods bundled in a separate class (you could also just add a `IRunnableContext fRunnableContext` field to the `TracingBlock` class and have the methods there), I would make it a record and create a new record object whenever the context is reset (which I assume happens rarely).
```suggestion
	private static record TracingOptionsManagerDelegate(IRunnableContext runnableContext) {
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
```diff
@@ -1092,7 +1092,11 @@ public class PDEUIMessages extends NLS {
 	public static String TracingTab_AttributeLabel_TracingOptions;
 	public static String TracingTab_AttributeLabel_TracingChecked;
 	public static String TracingTab_AttributeLabel_TracingNone;
+	public static String TracingTab_TracingOptionsManager_Progress;
```
HannesWell, line 1095: This has no value defined and is never called.
```suggestion
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
```diff
@@ -1092,7 +1092,11 @@ public class PDEUIMessages extends NLS {
...
 
+	public static String TracingBlock_initializing_tracing_options;
+	public static String TracingBlock_initializing_tracing_template;
+	public static String TracingBlock_initializing_tracing_template_table;
```
HannesWell, line 1099:
```suggestion
	public static String TracingBlock_initializing_tracing_options;
```
Because in all three of this cases the same method is called eventually for which the progress reporting happens I think it is sufficient to have one message here. The first one IMHO fits well. :)
```suggestion
	public static String TracingBlock_initializing_tracing_options;
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
```diff
@@ -58,6 +64,62 @@
...
+				final Map<String, String> result = new HashMap<>();
+
+				// due to a bug
+				// (https://github.com/eclipse-platform/eclipse.platform/issues/769),
+				// the task needs to be forked otherwise the UI
+				// does not show the progress indicator
+				fRunnableContext.run(true, false, monitor -> {
+					SubMonitor subMonitor = SubMonitor.convert(monitor, taskName, 1);
+					result.putAll(task.apply(subMonitor.split(1)));
+				});
```
HannesWell, line 107: It would be nice if there would be some kind of `ICallableWithProgress`.

HannesWell, line 107: Instead of a `Function` you could use a `BiFunction<TracingOptionsManager, IProgressMonitor, Map<String, String>>` and move the repetitive calls of `PDECore.getDefault().getTracingOptionsManager()` to this method. This would even allow to use a Method-Reference in one caller.
Furthermore if always the same task-name is used, we would save passing that as an argument.
```suggestion
		private Map<String, String> runShowingProgress(
				BiFunction<TracingOptionsManager, IProgressMonitor, Map<String, String>> task) {
			try {
				final Map<String, String> result = new HashMap<>();
				// due to a bug
				// (https://github.com/eclipse-platform/eclipse.platform/issues/769),
				// the task needs to be forked otherwise the UI
				// does not show the progress indicator
				fRunnableContext.run(true, false, monitor -> {
					SubMonitor subMonitor = SubMonitor.convert(monitor,
							PDEUIMessages.TracingBlock_initializing_tracing_options, 1);
					result.putAll(task.apply(PDECore.getDefault().getTracingOptionsManager(), subMonitor.split(1)));
				});
```

HannesWell (commented):
> I made some adjustments though:
> 
>     * I didn't add any method to make sure the template is initialized ( _e.g._ `ensureInitialized`) because I had no need for it
> 
> 
>     * I inverted the `if`-logic in the method to reduce the [cognitive complexity (nesting level)](https://docs.sonarsource.com/sonarqube/latest/user-guide/metric-definitions/).

That's fine.

>     * I converted the stream into a `for`-loop to properly report the progress since one needs to call `monitor.split(1)` for every model and not for every _filtered_ (non-null) model.

Good catch!




> I did **not** make the tasks/monitors cancelable for now because that requires to rethink how the UI behaves once the task has been canceled (probably some elements need to be disabled) and also where to handle the proper exceptions in the code. I'd like to leave that for a future PR since these changes already provide a good enough performance and usability improvement even without the task being cancelable

Sounds reasonable. If you one day make it cancalable, the Tracing-tab could look similar to the disabled tracing state, but the `Enable tracing` check-box state should not be affected by that. The editor could then contain a button instead to load the tracing options again.


With the mentioned changes this would be perfectly ready from PDE side.

> 
> This PR is blocked by [eclipse-platform/eclipse.platform#766](https://github.com/eclipse-platform/eclipse.platform/pull/766)

I'll try to participate there over the weekend.

fedejeanne:
> Sounds reasonable. If you one day make it cancalable, the Tracing-tab could look similar to the disabled tracing state, but the `Enable tracing` check-box state should not be affected by that. The editor could then contain a button instead to load the tracing options again.

Thank you! I'll keep that in my for the next PR :-)
 
> I'll try to participate there over the weekend.

Thank you!

fedejeanne:
Test failures documented in #554 and #844

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TracingOptionsManager.java
fedejeanne: Done, thanks for the tip!

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
fedejeanne: Deleted

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
fedejeanne: Done

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
fedejeanne: Done

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
fedejeanne: I made it a record, thank you for the tip!

HannesWell (commented):
Thanks for the adjustments.

I just have two non-critical remarks and then this is ready from the PDE site.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
```diff
@@ -58,6 +63,55 @@
...
 public class TracingBlock {
 
+	/**
+	 * This class encapsulates all calls to the tracing options manager and runs them using a
+	 */
+	private static record TracingOptionsManagerDelegate(IRunnableContext fRunnableContext) {
+
+		Map<String, String> getTracingTemplateCopy() {
+			return runShowingProgress(
+					(tracingOptionsManager, monitor) -> tracingOptionsManager.getTracingTemplateCopy(monitor));
```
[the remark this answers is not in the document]
fedejeanne: If it's ok to you, I'd rather keep it as it is: I find method references with parameters a bit harder to read so I usually go for the "verbose" option. The current style also matches the style of the other 2 calls in the class so that's another reason to keep it like this.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
```diff
@@ -58,6 +63,55 @@
...
+				return result;
+			} catch (InvocationTargetException | InterruptedException e) {
+				PDEPlugin.logException(e);
+			}
+
+			// This should never happen since the task can not be canceled
+			// (yet). If it does happen then some exception occurred while
+			// calling the TracingOptionsManager and all bets are off (i.e. NPEs
+			// will follow)
+			return null;
```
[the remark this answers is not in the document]
fedejeanne: Good point, I changed it.

Better to have a descriptive exception than to let it throw a follow-up NPE. 

For the record: I will tackle the issue of showing the exception to the user in the UI when I make the task cancelable (in another PR) since then it will be much more likely that an exception happens (specifically: [an `OperationCanceledException` wrapped inside a `InterruptedException`](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1259))

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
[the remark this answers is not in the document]
HannesWell: If you insist that's fine in this case :)

HannesWell (approved):
Great. So this looks ready from a PDE point of view.

Should we wait until your related Platform PR ist merged, before merging this?

fedejeanne:
Yes please, let's wait until https://github.com/eclipse-platform/eclipse.platform/pull/766 is merged otherwise this PR will introduce a bug.

fedejeanne:
@HannesWell https://github.com/eclipse-platform/eclipse.platform/pull/766 is merged --> this PR is good to go :)

HannesWell:
Great!
Thanks for this and the platform fix.
Nice work :)

Issue #676, opened 2023-07-22
Provide support for copying the location of a bundle or feature to the clipboard


merks:
For orbit-simrel, when authoring maven location dependencies with BND instructions, one wants to quickly review the contents of the wrapped bundle that's produced during target resolution:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/maven-bnd

It would be very convenient to be able to quickly determine the location of a bundle to inspect that specific jar:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/0caa5dc8-a355-417b-83c3-e34384d1c8e4)

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/90826b62-cfe4-409f-a9c0-e4094ac02dad)

One way to use this then is Alt  F  .  Ctrl-V Enter, so open an editor, which could be the system editor but also works for the BND editor, which is a really great formatted view of the MANIFEST.MF:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/85058400-f8ec-4d7f-bf4b-8b2d60b2c81e)

laeubi:
Just in case if you activate the target and search in the target platform state for your bundle you can also double-click to jump to the bundle directly.

PR #677, opened 2023-07-22
Support copying the location of a bundle or feature to the clipboard

merks:
https://github.com/eclipse-pde/eclipse.pde/issues/676

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/CopyLocationAction.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/CopyLocationAction.java
@@ -1,0 +1,81 @@
+/*******************************************************************************
+ *  Copyright (c) 2023 Eclipse Contributors and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.pde.internal.ui.shared.target;
+
+import java.net.URI;
+import java.nio.file.Path;
+
+import org.eclipse.equinox.frameworkadmin.BundleInfo;
+import org.eclipse.jface.action.Action;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.viewers.StructuredViewer;
+import org.eclipse.osgi.service.resolver.BundleDescription;
+import org.eclipse.pde.core.target.TargetBundle;
+import org.eclipse.pde.core.target.TargetFeature;
+import org.eclipse.swt.dnd.Clipboard;
+import org.eclipse.swt.dnd.TextTransfer;
+import org.eclipse.swt.dnd.Transfer;
+import org.eclipse.ui.ISharedImages;
+import org.eclipse.ui.PlatformUI;
+
+public class CopyLocationAction extends Action {
+
+	private StructuredViewer fViewer;
+
+	private String fLocation;
+
+	public CopyLocationAction(StructuredViewer viewer) {
+		fViewer = viewer;
+
+		setText(Messages.CopyLocationAction_copyLocation);
+		ISharedImages workbenchImages = PlatformUI.getWorkbench().getSharedImages();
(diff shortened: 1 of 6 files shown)


merks:
@HannesWell 

FYI, I squash and merge this before the end of the day, unless you cry out before then...

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/CopyLocationAction.java
```diff
@@ -0,0 +1,71 @@
...
+		fViewer = viewer;
+
+		setText(Messages.CopyLocationAction_copyLocation);
+		ISharedImages workbenchImages = PlatformUI.getWorkbench().getSharedImages();
+		setImageDescriptor(workbenchImages.getImageDescriptor(ISharedImages.IMG_TOOL_COPY));
+		setDisabledImageDescriptor(workbenchImages.getImageDescriptor(ISharedImages.IMG_TOOL_COPY_DISABLED));
+
+		viewer.addPostSelectionChangedListener(event -> {
+			fLocation = null;
+			var selection = event.getStructuredSelection();
```
HannesWell, line 42: Is there a specific reason why you used `var` here and below instead of the actual type?
Personally I would have just used the type, since it isn't a complex generics chain here and I would say that it is easier to understand with just the type written out.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
```diff
@@ -316,8 +316,9 @@ public void run() {
 				fTree.collapseAll();
 			}
 		});
+		fMenuManager.add(new CopyLocationAction(fTree));
 		Menu contextMenu = fMenuManager.createContextMenu(tree);
-		tree.setMenu(contextMenu);
+		fTree.getTree().setMenu(contextMenu);
```
HannesWell, line 321: Nice catch.

HannesWell (approved):
This is is nice. Looks good and works well. 👍🏽

It would be great to have somehing similar for the Target Platform State view, where the actual bundle's location is only displayed but one cannot copy it.
For PDE's Plug-in Dependencies Classpath Container its the same.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/CopyLocationAction.java
merks: Oh, oh.  We need a var-policy!  To use or when to use, that is the question?  😶

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
merks: That took much longer to figure out than I care to admit.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/CopyLocationAction.java
merks: I will change it if you wish...

HannesWell: > Oh, oh. We need a var-policy! To use or when to use, that is the question? 😶

That would be good, yes. 😄
Personally I only use var when the type is too 'long' (e.g. a complex, nested generic type) or something internal that generates a discouraged-access warning. Then one can often use the fully-qualified-name in the constructor and `var` for the type and suppress the warning for the variable and often get that in one line:
`var name = new org.eclipse.foo.bar.internal.ClassName()`
The downside of var is that one cannot read the type from the variable declaration anymore. If one calls a constructor on the RHS of the assignment it is still simple to figure out, but the return type of methods is often hard/impossible to guess.
So I only use war in the cases above or if the type is really, really uninteresting.

But that's just my personal preference, but I think it would be good to have a general agreement about the usage of var in the Eclipse SDK. Since we don't have that now, I would leave that to your discretion since it is your change :)

merks:
> This is is nice. Looks good and works well. 👍🏽
> 
> It would be great to have somehing similar for the Target Platform State view, where the actual bundle's location is only displayed but one cannot copy it. For PDE's Plug-in Dependencies Classpath Container its the same.

This latest update reuses the action for the Platform State view too.

Issue #578, opened 2023-04-15
Improve error message for Import-Package resulution failure due to missing mandatory attributes and provide a quick-fix


HannesWell:
If one imports a package that is exported with mandatory attributes the error message in the manifest should show an enhanced error message indicating that the mandatory attribute is missing and should provide a quick-fix to add the mandatory attributes with the value from the provider.

A Minimal example `MANIFEST.MF`s:
- Provider:
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-SymbolicName: foo.provider
Bundle-Version: 1.0.0.qualifier
Export-Package: org.foo;attr=val;mandatory:=attr
```
- Consumer:
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-SymbolicName: foo.consumer
Bundle-Version: 1.0.0.qualifier
Import-Package: org.foo
```

A quick-fix should change the package import to `Import-Package: org.foo;attr=val`.

IIRC the PDE-state provides information why a required capability failed to resolve so this hopefully reveals. If that information is not available, in case of a unresolved Import-Package we could check the state for providers of that package and see if they have mandatory attributes.
The case of multiple providers with different mandatory attributes or different attribute values should be handled too. The best would then probably be to provide a error resolution for each possible change that should fix the resolution error (maybe limiting the number of possible resolutions).

PR #678, opened 2023-07-23
Quick-fix for Import-Package resolution failure due to missing mandatory attributes

alshamams:
When a package is imported that has been exported with mandatory attributes, following are the possibilities:

1. the import package is missing a mandatory attribute, 
2. the package has a mandatory attribute, but it does not match the importing package's export declaration verb, 
3. it has a mandatory attribute with the matching key, but there is a value mismatch

This commit aims to provide a quick-fix that queries the mandatory attributes and supplies them at the import site.

Fixes: [#578](https://github.com/eclipse-pde/eclipse.pde/issues/578)


HannesWell:
Thank you @alshamams.
From a very coarse look this seems fine. I'll performe a more detailed review this/tomorrow evening.

alshamams:
> Thank you @alshamams. From a very coarse look this seems fine.
> 
> I'll performe a more detailed review this/tomorrow evening. Everyoneelse' review is of course welcome. :)

alshamams:
> Thank you @alshamams. From a very coarse look this seems fine.
> 
> I'll performe a more detailed review this/tomorrow evening. Everyoneelse' review is of course welcome. :)

Thank you @HannesWell !

ahmed82 (commented):
lgtm

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
```diff
@@ -1232,7 +1232,39 @@ private void validateImportPackage(IProgressMonitor monitor) {
...
+
+						Map<String, Object> exportAttributes = export.getAttributes();
+						Map<String, Object> exportDirectives = export.getDirectives();
+						if (exportDirectives != null) {
+							if (exportDirectives.containsKey(Constants.RESOLUTION_MANDATORY)) {
+								Map<String, Object> importAttributes = importSpec.getAttributes();
+								String[] mandatories = (String[]) exportDirectives.get(Constants.RESOLUTION_MANDATORY);
+								for (String mandatory : mandatories) {
+									if (importAttributes == null || !importAttributes.containsKey(mandatory)
+											|| importAttributes.get(mandatory).toString() != exportAttributes
```
HannesWell: You should (almost) never compare strings by their reference, since it is easy to have different string object with the same content.

HannesWell (changes requested):
I looked into this and in general this worked, nice job so far.
But there were a few places that needed fine-tuning (e.g. enhanced messages) or could be simplified/imported. I pushed another commit with those suggests.

What's currently missing is the handling in case there are multiple providers that export a package but with different attribute values.
In my suggestion commit I already made a change to get all exporters of a package, but it still has to be implemented that those multiple providers are considered in the error resolution.

This should be combined with the consideration of multiple mandatory/not-mandatory attributes.
I think it would be best to iterate over all exporters and create one resolution per exporter that each sets all mandatory attributes to the exporters value and set all non-mandatory attributes to the exporter's value, if the importer has another value.


Besides that a few test cases would be good. They could probably be added to `BundleErrorReporterTest`.

laeubi:
I think it would be great if this is already offered at the time when importing the package, maybe also add support for the import version as well.

HannesWell:
> I think it would be great if this is already offered at the time when importing the package, maybe also add support for the import version as well.

That would be great but also would make this task much greater since the proposal would need to be enhanced as well. We could certainly do that, but I suggest we keep this focused on enhancing the error messages/quick-fix and leave the enhancement of the Import-Package proposals for a follow up.
Not suggesting to add the attributes initially also has the 'advantage' that users are very aware that they add a package-import with attributes which is usually a sign that something special is happening.

alshamams:
Hi @HannesWell ,
I have incorporated the changes as per your suggestion. Kindly look through and let me know. 

Can you also have a look at the resolution window for the Quick Fix and suggest possible improvements, if any? Thanks in advance.

HannesWell:
Thanks. I will review it as soon as possible (probably in the next days).

laeubi:
> We could certainly do that, but I suggest we keep this focused on enhancing the error messages/quick-fix and leave the enhancement of the Import-Package proposals for a follow up.

Sure but I think the main problem is that an invalid package import is added at the first place.

@alshamams you should rebase (and maybe squash) your commits here.

alshamams:
Hi @laeubi , @HannesWell,
I have just rebased my commit. Kindly review and let me know.

HannesWell:
> Hi @laeubi , @HannesWell, I have just rebased my commit. Kindly review and let me know.

Sorry for the delay. Will review this ASAP.

alshamams:
Hi @laeubi , @HannesWell,
Just wanted to remind you that I await your review comments. Do let me know.🙂

Issue #668, opened 2023-07-13
Count is not yet updated in plug-ins tab


Dinesh0723:
Steps to reproduce:

1. Open run/debug configuration and create new configuration.
2. Open plug-ins tab and change the value of "Launch with:" combo box to "Plug-ins selected below".
3. Select the few plugins from Target Platform/Workspace check box and check the Only show selected checkbox.
4. Then uncheck & check the checkbox from the root level (target platform/workspace check box).
5. Count remains still 0 (It should be updated properly based on the child count).

PR #691, opened 2023-08-01
Display correct plugin selection count under only-show-selected option

alshamams:
While configuring launch/debug, the number of plugins selected (under plug-ins tab) is not updated when
 i) only show selected option is on,
ii) top level checkbox is used instead of individual checkboxes for plugins.

This PR identifies the check state of the top level controls and makes sure the 
plugin count is computed / reset appropriately.

Fixes: [668](https://github.com/eclipse-pde/eclipse.pde/issues/668)

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
@@ -512,6 +512,16 @@ public abstract class AbstractPluginBlock {
 			Object element = event.getElement();
 			if (element instanceof IPluginModelBase) {
 				handleCheckStateChanged(event);
+			} else if (element instanceof NamedElement namedElement) {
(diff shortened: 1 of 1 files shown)


gireeshpunathil (approved):

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
```diff
@@ -512,6 +512,17 @@ protected void createPluginViewer(Composite composite, int span, int indent) {
 			Object element = event.getElement();
 			if (element instanceof IPluginModelBase) {
 				handleCheckStateChanged(event);
+			} else if (element instanceof NamedElement) {
+				NamedElement namedElement = (NamedElement) element;
```
gireeshpunathil, line 516: optional suggestion to apply the Java 16 instance optimisation:
```suggestion
			} else if (element instanceof NamedElement namedElement) {
```

Issue #429, opened 2023-01-03
System packages not found by validation


ptziegler:
After updating to the Eclipse 4.26 (from 4.25), I get the following error when trying to launch our product:

![image](https://user-images.githubusercontent.com/70652978/210325866-4c4c047d-de7d-42f3-80bf-62283ad63272.png)

Those packages come from the JDK, so they should be available to the product.

![image](https://user-images.githubusercontent.com/70652978/210326585-a1222b24-5dad-4ab5-98b7-a9afd30ea268.png)

Tested with:
OpenJDK 17.0.4+7 & 17.0.5+8
Eclipse SDK 4.26 & 4.27.I20230102-1800

Example:
[eclipse4.26-validation-error..zip](https://github.com/eclipse-pde/eclipse.pde/files/10335297/eclipse4.26-validation-error.zip)

ptziegler:
Note: The product starts normally when I click on "continue".
I assume this is a false positive?

HannesWell:
This is a resolution error in the Target-Platform. I noticed that as well already.
When you look into the `Target Platform State` view, you will notice that the xml.resolver bundle (and probably also the other one) cannot be resolved due to the same errors (you can filter for unresolved Plug-ins in the burger-menu).

I can try to have a look at this soon.

ptziegler:
I'd be great if you can find some time to look into this.
As I said, it's more of an annoyance rather than a problem (and even then, I can always go back to the 4.25), but I assume more and more people will run into a similar problem, as they update their Eclipse installations.

HannesWell:
I started to dig into this problem with a simple target:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="xmlResolverTest">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/releases/2022-12/"/>
			<unit id="org.eclipse.osgi" version="0.0.0"/>
			<unit id="org.apache.xml.resolver" version="0.0.0"/>
		</location>
	</locations>
</target>
```
Note (to myself, took me far too much time), the osgi system.package is important so that other bundles can resolve osgi.ee capabilities and JRE-packages!
And an additional Plug-in in a debugged Eclipse Workspace that just has a simple manifest:
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
[...]

PR #693, opened 2023-08-03
Enhance computation of system-package provided by a ExecutionEnvironment

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/429
- Only consider EE of running VM in launch validation to match the behavior later at runtime
- For the PDEState the EE's system-package is computed with the following distinction:
  - for a EE that corresponds to a Java-9 or higher release, get the 'best matching' VMInstall and query the system-packages for the EE's   java release version (which is possible for modular JVMs)
  - for a Java-8 or lower release check if there is a VMInstall selected/strictly compatible to that version and query the system-packages from that VM (i.e. its rt.jar and other libraries)
  - if there is no VMInstall for a for a Java-8 or lower release, combine the hard-coded list of 'java.*' packages with the set of
  non-java packages provided by the default VMInstall of the workspace. The default VMInstall of the WS is at the same time the JVM selected in the active target definition (which sets its selection as WS default upon loading).
  - In general the available system-packages are only really known at runtime. Therefore at any other time, e.g. in the workspace one can only guess/assume what packages are available and the implemented algorithm seems to be the best trade-off between probable accuracy, convenience and the possibility to let the user control the result.

- Listen to changes of available VMInstalls and ExecutionEnvironment defaults. Re-read them on changes and re-resolve the PDE-State again with a subsequent re-validation of all Plugin projects.

@laeubi in case one only has one JRE in the WS respectively none set as default of a EE, this is equivalent to your suggestion to just use the TP's JRE. This is convenient, but at the same time allows to be more specific if desired by providing/setting a JRE for an EE.

In my local testing that replicated the guice case, this worked well.

This is currently a draft since the list of java-packages still has to be completed. I'm on vacation for a week now, but will complete this once I'm back.

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2004, 2021 IBM Corporation and others.
+ *  Copyright (c) 2004, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Hannes Wellmann - Enhance computation of system-package provided by a ExecutionEnvironment
  *******************************************************************************/
 package org.eclipse.pde.internal.build.site;
 
@@ -28,10 +29,14 @@ import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Function;
 import java.util.jar.JarFile;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
@@ -44,6 +49,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.JavaRuntime;
+import org.eclipse.jdt.launching.LibraryLocation;
 import org.eclipse.jdt.launching.environments.IExecutionEnvironment;
 import org.eclipse.osgi.service.resolver.BundleDescription;
 import org.eclipse.osgi.service.resolver.ExportPackageDescription;
@@ -388,7 +394,6 @@ public class PDEState implements IPDEBuildConstants, IBuildPropertiesConstants {
 
 		// initialize profileManager and get the JRE profiles
 		String[] javaProfiles = getJavaProfiles();
-		String systemPackages = null;
 		String ee = null;
 
 		for (Config aConfig : configs) {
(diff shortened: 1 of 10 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -481,14 +484,215 @@ public void resolveState() {
...
+						zip.stream().filter(e -> !e.isDirectory()).map(ZipEntry::getName).filter(n -> !n.startsWith("META-INF/")).map(n -> { //$NON-NLS-1$
+							int i = n.lastIndexOf('/');
+							return n.substring(0, i);
+						}).distinct().map(n -> n.replace('/', '.')).forEach(packages::add);
+					} catch (IOException e) {
+						ILog.get().error("Failed to read packages in JVM library for " + vm + ", at " + path, e); //$NON-NLS-1$ //$NON-NLS-2$
+					}
+				}
+			}
+			return packages;
```
HannesWell, line 690: @iloveeclipse can you tell if there is already a way to query JDT for the list of packages provided by a VMInstall?
Something internal would be fine for now.

iloveeclipse: I honestly don't know. Look around uses of `org.eclipse.jdt.internal.compiler.util.JRTUtil` / `org.eclipse.jdt.internal.compiler.util.JrtFileSystem`.

iloveeclipse: I believe you wanted to contribute it once, see https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/175

HannesWell:
I now enhanced this:
- Added a test-case and adapt existing test cases
- Add a listener to be notified about changes in the `Java -> Installed JREs (->Execution Environment)` preferences and trigger a re-read of system-packages then followed by a re-resolve and re-validation of all Project Manifests.
- Use EE profiles file for Equinox's o.e.osgi for those cases if they contain the correct list of java.* packages (all other packages are ignored).
  - The listed java.* packages are in all cases, except for `JRE-1.1` and `J2SE-1.2`, de-facto the same as those I extracted from ancient JDKs. The only difference is that I found the packages [java.awt.dnd.peer, java.awt.peer], but they seem not to be interesting for application developers. At least the doc says the following, which is why I assume it is fine to skip them.
```
The peer interfaces are intended only for use in porting the AWT. They are not intended for use by
application developers, and developers should not implement peers nor invoke any of the peer
methods directly on the peer instances.
```

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
HannesWell: > I honestly don't know. Look around uses of `org.eclipse.jdt.internal.compiler.util.JRTUtil` / `org.eclipse.jdt.internal.compiler.util.JrtFileSystem`.

Sorry I forgot the empathize that this about non-modular (Java-1.8 or lower) JREs.

> I believe you wanted to contribute it once, see [eclipse-jdt/eclipse.jdt.debug#175](https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/175)

Yes that's right. I develop that in parallel to this change and are about to submit a PR for this.

HannesWell: > I believe you wanted to contribute it once, see [eclipse-jdt/eclipse.jdt.debug#175](https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/175)

Here we go: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294

HannesWell:
@gireeshpunathil you maybe want to back-port this If we don't encounter problems with this in the next week.

gireeshpunathil:
thanks @HannesWell. sure, will do.

PR #707, opened 2023-08-11
Code Cleanup: Java 16 instanceof pattern matching

alshamams:
The commit tries to implement a code cleanup in pattern matching using instanceof operator according to Java 16 guidelines, making the code more concise. 
The change has been applied in org.eclipse.pde.api.tools project. 

Kindly review @akurtakov @iloveeclipse @vik-chand @gireeshpunathil and do let me know in case of any edits.


akurtakov:
Looks good. Thanks for this one.

PR #710, opened 2023-08-16
ApiAnalysisApplication should wait until jobs are done

laeubi:
As part of the build probabbly other jobs are sceduled, the ApiAnalysisApplication should wait for all jobs to complete before investigate error markers to prevent problems with outdated states.

Currently the ApiAnalysisApplication seem flaky as discovered here:
- https://github.com/eclipse-equinox/equinox/pull/246#issuecomment-1658189309

@HannesWell @vik-chand please give approval for M3 this should not affect any IDE users but can improve the situation for Tycho users.


HannesWell (approved):
Looks good to submit.

> this should not affect any IDE users but can improve the situation for Tycho users.

Yes, I only Tycho as user of the `APIAnalyisApplication`.

PR #712, opened 2023-08-17
Code Cleanup: Java 16 instanceof pattern matching in pde.core

alshamams:
The commit tries to implement a code cleanup in pattern matching using instanceof operator according to Java 16 guidelines, making the code more concise.
The change has been applied in org.eclipse.pde.core project.

Kindly review @akurtakov @iloveeclipse @vik-chand @gireeshpunathil and do let me know in case of any edits.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionPointSchemaBuilder.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionPointSchemaBuilder.java
@@ -70,6 +70,7 @@ public class ExtensionPointSchemaBuilder extends IncrementalProjectBuilder {
 			}
 
 			if (resource instanceof IFile candidate) {
+				// see if this is it
 				if (isSchemaFile(candidate)) {
 					// That's it, but only check it if it has been added or changed
 					if (delta.getKind() != IResourceDelta.REMOVED) {
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionPointSchemaBuilder.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionPointSchemaBuilder.java
@@ -69,9 +69,7 @@ public class ExtensionPointSchemaBuilder extends IncrementalProjectBuilder {
 				return true;
 			}
 
-			if (resource instanceof IFile) {
-				// see if this is it
-				IFile candidate = (IFile) resource;
+			if (resource instanceof IFile candidate) {
 				if (isSchemaFile(candidate)) {
 					// That's it, but only check it if it has been added or changed
 					if (delta.getKind() != IResourceDelta.REMOVED) {
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureConsistencyChecker.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureConsistencyChecker.java
@@ -48,9 +48,7 @@ public class FeatureConsistencyChecker extends IncrementalProjectBuilder {
 		public boolean visit(IResourceDelta delta) {
 			IResource resource = delta.getResource();
 
-			if (resource instanceof IProject) {
-				// Only check projects with feature nature
-				IProject project = (IProject) resource;
+			if (resource instanceof IProject project) {
 				try {
 					return (project.hasNature(PDE.FEATURE_NATURE));
 				} catch (CoreException e) {
@@ -58,9 +56,7 @@ public class FeatureConsistencyChecker extends IncrementalProjectBuilder {
 					return false;
 				}
 			}
-			if (resource instanceof IFile) {
-				// see if this is it
-				IFile candidate = (IFile) resource;
+			if (resource instanceof IFile candidate) {
 				if (isManifestFile(candidate)) {
 					// That's it, but only check it if it has been added or changed
 					if (delta.getKind() != IResourceDelta.REMOVED) {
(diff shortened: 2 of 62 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionPointSchemaBuilder.java
```diff
@@ -69,9 +69,7 @@ public boolean visit(IResourceDelta delta) {
 				return true;
 			}
 
-			if (resource instanceof IFile) {
-				// see if this is it
-				IFile candidate = (IFile) resource;
+			if (resource instanceof IFile candidate) {
```
gireeshpunathil, line 72: pls retain the comments

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureConsistencyChecker.java
```diff
@@ -48,19 +48,15 @@ public DeltaVisitor(IProgressMonitor monitor) {
 		public boolean visit(IResourceDelta delta) {
 			IResource resource = delta.getResource();
 
-			if (resource instanceof IProject) {
-				// Only check projects with feature nature
-				IProject project = (IProject) resource;
+			if (resource instanceof IProject project) {
```
gireeshpunathil, line 51: same as above, retain the comments

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureConsistencyChecker.java
```diff
@@ -48,19 +48,15 @@ public DeltaVisitor(IProgressMonitor monitor) {
...
 					return (project.hasNature(PDE.FEATURE_NATURE));
 				} catch (CoreException e) {
 					PDECore.logException(e);
 					return false;
 				}
 			}
-			if (resource instanceof IFile) {
-				// see if this is it
-				IFile candidate = (IFile) resource;
+			if (resource instanceof IFile candidate) {
```
gireeshpunathil, line 59: ditto

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/UpdateSiteBuilder.java
```diff
@@ -45,19 +45,15 @@ public DeltaVisitor(IProgressMonitor monitor) {
 		public boolean visit(IResourceDelta delta) {
 			IResource resource = delta.getResource();
 
-			if (resource instanceof IProject) {
-				// Only check projects with feature nature
-				IProject project = (IProject) resource;
+			if (resource instanceof IProject project) {
```
gireeshpunathil, line 48: ditto

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/UpdateSiteBuilder.java
```diff
@@ -45,19 +45,15 @@ public DeltaVisitor(IProgressMonitor monitor) {
...
 					return (project.hasNature(PDE.SITE_NATURE));
 				} catch (CoreException e) {
 					PDECore.logException(e);
 					return false;
 				}
 			}
-			if (resource instanceof IFile) {
-				// see if this is it
-				IFile candidate = (IFile) resource;
+			if (resource instanceof IFile candidate) {
```
gireeshpunathil, line 56: ditto

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/ProjectModifyOperation.java
```diff
@@ -517,9 +516,7 @@ private void configureManifest(IBundleProjectDescription description, IBundlePro
 				bmodel.getBundleModel().getBundle().setHeader(Constants.BUNDLE_MANIFESTVERSION, "2"); //$NON-NLS-1$
 			}
 		}
-		if (pluginBase instanceof IFragment) {
-			// host specification
-			IFragment fragment = (IFragment) pluginBase;
+		if (pluginBase instanceof IFragment fragment) {
```
gireeshpunathil, line 519: ditto

akurtakov:
@alshamams It's good to open a bug against jdt for removed comments.
@jjohnstn Would you please take a look at why this cleanup doesn't keep comments?

HannesWell:
> @alshamams It's good to open a bug against jdt for removed comments.

IIRC some other JDT clean-ups like 'add braces' has a similar issue for years (if not decades), nevertheless it would be great if this is fixed.

jjohnstn:
> @alshamams It's good to open a bug against jdt for removed comments. @jjohnstn Would you please take a look at why this cleanup doesn't keep comments?

This is a no-win area for cleanups.  The compiler treats a comment preceding a statement as belonging to the statement.  When a statement is moved, it's comments generally go along and when removed, it's comments are removed.  There is a way to preserve a comment but this isn't done for removals because there is no way to know if the comment really belongs to the statement or is a general logic comment like the ones you have above.  It can never get it right not knowing the context.  I was thinking of adding an option to preserve comments on remove but this is a large-scale undertaking which might require changes to all existing cleanups.  I think in this particular case, defaulting to keeping the comment might be reasonable since the cast statement is a trivial operation not usually worth documenting.  I will open an enhancement issue.

gireeshpunathil:
> There is a way to preserve a comment but this isn't done for removals because there is no way to know if the comment really belongs to the statement or is a general logic comment like the ones you have above. It can never get it right not knowing the context.

@jjohnstn - is the code removal logic generic for all the clean-up functions? or is it specific to clean-up of `instanceof` optimisation? at least in this (instanceof) case it is reasonable to assert that the comment is not associated with the trivial casting, instead the code that follows. So I would retain the comment by default.

gireeshpunathil (approved):

jjohnstn:
> > There is a way to preserve a comment but this isn't done for removals because there is no way to know if the comment really belongs to the statement or is a general logic comment like the ones you have above. It can never get it right not knowing the context.
> 
> @jjohnstn - is the code removal logic generic for all the clean-up functions? or is it specific to clean-up of `instanceof` optimisation? at least in this (instanceof) case it is reasonable to assert that the comment is not associated with the trivial casting, instead the code that follows. So I would retain the comment by default.

@gireeshpunathil As I noted, the comment for the cast statement is unlikely to be a comment regarding the cast.  I posted #752 so the comment is no longer removed.  It has been merged for 4.30 M1.

akurtakov:
Thanks!

Issue #709, opened 2023-08-16
NullPointerException in PDEState.querySystemPackages


laeubi:
I have updates my SDK an now getting NPE

```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:208)
	at java.base/java.lang.String.join(String.java:3320)
	at org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(PDEState.java:548)
	at org.eclipse.pde.internal.build.site.PDEState.getSystemPackages(PDEState.java:491)
	at org.eclipse.pde.internal.core.TargetPlatformHelper.getSystemPackages(TargetPlatformHelper.java:391)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.initialize(ApiBaseline.java:295)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.initialize(ApiBaseline.java:243)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.resolveSystemLibrary(ApiBaseline.java:471)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.addApiComponents(ApiBaseline.java:418)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.restoreFrom(ApiBaseline.java:763)
	at org.eclipse.pde.api.tools.internal.ApiBaselineManager.loadBaselineInfos(ApiBaselineManager.java:242)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.loadBaselineInfos(ApiBaseline.java:735)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.getApiComponents(ApiBaseline.java:519)
	at org.eclipse.oomph.setup.pde.impl.APIBaselineFromTargetTaskImpl.isNeeded(APIBaselineFromTargetTaskImpl.java:263)
	at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.initNeededSetupTasks(SetupTaskPerformer.java:2285)
	at org.eclipse.oomph.setup.ui.SetupUIPlugin.performStartup(SetupUIPlugin.java:811)
	at org.eclipse.oomph.setup.ui.SetupUIPlugin$1$1.run(SetupUIPlugin.java:266)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

laeubi:
[...]

PR #714, opened 2023-08-17
Query others suitable VMs for system-packages if one cannot be read

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/709

@laeubi can you test this with your 'bad' vm installation?

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


HannesWell:
Ideally the VMs would even be sorted so that those with the most packages are choosen. This would avoid accidental use of reduced JVMs for example included in `jextract`.
But a caching as proposed in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294, would then be good.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -509,11 +509,10 @@ public static String querySystemPackages(IExecutionEnvironment environment, Prop
...
 		Collection<String> systemPackages;
 		if (releaseVersion <= 8) {
-			IVMInstall eeVM = bestVmInstallFor(environment, vms -> null); // Use selected VM or a perfect match
-			if (eeVM != null) {
-				// In case a VM is selected for an EE, query that VM and use its system-packages
-				systemPackages = querySystemPackages(eeVM, null);
-			} else {
+			systemPackages = vmInstallsFor(environment).filter(environment::isStrictlyCompatible) // Use only selected VM or perfect matches
+					.map(vm -> querySystemPackages(vm, null)) // In case a VM is selected for an EE, query that VM and use its system-packages
+					.filter(Objects::nonNull).findFirst().orElse(null);
```
laeubi, line 514: you might want to use
```suggestion
					.filter(Objects::nonNull).findFirst().orElseGet(()->alternative code path);
```

instead of extra check for `null` below.

laeubi (approved):
looks good

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
HannesWell: Done.

Issue #717, opened 2023-08-22
Warning on startup: No JVM system-packages available for environment 


jukzi:
Now i get Warnings on every IDE start. "No JVM system-packages available for environment "
And they are too many. Not even distinct. And why do i get that warnings at all? I did not try to use any 18+ stuff
see https://github.com/eclipse-pde/eclipse.pde/issues/709#issuecomment-1687500795
caused by https://github.com/eclipse-pde/eclipse.pde/pull/714

![image](https://github.com/eclipse-pde/eclipse.pde/assets/51790620/214767b8-83d2-45a7-b070-76c37c234cc6)

jukzi:
> That being said, in cases like yours the code should probably simply check if there is a compatible installed JRE in the Workspace and only issue a warning if at least one is present.

As far as i remember there is a warning anyway if not compatible JRE is found.

PR #718, opened 2023-08-22
Suppress warning about no JRE-packages if no compatible VM is available

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/717

Since this is a fix for a regression, it should be merged for RC1.
I plan to merge this latest today (Wednesday) evening.


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -496,7 +496,9 @@ public static String getSystemPackages(IExecutionEnvironment environment, Proper
 			systemPackages = profileProperties.getProperty(Constants.FRAMEWORK_SYSTEMPACKAGES);
```
jukzi, line 496: "use the those its profile" is not understandable

jukzi (changes requested):

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
HannesWell: Indeed. Should be fixed now.

jukzi (approved):
lets try

Issue #721, opened 2023-08-24
NPE in PDEState.querySystemPackages()


yuzawa-san:
I have been trying to build https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/ using the instructions in its readme which matches https://www.vogella.com/tutorials/EclipsePlatformDevelopment/article.html#creating-a-custom-eclipse-ide-build as well.

It fails in
```
[INFO] --- tycho-eclipserun:4.0.2:eclipse-run (dont_skip_me) @ org.eclipse.jdt.doc.isv ---
```
with this following error:

```
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.launching.IVMInstall.getLibraryLocations()" because "vm" is null
	at org.eclipse.jdt.launching.JavaRuntime.getLibraryLocations(JavaRuntime.java:2144)
	at org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(PDEState.java:649)
	at org.eclipse.pde.internal.build.site.PDEState.lambda$5(PDEState.java:540)
	at java.base/java.util.Optional.orElseGet(Optional.java:364)
	at org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(PDEState.java:523)
	at org.eclipse.pde.internal.build.site.PDEState.getSystemPackages(PDEState.java:493)
	at org.eclipse.pde.internal.core.TargetPlatformHelper.getSystemPackages(TargetPlatformHelper.java:391)
	at org.eclipse.pde.internal.core.TargetPlatformHelper.addEnvironmentProperties(TargetPlatformHelper.java:378)
	at org.eclipse.pde.internal.core.TargetPlatformHelper.getPlatformProperties(TargetPlatformHelper.java:365)
	at org.eclipse.pde.internal.core.MinimalState.initializePlatformProperties(MinimalState.java:251)
	at org.eclipse.pde.internal.core.PDEState.<init>(PDEState.java:70)
	at org.eclipse.pde.internal.core.PluginModelManager.initializeTable(PluginModelManager.java:617)
	at org.eclipse.pde.internal.core.PluginModelManager.getEntryTable(PluginModelManager.java:554)
	at org.eclipse.pde.internal.core.PluginModelManager.findEntry(PluginModelManager.java:1031)
	at org.eclipse.pde.core.plugin.PluginRegistry.findEntry(PluginRegistry.java:77)
[...]

PR #722, opened 2023-08-24
Handle missing default VMInstall when querying JVM system-packages

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/721

This handles the case that `JavaRuntime.getDefaultVMInstall()` returns null, which according to the docs happens
`when no defaultVM was set or when the default VM has been disposed.`

I also considered to just select any of the available VMs, maybe the 'latest' one, but since no default JRE should actually not happen (at least I get an error in the preferences if I don't select one), for me it seems to be better to warn the user about that instead of selecting just any VM.

@akurtakov should we merge this now to have this in RC1 or better wait for RC2?

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -532,14 +532,18 @@ public class PDEState implements IPDEBuildConstants, IBuildPropertiesConstants {
 				if (javaPackages == null) {
 					String profileSystemPackages = preJava9ProfileProperties.getProperty(Constants.FRAMEWORK_SYSTEMPACKAGES, ""); //$NON-NLS-1$
 					if (profileSystemPackages.isBlank()) {
-						return null;
+						return List.of();
 					}
 					javaPackages = COMMA.splitAsStream(profileSystemPackages).filter(p -> p.startsWith("java.")).toList(); //$NON-NLS-1$
 				}
 				IVMInstall targetVM = JavaRuntime.getDefaultVMInstall(); // Set by the Target-Definition if specified there
+				if (targetVM == null) {
+					LOGGER.warn("No default JRE installation selected"); //$NON-NLS-1$
+					return List.of();
+				}
 				Collection<String> targetVMSystemPackages = querySystemPackages(targetVM, null);
 				if (targetVMSystemPackages == null) {
-					return null;
+					return List.of();
 				}
 				Stream<String> targetVMNonJavaPackages = targetVMSystemPackages.stream().filter(p -> !p.startsWith("java.")); //$NON-NLS-1$
 


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -537,6 +537,10 @@ public static String querySystemPackages(IExecutionEnvironment environment, Prop
 					javaPackages = COMMA.splitAsStream(profileSystemPackages).filter(p -> p.startsWith("java.")).toList(); //$NON-NLS-1$
 				}
 				IVMInstall targetVM = JavaRuntime.getDefaultVMInstall(); // Set by the Target-Definition if specified there
+				if (targetVM == null) {
+					LOGGER.warn("No default JRE installation selected available"); //$NON-NLS-1$  
+					return null;
```
iloveeclipse, line 542: Returning null will just move NPE to `getSystemPackages()` where this null will be immediately dereferenced. Here and below, instead of null one should return empty string.

iloveeclipse (changes requested):
Please change to not return null in querySystemPackages() and we probably should trigger one extra build containing this patch for RC1.

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
iloveeclipse: Sorry, after opening this in the the IDE that we are inside lambda block.

iloveeclipse (approved):

gireeshpunathil (approved):

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
HannesWell: Good point. Thanks.
I changed this and all other two locations that returned null to return an empty list, which will later result in an empty string for the list of packages.

This was forgotten in https://github.com/eclipse-pde/eclipse.pde/pull/714.

iloveeclipse:
@akurtakov, @HannesWell : tests are green, I would propose to merge and re-trigger IBuild for RC1.
WDYT?

HannesWell:
> @akurtakov, @HannesWell : tests are green, I would propose to merge and re-trigger IBuild for RC1. WDYT?

Agree on merging. But isn't a I-Build triggered any way in about two hours?

HannesWell:
> > @akurtakov, @HannesWell : tests are green, I would propose to merge and re-trigger IBuild for RC1. WDYT?
> 
> Agree on merging. But isn't a I-Build triggered any way in about two hours?

Looking at the calender there is indeed no more I-build planned for RC1. So I agree on both. :)
Lets give @akurtakov a few more minutes to response.

akurtakov:
Please merge and trigger I-build.

iloveeclipse:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.29/125/

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -532,14 +532,18 @@ public static String querySystemPackages(IExecutionEnvironment environment, Prop
...
+						return List.of();
 					}
 					javaPackages = COMMA.splitAsStream(profileSystemPackages).filter(p -> p.startsWith("java.")).toList(); //$NON-NLS-1$
 				}
 				IVMInstall targetVM = JavaRuntime.getDefaultVMInstall(); // Set by the Target-Definition if specified there
+				if (targetVM == null) {
+					LOGGER.warn("No default JRE installation selected"); //$NON-NLS-1$
+					return List.of();
+				}
 				Collection<String> targetVMSystemPackages = querySystemPackages(targetVM, null);
```
laeubi, line 544: I would rather let querySystemPackages return a stream instead instead it is even converted to a stream right after that... and having `null` as return value for collections is really ugly....

Issue #724, opened 2023-08-25
StringIndexOutOfBoundsException when creating a new PDE project


dhuebner:
I'm using the following setup:

```
Eclipse DSL Tools (includes Incubating components)

Version: 2023-09 M3 (4.29.0 M3)
Build id: 20230824-1840

```

When trying to create a new Plug-in project using the "New Project..." wizard, I'm getting following error:

```
java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 24
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4606)
	at java.base/java.lang.String.substring(String.java:2709)
	at org.eclipse.pde.internal.build.site.PDEState.lambda$7(PDEState.java:650)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.zip.ZipFile$EntrySpliterator.tryAdvance(ZipFile.java:565)
	at java.base/java.util.Spliterator.forEachRemaining(Spliterator.java:332)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(PDEState.java:651)
	at org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(PDEState.java:515)
	at org.eclipse.pde.internal.build.site.PDEState.getSystemPackages(PDEState.java:491)
[...]

PR #725, opened 2023-08-25
StringIndexOutOfBoundsException when creating a new PDE project

iloveeclipse:
Don't try to map class files in the root directories and catch whatever obscure exception may be thrown to avoid basic functionality to be broken.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/724

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -656,10 +656,10 @@ public class PDEState implements IPDEBuildConstants, IBuildPropertiesConstants {
 					try (ZipFile zip = new ZipFile(path.toFile())) {
 						// Collect names of all directories that contain a .class file
 						zip.stream().filter(e -> !e.isDirectory()).map(ZipEntry::getName) //
-								.filter(n -> n.endsWith(".class")) //$NON-NLS-1$
+								.filter(n -> n.endsWith(".class") && n.indexOf('/') > 0) //$NON-NLS-1$
 								.map(n -> n.substring(0, n.lastIndexOf('/'))) //
 								.forEach(classFileDirectories::add);
-					} catch (IOException e) {
+					} catch (Exception e) {
 						LOGGER.error("Failed to read packages in JVM library for " + vm + ", at " + path, e); //$NON-NLS-1$ //$NON-NLS-2$
 					}
 				}


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
```diff
@@ -656,10 +656,10 @@ private static Collection<String> querySystemPackages(IVMInstall vm, IExecutionE
 					try (ZipFile zip = new ZipFile(path.toFile())) {
 						// Collect names of all directories that contain a .class file
 						zip.stream().filter(e -> !e.isDirectory()).map(ZipEntry::getName) //
-								.filter(n -> n.endsWith(".class")) //$NON-NLS-1$
+								.filter(n -> n.endsWith(".class") && n.indexOf('/') > 0) //$NON-NLS-1$
```
HannesWell, line 659: I assume an entry-name like `/MyClass.class` is not possible or at least not relevant.

HannesWell (approved):
Looks good.
Thanks Andrey!

build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
dhuebner: Is`/MyClass.class` not just a default package case? It is still a valid case I think.

iloveeclipse: We don't need to add "default" package to system provided packages. 
We are talking about something that PDE needs to address via "import package". How PDE should import package `""` and how relevant that can be for anyone?

I wonder which "old" JRE Denis has and if that is something completely wrong JRE but automatically found & configured by the new "autodetect JRE installations" job.

iloveeclipse: @dhuebner : would be nice if you could check that and which exact JRE (or something that looks like JRE) is causing the trouble.

dhuebner: @iloveeclipse 
I think it is the bundled JRE version:
`-vm
/Users/dhuebner/Downloads/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.justj.openjdk.hotspot.jre.full.macosx.aarch64_17.0.8.v20230801-1951/jre/lib/libjli.dylib
`

iloveeclipse: Moment, this is Java 17, we are supposed to see only Java < 8? 
@HannesWell : that is at least not what the original PR expected?

HannesWell: > Is`/MyClass.class` not just a default package case? It is still a valid case I think.

But even if a class is in the jar root, I assume the entry is just named `MyClass.class`, i.e. without leading slash. But I don't know if that applies for all implementations (haven't checked the docs yet).

> We don't need to add "default" package to system provided packages. We are talking about something that PDE needs to address via "import package". How PDE should import package `""` and how relevant that can be for anyone?
> 
Yes exactly, that's not relevant for that reason.

> Moment, this is Java 17, we are supposed to see only Java < 8? @HannesWell : that is at least not what the original PR expected?

Yes, this code path is only entered if `JavaRuntime.isModularJava(vm)` returns false.
This only happens if the vm is not an `AbstractVMInstall`, `AbstractVMInstall.getJavaVersion()` returns null or a value below 1.8 respectively 8. Maybe the VMInstall type is special or not correctly configured?

Any way I think this change is fine to merge as it should not harm valid installations.
Nevertheless finding the root cause would be interesting for https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294.

iloveeclipse: > Nevertheless finding the root cause would be interesting for [eclipse-jdt/eclipse.jdt.debug#294](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294).

See https://github.com/eclipse-pde/eclipse.pde/issues/724#issuecomment-1693201112

iloveeclipse:
I think this is a must have for RC2, so +1 from me.

iloveeclipse:
I believe we have enough votes for merge, so let's merge.

Issue #689, opened 2023-07-31
PreferenceDialog: NPE in ApiBaseline


jukzi:
JDK was updated now i can't configure another because of NPE in PreferenceDialog
```
java.lang.NullPointerException
	at java.base/java.util.Hashtable.put(Hashtable.java:476)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.initialize(ApiBaseline.java:307)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.initialize(ApiBaseline.java:243)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.resolveSystemLibrary(ApiBaseline.java:477)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.rebindVM(ApiBaseline.java:1003)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.vmAdded(ApiBaseline.java:972)
	at org.eclipse.jdt.launching.JavaRuntime.fireVMAdded(JavaRuntime.java:2923)
	at org.eclipse.jdt.launching.VMStandin.convertToRealVM(VMStandin.java:152)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsPreferencePage.getVersionFromRealVM(JREsPreferencePage.java:283)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsPreferencePage.supportsCurrentCompliance(JREsPreferencePage.java:264)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsPreferencePage.validateComplianceAndJRE(JREsPreferencePage.java:196)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsPreferencePage.isValid(JREsPreferencePage.java:190)
	at org.eclipse.jface.preference.PreferenceDialog.isCurrentPageValid(PreferenceDialog.java:869)
	at org.eclipse.jface.preference.PreferenceDialog$5.lambda$0(PreferenceDialog.java:659)
```

PR #726, opened 2023-08-28
fix NPE in ApiBaseline.initialize #689

jukzi:
https://github.com/eclipse-pde/eclipse.pde/issues/689

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
@@ -299,7 +299,9 @@ public class ApiBaseline extends ApiElement implements IApiBaseline, IVMInstallC
 		}
 		if (!(this instanceof WorkspaceBaseline)) {
 			Dictionary<String, Object> dictionary = new Hashtable<>();
-			dictionary.put(Constants.FRAMEWORK_SYSTEMPACKAGES, value);
+			if (value != null) {
+				dictionary.put(Constants.FRAMEWORK_SYSTEMPACKAGES, value);
+			}
 			value = profile.getProperty(Constants.FRAMEWORK_EXECUTIONENVIRONMENT);
 			if (value != null) {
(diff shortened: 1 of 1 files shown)


HannesWell:
@jukzi this isn't a regression from a recent change is it?

jukzi:
probably not a recent regression

HannesWell:
Ok.
The fix itself looks good and is straight forward. But I'm a bit concerend that it might cover a more fundamental problem.
I havn't encountered such issue yet, but with the emhanced system-package computation many special VM installs poped up that I didn't encountered either.

@akurtakov or the others what do you think? Should this be merged now or for the december release?

merks:
It certainly can't break anything to prevent an NPE, and the NPE might prevent other valid things form working (like with saw with the DSL package for m3).

If you're concerned about masking problems (a legitimate concern) and the null is completely unexpected, you might add a warning/info log message too.

laeubi (approved):
looks good and is similar to other checks below.
Maybe it would even be better to have a `method putIfNotNull(dictonary, avalue)` on the long term...

HannesWell (approved):
> If you're concerned about masking problems (a legitimate concern) and the null is completely unexpected, you might add a warning/info log message too.

Yes that's my concern.
The method `PDEState.getSystemPackages()`, that produces the `value` (`TargetPlatformHelper.getSystemPackages()` just calls it), already issues a warning in that case :

https://github.com/eclipse-pde/eclipse.pde/blob/54ae301ae56ebfb1216db0dd0d4e124a49189c44/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java#L498-L503

So I think this is good to go now.
And until the final submission of RC2 on friday we still have a few days to test I-builds.

merks (approved):
+1

akurtakov:
Is it a regression in this release? In general this has to be brought to PMC for discussion. @theMcQ @vogella @tjwatson @jarthana  wdyt? 
To me it looks safe and worth pushing for this release.

jukzi:
BTW there is no hurry for me to merge it. Today i am only producing pull commits to test the IDE :-)

vogella:
+1 for merge from PMC member

Александър Куртаков ***@***.***> schrieb am Di., 29. Aug.
2023, 10:11:

PR #731, opened 2023-09-04
POM and product version changes for 4.30 release

MohananRahul:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1289

### DIFF
--- a/ui/org.eclipse.pde/pom.xml
+++ b/ui/org.eclipse.pde/pom.xml
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.eclipse.pde</groupId>
     <artifactId>eclipse.pde</artifactId>
-    <version>4.29.0-SNAPSHOT</version>
+    <version>4.30.0-SNAPSHOT</version>
     <relativePath>../../</relativePath>
   </parent>
   <artifactId>org.eclipse.pde</artifactId>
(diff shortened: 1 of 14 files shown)


gireeshpunathil (approved):

ui/org.eclipse.pde/pom.xml
```diff
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.eclipse.pde</groupId>
     <artifactId>eclipse.pde</artifactId>
-    <version>4.29.0-SNAPSHOT</version>
+    <version>4.30.0-SNAPSHOT</version>
```
HannesWell, line 17: Don't we need a service Version increment in this Plugin (and probably also in the Test-Plugins)?
I expected the build will fail with a baseline discrepancy as soon as it is set to 4.29.

merks: I think this is like the following commit:

https://github.com/MohananRahul/eclipse.pde/commit/d2ee47fec4b0f41f261a80a7ae8d7b834fbea75e#diff-ca105ebeeb15c9f1f44a80adf837a1e1e1162bdb6a9bc0286faf8136528b8df4

These more increments on the feature/poms happen at the start of each cycle...

HannesWell: > These more increments on the feature/poms happen at the start of each cycle...

I think that ideally this happens at the same time so that the build always works.

> These more increments on the feature/poms happen at the start of each cycle...\n\nActually there is a workflow that performs these tasks and check if increments are required + automatically perform increments see:\n\nhttps://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/updateRelease.yml

Yes that would probably simplify the releng Task and thus speed it up. A win for everyone. :)

PR #734, opened 2023-09-06
Comparator errors in I20230906-0400

MohananRahul:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1318


sravanlakkimsetti (approved):

gireeshpunathil (approved):

HannesWell:
> For initial comparator errors version bumps are needed as this will result in qualifier change only which is not enough for opening new stream.

Could you please try the next time to use the job for the automated next-release preparation to do the parent-version increment, qualifier-update and needed version bump in one change instead of three as asked in https://github.com/eclipse-pde/eclipse.pde/pull/731#discussion_r1315448066?

This would probably make that process much smother for everybody. :)

PR #741, opened 2023-09-15
Rewrite smartimport test as plain JUnit test

akurtakov:
This removes RedDeer dependency which has been announced to drop from simrel.


akurtakov:
Merging. It's surprising how much simpler a test can become :)

HannesWell:
This is indeed really great. Thanks for this!
And it also shows that splitting the core implementation from the UI can drastically simplify testing.
I created a little follow up with #743 with a few more clean-ups.

laeubi:
> This removes RedDeer dependency which has been announced to drop from simrel.

Just wondering even if they dropped we should be able to use it anyways from an older release? Or does Platform is only allowed to use what is on simrel?

akurtakov:
@laeubi In theory using from old release is possible and Platform is not limited to use only what's in simrel. In practice as soon as some user facing label/title/etc. changes Red Deer functionality breaks as it can no longer match widgets and one has to change Red Deer itself to make the fix (I've went through that multiple times in https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/ which is heavy Red Deer user). For that reason I decided to rewrite the single plugin using it to be plain JUnit and simplify Platform infrastructure which is already far more complicated than wanted.

PR #743, opened 2023-09-16
Further clean-ups and simplifications of smartimport-tests

HannesWell:
Generalize redundant code into the template class.

Replace the o.e.pde.ui.tests.smartimport/pom.xml by the still required pom.model. configuration in the build.properties (many properties were obsolete anyway).


akurtakov (commented):
Now imagine someone to convert it to parameterized test so it becomes single class

HannesWell:
> Now imagine someone to convert it to parameterized test so it becomes single class

Good idea! And since that test-plugin then only contains one class, I think we can just move that test and its resources to the pde.ui.tests Plugin.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/ui/tests/smartimport/ProjectSmartImportTest.java
```diff
@@ -0,0 +1,132 @@
...
+		// Copy imported projects to temp-directory to not pollute this project
+		Path source = Path.of("tests/smartImport").toRealPath();
+		Path target = workingDirectory.getRoot().toPath();
+		try (Stream<Path> files = Files.walk(source).filter(Files::isRegularFile)) {
+			for (Path file : (Iterable<Path>) files::iterator) {
+				Path targetFile = target.resolve(source.relativize(file));
+				Files.createDirectories(targetFile.getParent());
+				Files.copy(file, targetFile);
+			}
+		}
```
HannesWell, line 81: This will not work in I-builds since the test-plugin is a jar there.
Will need to generalize this to extract bundle entries, like it is done in other places.

akurtakov:
I strongly believe that incremental improvements are better than trying to do everything it once. E.g. push PR that makes this one single class test and leave the move to pde.ui for another PR . Over the years many such "small" improvements got lost after staling in a bigbang change .

HannesWell:
> I strongly believe that incremental improvements are better than trying to do everything it once. E.g. push PR that makes this one single class test and leave the move to pde.ui for another PR . Over the years many such "small" improvements got lost after staling in a bigbang change .

In general I agree, but in this case I had the changes ready locally and just wanted to submit https://github.com/eclipse-pde/eclipse.pde/pull/745 before to not have to adjust the test-suites twice.

For me this is now ready.

laeubi:
> Now imagine someone to convert it to parameterized test so it becomes single class

parameterized **JUnit5** test :-)

> I think we can just move that test and its resources to the pde.ui.tests Plugin

If its a plain JUnit test you can even move it inside the plugin that is tested into a separate `src_test` folder and execute if with plain surefire :-)

HannesWell:
> parameterized JUnit5 test :-)

Migrating to JUnit5 is really something I would lime to do (in the near? future).
But can we do that already? IIRC the I-build Tests had some issues with JUnit5, but I could be wrong.

akurtakov:
No issue with using JUnit 5 AFAIK. https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/tests/org.eclipse.e4.ui.tests.css.core is one bundle that is exclusively JUnit 5 including @Suite usage.

laeubi:
> IIRC the I-build Tests had some issues with JUnit5, but I could be wrong.

Then probably we should open an issue about the issues to track the possible issue :-)

PR #745, opened 2023-09-17
Merge AllPDEMinimalTests into AllPDETests and merge minimal tests

HannesWell:
This prevents (and resolves already existing) discrepancies between the tests run in the verification-builds and I-builds. All tests executed in verification builds should also run in I-builds.
Only those that cannot succeed in verification builds are dynamically disable using JUnit-Assume.

With that just the AllPDETests test-suite can be executed.

The discrepancies were about to get more in https://github.com/eclipse-pde/eclipse.pde/pull/497 or https://github.com/eclipse-pde/eclipse.pde/pull/743 because people just check the verification builds, if their test was really executed.

@vik-chand you split up those tests a while ago and I saw that you also tried to use Assumptions (at least they were in the tests as comments). Can you tell why you didn't continue with that approach?


HannesWell:
> This pull request removes 575 and adds 610 tests. Note that renamed tests count towards both.

I double checked it and effectively this change only adds new tests that are executed and removes none (some are effectively re-named).

Issue #311, opened 2022-08-25
Run/ Debug Configurations tab Plug-ins


cPecican:
When I am changing Start Level/ Auto-Start for some plug-ins and apply the changes, after I close and reopening the popup all the changes in the Start Level/ Auto-Start are disappearing.
**This only happens if `Only show selected` is checked.**
If I am not checking the option "Only show selected", then all of the above are working fine.

Dinesh0723:
Could you please let me know the eclipse version.

cPecican:
When I opened the case I was on 2022.06, now I am using 2022.12, but the behavior persist.
![image](https://github.com/eclipse-pde/eclipse.pde/assets/85338228/46510500-bb85-452a-bb10-4739b9a158e0)
Reopening the popup:
![image](https://github.com/eclipse-pde/eclipse.pde/assets/85338228/bf963090-3ea3-4254-8bec-9519d55ece59)

mickaelistria:
Please try upcoming 4.28 release https://download.eclipse.org/eclipse/downloads/drops4/S-4.28RC2a-202306050440/

cPecican:
Still the same behavior.

cPecican:
I've updated my IDE to 2023.06 and still not working.

PR #746, opened 2023-09-18
Run/ Debug Configurations tab Plug-ins issue #311

lathapatil:
In Run/Debug configurations when changing Start Level/ Auto-Start for some plug-ins and apply the changes,
and closing and reopening the popup will make changes to disappear provided "Only show selected" is selected.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/311

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
@@ -1113,12 +1113,12 @@ public abstract class AbstractPluginBlock {
 	}
 
 	protected final void initializePluginsState(Map<IPluginModelBase, String> selectedPlugins) {
+		fPluginTreeViewer.setCheckedElements(selectedPlugins.keySet().toArray());
 		for (Entry<IPluginModelBase, String> entry : selectedPlugins.entrySet()) {
 			IPluginModelBase model = entry.getKey();
 			setText(model, entry.getValue());
 		}
 
-		fPluginTreeViewer.setCheckedElements(selectedPlugins.keySet().toArray());
 		countSelectedModels();
 		resetGroup(fWorkspacePlugins);
 		resetGroup(fExternalPlugins);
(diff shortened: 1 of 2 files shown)


HannesWell:
@lathapatil thanks for this fix.

In general it looks good to me, but I couldn't test it yet.
I'll test it probably tomorrow or Wednesday, unless another committer does it before.

alshamams:
Hi @lathapatil ,
I have tested the changes made, and it seems to be working.

lathapatil:
> Hi @lathapatil , I have tested the changes made, and it seems to be working.

@alshamams Thanks for the confirmation

gireeshpunathil (approved):

gireeshpunathil:
what does this mean `this PR removes 1 test: AllPDETests org.eclipse.pde.ui.tests.launcher.AllLauncherTests org.eclipse.pde.ui.tests.launcher.FeatureBasedLaunchTest ‑ Unknown test` ? @HannesWell  - do you know? is this of some concern / blocker from landing?

HannesWell:
I also tried it and it works as desired and fixes the mentioned issue.

But since `CachedCheckboxTreeViewer` is also used through `FilteredCheckboxTree` in other places, could you please verify that those other users are negatively impacted by this change? Although in general it looks reasonable what you have done.

> what does this mean `this PR removes 1 test: AllPDETests org.eclipse.pde.ui.tests.launcher.AllLauncherTests org.eclipse.pde.ui.tests.launcher.FeatureBasedLaunchTest ‑ Unknown test` ? @HannesWell - do you know? is this of some concern / blocker from landing?

I think this is just a false note. Probably the check gets confused due to the random/unrelated failure. So this is not a blocker.

laeubi:
> could you please verify that those **other users are negatively impacted** by this change

Didn't know that it is a goal to "negatively impact other users" :-D

HannesWell:
Thank you @lathapatil.

> > could you please verify that those **other users are negatively impacted** by this change
> 
> Didn't know that it is a goal to "negatively impact other users" :-D

Oh course I meant not impacted, it was a little late.😅

gireeshpunathil:
I landed it based on the assumption that the suggestion was for a future PR. hope nothing is broken!

HannesWell:
> I landed it based on the assumption that the suggestion was for a future PR. hope nothing is broken!

We'll see. In case it is, it just needs to be fixed. :)

PR #750, opened 2023-09-20
TarFile: avoid Implicit narrowing conversion

jukzi:
skip() uses long integer

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/util/TarFile.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/util/TarFile.java
@@ -33,7 +33,7 @@ public class TarFile implements Closeable {
 		private int nextEntry = 0;
 		private int nextEOF = 0;
 		private int filepos = 0;
-		private int bytesread = 0;
+		private long bytesread = 0;
 		private TarEntry firstEntry = null;
 		private String longLinkName = null;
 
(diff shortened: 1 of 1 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/util/TarFile.java
```diff
@@ -33,7 +33,7 @@ private static class TarInputStream extends FilterInputStream {
 		private int nextEntry = 0;
 		private int nextEOF = 0;
 		private int filepos = 0;
-		private int bytesread = 0;
+		private long bytesread = 0;
```
laeubi, line 36: does this not also apply to filepos as well then?

laeubi: But this then seems like codesmell, if the position in the file is limited to int, how can I ever read more bytes than an int can hold?

jukzi: Due to the "Integer.decode(size.toString()).intValue();" the max filesize for files read with this class is limited to int size. 
This commit is only to get rid of the false positive warning.

laeubi: If it is a false positive there should be nothing to change (except some kind of suppress the warning itself). If we think there is a problem then it should be fixed e.g. by parsing the size as a long instead.

jukzi: I don't like to suppress warnings if its possible to change the code so that even the dump checker, who doesn't understand the whole workflow is happy.

laeubi: If the checker is dumb(?) we better disable it.
The purpose of such checks is to find problems in the code and fix them (in this case obviously long is better for handling possibly large files) than making the checker happy and give a false impression of security.

Issue #720, opened 2023-08-23
allow using units from referenced updatesites


laeubi:
Currently PDE does not allow to use a unit from a referenced update-site, even though it will probably load dependencies from such sites.

laeubi:
I have tested this now in more details and it seems that even requirements are not resolved from referenced updatesites even if the UI suggests this.

PR #753, opened 2023-09-21
Enable referenced repositories when resolving the target content

laeubi:
Currently referenced repositories are ignored, that prevents sharing of items that are already available in the referenced repository.

This enables searching references (if possible) to find IUs and dependencies in these referenced repositories as well.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/720

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -1008,14 +1009,17 @@ public class P2TargetUtils {
 		}
 
 		int repoCount = repos.length;
-		SubMonitor subMonitor = SubMonitor.convert(monitor, repoCount);
+		SubMonitor subMonitor = SubMonitor.convert(monitor, repoCount * 2);
 
-
+		Set<IRepositoryReference> seen = new HashSet<>();
 		List<IMetadataRepository> result = new ArrayList<>(repoCount);
+		List<IMetadataRepository> additional = new ArrayList<>();
 		MultiStatus repoStatus = new MultiStatus(PDECore.PLUGIN_ID, 0, Messages.IUBundleContainer_ProblemsLoadingRepositories, null);
 		for (int i = 0; i < repoCount; ++i) {
 			try {
-				result.add(manager.loadRepository(repos[i], subMonitor.split(1)));
+				IMetadataRepository repository = manager.loadRepository(repos[i], subMonitor.split(1));
+				result.add(repository);
+				addReferences(repository, additional, seen, manager, subMonitor.split(1));
 			} catch (ProvisionException e) {
 				repoStatus.add(e.getStatus());
 			}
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1008,14 +1008,17 @@ static IQueryable<IInstallableUnit> getQueryableMetadata(URI[] repos, IProgressM
 		}
 
 		int repoCount = repos.length;
-		SubMonitor subMonitor = SubMonitor.convert(monitor, repoCount);
-
+		SubMonitor subMonitor = SubMonitor.convert(monitor, repoCount * 2);
 
+		Set<IRepositoryReference> seen = new HashSet<>();
 		List<IMetadataRepository> result = new ArrayList<>(repoCount);
+		List<IMetadataRepository> additional = new ArrayList<>();
```
HannesWell, line 1015: Can you tell the reason why the references are collected in a separate list? To always add them last?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1024,12 +1027,33 @@ static IQueryable<IInstallableUnit> getQueryableMetadata(URI[] repos, IProgressM
 		if (result.size() != repos.length) {
 			throw new CoreException(repoStatus);
 		}
+		result.addAll(additional);
```
HannesWell, line 1030: `additional` could contain references that are already present in `result`. Making result a `LinkedHashSet` would avoid this.

HannesWell (commented):
Thanks for working on this.

To me it looks like that referenced artifact repositories also have to be considered in getArtifactRepositories(). I assume otherwise only the IU is found, but not the corresponding artifact, when using the slicer.

Would it be possible to add tests for this?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
laeubi: The reason is that the code explicitly checks that all given URLs could be loaded (some lines below), this could maybe archived smarter but having to list do not hurt much...

laeubi:
> To me it looks like that referenced artifact repositories also have to be considered in getArtifactRepositories(). I assume otherwise only the IU is found, but not the corresponding artifact, when using the slicer.

I can enable this for slicer as well, but references are handled already in the `ProvisioningContext` so from PDE we only need the metadata to initially find references.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
laeubi: I now unify this as the last step.

HannesWell:
> > To me it looks like that referenced artifact repositories also have to be considered in getArtifactRepositories(). I assume otherwise only the IU is found, but not the corresponding artifact, when using the slicer.
> 
> I can enable this for slicer as well, but references are handled already in the `ProvisioningContext` so from PDE we only need the metadata to initially find references.

OK great, if that works that's perfectly fine.
It should just be avoided that that for an IU found by the Planner or Slicer, the corresponding artifact is not found later if the TP is loaded.

laeubi:
I'll merge this now so we can get early feedback if it is causing any issue.

PR #762, opened 2023-09-22
Using Java 17 as default instead of Java 1.8 in the

vogella:
org.eclipse.pde.ui.templates.tests

As we require Java 17 in the latest release, seems reasonable to set this also as default for the tests.


HannesWell:
> As we require Java 17 in the latest release, seems reasonable to set this also as default for the tests.

But users might target older releases that don't require Java-17. I think this should be more dynamic.
What do you think about using the version of the workspace's default JVM (obtained from 
`JavaRuntime.getDefaultVMInstall()` casted to `IVMInstall2`, then you can call `getJavaVersion()`)?
This would also avoid the need to update that value later if a more recent release than 17 is used.

akurtakov:
What is the status of this one?

vogella:
Currently not working on this one.

Issue #733, opened 2023-09-06
JUnit Plug-in tests cannot be started if an older target platform is used 


vogella:
If I use 			
````
<repository location="https://download.eclipse.org/eclipse/updates/4.6"/> 
````

PDE plug-in tests will not work with the following error message:

```
!ENTRY org.eclipse.pde.junit.runtime 4 0 2023-09-06 08:20:51.713
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.pde.junit.runtime [97]
  Unresolved requirement: Require-Bundle: org.eclipse.core.runtime; bundle-version="[3.29.0,4.0.0)"
  Unresolved requirement: Require-Bundle: org.eclipse.jdt.junit.runtime; bundle-version="[3.5.0,4.0.0)"
    -> Bundle-SymbolicName: org.eclipse.jdt.junit.runtime; bundle-version="3.7.200.v20230627-0107"; singleton:="true"
       org.eclipse.jdt.junit.runtime [91]
         Unresolved requirement: Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=11))"

	at org.eclipse.osgi.container.Module.start(Module.java:444)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1620)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1599)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1571)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1514)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:340)
```

If I switch to a recent release, like I can start JUnit Plug-in tests.

```
			<repository location="https://download.eclipse.org/releases/2023-09"/>
```

Full target platform:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
[...]

PR #765, opened 2023-09-22
Lower requirements of pde.junit.runtime and simplify/clean-up its code

HannesWell:
This PR aims to lower the lower requirements of pde.junit.runtime to Java-1.8 and Eclipse Indigo (supporting 12 year old Eclipse seems sufficient to me).

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/733

The second commit applies those lowering of requirements and the first one cleans up and simplifies the code (yes even with those low requirements that's possible).
I checked it with a Juno TP and the Plugin complied successfully. My only attempt to start a JUnit Plugin Test failed, but this might be due to some missing Plugins in my TP and I didn't investigate this further. 
Of course it would be ideal to have a JUnit test for that, but it will probably need some significant time to download a Indigo TP since I had the impression that the P2-Repo has been moved to some slower server.

@vogella FYI


ui/org.eclipse.pde.junit.runtime/readme.txt
```diff
@@ -0,0 +1,18 @@
+ This PDE JUnit runtime is added to test-runtimes launched from an Eclipse workspace and allows JUnit tests to be run with an OSGi runtime.
```
akurtakov, line 1: It would be nice to make this README.md so it's auto rendered by github when viewing this folder.

HannesWell: Good suggestion. Done.

HannesWell:
Looks like the Maven-Toolchain is not correctly configured for the GH runners:
`Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.2:compile (default-compile) on project org.eclipse.pde.junit.runtime: useJDK = BREE configured, but no toolchain of type 'jdk' with id 'JavaSE-1.8' found. See https://maven.apache.org/guides/mini/guide-using-toolchains.html -> [Help 1]`

Although the `setup-java` action says it's creating the toolchains.xml.
IIRC I noticed a similar issue in the past that the toolchains.xml was completly overwritten by each java version so that only the last one was eventually contained.
```
Run actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0
  with:
    java-version: 8
  11
  17
  
    distribution: temurin
    cache: maven
    java-package: jdk
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
    token: ***
Installed distributions
  Resolved Java 8.0.382+5 from tool-cache
  Setting Java 8.0.382+5 as the default
  Creating toolchains.xml for JDK version 8 from temurin
  Writing to /home/runner/.m2/toolchains.xml
  
  Java configuration:
    Distribution: temurin
    Version: 8.0.382+5
    Path: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.382-5/x64
  
  Resolved Java 11.0.20+1 from tool-cache
  Setting Java 11.0.20+1 as the default
  Creating toolchains.xml for JDK version 11 from temurin
  Overwriting existing file /home/runner/.m2/toolchains.xml
  
  Java configuration:
    Distribution: temurin
    Version: 11.0.20+1
    Path: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/11.0.20-1/x64
  
  Resolved Java 17.0.8+1 from tool-cache
  Setting Java 17.0.8+1 as the default
  Creating toolchains.xml for JDK version 17 from temurin
  Overwriting existing file /home/runner/.m2/toolchains.xml
  
  Java configuration:
    Distribution: temurin
    Version: 17.0.8+1
    Path: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.8-1/x64
```

laeubi:
> IIRC I noticed a similar issue in the past that the toolchains.xml was completly overwritten by each java version so that only the last one was eventually contained.

Can you open an issue about that at setup-java action? You might want to use the upload action or similar to get the content of the generated file directly, maybe the problem is more that the version is not as expected, e.g. Tycho requires that either the version is specified as `1.8` or you have the id set to `JavaSE-1.8` at least [the ID is different](https://github.com/actions/setup-java#maven-options) and maybe it is using `8` as a version string as well...

HannesWell:
> You might want to use the upload action or similar to get the content of the generated file directly, maybe the problem is more that the version is not as expected, e.g. Tycho requires that either the version is specified as `1.8` or you have the id set to `JavaSE-1.8` at least [the ID is different](https://github.com/actions/setup-java#maven-options) and maybe it is using `8` as a version string as well...

Good point! I believe to have same dark memories that this was a problem, but lets see what the log says.

HannesWell:
> > You might want to use the upload action or similar to get the content of the generated file directly, maybe the problem is more that the version is not as expected, e.g. Tycho requires that either the version is specified as `1.8` or you have the id set to `JavaSE-1.8` at least [the ID is different](https://github.com/actions/setup-java#maven-options) and maybe it is using `8` as a version string as well...
> 
> Good point! I believe to have same dark memories that this was a problem, but lets see what the log says.

Yep, that's the problem, the version is 8.
```
<?xml version="1.0"?>
<toolchains xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
  <toolchain>
    <type>jdk</type>
    <provides>
      <version>8</version>
      <vendor>temurin</vendor>
      <id>temurin_8</id>
    </provides>
    <configuration>
      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.382-5/x64</jdkHome>
    </configuration>
  </toolchain>
[...]
</toolchains>
```
Created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1399 to set the ids.
Good that someone updated the documentation about that recently: https://github.com/eclipse-tycho/tycho/pull/2379

HannesWell:
> > IIRC I noticed a similar issue in the past that the toolchains.xml was completly overwritten by each java version so that only the last one was eventually contained.
> 
> Can you open an issue about that at setup-java action? You might want to use the upload action or similar to get the content of the generated file directly, maybe the problem is more that the version is not as expected, e.g. Tycho requires that either the version is specified as `1.8` or you have the id set to `JavaSE-1.8` at least [the ID is different](https://github.com/actions/setup-java#maven-options) and maybe it is using `8` as a version string as well...

Thanks again for this hint.

Issue #730, opened 2023-08-31
Remove irrelevant attributes of plugin element in feature.xml


HannesWell:
For Plug-ins included in a feature the Feature Editor adds many attributes to the `plugin` element in a `feature.xml`.
For a regular `Plugin` in the simplest form the element looks like 
```
   <plugin
         id="foo.bar"
         download-size="0"
         install-size="0"
         version="0.0.0"
         unpack="false"/>
```
For a Fragment it is
```
   <plugin
         id="foo.bar.fragment"
         download-size="0"
         install-size="0"
         version="0.0.0"
         fragment="true"
         unpack="false"/>
```
Like for the `version` attribute, the value of the `install/download-size` are currently computed at build time and set in the final build jar of the feature.
But many of those attributes are unnecessary:
- `install/download-size` is also encoded in the P2 metadata and P2 only reads those. As far as I know nobody is reading those values from a feature
- `unpack` - if a bundle is unpacked during installation by P2 is controlled inside the Plugin with a Manifest header `Eclipse-BundleShape: dir`. The value of this attribute is read nowhere AFAIK.
- `fragment` - if a bundle is a fragment or not is controlled within the Plugin with a Manifest header `Fragment-Host` and a corresponding Bundle-Symbolic name as value. The value of this attribute is read nowhere AFAIK.

Since all these attributes are unnecessary respectively redundant they should all be removed. This means PDE should ignore them when it reads existing `feature.xml` files and should not write them writing the `feature.xml` file (yes, this will probably significantly shorten existing `feature.xml` on the next change).
Consequently all code associated (to read, write, etc) with those attributes should be removed and if necessary replaced by better means (e.g. the icon displayed for Plugins/Feature should be selected based on if it is a Plugin/Feature according to the manifest).

The attributes `os`, `ws`, `nl`, `arch` are currently considered by P2 as far as I can tell. On the long run we could consider to even derive those values from the value of the `Eclipse-PlatformFilter` or `Require-Capability` header and just install what can resolve in the current environment. But that's another discussion and out of the scope of this issue.

HannesWell:
This seems to be a good first issue and everyone that wants to become more familiar with the PDE code-base is invited to work on this.
If you decide to work on this, please just let us and everyone know that to prevent duplicated work.
If nobody steps up within the next one or two month I will take care of this by myself.

merks:
Indeed, these attributes are mostly confusing and misleading.

vik-chand:
[...]

PR #770, opened 2023-09-27
Remove unused attributes in auto-generated feature.xml

alshamams:
This commit removes some of the unused attributes of plugin element, which are redundant and are never read from a feature.xml file: install-size, download size and unpack attributes.

Fixes: [730](https://github.com/eclipse-pde/eclipse.pde/issues/730)

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/feature/FeatureDataTestCase.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/feature/FeatureDataTestCase.java
@@ -41,10 +41,7 @@ public class FeatureDataTestCase {
 	public void testSerialization() throws Exception {
 		String expected = String.format("<data%n" +
 				FEATURE_INDENT + "id=\"id.1\"%n" + FEATURE_INDENT + "os=\"win32\"%n" + FEATURE_INDENT
-				+ "arch=\"x86_64\"%n" + FEATURE_INDENT
-				+ "download-size=\"1000\"%n" +
-				FEATURE_INDENT + "install-size=\"0\"/>%n"
-				);
+				+ "arch=\"x86_64\"/>%n");
 		FeatureData data = newFeatureData();
 		String actual = toXml(data);
 		assertEquals(expected, actual);
@@ -54,16 +51,13 @@ public class FeatureDataTestCase {
 		assertEquals(data.getOS(), data2.getOS());
 		assertEquals(data.getArch(), data2.getArch());
 		assertEquals(data.getFilter(), data2.getFilter());
-		assertEquals(data.getDownloadSize(), data2.getDownloadSize());
 	}
 
 	@Test
 	public void testSerializationWithLdapFilter() throws CoreException {
 		String expected = String.format("<data%n" + FEATURE_INDENT + "id=\"id.1\"%n" + FEATURE_INDENT + "os=\"win32\"%n"
 				+ FEATURE_INDENT + "arch=\"x86_64\"%n" + FEATURE_INDENT
-				+ "filter=\"(&amp; (osgi.ws=win32) (osgi.os=win32) (osgi.arch=x86))\"%n" + FEATURE_INDENT
-				+ "download-size=\"1000\"%n" + FEATURE_INDENT
-				+ "install-size=\"0\"/>%n");
+				+ "filter=\"(&amp; (osgi.ws=win32) (osgi.os=win32) (osgi.arch=x86))\"/>%n");
 		FeatureData data = newFeatureData();
 		data.setFilter("(& (osgi.ws=win32) (osgi.os=win32) (osgi.arch=x86))");
 		String actual = toXml(data);
@@ -79,7 +73,6 @@ public class FeatureDataTestCase {
 		data.setLabel("Feature1");
 		data.setArch("x86_64");
 		data.setOS("win32");
-		data.setDownloadSize(1000);
 		return data;
 	}
 
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureErrorReporter.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureErrorReporter.java
@@ -31,7 +31,6 @@ import org.eclipse.pde.internal.core.builders.IncrementalErrorReporter.VirtualMa
 import org.eclipse.pde.internal.core.ibundle.IBundlePluginModel;
 import org.eclipse.pde.internal.core.ibundle.IManifestHeader;
 import org.eclipse.pde.internal.core.ifeature.IFeatureModel;
-import org.eclipse.pde.internal.core.util.CoreUtility;
 import org.eclipse.pde.internal.core.util.IdUtil;
 import org.w3c.dom.Attr;
 import org.w3c.dom.Element;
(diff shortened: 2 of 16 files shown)


alshamams:
I did not remove the fragment attribute, as it causes warnings in feature.xml compiler. I will investigate where the compiler is processing that, and create a separate PR to take that out

HannesWell:
@alshamams thanks for working on this.
It looks like this change can go much further and remove all modelling of all attributes mentioned in https://github.com/eclipse-pde/eclipse.pde/issues/730#issue-1876399706 as well as their reading and writing and the UI support for them.
Eventually they should not be written, read or displayed in the UI. For existing Features (e.g. from a target-platform) those attributes should just be ignored. New or modified features should have them removed (which I assume happens automatically if they are not read and written on changes, so you don't have to add some logic to actively remove them).
To make it short, anything that belongs to those attributes should be removed.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/feature/FeatureDataTestCase.java
```diff
@@ -42,9 +42,7 @@ public void testSerialization() throws Exception {
 		String expected = String.format("<data%n" +
 				FEATURE_INDENT + "id=\"id.1\"%n" + FEATURE_INDENT + "os=\"win32\"%n" + FEATURE_INDENT
 				+ "arch=\"x86_64\"%n" + FEATURE_INDENT
-				+ "download-size=\"1000\"%n" +
-				FEATURE_INDENT + "install-size=\"0\"/>%n"
-				);
+				+ "download-size=\"1000\"/>%n");
```
gireeshpunathil, line 45: this can now go too?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/feature/FeatureDataTestCase.java
```diff
@@ -54,16 +52,14 @@ public void testSerialization() throws Exception {
...
 	}
 
 	@Test
 	public void testSerializationWithLdapFilter() throws CoreException {
 		String expected = String.format("<data%n" + FEATURE_INDENT + "id=\"id.1\"%n" + FEATURE_INDENT + "os=\"win32\"%n"
 				+ FEATURE_INDENT + "arch=\"x86_64\"%n" + FEATURE_INDENT
 				+ "filter=\"(&amp; (osgi.ws=win32) (osgi.os=win32) (osgi.arch=x86))\"%n" + FEATURE_INDENT
-				+ "download-size=\"1000\"%n" + FEATURE_INDENT
-				+ "install-size=\"0\"/>%n");
+				+ "download-size=\"1000\"/>%n");
```
gireeshpunathil, line 62: ditto

HannesWell:
Thanks @alshamams, after a quick look this looks much better. 👍🏽
I'm away over the weekend until mid of next week, but will review it in detail when back.
Of course everybody else review is also welcome.

HannesWell (commented):
I reviewed this in detail and everthing that was removed is fine.
There was just a little bit more to remove.
In `FeaturePlugin` the `fFragment` field can be removed and instead `isFragment()` returns if the PluginBase model is a fragment. The returned value is only considered to select the icon of an included Plugin in the Feature editor.
This also removes the `fragment` attribute, which should not be read or written anymore.

The method `FeatureExportOperation.setAdditionalAttributes()` can be removed because all overrides are now removed and the texts for all removed elements in the PluginDetailsSection in the Feature Edeitor can be removed too.

For all of that I have pushed another commit at the tip of your branch.

I also verified that with your change and all the unnecessary attributes are not written and displayed anymore. 👍🏽

Please address the only remaining comment and squash all your commits and mine into one commit with a reasonable overall message.
Then I think this should be ready for submission.

@laeubi can you tell if Tycho already handles features with all those features removed properly?

szarnekow:
@HannesWell Do I read this part as you intended?

> In FeaturePlugin the fFragment field can be removed

Is this a breaking change? Would existing feature.xml files no longer be understood by the newest revision of the parser?

Why I am asking this:
We build a product based on Eclipse, compatible with many Eclipse versions. We use multiple target platforms in our build as a sanity check to ensure that we can compile against old and new versions before we run our tests against different TPs, too. Our workspace doesn't have any warnings.
Old versions require the fragment property to be present in the feature.xml if the feature includes a fragment. If the new version doesn't allow this anymore, it will not be source-compatible and require a major version increment. Would it be worth it?

HannesWell:
> > In FeaturePlugin the fFragment field can be removed
> 
> Is this a breaking change? Would existing feature.xml files no longer be understood by the newest revision of the parser?

It shouldn't be a breaking change. For existing `feature.xml` the now unsupported attributes are simply ignored. And the other way round if e.g. the `download-size` attribute is absent a current Eclipse simply shows a `0` in the Editor.
You can verify that by simplify removing the attributes to be discontinued in a current workspace.

> Our workspace doesn't have any warnings.
> Old versions require the fragment property to be present in the feature.xml if the feature includes a fragment.

Yes a missing `fragment` attribute leads to a warning.
If that isn't possible it also isn't a problem to keep the fragment attribute (or the others) as they are then simply ignored. Those attributes are not even reported as Illegal attributes since the validation is (for now) kept intact:
https://github.com/eclipse-pde/eclipse.pde/blob/6b0f0ed23599ae076e6463931e85fc38f6fe7dff/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureErrorReporter.java#L124-L126

Some time later we can consider to remove even that so that a Illegal attribute will be issued for them if they are present.
The only inconvenience is that each time you modify a `feature.xml` through the editor all unknown attributes get removed.

But what makes me wonder, why can't you just use the latest Eclipse release?
Once a Feature is build and is in the TP it should not be validated anymore by the `FeatureErrorReporter`, so downstream consumers should not be affected.

HannesWell:
I also checked Tycho. A feature.xml without `install-size`, etc. can be build even with a current Tycho. But at the moment Tycho then still adds the install/download-size into the build feature.xml.
That should be adjusted.

alshamams:
Hi @HannesWell ,
I have addressed your latest review comment and squashed the commits. Kindly have a look and let me know.

vogella:
@HannesWell have your concerns be adressed?

HannesWell:
> Hi @HannesWell , I have addressed your latest review comment and squashed the commits. Kindly have a look and let me know.

Thanks for the update @alshamams! Your change looks good now. 👍🏽
I also checked the code-base of P2 for references to the attributes about to be removed and found references to `unpack`

But while checking the P2 code-base I found some references to the attributes about to be removed, i.e. the `unpack` attribute, where I'm not yet entirely sure if it is absolutely not used. I already checked with Tycho that it has no effect, but I have to figure out if the same applies for all use cases in P2. I'm working on it but would like to have confidence about it before submitting this.
Thus I would like to keep this open probably for a few more days, but you can probably consider this work as done.

HannesWell:
A quick update on this one.
The next Tycho release will handle `feature.xml` without the attributes to be removed properly (i.e. does not add the inferred sizes).
For Tycho 5 there is no full aggreement yet how to handle those attributes, but since its release is further in the future that's not a concern for this one.

What's currently blocking the submission of this one is that `CoreUtility.guessUnpack()` is currently used in P2 and we need to remove its usage before we can remove it here. Unfortunately the usage is quite convoluted and I didn't went 100% through it yet.
I'll be at EclipseCon from tomorrow on and will therefore not be able to continue the work on the P2 site until the end of the coming week. If any body of you attends as well, I'll be glad to have a chat about this. :)

vogella:
Is there an issue to follow the work of CoreUtility.guessUnpack()? 

Maybe remove the UI elements as a first step?

image

HannesWell (approved):
> Is there an issue to follow the work of CoreUtility.guessUnpack()?

I just created https://github.com/eclipse-equinox/p2/pull/368 to some related work on P2. But it looks like PDE-build again uses some parts of that again.
However the good news is that I confused `CoreUtility.guessUnpack()` with another Utility class in P2 and it is in fact not used and this change is ready for submission now.


Thank you @alshamams for your work on this issue.

alshamams:
> > Is there an issue to follow the work of CoreUtility.guessUnpack()?
> 
> I just created [eclipse-equinox/p2#368](https://github.com/eclipse-equinox/p2/pull/368) to some related work on P2. But it looks like PDE-build again uses some parts of that again. However the good news is that I confused `CoreUtility.guessUnpack()` with another Utility class in P2 and it is in fact not used and this change is ready for submission now.
> 
> Thank you @alshamams for your work on this issue.

Thanks @HannesWell !

vogella:
Thanks @alshamams and @HannesWell , always great if we get rid of outdated stuff.

PR #781, opened 2023-10-04
Update version ranges to work with biz.aQute.bndlib 7.0.0

HannesWell:
Prepare for the upcoming BND-lib 7.1 release.

Lets see if widening the version range is sufficient. Locally it at least compiled successfully.

In the context of https://github.com/eclipse-pde/eclipse.pde/discussions/774.


pom.xml
```diff
@@ -37,6 +37,13 @@
     <module>build</module>
     <module>features</module>
   </modules>
+  <repositories>
+	  <repository>
+		  <id>bnd-snapshot</id>
+		  <layout>p2</layout>
+		  <url>https://bndtools.jfrog.io/bndtools/update-snapshot</url>
+	  </repository>
+  </repositories>
```
HannesWell, line 46: This is only temporary until bnd-lib 7.1 is released and it is included in the sdk-prereq-target.
Has to be reverted before this can be merged.

HannesWell:
Lets see if this really works once bnd-lib is updated in the TP.

Issue #785, opened 2023-10-07
BaseApiAnalyzer should have a way to discover the source without a java project


laeubi:
I'm currently working on an enhance ApiTools Mojo that works on the build jar file and therefore does not require recompilation of the whole project for API analysis with potentially different settings (see for example https://github.com/eclipse-pde/eclipse.pde/issues/782).

This already works quite well but has the problem that `org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer` currently requires the source for some features (e.g. giving the location of the API problem).

See for example `org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.createCompatibilityProblem(IDelta, IApiComponent, IApiComponent)` or `org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.checkSinceTags(Delta, IApiComponent)`

What I would like to have is a way to tell the `BaseApiAnalyzer` (e.g. with a method `setComponetnSource(...)`) where the source is located and then the AST is used (I can also supply the full classpath if that would be required) instead of a full java project.

Another approach would be to construct a complete new java project from the existing one but I currently don't see how this can be archived without triggering all kind of unwanted stuff e.g. compilation, execution of additional builders and so on. Currently 
[ApiAnalysisApplication](https://github.com/eclipse-pde/eclipse.pde/blob/c9cbfbead1adb2dd9af108479c990923fdbc3da0/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiAnalysisApplication.java#L314) do so by performing a lot of non trivial operations including restoring project states and of course it always requires a `.project` file in the version control system.

@iloveeclipse @vik-chand can you probably help here?

laeubi:
I even wonder if one can discover the line/name from the class debug symbols as well...

PR #786, opened 2023-10-08
Make BaseApiAnalyzer#getJavaProject protected

laeubi:
Currently there is no way to pass the project to the analyzer without having an project ProjectComponent, but users of BaseApiAnalyzer might have other means to find the project in question.

This makes the BaseApiAnalyzer#getJavaProject() protected so implementors can override this (e.g. Tycho).

This relates to
- https://github.com/eclipse-pde/eclipse.pde/issues/785


laeubi:
Testfailure is already documented here:
- https://github.com/eclipse-pde/eclipse.pde/issues/553

PR #788, opened 2023-10-08
Wait for jobs to complete and perform up to three retries

laeubi:
Currently exporting the API description sometimes fails because there could be a race between a resource change event and disposal of the api component.

This changes the following things:
- the API description is performed as the very first step
- if generation fails wait until all running jobs are done
- then retry up to three times

This should fix https://github.com/eclipse-pde/eclipse.pde/issues/553


HannesWell:
Thanks for working on this. From a coarse view this looks good to me (I'm not so much into API-Tools code).

But I wonder why the builds didn't start? Some infra-structure issue? Could you please just push again to trigger them again?

Issue #207, opened 2022-07-13
[target] consider other locations in planner mode


laeubi:
Currently in planner mode (if one selects include required software) PDE only takes into account other IU locations. It would be useful to either automatically or just always include content from other locations as well see for example tycho currently works this way and it becomes quite handy e.g. in combination with providing a missing bundle by a maven or directory target.

jonahgraham:
A real world example of where this is helpful is explained in #675.

HannesWell:
Support for this would be good.

In https://github.com/eclipse-pde/eclipse.pde/issues/675#issuecomment-1646525555 @merks suggested
> I suppose PDE might be made smarter such that it resolves all non-IUBundleContainer containers first, and is able to synthesis a p2 repository with all the resolved IUs and make those available when resolving the remaining IUBundleContainers. That sounds quite non-trivial...

In https://github.com/eclipse-pde/eclipse.pde/issues/675#issuecomment-1646538691 @laeubi responded
> > I suppose PDE might be made smarter such that it resolves all non-IUBundleContainer containers first, and is able to synthesis a p2 repository with all the resolved IUs and make those available when resolving the remaining IUBundleContainers. That sounds quite non-trivial...
> 
> Actually it is not _that_ complex but would require a two phase resolve operation, first collect and resolve all direct items and then perform the planner operation afterwards, that's how Tycho do it. But all this is _behavioral_ change, probably _breaking_ change and that's where it becomes not so trivial for anyone not liking getting punished for that.

@laeubi can you elaborate in more detail how a two phase resolution does solve this issue? At any point the content of other locations must be considered, but I don't understand how a two phase resolution would solve this auto-magically?
And in which cases would that be a breaking change? I always assumed that the result of the slicer is a sub-set of the planner?
Or do you mean a resolution in the PDEState? In general this has to be distinguished.

At the same time Ed's suggestion seems to be, at least from a theoretical point-of-view, relatively simple. I cannot access how difficult this would be to implement with P2.

merks:
One of the key components that does this/similar work for Oomph's targlets is org.eclipse.oomph.targlets.internal.core.TargletContainer.ComposedTargetContent.  (Did you know that Navigate -> Open Discovered Type.. can be used to find any Java top-level class in any Eclipse Git repository?)

[...]

PR #792, opened 2023-10-08
Consider other locations when resolving IUs from updatesites

laeubi:
Currently the IULocation can only consider units from other locations if these are also of type IU location but other target types might be able to supply items in a way not know to the IULocation.

This adds the necessary infrastructure to support this usecase by using the adapter pattern.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/207
Requires https://github.com/eclipse-equinox/p2/pull/348

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/InstallableUnitGenerator.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/InstallableUnitGenerator.java
@@ -1,0 +1,149 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.core.target;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+import java.util.stream.Stream;
+
+import org.eclipse.equinox.frameworkadmin.BundleInfo;
+import org.eclipse.equinox.internal.p2.publisher.eclipse.FeatureManifestParser;
+import org.eclipse.equinox.internal.p2.publisher.eclipse.FeatureParser;
+import org.eclipse.equinox.p2.metadata.IInstallableUnit;
+import org.eclipse.equinox.p2.publisher.IPublisherInfo;
+import org.eclipse.equinox.p2.publisher.PublisherInfo;
+import org.eclipse.equinox.p2.publisher.PublisherResult;
+import org.eclipse.equinox.p2.publisher.eclipse.BundlesAction;
+import org.eclipse.equinox.p2.publisher.eclipse.Feature;
+import org.eclipse.equinox.p2.publisher.eclipse.FeaturesAction;
+import org.eclipse.equinox.p2.query.QueryUtil;
+import org.eclipse.osgi.framework.util.CaseInsensitiveDictionaryMap;
+import org.eclipse.osgi.service.resolver.BundleDescription;
+import org.eclipse.pde.core.IModel;
+import org.eclipse.pde.core.target.TargetBundle;
+import org.eclipse.pde.core.target.TargetFeature;
+import org.eclipse.pde.internal.core.ifeature.IFeature;
+import org.eclipse.pde.internal.core.ifeature.IFeatureModel;
+import org.xml.sax.SAXException;
+
+/**
+ * Class to generate {@link IInstallableUnit}s from target objects
+ */
+@SuppressWarnings("restriction")
+public class InstallableUnitGenerator {
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetBundle[] bundles, TargetFeature[] features) {
+		return Stream.concat(generateInstallableUnits(bundles), generateInstallableUnits(features));
+	}
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetBundle[] bundles) {
+		if (bundles == null || bundles.length == 0) {
+			return Stream.empty();
+		}
+		return Arrays.stream(bundles).flatMap(InstallableUnitGenerator::generateInstallableUnits);
+	}
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetBundle targetBundle) {
+		BundleInfo bundleInfo = targetBundle.getBundleInfo();
+		if (bundleInfo != null) {
+			String manifest = bundleInfo.getManifest();
+			if (manifest != null) {
+				try {
+					Manifest parsed = new Manifest(new ByteArrayInputStream(manifest.getBytes(StandardCharsets.UTF_8)));
+					Attributes mainAttributes = parsed.getMainAttributes();
+					CaseInsensitiveDictionaryMap<String, String> headers = new CaseInsensitiveDictionaryMap<>(
+							mainAttributes.size());
+					Set<Entry<Object, Object>> entrySet = mainAttributes.entrySet();
+					for (Entry<Object, Object> entry : entrySet) {
+						headers.put(entry.getKey().toString(), entry.getValue().toString());
+					}
+					PublisherInfo publisherInfo = new PublisherInfo();
+					publisherInfo.setArtifactOptions(IPublisherInfo.A_INDEX);
+					BundleDescription bundleDescription = BundlesAction.createBundleDescription(headers, null);
+					IInstallableUnit iu = BundlesAction.createBundleIU(bundleDescription,
+							BundlesAction.createBundleArtifactKey(bundleDescription.getSymbolicName(),
+									bundleDescription.getVersion().toString()),
+							publisherInfo);
+					return Stream.of(iu);
+				} catch (IOException e) {
+					// can't use it then...
+				}
+			}
+		}
+		return Stream.empty();
+	}
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature[] features) {
+		if (features == null || features.length == 0) {
+			return Stream.empty();
+		}
+		return Arrays.stream(features).flatMap(InstallableUnitGenerator::generateInstallableUnits);
+	}
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature targetFeature) {
+		String location = targetFeature.getLocation();
+		if (location != null) {
+			Feature feature = new FeatureParser().parse(new File(location));
+			if (feature != null) {
+				feature.setLocation(location);
+				return generateInstallableUnits(feature);
+			}
+		}
+		IModel model = targetFeature.getFeatureModel();
+		if (model instanceof IFeatureModel featureModel) {
+			IFeature feature = featureModel.getFeature();
+			if (feature != null) {
+				StringWriter stringWriter = new StringWriter();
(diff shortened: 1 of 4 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1547,6 +1551,24 @@ private URI[] getMetadataRepositories(ITargetDefinition target) throws CoreExcep
...
+	private List<IInstallableUnit> getAdditionalProvisionIUs(ITargetDefinition target) {
+		List<IInstallableUnit> result = new ArrayList<>();
+		ITargetLocation[] containers = target.getTargetLocations();
+		if (containers != null) {
+			for (ITargetLocation container : containers) {
+				if (container instanceof IUBundleContainer) {
+					// this is already handled by getMetadataRepositories(..)
+					continue;
+				}
+				IInstallableUnitQueryable queryable = Adapters.adapt(container, IInstallableUnitQueryable.class);
```
HannesWell, line 1563: Instead of Using `Adapters` and requiring a special sub-class of `IQueryable`, can't we use the existing `ITargetLocation.getBundles()/getFeatures()` and wrap the returned `TargetBundle/TargetFeature` instance to an `IInstallableUnit`?
Maybe this would require to change the order in `TargetDefinition.resolve()` in which currently IU-Locations are resolved first.
This would also have the advantage that there are no adjustments in other Locatation-Types (like Maven Locations) necessary.

laeubi: > Instead of Using Adapters and requiring a special sub-class of IQueryable, can't we use the existing ITargetLocation.getBundles()/getFeatures() and wrap the returned TargetBundle/TargetFeature instance to an IInstallableUnit?

Not really as there is no guarantee that the set of *resolved* bundles an features is the full set of available items.
Just assume the IU case:

- I have an update site that contains 100 IUs
- My location selects only one that has no dependencies so `getBundle()` will return one IU
- No there is another location that has a requirement to one of the not selected items
- --> resolve error as the other 99 IU are not visible through this approach

Beside that there would be a risk of endless recursion if `A` want the bundles of `B` and B then in return queries `A`.

> Maybe this would require to change the order in TargetDefinition.resolve() in which currently IU-Locations are resolved first.

This will require defiantly changes there, I'm actually not very happy with all this specific handling of concrete types but that's another story.

> This would also have the advantage that there are no adjustments in other Locatation-Types (like Maven Locations) necessary.

I don't think its possible to solve this without changing the implementation of location types as there currently is no way to know what else *might* be available.

laeubi:
This is now ready as the first minimal increment. It is likely that there are some changes one like to adjust sometime but at least this allows to resolve an `IU` location when using a `Directory` location that supply a missing bundle as in this example:


```
<target name="target_require_other">
   <locations>
      <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
         <repository location="https://download.eclipse.org/linuxtools/updates-docker-nightly/"/>
         <unit id="com.google.guava" version="0.0.0"/>
      </location>
      <location path="${project_loc:/test.pde}/lib" type="Directory"/>
      </locations>
</target>
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
HannesWell: > > This would also have the advantage that there are no adjustments in other Locatation-Types (like Maven Locations) necessary.
> 
> I don't think its possible to solve this without changing the implementation of location types as there currently is no way to know what else _might_ be available.

Agree on that, but nevertheless I think this could all be solved in PDE.
As far as I can overview it, what's effectively done, is to collect all target-bundles and -features of all location and wrap them into IUs.
But I don't see a need why this has to be solved via the Adapter-Approach (in general I find that concept bad because it effectively defines an API but does not look like one and is hard to find).
Why can't we just add an Interface like `IIUProviderTargetLocation` (or whatever is suitable) to **PDE** that has a method `getAllBundles()` and `getAllFeature()`? Then this could be called here without the indirection via the Adapter and an `IQueryable`?
Sure, we probably have to define another interface but I think it should be defined in PDE instead of P2, because from what I see now it is just used for target-locations and has no value for P2.

This would also have the advantage that for example in M2E or Oomph the internal `InstallableUnitGenerator` has not to be referenced or reimplemented to provide the IUs.

Thus I think https://github.com/eclipse-equinox/p2/pull/348 is not necessary and should be reverted to not unnecessarily add extra interfaces to P2 that only serve PDE and make the integration more complex.

Or is there something in the bigger picture that I fail to see?

laeubi: Adapter pattern is widely used in Eclipse and PDE and makes the code more flexible (e.g. you don't need to actually implement that in the class itself but can have a separate wrapper/adapter) and less coupled so I don't se a reason not using it here. Especially as I don't want to introduce an "API" this is completely internal behavior of the `IUBundleLocation` and not of general interest for others.

That's also why I don't want something like a `IIUProviderTargetLocation` or similar as this will only manifest the already bad design of existing Target Location API design that requires some materialized bundles / features while in fact we want to operate on the **metadata** of the IUs. That's also a reason why PDE is ultra-slow in resolving such locations compared to Tycho.

> This would also have the advantage that for example in M2E or Oomph the internal InstallableUnitGenerator has not to be referenced or reimplemented to provide the IUs.

P2 and PDE actually require to reference "internal" stuff all the way and m2e was always requiring latest PDE to support best integration so I don't really  mind that. For Targelts I also can thing that Oomph already operates on P2 InstallableUnits and don't need a converter anyway.

Most important, these changes are more intermediate solutions to lower the pain we currently see, my main goal is to rework how PDE TargetLocations are implemented in general to using the OSGi Repository API that will be much more flexible and complete than anything we can archive with the current stuff where we must respect so much legacy..

laeubi:
Testfailure seems unrelated, build is otherwhise green...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
HannesWell: > Especially as I don't want to introduce an "API" this is completely internal behavior of the `IUBundleLocation` and not of general interest for others.

But effectively you make this an API here, because if you remove the code something does not work as before. Behavior is also API although the API-Tools or similar checks don't cover that. That's why I don't like the Adapter approach. It often does not feel like an API but effectively it is. Any way.


> Most important, these changes are more intermediate solutions to lower the pain we currently see, my main goal is to rework how PDE TargetLocations are implemented in general to using the OSGi Repository API that will be much more flexible and complete than anything we can archive with the current stuff where we must respect so much legacy..

Acknowledged that and that sounds like a great plan. But from my experience resolving does not download artifacts, but it still is not that fast. :) I assume it works on the Plan the Planner creates to update the Target's profile.

Back to topic, I'm still convinced we can just use `IQueryable<IInstallableUnit>` here.
The only drawback is that `ExpressionMatchQuery` does not perform an actual type-check if the operand is `ExpressionUtil.TRUE_EXPRESSION`. Therefore an explicit type-check is necessary to filter IUs and we cannot rely on the generic type. But thanks to the new `IQueryResult.stream()` method this is quite simple. :)

I'm about to push another commit on top of your branch to demonstrate it.
With that I would be more than happy to have this great new feature. :)

laeubi:
Using a Generic Type is simply bad and fixing this later on up by adding filter only complicates the whole story for absolutely no gain and will limit our ability to later, also this discussion about "API" or "not API" does not bring any value this is not API this is a decoupled way of using optional features and if we remove it and later replace it by some other means things have to be adjusted or one can use another Adapter that bridges old -> new behavior as done already in the past when migrating to the more feature rich target locations..

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
laeubi: > But from my experience resolving does not download artifacts

"Resolving" a target location requires to download every item to disk because the `TargetBundle` requires a physical file,for `TargetFeature` I have already lifted that requirement to support virtual features but if  a feature is not virtual it again requires a physical download as you can't recover the feature model from the meta-data in almost all cases.

laeubi:
@HannesWell I now reworked the whole part so it automatically queries the locations and converts them I hope this now addresses your concerns.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/InstallableUnitGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+									bundleDescription.getVersion().toString()),
+							publisherInfo);
+					return Stream.of(iu);
+				} catch (IOException e) {
+					// can't use it then...
+				}
+			}
+		}
+		return Stream.empty();
+	}
```
HannesWell, line 95: Using `Stream.mapMulti()` would save the creation of Streams with one element.

```suggestion
	private static Stream<IInstallableUnit> generateInstallableUnits(TargetBundle[] bundles) {
		if (bundles == null || bundles.length == 0) {
			return Stream.empty();
		}
		return Arrays.stream(bundles).mapMulti(InstallableUnitGenerator::generateInstallableUnits);
	}

	private static void generateInstallableUnits(TargetBundle targetBundle, Consumer<IInstallableUnit> downStream) {
		BundleInfo bundleInfo = targetBundle.getBundleInfo();
		if (bundleInfo != null) {
			String manifest = bundleInfo.getManifest();
			if (manifest != null) {
				try {
					Manifest parsed = new Manifest(new ByteArrayInputStream(manifest.getBytes(StandardCharsets.UTF_8)));
					Attributes mainAttributes = parsed.getMainAttributes();
					CaseInsensitiveDictionaryMap<String, String> headers = new CaseInsensitiveDictionaryMap<>(
							mainAttributes.size());
					Set<Entry<Object, Object>> entrySet = mainAttributes.entrySet();
					for (Entry<Object, Object> entry : entrySet) {
						headers.put(entry.getKey().toString(), entry.getValue().toString());
					}
					PublisherInfo publisherInfo = new PublisherInfo();
					publisherInfo.setArtifactOptions(IPublisherInfo.A_INDEX);
					BundleDescription bundleDescription = BundlesAction.createBundleDescription(headers, null);
					IInstallableUnit iu = BundlesAction.createBundleIU(bundleDescription,
							BundlesAction.createBundleArtifactKey(bundleDescription.getSymbolicName(),
									bundleDescription.getVersion().toString()),
							publisherInfo);
					downStream.accept(iu);
				} catch (IOException e) {
					// can't use it then...
				}
			}
		}
	}
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/InstallableUnitGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature[] features) {
```
HannesWell, line 97:
```suggestion
	private static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature[] features) {
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/InstallableUnitGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+		if (features == null || features.length == 0) {
+			return Stream.empty();
+		}
+		return Arrays.stream(features).flatMap(InstallableUnitGenerator::generateInstallableUnits);
+	}
+
+	public static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature targetFeature) {
```
HannesWell, line 104:
```suggestion
	private static Stream<IInstallableUnit> generateInstallableUnits(TargetFeature targetFeature) {
```

HannesWell (approved):
> @HannesWell I now reworked the whole part so it automatically queries the locations and converts them I hope this now addresses your concerns.

Yes, this looks perfectly fine, thanks. 👍🏽 
I'm glad we came to a better solution that has less requirements and also works without adjustments in other target-locations. :)

Just two small style-remarks below, but nothing critical.

Issue #600, opened 2023-05-20
OSGi Metadata Automatic Generation: export dialog doesn't list projects that have auto gen enabled.


niport:
Version: 2023-06 M2 (4.28.0 M2)
Build id: 20230504-1428
OS: Windows 10

The dialog `File > Export... > Deployable Plug-ins and Fragments` doesn't show any of the plugins that have been created with the `"Generate OSGi metadata automatically"` option enabled.

The same happens if the same dialog is opened by clicking on the export button present on the upper right corner of the pde.bnd editor.

![image](https://github.com/eclipse-pde/eclipse.pde/assets/36211095/8cfad555-ab7a-4613-adb3-43e94bd9a94c)

PR #799, opened 2023-10-13
List OSGI plugins in the export list

alshamams:
The commit fixes OSGi plugins (that have been created using standard OSGi framework with 'Generate OSGi metadata automatically' feature) not being displayed in the export dialog box. This happens because an OSGi project has a pde.bnd file, unlike the eclipse plugins, which has build.properties. Therefore the previous behaviour was to check for the presence of a build.properties file for validating a plugin project. This assumption breaks with the new type.

The commit relies on the following assumptions in case of OSGi template projects:
 - The location of the pde.bnd file is always under the root folder of the plugin project
 - The filename is always constant, namely, pde.bnd.
 
Fixes: #600

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/exports/PluginExportWizardPage.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/exports/PluginExportWizardPage.java
@@ -27,6 +27,7 @@ import org.eclipse.pde.core.plugin.PluginRegistry;
 import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.core.PDECore;
 import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.bnd.BndProjectManager;
 import org.eclipse.pde.internal.ui.IHelpContextIds;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
 import org.eclipse.pde.internal.ui.util.PersistablePluginObject;
@@ -53,7 +54,8 @@ public class PluginExportWizardPage extends BaseExportWizardPage {
 		for (int i = 0; i < projects.length; i++) {
 			if (!WorkspaceModelManager.isBinaryProject(projects[i]) && WorkspaceModelManager.isPluginProject(projects[i])) {
 				IModel model = PluginRegistry.findModel(projects[i]);
-				if (model != null && isValidModel(model) && hasBuildProperties((IPluginModelBase) model)) {
+				if (model != null && isValidModel(model) && (hasBuildProperties((IPluginModelBase) model)
+						|| hasPdeBndDescriptor(projects[i]))) {
 					result.add(model);
 				}
 			}
@@ -69,6 +71,15 @@ public class PluginExportWizardPage extends BaseExportWizardPage {
 	private boolean hasBuildProperties(IPluginModelBase model) {
 		File file = new File(model.getInstallLocation(), ICoreConstants.BUILD_FILENAME_DESCRIPTOR);
 		return file.exists();
+	}
+
+	private boolean hasPdeBndDescriptor(IProject project) {
+		try {
+			return BndProjectManager.getBndProject(project).isPresent();
+		} catch (Exception e) {
+			return false;
+		}
+
 	}
 
 	@Override


gireeshpunathil:
the second assumption (instruction file name is always `pde.bnd`) is definitely right due to:
https://github.com/eclipse-pde/eclipse.pde/blob/27cf508efead9b8bfbb8bef99b232f4a3d8ee453/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/natures/BndProject.java#L24-L26

the first assumption (the project structure) looks to be correct too, but pinging @laeubi to confirm just in case.

laeubi:
Yes `pde.bnd` in the project root is a "marker file" that should always be constant.

gireeshpunathil (approved):

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/exports/PluginExportWizardPage.java
```diff
@@ -71,6 +72,16 @@ private boolean hasBuildProperties(IPluginModelBase model) {
 		return file.exists();
 	}
 
+	private boolean hasPdeBndDescriptor(IPluginModelBase model) {
+		File bin = new File(model.getInstallLocation());
+		if (bin.exists()) {
+			File file = new File(bin.getParent(), ICoreConstants.BND_FILENAME_DESCRIPTOR);
+			return file.exists();
+		}
+		return false;
```
HannesWell, line 81: Wouldn't it be better to operate on `IResources` here?
I.e. check if `model.getUnderlyingResource()` returns non null and then check for the `pde.bnd` file of that resource's project?

HannesWell: Furthermore it is actually not necessary to check for the existance of the directory and file.
`File`, just like `Path` or `IFile/IFolder/...` are just handles/pointers and one can create them regardless of if they exist.
Just checking if the finally constructed `(I)File` is an existing file saves one call to the file-system (given the file exists).

laeubi: If you have a project handle one can also use `org.eclipse.pde.internal.core.natures.BndProject.isBndProject(IProject)` (that will only validate the nature) or `org.eclipse.pde.internal.core.bnd.BndProjectManager.getBndProject(IProject)` if also want to validate is has a *valid* instruction file.

laeubi: > Wouldn't it be better to operate on IResources here?
I.e. check if model.getUnderlyingResource() returns non null and then check for the pde.bnd file of that resource's project?

As far as I can see on the callerside the project is already known!

alshamams: > Furthermore it is actually not necessary to check for the existance of the directory and file. `File`, just like `Path` or `IFile/IFolder/...` are just handles/pointers and one can create them regardless of if they exist. Just checking if the finally constructed `(I)File` is an existing file saves one call to the file-system (given the file exists).

I have implemented the changes suggested by @laeubi , which has made these changes unnecessary, and also doesn't attempt file access.

alshamams: > If you have a project handle one can also use `org.eclipse.pde.internal.core.natures.BndProject.isBndProject(IProject)` (that will only validate the nature) or `org.eclipse.pde.internal.core.bnd.BndProjectManager.getBndProject(IProject)` if also want to validate is has a _valid_ instruction file.

I have implemented the changes, kindly look through and let me know @laeubi .

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ICoreConstants.java
```diff
@@ -60,6 +60,9 @@ public interface ICoreConstants {
 	/** Constant for the string <code>build.properties</code> */
 	String BUILD_FILENAME_DESCRIPTOR = "build.properties"; //$NON-NLS-1$
 
+	/** Constant for the string <code>pde.bnd</code> */
+	String BND_FILENAME_DESCRIPTOR = "pde.bnd"; //$NON-NLS-1$
```
laeubi, line 64: There is already `org.eclipse.pde.internal.core.natures.BndProject.INSTRUCTIONS_FILE` so you should at laest reference it here
```suggestion
	String BND_FILENAME_DESCRIPTOR = BndProject.INSTRUCTIONS_FILE;
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ICoreConstants.java
```diff
@@ -60,6 +60,9 @@ public interface ICoreConstants {
 	/** Constant for the string <code>build.properties</code> */
 	String BUILD_FILENAME_DESCRIPTOR = "build.properties"; //$NON-NLS-1$
 
+	/** Constant for the string <code>pde.bnd</code> */
```
laeubi, line 63: This seems not used anymore in your PR maybe revert this change?

alshamams:
> @alshamams can you please rebase/squash and cleanup the mentioned no longer needed part? The I think this can be merged.

Hi @laeubi ,
I have reverted the change and rebased after squashing.

laeubi (approved):
For me it looks all sane @HannesWell  @gireeshpunathil any concerns to merge?

HannesWell (approved):
No concern from me, just a minor, Not required suggestion.

HannesWell:
Thanks @alshamams please squash all commits into one, then we can submit this.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/exports/PluginExportWizardPage.java
```diff
@@ -71,6 +73,15 @@ private boolean hasBuildProperties(IPluginModelBase model) {
 		return file.exists();
 	}
 
+	private boolean hasPdeBndDescriptor(IProject project) {
+		try {
+			return !BndProjectManager.getBndProject(project).isEmpty();
```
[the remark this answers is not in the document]
alshamams: I have made the changes @HannesWell .

alshamams:
> Thanks @alshamams please squash all commits into one, then we can submit this.

Done @HannesWell

HannesWell (approved):
Thanks @alshamams.

In general you can actually always squash locally or just amend existing commits and force push as one commit (unless you eventually want to make a change with multiple commits that are self-contained by themself).
After force-push you can still see in the Github UI what as changed by clicking the `Compare` button in the feed:

grafik

Since Gireesh and Christoph were fine with this change too, I'll submit it as soon as the build completes.

Issue #798, opened 2023-10-12
Creation of a plugin project with Eclipse as target throws errors if automatic metadata generation option is left unchecked


alshamams:
Steps to recreate:
 - Create a plugin project.
 - Select 'Generate OSGi metadata automatically' in the 'target to run with' option
 - Finish
 - Create another plugin project
 - Select Eclipse in the 'target to run with' option
 - Select a standard plugin template
 - Finish
This results in plugin creation error:
Cannot invoke "org.eclipse.pde.core.plugin.IPluginModelBase.getPluginBase()" because "model" is null

alshamams:
I plan on taking this issue up as soon as possible.

laeubi:
> This results in plugin creation error:
Cannot invoke "org.eclipse.pde.core.plugin.IPluginModelBase.getPluginBase()" because "model" is null

Can you add the full stack trace for reference here?

Issue #782, opened 2023-10-05
API tooling goes amok in Jenkins, need some extra logging


iloveeclipse:
Something similar to https://github.com/eclipse-pde/eclipse.pde/issues/771 (but not on Java 21), reported for this change (which doesn't break or add or remove any API): https://github.com/eclipse-platform/eclipse.platform.ui/pull/1097.
Full maven log: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-1097/4/consoleFull

Relevant part says that most of JFace API was removed. !?!?

I can only assume that something completely weird happening in the API application, like not finding any classes or comparing "empty" state with baseline etc.

Since we can't find the root cause, we should at least provide more data while reporting API errors.

This could be: 
* exact Java version used,
* some baseline description (path/content or something like that)
* whatever else that could be of interest from API tooling point of view

```
15:12:24.531 [INFO] --- tycho-p2-extras:4.0.3:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.jface ---
15:12:24.539 [INFO] 
15:12:24.539 [INFO] --- tycho-apitools:4.0.3:verify (verify) @ org.eclipse.jface ---
15:12:24.541 [INFO] Resolve API baseline for org.eclipse.platform:org.eclipse.jface:eclipse-plugin:3.32.0-SNAPSHOT
15:12:25.211 [INFO] Resolve API tools runtime from MavenRepositoryLocation [location=https://download.eclipse.org/eclipse/updates/4.30-I-builds/]...
383 API ERRORS
[API ERROR] File MANIFEST.MF at line 5: The major version should be incremented in version 3.32.0, since API breakage occurred since version 3.31.0 - The type org.eclipse.jface.action.IMenuListener2 in org.eclipse.jface_3.32.0 is no longer an API
- The type org.eclipse.jface.dialogs.MessageDialog in org.eclipse.jface_3.32.0 is no longer an API
- The type org.eclipse.jface.viewers.AbstractListViewer in org.eclipse.jface_3.32.0 is no longer an API
- The type org.eclipse.jface.action.StatusLineManager in org.eclipse.jface_3.32.0 is no longer an API
- The type org.eclipse.jface.viewers.TreeSelection in org.eclipse.jface_3.32.0 is no longer an API
- The type org.eclipse.jface.bindings.keys.SWTKeyLookup in org.eclipse.jface_3.32.0 is no longer an API
[...]

PR #800, opened 2023-10-13
Toggle OSGi and Eclipse plugin creation options appropriately

alshamams:
If one creates plugins using the automatic manifest generation feature,
subsequent plugin projects created with Eclipse as the target is throwing errors, due to the 'Generate OSGi metadata automatically' button not being unchecked. This commit tries to resolve this issue by unchecking the feature whwnever an Eclipse plugin project using templates is created.

Fixes: #798


gireeshpunathil:
change looks good, build is failing due to issue mentioned in #782

laeubi:
I think this is better fixed with:
- https://github.com/eclipse-pde/eclipse.pde/pull/829

laeubi:
@alshamams I just merged a fix for this can you verify if it solves your problem? In the meanwhile I'll close this PR.

PR #806, opened 2023-10-19
Setting BUNDLE_ROOT_PATH preference to an empty string gives an error

laeubi:
Currently one must either specify a folder or delete the preference BUNDLE_ROOT_PATH or PDE gives an error message.

Instead we should interpret an empty BUNDLE_ROOT_PATH as the explicit intend of the user to use the project as a bundle root.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/PDEProject.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/PDEProject.java
@@ -60,6 +60,9 @@ public class PDEProject {
 		if (node != null) {
 			String string = node.get(BUNDLE_ROOT_PATH, null);
 			if (string != null) {
+				if (string.isBlank()) {
+					return project;
+				}
 				IPath path = IPath.fromPortableString(string);
 				return project.getFolder(path);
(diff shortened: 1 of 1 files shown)


HannesWell:
> Instead we should interpret an empty BUNDLE_ROOT_PATH as the explicit intend of the user to use the project as a bundle root.

Does a single dot work? Often that is treated as 'this' repository.

laeubi:
> > Instead we should interpret an empty BUNDLE_ROOT_PATH as the explicit intend of the user to use the project as a bundle root.
> 
> Does a single dot work? Often that is treated as 'this' repository.

Maybe but I don't see a reason to let the user do some guesswork what mightbe the right thing here if we can fix it easily without doubt.

HannesWell:
> > > Instead we should interpret an empty BUNDLE_ROOT_PATH as the explicit intend of the user to use the project as a bundle root.
> > 
> > Does a single dot work? Often that is treated as 'this' repository.
> 
> Maybe but I don't see a reason to let the user do some guesswork what mightbe the right thing here if we can fix it easily without doubt.

It was more ment as an alternative.
If both work, even better.

HannesWell (approved):

PR #812, opened 2023-10-22
Rework TargetDefinition#resolve

laeubi:
Currently the flow of action is quite hard to understand, also synchronizers are performed first and can stop resolving other locations even if these do not fail.

This changes it in a way where we first collect all synchronizers but perform any other location resolving first. Then all remaining items are resolved and added to the status that is then finally resolved to include all target states.

@merks @HannesWell this is a prerequisite for making 
- https://github.com/eclipse-pde/eclipse.pde/pull/792

work more smooth, also it is now much more clear what happens when...


merks (approved):
It generally looks simpler.

HannesWell:
Looks perfectly fine. 👍🏽 
Just as a remark, calling `SubMonitor.checkCanceled()` is not necessary when `SubMonitor.split()` is called, since the latter calls the former any way.
But I'm currently working on a larger clean-up of that class and will include that as well.

Issue #701, opened 2023-08-08
Support for automatic manifest generation: Syntax highlighting in bnd files


mikerumpf:
Currently the pde.bnd tab has no kind of syntax highlighting.

This makes reading and maintaining the content hard.

(Maybe this could be reused from the bndtools editor which offers that feature already.)

laeubi:
> (Maybe this could be reused from the bndtools editor which offers that feature already.)

Sadly the BndEditors syntax highlight is tightly coupled to BndTools i asked @juergen-albert about that already but until now there is no progress in this regard.

laeubi:
I'm currently looking into maybe providing a Language Server for BND Syntax:
- https://github.com/bndtools/bnd/issues/5833

@mickaelistria do you know how one would integrate LSP into Eclipse Source Pages in this case?

This is currently using `org.eclipse.pde.internal.ui.editor.MultiSourceEditor.addSourcePage(String)` that extends `TextEditor` but should probably need `ExtensionBasedTextEditor` for LSP support, can you probably help here?

mickaelistria:
Yes, the trick is to make the Source page use ExtensionBasedTextEditor instead of just TextEditor; then all LSP or TM4E configuration should work. It's similar to what was done in m2e.
I don't think I can dedicate time to that work at the moment, sorry.

laeubi:
Okay then I need to check if we can do this for the BND part at least, thinking a bit further `MANIFEST.MF` is similar thing that might be handled by a LS....

mickaelistria:
Every text file format should be handled by a LS.

laeubi:
> Yes, the trick is to make the Source page use ExtensionBasedTextEditor instead of just TextEditor

I'm currently try that out but found that `ExtensionBasedTextEditor` is actually internal API, is this intentional?

mickaelistria:
When creating the generic editor, we tried to remain protective and to make API as little as possible, so didn't export most of its content. Now that there are clear needs from accessing this class, I think it would make sense to make it more officially public.

laeubi:
> When creating the generic editor, we tried to remain protective and to make API as little as possible, so didn't export most of its content. Now that there are clear needs from accessing this class, I think it would make sense to make it more officially public.

What I noticed is that there are mostly two ways:

1. The official one using extension point and having stand alone editor
2. source pages in form editors

[...]

PR #818, opened 2023-10-25
Use the generic editor as the source editor for bnd instructions

laeubi:
Currently a plain TextEditor is used, but the ExtensionBasedTextEditor is much more powerful.

@mickaelistria I was able  to use `ExtensionBasedTextEditor` now and register the completion proposals via extension registry so it seems to work at the basics, can you take a look if it looks sane or I'm doing anything wrong here/not recommended?

Relates to https://github.com/eclipse-pde/eclipse.pde/issues/701

### DIFF
--- a/ui/org.eclipse.pde.ui/plugin.xml
+++ b/ui/org.eclipse.pde.ui/plugin.xml
@@ -2475,4 +2475,11 @@
 			</enablement>
          </commonWizard>
       </extension>
+      <extension
+            point="org.eclipse.ui.genericeditor.contentAssistProcessors">
+         <contentAssistProcessor
+               class="org.eclipse.pde.internal.ui.editor.bnd.BndAutoCompleteProcessor"
+               contentType="org.eclipse.pde.bndInstructions">
+         </contentAssistProcessor>
+      </extension>
 </plugin>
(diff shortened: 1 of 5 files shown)


mickaelistria:
While I didn't spot anything obviously wrong, I recommend you compare it with how the TargetEditor is implemented. It also does use the generic editor in a tab. Not sure which approach is better for this case.

gireeshpunathil:
will this have any implications (positive or negative) on the (pending) auto-completion feature? or is it independent of the editor type?

laeubi:
> will this have any implications (positive or negative) on the (pending) auto-completion feature? or is it independent of the editor type?

Depends on what specific auto-completion feature you refer to, this has two purposes:

1. GenericEditor can be extended even by other (non pde) plugins
2. There is ongoing discussions about having a [BND LanguageServer](https://github.com/bndtools/bnd/issues/5833) so I hope if we get this working to integrate the bnd instructions source page with the language server to [get autocompletion/validation](https://github.com/eclipse-pde/eclipse.pde/issues/701) and so on in an PDE independent way what can be shared across tools (PDE/BNDtools/VSCode/Theia/...)

laeubi:
> I recommend you compare it with how the TargetEditor is implemented. It also does use the generic editor in a tab

Good pointer, I checked this and it only slightly differs a bit as the Target editor has a bit more "control", so for the ManifestEdior things are a bit abstracted why it seems that I need to implement `IFormPage` ...

PR #819, opened 2023-10-25
Show only single editor when open plain BND files

laeubi:
Currently the manifest editor shows an error when one opens a plain bnd file. This is because no "manifest first" parts are found and the bnd form page is only added for pde.bnd files.

This now fixes the issue by making manifest-first files completely optional and show a plain bnd editor instead in case of only a bnd file.


laeubi:
requires 
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1249

to see whats going on ...

PR #821, opened 2023-10-26
Language Server Experiments

laeubi:
Some playground for integration of Language Servers in PDE as in
- https://github.com/bndtools/bnd/issues/5833
- https://github.com/eclipse-pde/eclipse.pde/issues/701

Actually a LS would also be good for any PDE artifacts:

- MANIFEST.MF
- build.properties
- products
- features
- categories
- ...


vogella:
Using language server in PDE would be amazing. Thanks Christopher for working on this.

PR #824, opened 2023-10-26
Support names for Updates repositories in product definition

jcompagner:
2 editors are changed: product file editor and then the repository section but also the category editor which uses the same stuff (code is very simular)

UI part for https://github.com/eclipse-equinox/p2/issues/345

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/RepositoryInfo.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/RepositoryInfo.java
@@ -26,9 +26,11 @@ public class RepositoryInfo extends ProductObject implements IRepositoryInfo {
 	private static final long serialVersionUID = 1L;
 
 	public static final String P_LOCATION = "location"; //$NON-NLS-1$
+	public static final String P_NAME = "name"; //$NON-NLS-1$
 	public static final String P_ENABLED = "enabled"; //$NON-NLS-1$
 
 	private String fURL;
+	private String fName;
 	private boolean fEnabled = true; // enabled unless specified otherwise
 
 	public RepositoryInfo(IProductModel model) {
@@ -50,6 +52,20 @@ public class RepositoryInfo extends ProductObject implements IRepositoryInfo {
 	}
 
 	@Override
+	public String getName() {
+		return fName;
+	}
+
+	@Override
+	public void setName(String name) {
+		String old = fName;
+		fName = name;
+		if (isEditable()) {
+			firePropertyChanged(P_NAME, old, fName);
+		}
+	}
+
+	@Override
 	public boolean getEnabled() {
 		return fEnabled;
 	}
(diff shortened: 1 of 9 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/RepositoryInfo.java
```diff
@@ -78,6 +95,7 @@ public void write(String indent, PrintWriter writer) {
 		if (isURLDefined()) {
 			writer.print(indent + "<repository location=\"" + fURL + "\""); //$NON-NLS-1$ //$NON-NLS-2$
 			writer.print(" enabled=\"" + fEnabled + "\""); //$NON-NLS-1$//$NON-NLS-2$
+			writer.print(" name=\"" + fName + "\""); //$NON-NLS-1$//$NON-NLS-2$
```
HannesWell, line 98: Can you please move the name as second Argument before enabled?
The Name is IMO of higher interest since most of the time enabled is true any ways.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/RepositoryResult.java
```diff
@@ -0,0 +1,4 @@
+package org.eclipse.pde.internal.ui.dialogs;
```
HannesWell, line 1: Please move this into the dialog class having this record standalone is Not necessary.

HannesWell (changes requested):
Thanks @jcompagner for this PR.
I had only time for a quick look yet, but it looks good so far.

But please revert all the changes in the launch config. Those are not related.

HannesWell (approved):
Thanks @jcompagner for this PR.

I did a few cosmetic changes, unified the order of some elements and moved the URL and name trimming to the canonical constructor of the `RepositoryResult` record.

Furthermore I improved the handling of empty names so that in case one leaves the name blank the name element is not written. This is especially important since no existing product has the name element and will therefore first have a null name read when being loaded the next time. Without consideration the name `null` would be set and we would suddenly have many repos named `null`. Long story short, if no name element exists none will be added automatically but one actively has to add a name. In order to remove an existing name, one can just clear the name field.

With that this is ready.

> product file editor and then the repository section but also the category editor which uses the same stuff (code is very simular)

Indeed I'm tempted to unify this code as far as possible but this is something for follow up PRs.
In general I would also like to make the dialog to enter the name and the URL obsolete and let users just enter the name and URL in the table directly.

HannesWell:
I also checked the processing of category.xml files in p2 and it looks like that is already handled, so nothing needs to be done for that part:

https://github.com/eclipse-equinox/p2/blob/e55ba98eadc7e7d9d1e8630ad48187a62546b230/bundles/org.eclipse.equinox.p2.updatesite/src/org/eclipse/equinox/internal/p2/updatesite/CategoryParser.java#L695-L703

PR #827, opened 2023-10-27
Touch bundles affected by the signature change in MultiSourceEditor

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1480

### DIFF
--- a/ds/org.eclipse.pde.ds.ui/forceQualifierUpdate.txt
+++ b/ds/org.eclipse.pde.ds.ui/forceQualifierUpdate.txt
@@ -1,2 +1,3 @@
 # To force a version qualifier update add the bug here
 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184
+https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1480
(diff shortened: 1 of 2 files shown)


szarnekow:
@iloveeclipse if a touch is necessary because of the presence or absence of synthetic bridge methods, I‘d expect a changed Manifest dependency since the code is no longer binary compatible. Or did I misunderstand smth?

iloveeclipse:
You mean, *version bump* on dependency in two bundles? Or version bump of PDE?

szarnekow:
In fact it might require both. The binary method signature in PDE changed and - if the original method signature was called from somewhere - the dependency must be changed there, too

szarnekow:
I prefer Version bumps over NoSuchMethodError at runtime because of bogus version constraints. Why would you like that?

szarnekow:
When you say it’s not required, you imply that’s ok to have mismatched OSGi metadata, because ‘internal’ - is my understanding correct?

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/pull/830.
I've bumped all dependencies of affected bundles to prevent that someone might update bundles without updating dependencies. That should fix the problem for these bundles only.

For the 3rd party bundles I don't care, as the pde bundle itself already bumped minor version segment before.

mickaelistria:
> When you say it’s not required, you imply that’s ok to have mismatched OSGi metadata, because ‘internal’ - is my understanding correct?

Yes. As the internal packages are not meant for public consumption, there OSGi metadata can be unreliable or even incorrect; as long as it works internally on the cases supported by the project.
But again, anyone is welcome to do things better; but we cannot enforce the "proper API metadata" rule for internal code to all committers; that would drastically slow down development.

iloveeclipse:
And many thanks @szarnekow for pinging me, I was in harry this morning to get some useful SDK build so I can get our weekly platform tests running with fresh SDK.

PR #834, opened 2023-10-28
Use ExtensionBasedTextEditor as the base instead of TextEditor

laeubi:
Actually all PDE source pages can benefit from Generic Editor also some functionality relies on the PDESourcePage as a base.

This makes the PDESourcePage extend the ExtensionBasedTextEditor to account for this.

### DIFF
--- a/ds/org.eclipse.pde.ds.ui/META-INF/MANIFEST.MF
+++ b/ds/org.eclipse.pde.ds.ui/META-INF/MANIFEST.MF
@@ -16,7 +16,8 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.30.0,4.0.0)",
  org.eclipse.jface.text;bundle-version="[3.24.200,4.0.0)",
  org.eclipse.jdt.core;bundle-version="[3.36.0,4.0.0)",
  org.eclipse.jdt.ui;bundle-version="[3.31.0,4.0.0)",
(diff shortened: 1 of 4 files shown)


HannesWell:
> Actually all PDE source pages can benefit from Generic Editor also some functionality relies on the PDESourcePage as a base.

Can you please name the concrete benefits?

laeubi:
> > Actually all PDE source pages can benefit from Generic Editor also some functionality relies on the PDESourcePage as a base.
> 
> Can you please name the concrete benefits?

In the "classic" Texteditor everything has to be provided by code, the Generic Editor instead works on extensions what can be provided even by third party plugins:

grafik

One quite complex example are language servers, lsp4e provide several things based on the LanguageServerProtocol (e.g. autocompletions) therefore one can plugin LSPs into arbitrary editors (if the use the Generic Editor Framework), PDE already do this with `org.eclipse.pde.genericeditor.extension` for the target editor.

laeubi:
Another example is that we now can replace `org.eclipse.pde.internal.ui.editor.build.BuildSourceViewerConfiguration` by extension points to generic editor.

mickaelistria:
IIRC, by default, the Generic Editor brings some features that are not in the default text editor, such as matching bracket highlight/jump; some default completion based on words already existing in the docs or other basic things like that which comes for free. But as @laeubi mentions, the benefit is that is becomes much much easiler to enrich the editors with simple extensions now.

PR #850, opened 2023-11-01
Enable Javadoc generation in build

laeubi:
Currently javadoc is not generated as part of the build making javadoc errors hard to discover.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,9 @@ pipeline {
 				wrap([$class: 'Xvnc', useXauthority: true]) {
 					sh '''
 						mvn clean verify --batch-mode -Dmaven.repo.local=$WORKSPACE/.m2/repository \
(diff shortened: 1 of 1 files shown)


akurtakov:
Looks good to me. I'm pushing this one. I am still torn between generating all javadoc and skipping *.internal packages. The former is the right thing to do the later would allow us to more easily see improvements and have it actually fail the build (should be possible?) so the issue is actually seen in verification and not wait for I-build.

akurtakov (commented):
Looks good to me.

laeubi:
@akurtakov the javadoc plugin has an option to skip packages by wildcards:

https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#excludePackageNames

not sure if this is exactly what you have in mind.

akurtakov:
I know about it. I can't still make my mind which one is better yet.

PR #853, opened 2023-11-01
Schema provider is optional in SchemaInclude

laeubi:
It is allowed to not specify any schema provider in that case no additional search is performed.

Fix https://github.com/eclipse-platform/.github/issues/167


laeubi:
Testfailure are nown (see for example https://github.com/eclipse-pde/eclipse.pde/issues/554)

kthoms:
Thanks @laeubi , this change looks good.

PR #854, opened 2023-11-01
Usage of Plug-ins instead of Plugins in the product content section

vogella:
"Plug-ins" is the term which is also used if you switch your product to be plug-in based.


HannesWell (approved):
Thanks

PR #855, opened 2023-11-01
Using List.of in e3 view template

vogella:
Minimal change, just using the List.Of operator which I think looks nicer for the code example.


ui/org.eclipse.pde.ui.templates/templates_3.1/view/java/$className$.java
```diff
@@ -217,7 +217,7 @@ public void createPartControl(Composite parent) {
 		
 %if viewType == "tableViewer"
 		viewer.setContentProvider(ArrayContentProvider.getInstance());
-		viewer.setInput(new String[] { "One", "Two", "Three" });
+		viewer.setInput(List.of("One", "Two", "Three"));
```
szarnekow, line 220: Why is this an improvement?
At runtime this will cause an extra copy of the list into an array.

vogella:
Ok, lets abandon

PR #858, opened 2023-11-01
Shorter text in the target editor for location description

vogella:
Instead of "The following list of locations will be used to collect plug-ins for this target definition." we use "The following locations are used to collect plug-ins for this target definition." which is shorter but has the same information.


laeubi:
> The following list of locations will be used to collect plug-ins for this target definition.
> The following locations are used to collect plug-ins for this target definition.

> which is shorter but has the same information

But actually is wrong, features are also collected ;-)
So maybe you want to use the saved character to mention features as well.

vogella:
> > The following list of locations will be used to collect plug-ins for this target definition.
> > The following locations are used to collect plug-ins for this target definition.
> 
> > which is shorter but has the same information
> 
> But actually is wrong, features are also collected ;-) So maybe you want to use the saved character to mention features as well.

WDYT about: "Locations used to collect plug-ins and features for this target definition" basically removing "The following ... are used" which I think it not value add.

laeubi:
Yes something like that, I wonder if the text is helpful anyways for users, maybe one can remove it completely and having a better headline instead of "Locations" if that alone is not enough. Will give more space for what is really important, these form layouts are really taking already much space.

vogella:
> Yes something like that, I wonder if the text is helpful anyways for users, maybe one can remove it completely and having a better headline instead of "Locations" if that alone is not enough. Will give more space for what is really important, these form layouts are really taking already much space.

I agree to the removal but as this most likely requires fiddling with the layout, lets leave this for another commit. In general I would like to remove form layouts, they feel outdated to me.

PR #870, opened 2023-11-03
Running all JDT performance clean-ups on pde.ui

vogella:
Selected all performance JDT clean-ups and ran that on pde.ui.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/contentassist/TypePackageCompletionProcessor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/contentassist/TypePackageCompletionProcessor.java
@@ -134,11 +134,11 @@ public abstract class TypePackageCompletionProcessor implements IContentAssistPr
 						pName = completion.substring(0, period);
 					}
 					Image image = isInterface ? PDEPluginImages.get(PDEPluginImages.OBJ_DESC_GENERATE_INTERFACE) : PDEPluginImages.get(PDEPluginImages.OBJ_DESC_GENERATE_CLASS);
-					String label = cName;
+					StringBuilder label = new StringBuilder().append(cName);
 					if (pName != null)
-						label = label + " - " + pName; //$NON-NLS-1$
+						label.append(" - ").append(pName); //$NON-NLS-1$
 					String typeName = String.valueOf(Signature.toCharArray(Signature.getTypeErasure(proposal.getSignature())));
-					addProposalToCollection(c, startOffset, length, label, completion, image, project, typeName);
+					addProposalToCollection(c, startOffset, length, label.toString(), completion, image, project, typeName);
 				}
 			}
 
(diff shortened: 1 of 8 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/PDEFormEditor.java
```diff
@@ -480,11 +480,11 @@ public void doSaveAs() {
 			// Ignore
 		} catch (Exception e) {
 			String title = PDEUIMessages.PDEFormEditor_errorTitleProblemSaveAs;
-			String message = PDEUIMessages.PDEFormEditor_errorMessageSaveNotCompleted;
+			StringBuilder message = new StringBuilder().append(PDEUIMessages.PDEFormEditor_errorMessageSaveNotCompleted);
```
merks, line 483: Odd, I would have thought that this would be more performant (and would generated less byte code) with one less method call:
```
new StringBuilder(PDEUIMessages.PDEFormEditor_errorMessageSaveNotCompleted);
```

vogella: JDT clean-ups do not generated perfect code, could be enhanced with a follow-up commit. Mixing automatic clean-ups with manual adjustments should IMHO avoided

merks: It just seems ironic that a performance cleanup generates glaringly suboptimal code, both in terms of performance and verbosity.  I suppose it's because there are more append overloads than constructor overloads, so JDT would need to be smarter and more careful when generating a constructor call...

laeubi: @vogella you can just push another commit to your branch as well for the manual work (or revert it + bugreport to JDT).

i also don't see why this is more "performant" (regardless of if one needs to optimize performance for the error case). I think if you really care about performance the code can be rewritten in one constant logging and one with exception like this:

```
if (e.getMessage() == null) {
    PDEPlugin.logException(e, title, PDEUIMessages.PDEFormEditor_errorTitleProblemSaveAs);
} else {
    PDEPlugin.logException(e, title, PDEUIMessages.PDEFormEditor_errorTitleProblemSaveAs+ '  ' + e.getMessage());
```
}

laeubi: > Odd, I would have thought that this would be more performant (and would generated less byte code) with one less method call:

Even that is wasteful as for the case of message= null it created a `StringBuilder` from a constant to convert it into a string right after that...

merks: Indeed, my comment was just from looking at the one line in isolation of all context...

vogella:
I reverted the change in PDEFormEditor

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/contentassist/TypePackageCompletionProcessor.java
```diff
@@ -134,11 +134,11 @@ public void accept(CompletionProposal proposal) {
 						pName = completion.substring(0, period);
 					}
 					Image image = isInterface ? PDEPluginImages.get(PDEPluginImages.OBJ_DESC_GENERATE_INTERFACE) : PDEPluginImages.get(PDEPluginImages.OBJ_DESC_GENERATE_CLASS);
-					String label = cName;
+					StringBuilder label = new StringBuilder().append(cName);
```
szarnekow, line 137: I don't think this is an improvement in terms of performance or readability.

https://openjdk.org/jeps/280 as well as the optimized code that ECJ emits for these trivial cases (string + string + string)  outweigh the whatever was intended here.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/contentassist/TypePackageCompletionProcessor.java
```diff
@@ -214,14 +214,14 @@ public void acceptType(int modifiers, char[] packageName, char[] simpleTypeName,
 					// Accept search results from the JDT SearchEngine
 					String cName = new String(simpleTypeName);
 					String pName = new String(packageName);
-					String label = cName;
+					StringBuilder label = new StringBuilder().append(cName);
 					String replaceString = cName;
```
szarnekow, line 218: Another curious case. Why was `lable` StringBuilderified but replaceString wasn't?

vogella: I suggest to open a JDT UI bug for this cleanup.

vogella:
Agree with @szarnekow not a real improvement, closing

PR #871, opened 2023-11-03
Removes progress reporting from product configuration file wizard

vogella:
Reporting progress for the creation of the file with only one step reported after the file is created by before it it opened, seems unless.


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/product/BaseProductCreationOperation.java
```diff
@@ -67,11 +66,8 @@ public BaseProductCreationOperation(IFile file) {
 
 	@Override
 	protected void execute(IProgressMonitor monitor) throws CoreException, InvocationTargetException, InterruptedException {
-		monitor.beginTask(PDEUIMessages.BaseProductCreationOperation_taskName, 2);
 		createContent();
```
laeubi, line 69: I think the monitor should actually be splitted and passed down to createcontent and open files ...

laeubi:
@vogella can you please apply the suggestion and update the PR?

Issue #875, opened 2023-11-03
Reuse existing bundle pools


laeubi:
Currently PDE maintains its own bundle pool in the workspace, this has many drawbacks:

1. Multiple workspace download the same things twice
2. Oomph has its own bundle pool and download things as well
3. Tycho also

PDE should have some mean to access existing well known pools of P2 items to leverage already downloaded content. As we have hashes + signatures if in doubt one could even verify that data from such pools is "true"...

merks:
Note that I've verified manually, by instrumenting org.eclipse.equinox.internal.p2.artifact.repository.ArtifactRequest.setSourceRepository(IArtifactRepository),  that this very simple change will ensure that any artifact already available in the pool will be fetched in favor of the remote artifact.  So I see what comes from the pool:
```
file:///D:/Users/test32/.p2/pool/->osgi.bundle,bcpg,1.76.0
```
versus what comes from the remote site:
```
https://download.eclipse.org/eclipse/updates/4.29/R-4.29-202309031000->osgi.bundle,bcpg.source,1.76.0
```

HannesWell:
> Note that I've verified manually, by instrumenting org.eclipse.equinox.internal.p2.artifact.repository.ArtifactRequest.setSourceRepository(IArtifactRepository), that this very simple change will ensure that any artifact already available in the pool will be fetched in favor of the remote artifact. So I see what comes from the pool:

You are referring to https://github.com/eclipse-pde/eclipse.pde/pull/878, don't you?
So this does not add new artifacts to the shared bundle pool if they are not yet present and only required by PDE?
For that the profile of the target probably would have to be registered in the shared bundle-pool, wouldn't it?

merks:
Indeed, the shared bundle pool is managed by a shared agent that manages all the profiles, i.e., of all the installer-installed installation profiles and all targlet profiles.  Without such proper management, a pool would just grow indefinitely and could never be garbage collected.  The shared agent itself is implemented by and managed by a large stack of technology where the bundle pool is just the tip of a vast iceberg.

HannesWell:
> The shared agent itself is implemented by and managed by a large stack of technology where the bundle pool is just the tip of a vast iceberg.

Is that technology-stack part of P2 or Oomph?
If it is not yet part of P2, it might be worth to migrate it there as part of https://github.com/eclipse-equinox/p2/issues/373?

merks:
Here is what we are talking about "migrating":

[...]

PR #878, opened 2023-11-04
Add Oomph's bundle pools to PDE's artifact repositories for resolution

merks:
https://github.com/eclipse-pde/eclipse.pde/issues/875

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
@@ -18,6 +18,7 @@ package org.eclipse.pde.internal.core.target;
 
 import java.io.File;
 import java.net.URI;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -45,6 +46,7 @@ import org.eclipse.core.runtime.URIUtil;
 import org.eclipse.core.runtime.preferences.IPreferencesService;
 import org.eclipse.equinox.internal.p2.director.PermissiveSlicer;
 import org.eclipse.equinox.internal.p2.garbagecollector.GarbageCollector;
+import org.eclipse.equinox.internal.p2.repository.helpers.RepositoryHelper;
 import org.eclipse.equinox.p2.core.IAgentLocation;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.core.IProvisioningAgentProvider;
@@ -1395,7 +1397,7 @@ public class P2TargetUtils {
 	 * @exception CoreException
 	 */
 	private Collection<URI> getArtifactRepositories(ITargetDefinition target) throws CoreException {
-		Set<URI> result = new HashSet<>();
+		Set<URI> result = new LinkedHashSet<>();
 		ITargetLocation[] containers = target.getTargetLocations();
 		if (containers == null) {
 			containers = new ITargetLocation[0];
@@ -1418,7 +1420,16 @@ public class P2TargetUtils {
 			// get all the artifact repos we know in the manager currently
 			result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_ALL)));
 
-			// Add in the IDE profile bundle pool and all known workspaces
+			if (Boolean.parseBoolean(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$
+				try {
+					result.addAll(RepositoryHelper.getSharedBundlePools().stream().map(Path::toUri).toList());
+				} catch (Exception e) {
+					//$FALL-THROUGH$
+				}
+			}
+
+			// Add in the IDE profile bufindProfileReposndle pool and all known
+			// workspaces
 			findProfileRepos(result);
 			findWorkspaceRepos(result);
 		}


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1418,13 +1422,36 @@ private Collection<URI> getArtifactRepositories(ITargetDefinition target) throws
...
+				for (String poolLocation : Files.readAllLines(bundlePools, Charset.defaultCharset())) {
+					Path poolLocationPath = Path.of(poolLocation);
+					if (Files.isDirectory(poolLocationPath)) {
+						result.add(poolLocationPath.toUri());
+					}
+				}
+			}
+		}
+		return result;
+	}
```
HannesWell, line 1453: Using streams this could be a little bit more compact.

Furthermore I wonder if it would be good to always use UTF-8 to write that file?

```suggestion
	private Collection<URI> getSharedBundlePools() throws IOException {
		if ("true".equals(System.getProperty("pde.usePoolsInfo", "true"))) {
			Path bundlePools = Path.of(System.getProperty("user.home"), ".p2/pools.info");
			if (Files.isRegularFile(bundlePools)) {
				return Files.lines(bundlePools, Charset.defaultCharset()) //
						.map(Path::of).filter(Files::isDirectory) //
						.map(Path::toUri).toList();
			}
		}
		return List.of();
	}
```

HannesWell (commented):
Very nice!
I thought about that already multiple times too and assumed that this couldn't be that hard but I'm surprised that it is that easy (you just have to know how 😄)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
HannesWell: I forgot the try-with-resources. Updated the suggestion.

merks: I wish we did always use UTF-8 to write that file. 😱   

But I looked at the implementation and we use new FileReader(file).  😭 

It's rather tragic...

HannesWell: > It's rather tragic...

Indeed. Especially that with [JEP 400: UTF-8 by Default](https://openjdk.org/jeps/400) delivered in Java 18 UTF-8 will become the default default.
So maybe it would be good to have a migration path anyways (maybe write the encoding as comment in the first line and re-read again if it has not written accordingly).

merks: Indeed. I’m concerned how older versions can handle newly written changes. What a nightmare on Windows.

HannesWell: > Indeed. I’m concerned how older versions can handle newly written changes. What a nightmare on Windows.

Maybe this helps:
- https://softwareengineering.stackexchange.com/a/187174
- https://learn.microsoft.com/en-us/windows/win32/intl/using-byte-order-marks?redirectedfrom=MSDN

Or the poor man's variant of having a initial comment indicating the encoding. The absence could be treated as Windows default (IIRC `ISO_8859_1`).

merks: If I change the serialization form in any way not recognized by older versions of the libraries, I think I'm going to get bitten.   Here is a thought...

I write a file that's effectively created and initialized one time.  It contains the encoding used for all the other files.  It's written in UTF-8.  Older runtimes will just continue to use the default encoding as before, not looking at this file.  Newer runtimes will create the file if it doesn't exist and subsequently will use the encoding in that file from then on.  When eventually Java 21 changes the default to UTF-8 the newer runtimes will continue to use the recorded encoding that was used for the existing files.  In the future, brand-new shared agent would just record and use UTF-8 everywhere.

I've prototyped that in Oomph and the code here for PDE could be made safely compatible with that approach as follows:

image

What do you think?

merks: I did some extensive testing with a user home folder `D:\Users\tëst33\.p2\pool` with special characters.   The encoding used by default right now is `windows-1252`.  I tested that if `pools.info` is read with UTF-8 that it fails to determine the pool locations correctly.   With the extended logic, even if the system default encoding is UTF-8, it reads the encoding from `encoding.info` and then reads the `pools.info` correctly...

merks: I have a commit ready for Oomph to create the `encoding.info` and to use it as well to future proof against changes to the default encoding.  Of course the code here for PDE works with or without the `encoding.info`, as long as PDE is running with the same default encoding as is currently the case with Java 17 (and of course if it isn't it will just quietly not use the shared bundle pool).

laeubi: @merks I think you can quite easily hand this, simply read the file as byte array, then decode it at "windows-1252" and "utf-8", the one where you find a valid file is the one with correct encoding (in some case simply both will be the same)...

merks: The default encoding on Windows in Japan is Shift-JIS I think.  In China I think it's MS936.  I really don't know all the defaults in various countries.  I'd really like to avoid assumptions.  Furthermore, there are a number of files being read **and written** by this infrastructure in Oomph so I'd like to ensure they all continue to work as they do today.   So ideal to my thinking is to know the encoding and use exactly that correct encoding when reading and when writing so that no existing installations will suddenly just stop work...

laeubi: Good point, whatever solution is used, I think it would be good to add that reading support to P2 as suggested here:
- https://github.com/eclipse-equinox/p2/issues/373

and just consume this then from Oomph, PDE (and after the release from Tycho) as well, also it would be good to write the encoding file if it is not present yet so it is ensured the current encoding is recorded for status-quo.

merks: Given it just returns a list of URIs, it could be added almost anywhere.  Any specific suggestion in that regard?

laeubi: If we want to keep it internal I think `org.eclipse.equinox.internal.p2.repository.helpers.RepositoryHelper` would be suitable, otherwise I would just create a new class `P2Pool` in package `org.eclipse.equinox.p2.repository.artifact`.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1418,7 +1420,16 @@ private Collection<URI> getArtifactRepositories(ITargetDefinition target) throws
 			// get all the artifact repos we know in the manager currently
 			result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_ALL)));
 
-			// Add in the IDE profile bundle pool and all known workspaces
+			if ("true".equals(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
HannesWell, line 1423:
```suggestion
			if (Boolean.parse(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$
```
Would be a bit more flexible since `TRUE` would also be ok and is IMHO a bit more readable.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
merks: The latest changes have moved the logic to p2 and reuse it in PDE.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
merks: Done.

HannesWell (approved):
Looks good. Thank you.

For clarification: with this change PDE only re-uses artifacts from the shared bundle-pool in the user.home if they are present, but does not add missing ones, doesn't it?

So I would say this is a good step towards https://github.com/eclipse-pde/eclipse.pde/issues/875, but not a full resolution?

merks:
I'm not sure what you mean by "missing ones".   Oomph only has a single shared agent manager that is always located in ~/.p2.   There is a fair bit more flexibility to the agent manager than that,  but I expect pretty much no one in the world is using multiple agents.  So I'm not sure what important thing you think might be missing.

HannesWell:
> I'm not sure what you mean by "missing ones".

Artifacts not yet in the pool, that are included in a target and would need to be added to the pool.

merks:
Yes, it only allows p2 to read from the artifact/bundle pool, exactly like it does for all the other artifact repository URIs that are added by the existing artifact repository gathering logic. 

PDE would need some major design changes and restructuring before it could be reasonably trusted (by me) to write into and reuse the shared bundle pool.  For me, if that pool is corrupted, it would destroy more than 100 installations/workspaces.

HannesWell:
> PDE would need some major design changes and restructuring before it could be reasonably trusted (by me) to write into and reuse the shared bundle pool. For me, if that pool is corrupted, it would destroy more than 100 installations/workspaces.

Yes, I just wanted to make it clear (for myself and others) what this change is exactly doing. :)

merks:
> > PDE would need some major design changes and restructuring before it could be reasonably trusted (by me) to write into and reuse the shared bundle pool. For me, if that pool is corrupted, it would destroy more than 100 installations/workspaces.
> 
> Yes, I just wanted to make it clear (for myself and others) what this change is exactly doing. :)

It does only exactly what the title says.  😁

Issue #881, opened 2023-11-05
Avoid adding all known artifact repositories during target resolution


merks:
The following seems highly questionable to me:

https://github.com/eclipse-pde/eclipse.pde/blob/3f9410787872dabfff01e5108238caf66f395da8/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java#L1417-L1419

When resolving a target platform, one generally expects that primarily only the repositories mentioned in the target platform resource will be used.   But what the above line of code does is add all enabled repositories listed in Window -> Preferences -> Install/Update -> Available Software Sites.  That just seems very wrong.  

I think the intent here is likely that IRepositoryManager.REPOSITORIES_LOCAL should be used.  I.e., it's a performance optimization to provide local artifact repositories that might contain the artifacts and from which one could transfer those artifacts much more quickly than from a remote site.  Adding more remote sites can and often does hurt performance...

laeubi:
> I think the intent here is likely that IRepositoryManager.REPOSITORIES_LOCAL should be used. I.e., it's a performance optimization to provide local artifact repositories that might contain the artifacts and from which one could transfer those artifacts much more quickly than from a remote site

Sounds reasonable. I think the name useAdditional**Local**Artifacts() also implies this, I would vote for:

1. remove `useAdditionalLocalArtifacts()` as it always returns true and obviously never was implementede/enhanced
2. use unconditional `result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_LOCAL)));`

Especially as Tycho does not use something similar and we never got any complaint about something working in PDE but not for Tycho in this regards...

@HannesWell @vik-chand any concerns?

HannesWell:
> When resolving a target platform, one generally expects that primarily only the repositories mentioned in the target platform resource will be used. But what the above line of code does is add all enabled repositories listed in Window -> Preferences -> Install/Update -> Available Software Sites. That just seems very wrong.

Absolutely. At least this is only used to get the artifacts and not also the metadata, which would mean that always the latest repo available as update-site is used.

> * remove `useAdditionalLocalArtifacts()` as it always returns true and obviously never was implementede/enhanced
> 
> * use unconditional `result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_LOCAL)));`

Full support from my side.

[...]

PR #883, opened 2023-11-05
Add local artifact repositories for resolution

merks:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/881

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
@@ -1416,33 +1416,23 @@ public class P2TargetUtils {
 				result.addAll(getArtifactRepositories(referencedTargetDefinition));
 			}
 		}
-		if (useAdditionalLocalArtifacts()) {
-			// get all the artifact repos we know in the manager currently
-			result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_ALL)));
 
-			if (Boolean.parseBoolean(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$
-				try {
-					result.addAll(RepositoryHelper.getSharedBundlePools().stream().map(Path::toUri).toList());
-				} catch (Exception e) {
-					//$FALL-THROUGH$
-				}
+		// get all the local artifact repos we know in the manager currently
+		result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_LOCAL)));
+
+		if (Boolean.parseBoolean(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$
+			try {
+				result.addAll(RepositoryHelper.getSharedBundlePools().stream().map(Path::toUri).toList());
+			} catch (Exception e) {
+				//$FALL-THROUGH$
 			}
-
-			// Add in the IDE profile bufindProfileReposndle pool and all known
-			// workspaces
-			findProfileRepos(result);
-			findWorkspaceRepos(result);
 		}
-		return result;
-	}
 
-	/**
-	 * return whether or not to use local artifact repositories when provisioning the target
-	 */
-	private boolean useAdditionalLocalArtifacts() {
-		// XXX consider using a preference here or another strategy if users are able to spec
-		// what local repos are to be considered.
-		return true;
+		// Add in the IDE profile bundle pool and all known workspaces
+		findProfileRepos(result);
+		findWorkspaceRepos(result);
+
+		return result;
 	}
 
 	/**


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1416,33 +1416,24 @@ private Collection<URI> getArtifactRepositories(ITargetDefinition target) throws
...
-	}
 
-	/**
-	 * return whether or not to use local artifact repositories when provisioning the target
-	 */
-	private boolean useAdditionalLocalArtifacts() {
-		// XXX consider using a preference here or another strategy if users are able to spec
-		// what local repos are to be considered.
-		return true;
+		// Add in the IDE profile bufindProfileReposndle pool and all known
```
szarnekow, line 1431: `bufindProfileReposndle` ?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1416,33 +1416,24 @@ private Collection<URI> getArtifactRepositories(ITargetDefinition target) throws
...
+		// get all the local artifact repos we know in the manager currently
+		result.addAll(Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_LOCAL)));
 
-			// Add in the IDE profile bufindProfileReposndle pool and all known
-			// workspaces
-			findProfileRepos(result);
-			findWorkspaceRepos(result);
+		if (Boolean.parseBoolean(System.getProperty("pde.usePoolsInfo", "true"))) { //$NON-NLS-1$ //$NON-NLS-2$
+			try {
+				result.addAll(RepositoryHelper.getSharedBundlePools().stream().map(Path::toUri).toList());
```
szarnekow, line 1425: Do we want to fail to add any shared bundle pool if a single Path.toUri throws an exception? Or should the try-catch be on a basis of a single pool?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
merks: That crap was there before. 😱 

https://github.com/eclipse-pde/eclipse.pde/blob/e7d05ff7dc03989b3da04b075feb3962444e0612/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java#L1431

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
merks: It's so much prettier with streams.   😱 

Realistically 99.99% of the people (likely an huge under-estimate) have only one pool, and it is 0% likely that any of the paths cannot be converted to URIs because all hell would have broken loose long before p2/PDE ever reads them.  

So technically yes, but realistically, not really.

HannesWell: I would say that actually it _should_ never fail, otherwise the file is corrupted (e.g. wrong encoding) and it should be fixed.
So I would say it is ok to fail for all, but the catched exception should be logged instead so that a users notices the corruption: `PDECore.log(Status.error("Failed to load shared bundle-pool locations",e));` or similar

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
HannesWell: > That crap was there before. 😱

You added it in your previous commit 🙈
Before that it was: `// Add in the IDE profile bundle pool and all known workspaces`

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
merks: @laeubi was suggesting that maybe getSharedBundlePools should handle/log all the exceptions

https://github.com/eclipse-equinox/p2/pull/377#discussion_r1382541460

I tend to agree with him.  What about you?

In any case, this commit was about using `IRepositoryManager.REPOSITORIES_LOCAL` and removing  the condition.  😱

HannesWell: > I tend to agree with him. What about you?

If all exceptions are handled there, I'm fine with that as well. But that should include the conversion to URIs.
If Oomph also needs a URI, it would make sense to handle all exceptions there, if there a different demands it is better handled by the callers (since they need some handling any ways).

> In any case, this commit was about using IRepositoryManager.REPOSITORIES_LOCAL and removing the condition. 😱

Yes that's right. Nevertheless it would be good to find some kind of agreement on this quickly to reduce the trouble of multiple versions. 😅

merks: I have this change pending:

https://github.com/eclipse-equinox/p2/pull/379

As such the only possible exception here is from the toURI call, but note that the called method has done an isDirectory test so I don't think we need to worry about to URI failing for an existing directory and needing to log it...

merks: And no, I don't want to convert to URIs in that method because I think it's important that these things are locations in the file system which is no longer clear/obvious when they are returned as URIs.

I think at some point here I will need to draw the line and call it done.  Folks can continue do whatever they like.  One may wish to revisit a number of things including things like this:  🙈 

https://github.com/eclipse-pde/eclipse.pde/blob/e7d05ff7dc03989b3da04b075feb3962444e0612/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java#L1502-L1504

The chance that toURI will fail for a Path for which isDirectory is true, where the value is the bundle pool used for all installations is pretty much zero, so talking about the value of logging that information is probably wasted breath.  😬

HannesWell: > As such the only possible exception here is from the toURI call, but note that the called method has done an isDirectory test so I don't think we need to worry about to URI failing for an existing directory and needing to log it...

Makes sense.

Looking at the Java-doc of `Path.toUri()` I see only:
```
     * @throws  java.io.IOError
     *          if an I/O error occurs obtaining the absolute path, or where a
     *          file system is constructed to access the contents of a file as
     *          a file system, and the URI of the enclosing file system cannot be
     *          obtained
     *
     * @throws  SecurityException
     *          In the case of the default provider, and a security manager
     *          is installed, the {@link #toAbsolutePath toAbsolutePath} method
     *          throws a security exception.
```
Both are more fundamental problems that will probably apply to all paths and I don't expect any other than the default file-systems to be used.

HannesWell (approved):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
laeubi: >  I think it's important that these things are locations in the file system which is no longer clear/obvious when they are returned as URIs

A Path can be a remote filesystem or even a virtual one... actually `Path` = `URI` + `Browsing Capability` + (optional) `Watcher` ;-)

merks: That's true.   Even an archive has Paths to access it.

laeubi (approved):

Issue #884, opened 2023-11-06
New PDE test fails since I20231105-1800


iloveeclipse:
We have new 8 P2Tests failures since I20231105-1800 build on all platforms.

* https://download.eclipse.org/eclipse/downloads/drops4/I20231105-1800/testResults.php
* https://download.eclipse.org/eclipse/downloads/drops4/I20231105-0400/testresults/html/org.eclipse.pde.build.tests_ep430I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

```
[...]

Issue #801, opened 2023-10-14
[apitools] Use the AST/JDT to discover package names


laeubi:
As we have seen here:

- https://github.com/eclipse-pde/eclipse.pde/issues/782#issuecomment-1762550912
- https://github.com/eclipse-pde/eclipse.pde/issues/791

the `org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer` currently uses the folders of an output location to determine the available packages.

While this might be a good strategy for a binary artifact, it is actually not when we are analyze the sources especially as in `org.eclipse.pde.api.tools.internal.model.ProjectComponent.getApiTypeContainer(String, ProjectComponent, Map<IPath, IApiTypeContainer>)` there already is a reference to a `IPackageFragmentRoot` so better reuse what JDT offer than trying to be "smart" and reimplement the thing.

FYI @iloveeclipse @merks @vik-chand

PR #886, opened 2023-11-06
Assume bundle shape as jar by default

iloveeclipse:
- set PDE generator to assume "unpack" attribute from bundles to be "false" by default.
- additionally adopted testBug263272() to ignore "unpack=false" instructions from the feature.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/884

### DIFF
--- a/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
+++ b/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests Plug-in
 Bundle-SymbolicName: org.eclipse.pde.build.tests;singleton:=true
-Bundle-Version: 1.4.100.qualifier
+Bundle-Version: 1.4.200.qualifier
 Bundle-Activator: org.eclipse.pde.build.tests.Activator
 Export-Package: org.eclipse.pde.build.internal.tests;x-internal:=true,
  org.eclipse.pde.build.internal.tests.ant;x-internal:=true,
--- a/build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/p2/P2Tests.java
+++ b/build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/p2/P2Tests.java
@@ -629,8 +629,8 @@ public class P2Tests extends P2TestCase {
 		IFolder b = Utils.createFolder(buildFolder, "plugins/b");
 		IFolder c = Utils.createFolder(buildFolder, "plugins/c");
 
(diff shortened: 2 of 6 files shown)


iloveeclipse:
Is that really caused by my changes???

@laeubi : could be some tycho issue? Too old tycho? Too new?

```
15:15:07  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.4-SNAPSHOT:verify (verify) on project org.eclipse.pde.build: There are API errors:
15:15:07  [ERROR] src/org/eclipse/pde/build/Constants.java:21 The type org.eclipse.pde.build.Constants has been removed from org.eclipse.pde.build_3.12.200
15:15:07  [ERROR] src/org/eclipse/pde/build/IAntScript.java:27 The type org.eclipse.pde.build.IAntScript has been removed from org.eclipse.pde.build_3.12.200
15:15:07  [ERROR] src/org/eclipse/pde/build/IFetchFactory.java:50 The type org.eclipse.pde.build.IFetchFactory has been removed from org.eclipse.pde.build_3.12.200
15:15:07  [ERROR] META-INF/MANIFEST.MF:5 The major version should be incremented in version 3.12.200, since API breakage occurred since version 3.12.100
```

iloveeclipse:
OK, the problem is unrelated to my changes: https://github.com/eclipse-pde/eclipse.pde/pull/887

iloveeclipse:
Btw, the code **was** changed already in 4.30 (that's the reason manifest is not changed), so this is really a new issue...

iloveeclipse:
@laeubi: I wonder if API tools would also fail in the official build - and if yes, will that mean, we would not able to build SDK anymore with this PR?

I fear the change in PDE is needed/blocker for 4.30 M3, otherwise the PDE will build some garbage...

laeubi:
As far as I know (but @akurtakov or @sravanlakkimsetti might can explain better), that API errors will not fail the build bud but mark it as unstable as there is another api checker used...

Beside that I think only tests are affected as pdebuild is really not used anywhere anymore so apart from these test-failures everything should be fine, I'll try to fix that as soon as possible and then maybe we can have this included as bugfix to M3...

laeubi:
What seem to trigger this is that actually one package from that jar is exported by the `MANIFEST.MF` (`org.eclipse.pde.internal.build.tasks`) and because of
- https://github.com/eclipse-pde/eclipse.pde/issues/801

then API tools is confused because it can't discover the packages from `bin_ant` because this currently points to the wrong location.

merks:
You guys are really impressive.

HannesWell (approved):
Thank you Andrey for your Investigation and providing the fix.

> - set PDE generator to assume "unpack" attribute from bundles to be "false" by default.

That is in general a more realistic/better assumption, p2 uses the same default.

Please consider this a project-lead approval to submit this after the M3-deadline as soon as the build ist green.
And thanks to @laeubi for fixing Tycho.

PR #887, opened 2023-11-06
DUMMY PR

iloveeclipse:
Check API fails in https://github.com/eclipse-pde/eclipse.pde/pull/886


HannesWell:
Since https://github.com/eclipse-pde/eclipse.pde/pull/886 is submitted, I think we can close this.

Issue #888, opened 2023-11-06
PDE Jenkins build broken, API tools report errors


iloveeclipse:
See 
* "empty" PR with whitespace changes only in Java code: https://github.com/eclipse-pde/eclipse.pde/pull/887
* https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-887/1/

```
15:44:12  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.4-SNAPSHOT:verify (verify) on project org.eclipse.pde.build: There are API errors:
15:44:12  [ERROR] src/org/eclipse/pde/build/Constants.java:21 The type org.eclipse.pde.build.Constants has been removed from org.eclipse.pde.build_3.12.200
15:44:12  [ERROR] src/org/eclipse/pde/build/IAntScript.java:27 The type org.eclipse.pde.build.IAntScript has been removed from org.eclipse.pde.build_3.12.200
15:44:12  [ERROR] src/org/eclipse/pde/build/IFetchFactory.java:50 The type org.eclipse.pde.build.IFetchFactory has been removed from org.eclipse.pde.build_3.12.200
15:44:12  [ERROR] META-INF/MANIFEST.MF:5 The major version should be incremented in version 3.12.200, since API breakage occurred since version 3.12.100
15:44:12  [ERROR] -> [Help 1]
```

iloveeclipse:
@laeubi : do you see anything useful in https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-887/2/console ?

This reminds me on https://github.com/eclipse-pde/eclipse.pde/issues/782.
Is this Tycho fix may be not working for PDE for some reason?

laeubi:
The debug output looks normal:
```
[2023-11-06T16:15:46.825Z] [INFO] Resolve API tools runtime from MavenRepositoryLocation [location=https://download.eclipse.org/eclipse/updates/4.30-I-builds/]...
[2023-11-06T16:15:50.158Z] API Tools version: 1.3.200.v20231102-1258
[2023-11-06T16:15:50.158Z] Job Load Target Platform(24) scheduled...
[2023-11-06T16:15:50.158Z] Job Load Target Platform(24) aboutToRun...
[2023-11-06T16:15:50.158Z] Job Load Target Platform(24) running...
[2023-11-06T16:15:50.158Z] Job Updating workspace(6) scheduled...
[...]

PR #889, opened 2023-11-06
Enable tycho.apitools.debug & disable DetectVMInstallationsJob

iloveeclipse:
This doesn't fix the problem, but may be gives some hints

See https://github.com/eclipse-pde/eclipse.pde/issues/888

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,6 +21,8 @@ pipeline {
 						-Pbree-libs \
 						-Papi-check \
 						-Pjavadoc \
+						-DDetectVMInstallationsJob.disabled=true \
+						-Dtycho.apitools.debug \
 						-Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
 					'''
 				}


laeubi:
You need to add a change like in https://github.com/eclipse-pde/eclipse.pde/pull/887 or otherwise API check will be skipped.

iloveeclipse:
> You need to add a change like in #887 or otherwise API check will be skipped.

I would like to merge it "as is" to master. Or does it break anything?

laeubi:
> > You need to add a change like in #887 or otherwise API check will be skipped.
> 
> I would like to merge it "as is" to master. Or does it break anything?

It will just print more things to the log.

iloveeclipse:
> It will just print more things to the log.

OK, so fine to merge then, that is exact the purpose of this PR...

laeubi:
> OK, so fine to merge then, that is exact the purpose of this PR...

Nevertheless if we do not want **permanent** debug logging it might be better to include such temporal things in the affected PR...

iloveeclipse:
> Nevertheless if we do not want **permanent** debug logging it might be better to include such temporal things in the affected PR...

Once we have solution for #888 we can revert this commit.

Issue #890, opened 2023-11-07
Multiple "Referenced element is not defined" errors in many platform bundles


iloveeclipse:
Looks like a severe regression from aa208c02423ac8aa48f10a14d5c2003b2147d1ce / https://github.com/eclipse-pde/eclipse.pde/pull/845.

Last good build seem to be `I20231030-0450`

I see now many Plug-In errors like `Referenced element 'and' is not defined` reported for:
- org.eclipse.debug.ui
- org.eclipse.ui.console
- org.eclipse.tips.core
- org.eclipse.ui.intro
- org.eclipse.jdt.core.manipulation
- org.eclipse.jdt.ui
- org.eclipse.ui
- org.eclipse.ui.ide
- org.eclipse.ui.navigator
- org.eclipse.core.resources
- org.eclipse.core.expressions
- org.eclipse.pde.ui.tests
- org.eclipse.core.expressions.tests
- may be more, have not all bundles in the workspace

![image](https://github.com/eclipse-pde/eclipse.pde/assets/964108/8af36ede-b661-4da6-94ba-997d4b641fe3)

laeubi:
I don't see this error in

```
Version: 2023-12 (4.30)
Build id: I20231102-1800
```

iloveeclipse:
Looks like you need a clean build on affected projects. That seem to be the reason I haven't noticed the errors before - I always use last nightly build for my day work.

I can reproduce with Eclipse started on Java 21 and on Java 17, both in the old and in a fresh new workspace:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/964108/96dcb18c-7d4f-4937-a65a-e92fdf338f07)

merks:
I expect it's being caused by the include not being processed properly:
```
<include schemaLocation="schema://org.eclipse.core.expressions/schema/expressionLanguage.exsd"/>
```

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/e8c12b30-4cd7-4f24-aa62-3a621cfe3efb)

laeubi:
> Looks like you need a clean build on affected projects

This did the trick for me, will look into it.

merks:
I Oomphed a fresh installation with just the Platform UI clone and here a clean build does not fix the problem:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/ab165b5d-9773-48b0-a24c-74575989fdde)

laeubi:
[...]

PR #891, opened 2023-11-07
Use a default schema provider without path in case none is supplied

laeubi:
Even though no search path is provided we need to decode the URL and see if we can find the schema in the current workspace.

Also fixes an issue where schema can't be opened by double click on linux.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/890


laeubi:
@vik-chand @HannesWell please give +1 to merge this bugfix/regression regardless of M3 freeze.

iloveeclipse:
> please give +1 to merge this bugfix/regression regardless of M3 freeze.

Fixes for regressions do not need approval in M3

merks (approved):
Don’t wait for reviews.

laeubi:
Okay then lets merge as soon as CI checks have completed.

iloveeclipse (approved):

vik-chand:
thanks @laeubi  @iloveeclipse

Issue #764, opened 2023-09-22
ClasspathUpdaterTests fail on Java 21


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20230922-0140/testresults/html/org.eclipse.pde.ui.tests_ep430I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html

```
Classpath entry for JavaSE-17 is missing

java.lang.AssertionError: Classpath entry for JavaSE-17 is missing
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.findClasspathEntry(ClasspathUpdaterTest.java:202)
at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.assertClasspathAttribute(ClasspathUpdaterTest.java:174)
at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.test_ChangeSourceAttachment(ClasspathUpdaterTest.java:139)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:276)
```

HannesWell:
@haubi can you have a look at this?
If I select a JDK 20 (I have 21 not yet available) as default JVM the test still passes, so I assume the only JDK available in the workspace must be greater 17 in order to reproduce this failure.

iloveeclipse:
See errors reported by API tooling on https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1439 and https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1436.

They both run on Java 21 on Jenkins and both show same API errors, all reported on enums defined in JDT.

As I've wrote on https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1436, I suspect a problem with classpath resolving in API tooling.

It looks like if the API baseline would see java.lang.Enum from Java 8 or 11 in the **current** API (why???) and compare it with java.lang.Enum from Java 17 **in the API baseline** where it **additionally** implements Constable (since Java 12).

So it would be as if PDE is not able to recognize Java 21 as valid JDK and fails back to Java 11 (or 8). 
[...]

PR #893, opened 2023-11-07
ClasspathUpdaterTests fail on Java 21

iloveeclipse:
Don't hard code JRE version in test expectations. The test only wants to know if the classpath entries are there and in the right order, independently which JRE is used to run Eclipse / be on classpath.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/764

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/pom.xml
+++ b/ui/org.eclipse.pde.ui.tests/pom.xml
@@ -35,6 +35,7 @@
         <artifactId>tycho-surefire-plugin</artifactId>
         <version>${tycho.version}</version>
         <configuration>
+          <argLine>-DDetectVMInstallationsJob.disabled=true</argLine>
           <useUIHarness>true</useUIHarness>
           <useUIThread>true</useUIThread>
           <dependencies>
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathupdater/ClasspathUpdaterTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathupdater/ClasspathUpdaterTest.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.pde.ui.tests.classpathupdater;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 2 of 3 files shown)


HannesWell:
Thanks Andrey for providing a fix.

> Don't hard code JRE version in test expectations. The test only wants to know if the classpath entries are there and in the right order, independently which JRE is used to run Eclipse / be on classpath.

Thinking only shortly about this I wonder if the JRE version should always be derived from the BREE instead? Regardless if the JRE/JDK available is an exact match or a higher version.
I cannot say a.t.m. if this is a regression from the associated change or if this was always the case.

iloveeclipse:
> JRE version should always be derived from the BREE instead

This PR makes it irrelevant which version will be used, as explained in the commit message, the test isn't about Java versions but about classpath entries persistence/order.

HannesWell:
> > JRE version should always be derived from the BREE instead
> 
> This PR makes it irrelevant which version will be used, as explained in the commit message, the test isn't about Java versions but about classpath entries persistence/order.

Yes that's right.
Nevertheless it looks like that it (unintentionally) reveals a flaw in the ClasspathUpdater.

iloveeclipse:
> Nevertheless it looks like that it (unintentionally) reveals a flaw in the ClasspathUpdater.

Which one?

iloveeclipse:
The test expected to see **literally** `Java17` classpath container, and failed if that wasn't found on the system (because JDT found Java 21). So test failed.

iloveeclipse:
If there are no objections, I would like to merge. I want see a green build at some time :-)

HannesWell:
> The test expected to see **literally** `Java17` classpath container, and failed if that wasn't found on the system (because JDT found Java 21). So test failed.

Are we talking about the String that's written into the `.classpath` file or about the actual VMInstall it is resolved against?
Because in case of the former, I would expect in general (regardless of if it is the purpose of the adjusted tests or not) that if the Manifest.MF contains a BREE Java-17 that the .classpath also contains a Java-17 JRE-Container.
But I just returned home from work and didn't have much time yet to make a in depth analysis of the topic.

HannesWell:
With your steps from https://github.com/eclipse-pde/eclipse.pde/issues/764#issuecomment-1798965303 I was able to reproduce the failure in the workspace (run the Plugin-Test, set in the created Launch-Config the `Runtime-JRE` explicitly to Java-21 and add the VM argument `-DDetectVMInstallationsJob.disabled=true`).

The problem is that the test-project does not define a BREE in its MANIFEST.MF:
https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.ui.tests/tests/projects/classpathupdater/META-INF/MANIFEST.MF
Adding 
```
Bundle-RequiredExecutionEnvironment: 
 JavaSE-17
```
also fixed the test for me locally.

iloveeclipse:
So would you like to push your change instead?

HannesWell:
> So would you like to push your change instead?

Since it requires less code I would prefer that solution, if it works.
I can provide a separate PR, unless you want to apply the alternative solution in this PR. :)

iloveeclipse:
No, I don't have access to PC now, so please push your change.

PR #900, opened 2023-11-08
ExternalLibraryCache: fix javadoc

jukzi:


jukzi:
As far as i see that is the last javadoc issue in pde.

HannesWell (dismissed):
Looks like a save change.

It can be submitted for RC1 after M3 has been declared.
**So don't submit this before the announcement was made (expected for Friday).**

PR #892, opened 2023-11-07
Remove blank Javadoc

jukzi:
This commit cleans up Javadoc that does not add information. It resolves ecj warnings:
 `Javadoc: Description expected after ...`
It helps to prevent future empty javadoc by disabling missingJavaDoc warnings. This resolves
 `Javadoc: Missing ...`

The modification is a result of regular expression search&replace:

in files `*.java`
 `^[\s]*\*[\s]*(@return|@param[\s]*[^\s]+|@throws[\s]*[^\s]+)\R([\s]*\*[\s]*@|[\s]*\*/\R)`
 ->`$2`
 `^([\s]*\*[\s]*\R)([\s]*\*/\R)`
 ->`$2`
 `^[\S\t]*/\*\*\R[\s]*\*/\R`
 ->``

in files `org.eclipse.jdt.core.prefs`
 `org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc(Comments|Tags)\=[^\s]*`
 ->`org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc$1\=ignore`

excluded resources:
 apitools/org.eclipse.pde.api.tools.tests/test-builder
 apitools/org.eclipse.pde.api.tools.tests/test-classes
 apitools/org.eclipse.pde.api.tools.tests/test-source
 apitools/org.eclipse.pde.api.tools.tests/tests-deltas

PR #901, opened 2023-11-08
Revert "Remove blank Javadoc"

HannesWell:
Reverts eclipse-pde/eclipse.pde#892

That change was too large and cannot be reviewed in detail and thus it is not save to submit this so late in the release-cycle.
Since there is also no user benefit, there is not need to rush this into PDE now.

Therefore it is reverted.

At the very beginning of the next release cycle, the same change can be applied again.

@akurtakov FYI


akurtakov:
For the record. Changes should be reviewable so such a change would better be done per component/module.

PR #903, opened 2023-11-09
Fixes PDE preference page template

vogella:
The current template assumes that the plug-in has an activator. As we do not use activators anymore in the default plug-in template, we should also adjust this template.

Plus getting rid of NON-Javadoc

### DIFF
--- a/ui/org.eclipse.pde.ui.templates/templates_3.1/preferences/java/$pageClassName$.java
+++ b/ui/org.eclipse.pde.ui.templates/templates_3.1/preferences/java/$pageClassName$.java
@@ -1,9 +1,11 @@
 package $packageName$;
 
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jface.preference.*;
 import org.eclipse.ui.IWorkbenchPreferencePage;
 import org.eclipse.ui.IWorkbench;
-import $pluginClass$;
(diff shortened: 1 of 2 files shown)


HannesWell:
Can you please provide the steps to apply this?
Finding the templates in PDE's new project wizard is often too much trial-and-error.

vogella:
Ah, sorry for not providing steps:

In an existing plugin.xml add the extension org.eclipse.ui.preferencePages and select the template on the left bottom side.

Or to start from scretch: https://www.vogella.com/tutorials/EclipsePreferences/article.html#exercise-contribute-a-preference-page-to-the-eclipse-ide but in instead of adding the "Enter the data in plugin.xml similar to the following:" manually use the wizard and select the entry in the left bottom corner.

vogella:
If you still not find it, I can add screenshots, currently on my way to get a "Grippeimpfung". :-)

vogella:
@HannesWell fine to commit or do you prefer to wait for the next release?

HannesWell:
> @HannesWell fine to commit or do you prefer to wait for the next release?

Sorry had no time to check this.
I'll try to do it tomorrow, but I tend to wait until next release cycle since it looks like this is not a recent flaw that is fixed here.

vogella:
Mac build error looks unrelated

PR #905, opened 2023-11-10
Enable pde build tests

laeubi:
At least the test runtime comes up already ...

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -23,4 +23,33 @@
     <module>org.eclipse.pde.build</module>
     <module>org.eclipse.pde.build.tests</module>
   </modules>
+  
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-surefire-plugin</artifactId>
+        <configuration>
+          <dependencies>
+            <dependency>
+              <type>eclipse-plugin</type>
+              <artifactId>org.eclipse.osgi.compatibility.state</artifactId>
+              <version>0.0.0</version>
+            </dependency>
+            <dependency>
+              <type>p2-installable-unit</type>
+              <artifactId>org.eclipse.equinox.executable</artifactId>
+              <version>0.0.0</version>
+            </dependency>
+            <dependency>
+              <type>eclipse-feature</type>
+              <artifactId>org.eclipse.sdk</artifactId>
+              <version>0.0.0</version>
+            </dependency>
+          </dependencies>
+          <testRuntime>p2Installed</testRuntime>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
--- a/build/org.eclipse.pde.build.tests/build.properties
+++ b/build/org.eclipse.pde.build.tests/build.properties
@@ -20,5 +20,5 @@ bin.includes = META-INF/,\
                test.xml
 src.includes = about.html
 jars.compile.order = .
-pom.model.property.skipTests = true
+pom.model.property.skipTests = false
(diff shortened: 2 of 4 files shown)


HannesWell:
Enabling them is a good idea, but before this is submitted the tests should of course pass.

So fixing https://github.com/eclipse-pde/eclipse.pde/issues/884 is a prerequisite.

build/pom.xml
```diff
@@ -23,4 +23,33 @@
...
+              <version>0.0.0</version>
+            </dependency>
+            <dependency>
+              <type>eclipse-feature</type>
+              <artifactId>org.eclipse.sdk</artifactId>
+              <version>0.0.0</version>
+            </dependency>
+          </dependencies>
+          <testRuntime>p2Installed</testRuntime>
+        </configuration>
```
mickaelistria, line 51: Is this really necessary to have it installed at that stage (my concern is that test execution in requirements aren't so portable)?
Could it be replaced by some instructions in p2.inf?
Could it be replaced by creation of a TargetPlatform in the tests that would consist of running target?
Do you really need it as "p2installed", or would running this setting the right `testProduct/testApplication` to usual SDK values be sufficient? `p2installed` is usually useful only when there are crazy things like touchpoints or when you need the files to be organized on disk similarly to an installation.

laeubi: > Is this really necessary to have it installed at that stage (my concern is that test execution in requirements aren't so portable)?

I tried without but didn't work so well... I have no clue why the `compatibility` is not discovered automatically but it seems a common problem.

> Could it be replaced by some instructions in p2.inf?

Probably yes, but last time I tried it wasn't that obvious and introduced build cycles in aggregator.

> Could it be replaced by creation of a TargetPlatform in the tests that would consist of running target?

The problem is that the test don't start at all, the only good think is that one can easily try this standalone and play around by go into the folder `eclipse.pde/build/org.eclipse.pde.build.tests` and execute `mvn clean verify` and it is also quite fast.

> Do you really need it as "p2installed"

`p2installed` was just something I tried here because the test do special stuff, especially require that the plugins are exploded, that an installation location is set and so on, I haven't debugged this in detail and there might be mitigations but as pde-build is nothing I have experience with its hard for me to decide.

> p2installed is usually useful only when there are crazy things like touchpoints or when you need the files to be organized on disk similarly to an installation.

I fear that's the case here...

mickaelistria: > I have no clue why the compatibility is not discovered automatically but it seems a common problem.

It's a fragment of org.eclipse.osgi, and it's required whenever one invokes `Platform.getPlatformAdmin()`.

I added it (in p2.inf) and see that it indeed seems to require a full installation (using the simpleconfigurator/bundles.info file).

But I see the text.xml file contains 2 launches, one with a system property. Passing this system property may help running the tests here.
I'm actually curious as to whether PDE would handle this case as well without enabling p2.

mickaelistria:
I've opened https://github.com/eclipse-pde/eclipse.pde/pull/927 with further configuration to run the tests. I get 12 failures locally that I believe are all caused by the missing "Delta Pack".
Provisionign a "delta pack" shouldn't be a test pre-requisite, but instead be somehting the the test themselves would perform as part of their setup (in Java code), so we can run the tests with whichever build or script technology without having to rewrite those bits. I will try to work on it sometime later, but if anyone is willing to start this (not too difficult I hope) change in the meantime, it would be welcome and can be developed/merged independently from this current PR.

laeubi:
@mickaelistria thanks for that, I have added a comment in the PR regarding delta-pack.

laeubi:
Closing in favor of 
- https://github.com/eclipse-pde/eclipse.pde/pull/927

PR #908, opened 2023-11-11
New PDE test fails since I20231105-1800

iloveeclipse:
Adopted test expectations to changes made in
1765416677a731a710a9eb43524b4bd88f89b752, where PDE exports bundles as jars by default (unless "Eclipse-BundleShape: dir" is specified in manifest or other preconditions force bundle to be a directory).

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/884


HannesWell:
Thanks for keeping the tests green!

Issue #447, opened 2023-01-28
PDE API-Tools should add api.tools.annotations on a Plug-ins classpath automatically


HannesWell:
In order to use the PDE API-Tools annotations they have to be added from the current TP to the classpath of a Plugin by either `Import-Package`, `Required-Bundle` in the `MANIFEST.MF` or as `additional.bundles` in the `build.properties`. The latter is used by the `ApiQuickFixProcessor`, when converting the api-tools javadoc tags to annotations because it does not introduce a dependency at runtime (the annotations are only processed at build time).

In order to make using api-tools annotations even simpler the annotation should be added to a Plug-ins classpath using a `IClasspathContributor` created by the api-tools nature or builder. Then the a entry like `additional.bundles = org.eclipse.pde.api.tools.annotations` in the build.properties or even worse a entry in the `MANIFEST.MF` is not necessary anymore.
We already use that for the `OSGiAnnotationsClasspathContributor` and it works quite well.

If PDE supports that we have to check that Tycho supports that as well, but since Tycho just uses PDE's `APIFileGenerator` we can do that in PDE as well.

laeubi:
> If PDE supports that we have to check that Tycho supports that as well, but since Tycho just uses PDE's `APIFileGenerator` we can do that in PDE as well.

Tycho has a similar concept an the `tycho-apitools-plugin` must just provide a (plexus)component that extends `AbstractSpecificationClasspathContributor` to automatically enable this.

HannesWell:
In order to ease the usage of these annotations it would probably also be good to have a Quick-Fix and/or clean-up to replace the java-doc based annotations.

laeubi:
> In order to ease the usage of these annotations it would probably also be good to have a Quick-Fix and/or clean-up to replace the java-doc based annotations.

I would love to get open-rewrite working for that:
- https://github.com/openrewrite/rewrite-maven-plugin/issues/649#issuecomment-1792525034

@timtebeek is currently investigate that for [moderne.io](https://app.moderne.io/) and I'm currently investigate to have a JDT plugin for that:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/871
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/870

PR #912, opened 2023-11-12
Add a ClasspathContributor for API tools annotations

laeubi:
Currently it is quite cumbersome to use and discover the PDE own API tools annotations and requires some kind of workarounds.

This adds a classpath contributor that makes the annotations available like already done for ds and bundle annotations.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/447


laeubi:
Tycho change:
- https://github.com/eclipse-tycho/tycho/pull/3035

PR #913, opened 2023-11-12
Make annotations use CLASS retention policy

laeubi:
Currently the API Analysis annotations are only retained as SOURCE level what mean that they are not discoverable by tools that scan the final jar.

This changes the retention policy to CLASS so they are retained at the class level but not at runtime.


iloveeclipse:
Note, I would assume change on retention policy is similar to API level change, so it requires minor version segment bump.

laeubi:
> Note, I would assume change on retention policy is similar to API level change, so it requires minor version segment bump.

yes lets wait for the next stream ope

> Note, I would assume change on retention policy is similar to API level change, so it requires minor version segment bump.

Done I also lowered the EE requirement so the annotations can be used in the context of older Java levels, I think we not really benefit from any newer java language features here.

HannesWell:
> Currently the API Analysis annotations are only retained as SOURCE level what mean that they are not discoverable by tools that scan the final jar.

Do you plan to create such tools? Because at the moment the api-description ist generated from it, which makes the annotations then unnessecary. Of course having class Retention allows to generate the description from compiled classes instead of sources only.

laeubi:
> Do you plan to create such tools? Because at the moment the api-description ist generated from it, which makes the annotations then unnessecary. 

I would argue that actually the `api-description` is unnecessary if we can retain the same information in class-code :-)
e.g. `tycho-baseline-plugin` can be made to consider these additional information, `bnd-baseline-plugin` supports already API-Guardian so with some effort one can write a plugin, but require a special description file (that needs a parser as well and can be forgotten) seems not very useful these days.

HannesWell:
> > Do you plan to create such tools? Because at the moment the api-description ist generated from it, which makes the annotations then unnessecary.
> 
> I would argue that actually the `api-description` is unnecessary if we can retain the same information in class-code :-)

I think I have to correct myself. IIRC the .api_description mainly contains suppressed API errors and an optional decription message.
Therefore it is probably not possible to replace it only by annotations.

HannesWell (approved):

laeubi:
I created 
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4011

but I think its time to rething if we really want to use BREE Libs anymore or should start to simply rely on target feature of JVM.

laeubi:
I now created
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1603

so we can configure `useJDK=SYSTEM` for this module.

akurtakov:
Why switching to JavaSE 1.6? What/where will benefit from it? Latest javac can target 8+ which IMHO is the absolute minimum in Java nowadays.

HannesWell:
> Why switching to JavaSE 1.6? What/where will benefit from it? Latest javac can target 8+ which IMHO is the absolute minimum in Java nowadays.

But one can use e.g. Java-1.8 for compilation, which can target older version.
But I don't really disagree with that, nevertheless since this Plugin only contains annotations I don't know any benefit of requiring higher versions. If there are benefits in the future we have to discuss this again.

laeubi:
> Why switching to JavaSE 1.6

it is the minimum version one can use with -release option of modern JDKs and is enough for defining an annotation.

akurtakov:
My javac 21 gives me:
```
--release <release>
        Compile for the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
```
javac 17 gives:
```
--release <release>
        Compile for the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17
```

What am I missing?

laeubi:
> What am I missing?

Good question, I'll try to check later if really 1.6 is used as target, at least ECJ can compile down to Java 1.1 ...

akurtakov:
Using --target with ecj doesn't mean that --release will work with ecj as it depends on what's the JVM shipping. So on Java 21 --release can not work for pre Java 1.8 even with ecj.

laeubi:
> So these annotations retained at class level should be full replacement for the .api_description file.

Good to know, only thing is that the Api Tools need to consider them at class level where they now only use `.api_description` file.

Issue #915, opened 2023-11-13
Bogus "Plug-in reference '...' cannot be resolved" warnings


merks:
Oomph's workspace was warning free but with the current I-Build it has this bogus warning::

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/24357816-06e8-4e3d-9d4d-7f112640b2e7)

One can even double click on the included plugin to navigate to it.

I think this is a regression perhaps related to removal of support for some of the properties.

laeubi:
Just in case can you try remove the fragment property if it fix the warning?

merks:
Unfortunately not:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/bab95d8b-b655-487a-bd6d-4dea30165bb5)

merks:
We see the same problem in the SDK's own workspace too:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/c2fd64b6-8d7b-44c9-acff-c1cf49d09ac1)

HannesWell:
Fixed with https://github.com/eclipse-pde/eclipse.pde/pull/916.

PR #916, opened 2023-11-13
FeatureErrorReporter.validatePluginExists should recognized fragments

merks:
The validator currently produces an error when it find the referenced plugin but the plugin is a fragment. This is wrong behavior for the latest implementation that makes no distinction between plugins versus fragments.

https://github.com/eclipse-pde/eclipse.pde/issues/915


laeubi (approved):

HannesWell (approved):
Yes that should have been removed in https://github.com/eclipse-pde/eclipse.pde/pull/770.

Thanks Ed!

HannesWell:
Since this is a recent regression, lets merge this now.

PR #918, opened 2023-11-13
Add adapter for IProject > (bnd) Project

laeubi:
If one wants to write generic UI components then there is often the question to answer to what (bnd) project an Eclipse IProject belongs (if any).

This adds an Adapter that can adapt from IProject > (bnd) Project if the IProject is a PDE Bnd backed project.

See also
- https://bnd.discourse.group/t/announcement-bnd-and-pde-cooperation/372
- Corresponding bndtools change https://github.com/bndtools/bnd/pull/5885


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/PdeBndAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+public class PdeBndAdapter implements IAdapterFactory {
+
+	@Override
+	public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
+		if (adapterType == Project.class) {
+			if (adaptableObject instanceof IProject project) {
+				try {
+					return adapterType.cast(BndProjectManager.getBndProject(project).orElse(null));
+				} catch (Exception e) {
+					// can't adapt then...
```
szarnekow, line 34: Should this be logged at least?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/PdeBndAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.IAdapterFactory;
+import org.osgi.service.component.annotations.Component;
+
+import aQute.bnd.build.Project;
+
+@Component(service = IAdapterFactory.class, property = {
+		IAdapterFactory.SERVICE_PROPERTY_ADAPTABLE_CLASS + "=org.eclipse.core.resources.IProject",
+		IAdapterFactory.SERVICE_PROPERTY_ADAPTABLE_CLASS + "=aQute.bnd.build.Project" })
```
HannesWell, line 24: Isn't this an unidirectional adaption from IProject to (BND)-Project?
Therefore it does not look right that `SERVICE_PROPERTY_ADAPTABLE_CLASS` is used for both source and target type.
If I understood the doc correctly it should be:
```suggestion
		IAdapterFactory.SERVICE_PROPERTY_ADAPTER_NAMES + "=aQute.bnd.build.Project" })
```

HannesWell: Looking at https://github.com/bndtools/bnd/pull/5885/files#diff-798b65a162b9e5e53ee378226f1a31b3c72511f2d12c6aec6bea5c31ced1d891 seems to confirm my assumption, nevertheless I have the impression that the implementation and service properties don't really provide the opposite but the same direction. Or was it not the intend to provide an (BND-project->IProject adoption there?)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/PdeBndAdapter.java
laeubi: The `Exception` is sadly a burden from the BND project that uses throws Exception all the time :-\
In this case it actually *should* never happen but if it does something is wrong and caller can just give up as we can't decide whats going wrong.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/PdeBndAdapter.java
laeubi: > Or was it not the intend to provide an (BND-project->IProject adoption there?)

It should be `IProject` -> (bnd) `Project`

PR #927, opened 2023-11-20
Run some pde.build tests with Tycho

mickaelistria:
Missing tests require provisioning the "delta pack" (ie launchers for all Platforms)

### DIFF
--- a/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
+++ b/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
@@ -14,7 +14,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.apache.ant,
  org.eclipse.debug.core,
  org.eclipse.equinox.p2.publisher;bundle-version="1.1.0",
- org.eclipse.equinox.p2.publisher.eclipse;bundle-version="1.0.0"
+ org.eclipse.equinox.p2.publisher.eclipse;bundle-version="1.0.0",
+ org.eclipse.equinox.p2.repository.tools
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Import-Package: org.eclipse.equinox.frameworkadmin;version="2.0.0",
  org.eclipse.equinox.internal.p2.artifact.repository,
--- a/build/org.eclipse.pde.build.tests/META-INF/p2.inf
+++ b/build/org.eclipse.pde.build.tests/META-INF/p2.inf
@@ -1,0 +1,2 @@
+requires.0.namespace = org.eclipse.equinox.p2.iu
+requires.0.name = org.eclipse.osgi.compatibility.state
--- a/build/org.eclipse.pde.build.tests/build.properties
+++ b/build/org.eclipse.pde.build.tests/build.properties
@@ -20,5 +20,4 @@ bin.includes = META-INF/,\
                test.xml
 src.includes = about.html
 jars.compile.order = .
-pom.model.property.skipTests = true
 pom.model.property.code.ignoredWarnings = ${tests.ignoredWarnings}
--- a/build/org.eclipse.pde.build.tests/pom.xml
+++ b/build/org.eclipse.pde.build.tests/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.eclipse.pde</groupId>
+		<artifactId>eclipse.pde.build</artifactId>
+		<version>4.30.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>org.eclipse.pde.build.tests</artifactId>
+	<version>1.4.200-SNAPSHOT</version>
+	<packaging>eclipse-test-plugin</packaging>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>target-platform-configuration</artifactId>
+				<configuration>
+					<dependency-resolution>
+						<extraRequirements>
+							<requirement>
+								<!-- The product -->
+								<type>p2-installable-unit</type>
+								<id>org.eclipse.sdk.ide</id>
+								<versionRange>0.0.0</versionRange>
+							</requirement>
+						</extraRequirements>
+					</dependency-resolution>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-surefire-plugin</artifactId>
+				<configuration>
+					<product>org.eclipse.sdk.ide</product>
+					<testRuntime>p2Installed</testRuntime>
+					<testClass>org.eclipse.pde.build.tests.PDEBuildTestSuite</testClass>
+					<systemProperties>
+						<pde.build.includeP2>false</pde.build.includeP2>
+					</systemProperties>
+				</configuration>
+				<executions>
+					<execution>
+						<id>p2-tests</id>
+						<goals><goal>test</goal></goals>
+						<configuration>
+							<testClass>org.eclipse.pde.build.tests.P2TestSuite</testClass>
+							<systemProperties/>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+</project>


laeubi:
@mickaelistria from my previous debugging I think that pdebuild want something like we have in product with "include native launcher artifacts", not sure if such an option could be added to p2 installed test framework.

The second thing is, it requires that the "install location" is set and points to the root of the installation that contains the eclipse.exe

mickaelistria:
@laeubi thanks for looking at it. I don't think I'll spend more time on it, my goal was more to demonstrate how such strange cases can be handled and to bootstrap a solution. So if you're willing to continue working on this topic, re-using this patch or not, that would still be welcome.
About delta-pack, I believe that a proper solution is to have a `DeltaPackUtils.provisionDeltaPack(File destination)` method that would take care of invoking the p2 slicer to populate the destination.

akurtakov:
IIRC installing org.eclipse.equinox.executable feature supplies all the different launchers and somewhere in the code it handles that as if delta pack is available. It should have been properly cleaned up but no one really cares about pde.build so it is what it is.

laeubi:
@mickaelistria thats already helpful and I'll try to take a look into it later on after the december release, do oyu mind adding the `org.eclipse.equinox.executable` feature as suggested by @akurtakov ?

laeubi:
I take a look and actually the executable already is included in the p2 installed runtime (I assume the sdk product already pull that in ...) but it explicitly want to find `org.eclipse.equinox.executable` feature so is not happy with the default eclipse executable in the product itself.

akurtakov:
IIRC o.e.equinox.executable contains all arch specific executables and many tests are written with e.g. win32 being hardcoded.

laeubi:
> IIRC o.e.equinox.executable contains all arch specific executables and many tests are written with e.g. win32 being hardcoded.

I now was able to supply the build with this "deltapack" that it finds the launcher and bring it down to 9 failures... but it seems it also wants SWT items as well... so it does not only use the executables for test it also uses the generted repo for provision a target in the tests...

laeubi:
> and many tests are written with e.g. win32 being hardcoded.

That seems to be the problem for the remaining tests, they want win32 to be "installed" but if running on linux of course this will not happen... so extra requirements don't work... need to find how to trick P2 to still install them...

iloveeclipse:
> That seems to be the problem for the remaining tests, they want win32 to be "installed"

If that the only issue that prevents other tests to be executed in Jenkins, I see no problem to disable it on non-Windows platforms for now. A proper fix would be even to be platform independent, but I have no idea how much work that would be. Better have more tests immediately as no tests for long time.

laeubi:
> If that the only issue that prevents other tests to be executed in Jenkins, I see no problem to disable it on non-Windows platforms for now

That's an interesting idea, I'll push my change and see if it runs on the windows runner already...

> A proper fix would be even to be platform independent, but I have no idea how much work that would be

You can take a look at `org.eclipse.pde.build.internal.tests.AssembleTests` there it uses to build some test artefacts and these contain for example this:

https://github.com/eclipse-pde/eclipse.pde/blob/eeffb962b568b6208d1de4d8b897a78371a4eec2/build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/AssembleTests.java#L38-L39

so at a first glance it seems one needs instead a case for every platform (win/linux/mac) for this to run on different installs.

laeubi:
A quick try with only allow windows as environment failed, seems the P2 installed runtime is missing the fix to use the only defined runtime if it does not match the running target:

```
Cannot complete the install because one or more required items could not be found.
	Software being installed: Eclipse SDK 4.30.0.I20231123-1700 (org.eclipse.sdk.ide 4.30.0.I20231123-1700)
	Missing requirement for filter properties ~= $0: Equinox p2, headless functionalities 1.7.0.v20231112-1314 (org.eclipse.equinox.p2.core.feature.feature.group 1.7.0.v20231112-1314) requires 'org.eclipse.equinox.p2.iu; org.eclipse.equinox.security.linux [1.1.100.v20231012-1237,1.1.100.v20231012-1237]' but it could not be found
	Cannot satisfy dependency:
		From: Equinox p2, Provisioning for IDEs. 2.4.2200.v20231112-1314 (org.eclipse.equinox.p2.user.ui.feature.group 2.4.2200.v20231112-1314)
		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.core.feature.feature.group [1.7.0.v20231112-1314,1.7.0.v20231112-1314]
	Cannot satisfy dependency:
		From: Eclipse p2 Provisioning Developer Resources 2.4.2200.v20231112-1314 (org.eclipse.equinox.p2.user.ui.source.feature.group 2.4.2200.v20231112-1314)
		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.user.ui.feature.group [2.4.2200.v20231112-1314,2.4.2200.v20231112-1314]
	Cannot satisfy dependency:
		From: Eclipse SDK 4.30.0.I20231123-1700 (org.eclipse.sdk.ide 4.30.0.I20231123-1700)
		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.user.ui.source.feature.group [2.4.2200.v20231112-1314,2.4.2200.v20231112-1314]
```

I'll check if that can be fixed at Tycho ...

laeubi:
Here we see https://github.com/eclipse-pde/eclipse.pde/runs/18999626168 that there are some test that fail on 2/3 platforms that means they seem to work in windows!

Problem is if we disable them somehow for non windows they wont run on the i-builds as well or we need just another property to set like the p2 one...

laeubi:
Stupid enough some seem to have hardcoded LINUX in their assumptions:

> Unable to find plug-in: org.eclipse.swt.gtk.linux.x86_64_0.0.0. Please check the error log for more details.

🤯

laeubi:
And then there are some test that need different platform items... because they want to test some kind of building for another platform maybe (like Tycho supports), so to support this usecase I think one needs a new option in Tycho p2installed runtime to provision additional things...

mickaelistria:
@laeubi Is it OK if I close this one as I see you're making further progress with https://github.com/eclipse-pde/eclipse.pde/pull/933 ?

laeubi:
> A quick try with only allow windows as environment failed, seems the P2 installed runtime is missing the fix to use the only defined runtime if it does not match the running target:
> 
> ```
> Cannot complete the install because one or more required items could not be found.
> 	Software being installed: Eclipse SDK 4.30.0.I20231123-1700 (org.eclipse.sdk.ide 4.30.0.I20231123-1700)
> 	Missing requirement for filter properties ~= $0: Equinox p2, headless functionalities 1.7.0.v20231112-1314 (org.eclipse.equinox.p2.core.feature.feature.group 1.7.0.v20231112-1314) requires 'org.eclipse.equinox.p2.iu; org.eclipse.equinox.security.linux [1.1.100.v20231012-1237,1.1.100.v20231012-1237]' but it could not be found
> 	Cannot satisfy dependency:
> 		From: Equinox p2, Provisioning for IDEs. 2.4.2200.v20231112-1314 (org.eclipse.equinox.p2.user.ui.feature.group 2.4.2200.v20231112-1314)
> 		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.core.feature.feature.group [1.7.0.v20231112-1314,1.7.0.v20231112-1314]
> 	Cannot satisfy dependency:
> 		From: Eclipse p2 Provisioning Developer Resources 2.4.2200.v20231112-1314 (org.eclipse.equinox.p2.user.ui.source.feature.group 2.4.2200.v20231112-1314)
> 		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.user.ui.feature.group [2.4.2200.v20231112-1314,2.4.2200.v20231112-1314]
> 	Cannot satisfy dependency:
> 		From: Eclipse SDK 4.30.0.I20231123-1700 (org.eclipse.sdk.ide 4.30.0.I20231123-1700)
> 		To: org.eclipse.equinox.p2.iu; org.eclipse.equinox.p2.user.ui.source.feature.group [2.4.2200.v20231112-1314,2.4.2200.v20231112-1314]
> ```
> 
> I'll check if that can be fixed at Tycho ...

This will be addressed with
- https://github.com/eclipse-tycho/tycho/pull/3088

even though it does not help here...

laeubi:
> @laeubi Is it OK if I close this one as I see you're making further progress with #933 ?

I think we can leave it open as most discussions have taken place here until we found a final solution.

laeubi:
With this change I'm now able to run all test except one that is failing as it now gets more content than expected
- https://github.com/eclipse-tycho/tycho/pull/3091

I'll disable that test for verification builds maybe one more familiar with PDE builds (@iloveeclipse) is then able to make that test work in both scenarios...  The test in question is `org.eclipse.pde.build.internal.tests.ScriptGenerationTests.testBug238177()` what has very strict requirements on number and order of ant targets and seem to return one target more when executed with the p2 provisioned install that is used as a target platform.

laeubi:
So to summarize my findings here:

- pde tests is using p2 in test.xml to build a special local "updatesite"
- this is then used in some test as a target platform to resolve stuff
- the util method for that is searching for the launcher and then assume this is location where plugins folder exits to resolve more stuff including multiple plugins for multiple target platforms

Currently we "fake" this with the following workarounds:

1. Install the equinox launcher feature that make the Util method think the "deltapack" is the eclipse running the test
2. Install additional target environments so afterwards they are found in the plugin folders

So the best actually would be to rewrite the test so they use for example a release repository or even the ibuilds repository and then simply setup a target platform that provide in what is desired, this would make the test easier to run in different environments and not rely on ant/tycho special setups.

Just in case one want to look into this, a good start is `org.eclipse.pde.build.internal.tests.Utils.findExecutable()` all test calling this effectively later do a

```
if (!executable.equals(new File((String) properties.get("baseLocation"))))
			properties.put("pluginPath", executable.getAbsolutePath());
```

to set a plugin path that is then later passed to the ant build scripts and from there passed as an additional path to the targets to search for stuff...

HannesWell:
I think this is included in https://github.com/eclipse-pde/eclipse.pde/pull/980 and thus can be closed.
Nevertheless thanks for this work.

Issue #703, opened 2023-08-08
Support for automatic manifest generation: Feature request for converter of old style PDE bundles


mikerumpf:
There should be an easy way to convert old PDE projects into the new BND form.

This should
 - create the bnd file with settings and dependencies derived from the existing manifest, creating necessary package-info.java classes with the necessary annotations)
 - configure .project and .classpath files correctly
 - delete the old manifest and other OSGi metadata

PR #934, opened 2023-11-25
Provide and action to convert MANIFEST-first PDE projects to PDE automatic manifest

laeubi:
This now supports:

1. convert manifest to `pde.bnd` entries
2. convert `build.properties` to `pde.bnd` instructions
3. adding all required bundles as buildpath
4. creating package-info.java for export packages
5. converting library jars from `build.properties` (thanks @pkriens for the hints about `-make` instruction)

What is left but better handled in follow ups:

1. converting of old style handcrafted component.xml to annotations
2. ...

There is always a choice for the user I also added an option to place the manifest at the project root as discussed at [EclipseCon 2023](https://www.eclipsecon.org/2023) FYI @merks @juergen-albert @stbischof

grafik

Fix https://github.com/eclipse-pde/eclipse.pde/issues/703


laeubi:
See linked issue:
- https://github.com/eclipse-pde/eclipse.pde/issues/703

mickaelistria:
I've seen the issue, but I'm commenting the commit here. The message is IMO not clear enough because "old" per se doesn't mean something clear enough in this context.
If the goal is to "Provide and action to convert MANIFEST-first PDE projects to BND projects", let's just write it clearly.

laeubi:
I adjusted PR title and commit message.

mickaelistria:
> I adjusted PR title and commit message.

Thanks, that's much clearer and much more interesting with this fresher wording ;)

laeubi:
After the transformation it *should* update the UI immediate, but depending on the settings you might need to open the manifest editor again as the `MANIFEST.MF` gets deleted.

Also this is not 100% perfect as we need to test it on different settings but should cover the 80% case already and I hope to improve it in this release cycle with additional follow ups.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/natures/PluginProject.java
```diff
@@ -36,4 +38,19 @@ public void deconfigure() throws CoreException {
 		removeFromBuildSpec(PDE.MANIFEST_BUILDER_ID);
 		removeFromBuildSpec(PDE.SCHEMA_BUILDER_ID);
 	}
+
+	public static boolean isPluginProject(IProject project) {
+		if (project.isOpen() && hasRequiredNatures(project)) {
+			return true;
+		}
+		return false;
```
HannesWell, line 46:
```suggestion
		return project.isOpen() && hasRequiredNatures(project);
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/tools/ConvertAutomaticManifestAction.java
```diff
@@ -0,0 +1,80 @@
...
+	private static IProject toProject(Object object) {
+		if (object instanceof IJavaProject java) {
+			return java.getProject();
+		}
+		if (object instanceof IProject project) {
+			return project;
+		}
+		if (object instanceof IResource resource) {
+			return resource.getProject();
+		}
```
HannesWell, line 76: An `IProject` is also an `IResource` and just returns itself when `getProject()` is called on it.
```suggestion
		if (object instanceof IResource resource) {
			return resource.getProject();
		}
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -300,7 +300,7 @@ public void defaultVMInstallChanged(IVMInstall previous, IVMInstall current) {
 			}
 		});
 	}
-	private static final String PDE_MANIFEST_BUILDER = "org.eclipse.pde.ManifestBuilder"; //$NON-NLS-1$
+	public static final String PDE_MANIFEST_BUILDER = "org.eclipse.pde.ManifestBuilder"; //$NON-NLS-1$
```
HannesWell, line 303: The class org.eclipse.pde.internal.core.natures.PDE already provides the constant for this named `MANIFEST_BUILDER_ID`. Actually I should have used that here too and probably just oversaw the existing constant.
Created https://github.com/eclipse-pde/eclipse.pde/pull/941 to change that and I suggest you use that too.

HannesWell (commented):
That's very nice to have. 🚀 

I cannot say much about the details and just have a few minor style-related suggestions.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/natures/PluginProject.java
HannesWell: Actually the `PDE` class already has a method `hasPluginNature()` that does exactly, what this helper method does.
Therefore I suggest you just inline `project.isOpen() && PDE.hasPluginNature(project)` to the only caller or if the project actually should always be open and a not open project is an error to use `PDE.hasPluginNature(project)` only.

laeubi: It does not do exactly what this method does:

1. throw Core exception
2. do not check if project is open

laeubi:
> I cannot say much about the details and just have a few minor style-related suggestions.

I have now applied these and think its a good state to merge and improve on top of this to not make the change bigger, this also has the advantage that one can start using that to convert some projects and look what is the outcome.

PR #936, opened 2023-11-27
Deploy PDE artifacts to maven

laeubi:
As an alternative to
- https://github.com/eclipse-pde/eclipse.pde/pull/826

I think we maybe better should go the path of deploy PDE directly to the maven where we can consume than with maven target locations instead of create a new update site.

This has the following benefits:

1. No need to setup a download space and if we later decide its not good at all the snapshots get deleted after a while automatically
2. We can get more experience with this way of deployment and how it works out so PDE can become a blueprint for other projects to do so as well
3. This could be a first step to make it possible to release artifacts to central without the help of global deploy script
4. It would allow to see how deploying features to a maven repository works out

If this is in place, we can start to iron out the produced poms and start trying to strip PDE form the aggregator and instead consume it from the target.

FYI @HannesWell @akurtakov @mickaelistria @vogella

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,5 @@
+def deployBranch = 'master'
+
 pipeline {
 	options {
 		timeout(time: 40, unit: 'MINUTES')
@@ -12,12 +14,15 @@ pipeline {
 		maven 'apache-maven-latest'
 		jdk 'temurin-jdk17-latest'
 	}
+	environment {
+		MVN_GOALS = getMavenGoals()
+	}
 	stages {
 		stage('Build') {
 			steps {
 				wrap([$class: 'Xvnc', useXauthority: true]) {
 					sh '''
-						mvn clean verify --batch-mode -Dmaven.repo.local=$WORKSPACE/.m2/repository \
+						mvn clean ${MVN_GOALS} --batch-mode -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 						-Pbree-libs \
 						-Papi-check \
 						-Pjavadoc \
@@ -38,4 +43,11 @@ pipeline {
 			}
 		}
 	}
+}
+
+def getMavenGoals() {
+	//if(env.BRANCH_NAME == deployBranch) {
+		return "deploy -DdeployAtEnd=true"
+	//}
+	//	return "verify"
 }
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,19 @@
     <module>features</module>
     <module>org.eclipse.pde.doc.user</module>
   </modules>
+  
+  <distributionManagement>
+        <repository>
+            <id>repo.eclipse.org</id>
+            <name>PDE - Releases</name>
+            <url>https://repo.eclipse.org/content/repositories/pde/</url>
+        </repository>
+        <snapshotRepository>
+            <id>repo.eclipse.org</id>
+            <name>PDE - Snapshots</name>
+            <url>https://repo.eclipse.org/content/repositories/pde-snapshots/</url>
+        </snapshotRepository>
+    </distributionManagement>
 
   <!--
     To build individual bundles, we specify a repository where to find parent pom,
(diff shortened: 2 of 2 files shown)


akurtakov:
I fully support deploying to maven.
Having p2 repo is still mandatory IMO.

laeubi:
> Having p2 repo is still mandatory IMO but one shouldn't block the other.

You can have a p2 repo on maven as well, see https://github.com/eclipse-tycho/tycho/tree/master/demo/p2-maven-site

Beside that I think Eclipse/IBuild/SDK/Simrel/... will still include PDE anyways one way or another.

akurtakov:
> > Having p2 repo is still mandatory IMO but one shouldn't block the other.
> 
> You can have a p2 repo on maven as well, see https://github.com/eclipse-tycho/tycho/tree/master/demo/p2-maven-site
> 
> Beside that I think Eclipse/IBuild/SDK/Simrel/... will still include PDE anyways one way or another.

So if SDK/I-builds want to include PDE without building it, you propose adding it as maven dependency. Can we say "latest" in maven deps somehow as I don't envision editing target file to pick latest PDE?

laeubi:
> So if SDK/I-builds want to include PDE without building it, you propose adding it as maven dependency

Yes

> Can we say "latest" in maven deps somehow

Yes but it is [deprecated](https://www.baeldung.com/maven-dependency-latest-version#1-deprecated-syntax).

>  I don't envision editing target file to pick latest PDE?

In the end PDE becomes a third party dep and we do updates for these as well, at best there the just one feature mentioned and one might even use a version range there, that especially the things we need to explore here how it works out and where we need better support or additional features to support such use-cases, but I strongly believe that platform should be leading here even though I know others are already using such setups.

laeubi:
by the way I created https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3993 as it seems PDE is suffering build executors...

vogella:
I fully support this approach, moving to a pure Maven build is also what a lot of my clients desire and if we use this approach it will help to stabilise the process. P2 update sites are often considered as "extra" or "Eclipse only" technologies by my clients which normally also have a lot of "regular" Maven projects and Maven knowledge.

laeubi:
It seems some plugins use an older deploy plugin I have no clue why... will try to move this up otherwhise have to wait until master opens and maybe define it in the parent...

akurtakov:
> It seems some plugins use an older deploy plugin I have no clue why... will try to move this up otherwhise have to wait until master opens and maybe define it in the parent...

Please put it in the parent now. It's better to push such change now rather than multiplying it.

mickaelistria:
I also fully support deploying to Maven.
Just to make sure, are p2 metdata published with the artifacts? I think they must be published in general to handle cases where some p2.inf matters in the output installation.

mickaelistria:
>  I think they must be published in general to handle cases where some p2.inf matters in the output installation.

Actually, if the p2.inf is included in the artifact and is honored later by the PDE target resolution, and by Tycho target resolution and by p2-maven-site, there seems to be no need to store them in Maven repo as they can be inferred later.
@laeubi Can you confirm that embedded p2.inf are honored by those consumers?

mickaelistria:
> p2.inf is only considered at build time but not later on

IIRC, if the p2.inf is embedded at the right place in the artifact, the p2 publisher would always process it. So if client use the publisher, sharing the p2.inf is enough as every client should then regenerate the same p2 metadata.
But anyway, whatever works: if p2 metadata are published for this Maven artifacts directly (instead of assuming clients will regenerate them properly), that's good too.

laeubi:
> IIRC, if the p2.inf is embedded at the right place in the artifact, the p2 publisher would always process it. So if client use the publisher, sharing the p2.inf is enough as every client should then regenerate the same p2 metadata.

That's possible at least Tycho do not do any special processing if used as maven artifact in a target everything is delegated to P2 publisher. But p2.inf is not the only source in a Tycho build for p2 metadata.

> But anyway, whatever works: if p2 metadata are published for this Maven artifacts directly (instead of assuming clients will regenerate them properly), that's good too.

This is one thing we should/can check then as well if Tycho also would download such data and use it later on.

vogella:
Maybe we could we also directly publish to repo.eclipse.org?

PR #937, opened 2023-11-27
Remove blank Javadoc

jukzi:
This commit cleans up Javadoc that does not add information. It resolves ecj warnings:
 `Javadoc: Description expected after ...`
It helps to prevent future empty javadoc by disabling missingJavaDoc warnings. This resolves
 `Javadoc: Missing ...`

The modification is a result of regular expression search&replace:

in files `*.java`
 `^[\s]*\*[\s]*(@return|@param[\s]*[^\s]+|@throws[\s]*[^\s]+)\R([\s]*\*[\s]*@|[\s]*\*/\R)`
 ->`$2`
 `^([\s]*\*[\s]*\R)([\s]*\*/\R)`
 ->`$2`
 `^[\S\t ]*/\*\*\R[\s]*\*/\R`
 ->``

in files `org.eclipse.jdt.core.prefs`
 `org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc(Comments|Tags)\=[^\s]*`
 ->`org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc$1\=ignore`

excluded resources:
 apitools/org.eclipse.pde.api.tools.tests/test-builder
 apitools/org.eclipse.pde.api.tools.tests/test-classes
 apitools/org.eclipse.pde.api.tools.tests/test-source
 apitools/org.eclipse.pde.api.tools.tests/tests-deltas

This reverts commit 38481a9a389f985367b3a32c7439b97dc93f4371.


HannesWell (approved):
Thank you, it is great to have this eventually cleaned up. Especially cleaning up the templates hopefully help to avoid further spread of this bad pattern of empty/useless doc.

I went through all the changes and as far as I can say it, they look all good.
There are still some parameters left that only have a single word as description and are therefore probably useless. But I don't think it is worth the time to inspect them all.

@jukzi please merge this as soon as the master branch opens again.
But I assume once the baseline is set to the next release you will need to bump a lot of versions.
I think it is simple to prepare this now and just bump the service version of every affected plugin now. If there are then other bumps in the meantime potential conflicts are faster to resolve than letting waiting for the CI to fail for each project.

HannesWell:
> But I assume once the baseline is set to the next release you will need to bump a lot of versions.
> I think it is simple to prepare this now and just bump the service version of every affected plugin now. If there are then other bumps in the meantime potential issues are faster to resolve than letting waiting for the CI to fail for each project.

Looks like this already happened. In order to also contribute a bit to this I created the version bumps in a separate commit before the javadoc clean-up. Locally the build passed with that. So with that it should be ready.
@jukzi Please submit this as soon as the master branch opens, which I hope happens very soon.

jukzi:
> I created the version bumps

@HannesWell Thanks, i appreciate your work!!

PR #940, opened 2023-11-27
Support sub bundles build

laeubi:
Currently PDE only builds the "main bundle" but not sub-bundles, this adds support for building these as well.

### DIFF
--- a/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.17.300.qualifier
+Bundle-Version: 3.17.400.qualifier
 Bundle-Activator: org.eclipse.pde.internal.core.PDECore
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/BndBuilder.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/BndBuilder.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.core.bnd;
 
+import java.io.File;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -42,6 +44,8 @@ import org.eclipse.pde.internal.core.project.PDEProject;
 
 import aQute.bnd.build.Project;
 import aQute.bnd.build.ProjectBuilder;
+import aQute.bnd.osgi.Builder;
+import aQute.bnd.osgi.Jar;
 
 public class BndBuilder extends IncrementalProjectBuilder {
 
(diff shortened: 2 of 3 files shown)


ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.17.300.qualifier
+Bundle-Version: 3.17.400.qualifier
```
iloveeclipse, line 5: Why have you changed it? This causes now API error because it was already changes for the 4.31 release.

Please consider quickly revert this before next IBuild kicks in.

iloveeclipse: I've set https://github.com/eclipse-pde/eclipse.pde/pull/948

laeubi: > Why have you changed it? This causes now API error because it was already changes for the 4.31 release.

Why it not failed this PR?

akurtakov: I guess this is yet another API tools mystery.

iloveeclipse: There was another issue with API error not being reported, see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/914 .
Note that https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/914 is reported in SDK build and in the IDE.
So it could be some new tycho build issue.

laeubi: > There was another issue with API error not being reported

The green build reports it as a **warning**

> [2023-11-29T07:02:13.259Z] [WARNING] [API WARNING] File MANIFEST.MF at line 5: The service version is increased unnecessarily since either the major or minor or service version is already increased (location: /home/jenkins/agent/workspace/eclipse.pde_PR-940/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF)

> So it could be some new tycho build issue.

API tools is provided by PDE not Tycho, Tycho just provides a wrapper around it, so either it is configured differently `WARNING` vs `ERROR` or the configuration is not picket up by Tycho @vik-chand might better know if/how/where this is to be configured.

iloveeclipse: > Tycho just provides a wrapper around it

That is what I've meant with SDK/IDE build showing error.

I have no idea if it is the wrapper or the tycho config or whatever else. 
API tooling in the IDE works.

laeubi: I have now debugged the case and the problem is that `ApiProblem.getSeverity()` **always** return the default as it passes `null` as a project... Why it works in IDE/SDK ist because here Error Markers are used and when ApiTools converts `ApiProblem` > `Error Marker` it does _IGNORE_ `ApiProblem.getSeverity()` and instead computes the value again but this time using current build project... will replicate this behaviour with Tycho as well ...

laeubi: Actually ApiProblem.getSeverity() is only used in two places:

1. ApiProblem.toString (what will print wrong value then or at least ignores the project settings)
2. In a `org.eclipse.pde.api.tools.problems.tests.ApiProblemTests.testGetSeverity() `

so actually I think that method can completely be deleted to prevent further confusion...

PR #945, opened 2023-11-28
Add ignore pattern for the javadoc parts of the doc bundle

laeubi:
As the javadoc changes every build because of timestamp we can not reliable baseline compare it.


akurtakov:
Jenkins succeeded and honestly I'm not totally sure what's needed here so I trust @laeubi .

PR #952, opened 2023-11-29
Add missing attributes

laeubi:


ui/org.eclipse.pde.ui/plugin.xml
```diff
@@ -2536,7 +2537,7 @@
          name="Bnd Run Descriptor Location">
       <targetLocation
             locationFactory="org.eclipse.pde.internal.ui.bndtools.RunDescriptorTargetLocation$Factory"
-            type="BndRunDescriptorLocation">
+            type="BndRunDescriptorLocation" canUpdate="false">
```
iloveeclipse, line 2540: Would be nice to have consistent formatting in same PR.

Issue #874, opened 2023-11-03
Unable to resolve target definition when non-IU locations are used


ptziegler:
As a follow-up on #867 and based on the example in #675.

Using the nightly build, the following target-platform definition is able to resolve without problems (i.e. the p2 Guava bundle can use the Maven bundle to resolve its dependencies).

```
<target name="linuxtools" sequenceNumber="0">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/linuxtools/update-docker-5.12.0/"/>
			<unit id="com.google.guava" version="0.0.0"/>
		</location>

		<location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" label="DirectFromMaven" missingManifest="error" type="Maven">
			<dependencies>
				<dependency>
					<groupId>com.google.guava</groupId>
					<artifactId>failureaccess</artifactId>
					<version>1.0.1</version>
					<type>jar</type>
				</dependency>
			</dependencies>
		</location>
	</locations>
</target>
```

The following does **not** work (with FailureAccess 1.0.2 instead of 1.0.1):

```
<target name="linuxtools" sequenceNumber="0">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/linuxtools/update-docker-5.12.0/"/>
			<unit id="com.google.guava" version="0.0.0"/>
		</location>

		<location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" label="DirectFromMaven" missingManifest="error" type="Maven">
			<dependencies>
				<dependency>
					<groupId>com.google.guava</groupId>
					<artifactId>failureaccess</artifactId>
					<version>1.0.2</version>
					<type>jar</type>
				</dependency>
			</dependencies>
		</location>
	</locations>
</target>
```

Setting the Guava version doesn't help, either (with Guava 32.1.2.jre and FailureAccess 1.0.2):

```
<target name="linuxtools" sequenceNumber="0">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/linuxtools/update-docker-5.12.0/"/>
[...]

PR #957, opened 2023-11-30
Create reproducer test case for issue #874

ptziegler:
Consider the following situation: The SimRel 2023-12 repository contains FasterXML Jackson bundles with version 2.16.0, with the exception of one bundle that is still in 2.15.3.

In order to ensure that all those bundles have the same version, the one in 2.15.3 is also contributed via Maven Central (imitated by a local directory) in 2.16.0.

Trying to resolve this target definition fails with: No repository found containing:
  osgi.bundle,
  com.fasterxml.jackson.module.jackson-module-jaxb-annotations,
  2.16.0

Meaning PDE knows that there is a newer version available, it just doesn't know where to find it...
Note that one can work around this problem by explicitly adding the missing bundle to the IU location.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/WizardElement.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/WizardElement.java
@@ -65,6 +65,26 @@ public class WizardElement extends NamedElement implements IPluginContribution {
 		return description;
 	}
 
+	public Boolean getFlag(String name) {
+		if (configurationElement != null) {
+			String value = configurationElement.getAttribute(name);
+			if (value != null) {
+				return Boolean.valueOf(value.equalsIgnoreCase("true")); //$NON-NLS-1$
+			}
+		}
+		return null;
+	}
+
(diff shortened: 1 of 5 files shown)


ptziegler:
> Cannot invoke "org.eclipse.equinox.internal.p2.repository.Transport.download(java.net.URI, java.io.OutputStream, org.eclipse.core.runtime.IProgressMonitor)" because the return value of "org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.getTransport()" is null

Of course the test doesn't work during the Jenkins run... :angry:

laeubi:
> Of course the test doesn't work during the Jenkins run... 😠 it could you need an additional test requirement for the ECF transport.

laeubi:
@ptziegler this can happen if if you push a commit that is even with master (e.g. you did not push any changes).

> Several test are disabled during the Jenkins build via PDETestCase.assumeRunningInStandaloneEclipseSDK() for similar reasons. So I assume it's simply not possible to connect to a "real" repository during the test.

Actually it should be possible, maybe we should rethink that but of course it could slow down the build a bit.

Issue #958, opened 2023-12-01
Project 'org.eclipse.pde.ds.tck' is missing required source folder: 'tck/OSGI-INF/impl-src'


jukzi:
When settting up a oomphed workspace it results in error:
![image](https://github.com/eclipse-pde/eclipse.pde/assets/51790620/62a86efe-5a08-4374-8bdd-cd26bfa1bd75)

jukzi:
Oomphing workspace did work a month ago

merks:
I'll await @laeubi before committing such a change...

merks:
I might as well create a PR, so if this is good like this, we can get it checked in faster...

PR #959, opened 2023-12-01
Create /org.eclipse.pde.ds.tck/tck/OSGI-INF/impl-src/.gitignore

merks:
Move it from /org.eclipse.pde.ds.tck/tck/.gitignore to ensure that the source folder always exists and that there are no errors about a non-existing folder on the build path.

https://github.com/eclipse-pde/eclipse.pde/issues/958


merks:
@laeubi  This seems to do the trick.

image

laeubi (approved):

laeubi:
Thanks for taking care!

PR #962, opened 2023-12-01
Add non-IU artifact repositories to the planner resolution #874

ptziegler:
Given that units from non-IU locations are considered when resolving IUs, it may happen that units are contained in the resolution path, which are not contained by any repository. This may break previously functional target definitions.

To resolve this problem, an in-memory artifact and metadata repository is created, based on the metadata of the non-IU locations. Those repositories are then added to the provisioning context for further processing.

Amends dd321b56a92b8e86f98de6f0dae25fbc43e4f825


laeubi (commented):
I really like this approach, just a few comments and Jenkins reports some compiler warnings (even though they are not displayed correctly here).

laeubi:
@jukzi as you are recently concerned about warnings, this is an example where one can see new warnings reported by Jenkins:
grafik
it is also shown in the code review (even though for some reason missing the message)
grafik
and the overview
grafik


@ptziegler sorry for the offtopic :-)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/FileArtifactRepository.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	@Override
+	public IQueryResult<IArtifactKey> query(IQuery<IArtifactKey> query, IProgressMonitor monitor) {
+		return new QueryResult<>(artifacts.keySet().stream().map(IArtifactDescriptor::getArtifactKey).iterator());
+	}
+
+	@Override
+	public boolean contains(IArtifactDescriptor descriptor) {
+		return artifacts.containsValue(descriptor);
```
jukzi, line 143: solve all warnings, please

jukzi (dismissed):

laeubi:
> @laeubi could we fail the verification if new warnings are introduced?

Yes we can
- https://github.com/eclipse-pde/eclipse.pde/pull/964

ptziegler:
I've cleaned up the FileArtifactRepository and so far, everything should work. In case of artifical features, I create an in-memory jar only containing the `feature.xml`. Otherwise the location URL/path is used.

Note that I've renamed the repositories from File*Repository to Virtual*Repository, given that they don't work on files anymore, so their name makes little sense anymore.

What I still need to figure out is how to properly inject the repositories into the provisioning context.
Reflection works, but it's not the correct solution...

I probably need proper setter methods... given that the indented extension points are not an option.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/FileArtifactRepository.java
```diff
@@ -0,0 +1,179 @@
...
+	private static final String DESCRIPTION = """
+			In-Memory repository created for a single Non-IU repository, used
+			during the planner resolution of a real IU repository, in order to
+			grant access to external, installable units.
+			""";
+	private static final String SCHEME = "memory://"; //$NON-NLS-1$
+	private final Map<IArtifactDescriptor, File> artifacts = new HashMap<>();
+
+	public FileArtifactRepository(IProvisioningAgent agent, ITargetLocation targetLocation) {
+		super(agent, NAME, null, null, URI.create(SCHEME + UUID.randomUUID()), DESCRIPTION, null, null);
```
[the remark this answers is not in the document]
ptziegler: The type should probably be the 3rd argument, rather than a part of the name. The rest I've included.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/FileArtifactRepository.java
```diff
@@ -0,0 +1,179 @@
...
+			File featureFile = new File(featureLocation);
+			if (!featureFile.exists()) {
+				PDECore.log(Status.warning("Feature file doesn't exist: " + featureFile));
+				continue;
+			}
+			IArtifactKey artifactKey = FeaturesAction.createFeatureArtifactKey(targetFeature.getId(),
+					targetFeature.getVersion());
+			IArtifactDescriptor artifactDesriptor = new ArtifactDescriptor(artifactKey);
+			if (artifacts.containsKey(artifactDesriptor)) {
+				PDECore.log(Status.warning("Artifact already exists: " + artifactDesriptor));
```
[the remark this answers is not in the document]
ptziegler: Then I can probably get rid of the if-check entirely...

laeubi:
> What I still need to figure out is how to properly inject the repositories into the provisioning context.
Reflection works, but it's not the correct solution...

Extending the Provisiong context like you did seems a proper soloution to me...

ptziegler:
> > What I still need to figure out is how to properly inject the repositories into the provisioning context.
> > Reflection works, but it's not the correct solution...
> 
> Extending the Provisiong context like you did seems a proper soloution to me...

Overwriting the method probably works, then, but certainly not the way it's currently done: :P 
```
Field f = oldQueryable.getClass().getDeclaredField("repositories");
f.setAccessible(true);
artifactRepositories.addAll((List<IArtifactRepository>) f.get(oldQueryable));
```

ptziegler:
> > > What I still need to figure out is how to properly inject the repositories into the provisioning context.
> > > Reflection works, but it's not the correct solution...
> > 
> > 
> > Extending the Provisiong context like you did seems a proper soloution to me...
> 
> Overwriting the method probably works, then, but certainly not the way it's currently done: :P
> 
> ```
> Field f = oldQueryable.getClass().getDeclaredField("repositories");
> f.setAccessible(true);
> artifactRepositories.addAll((List<IArtifactRepository>) f.get(oldQueryable));
> ```

Found the solution already: I can simply use a CompoundQueryable

laeubi:
```
ProvisioningContext context = new ProvisioningContext(agent) {
	@Override
	public IQueryable<IArtifactRepository> getArtifactRepositories(IProgressMonitor monitor) {
	 	... enhance here ...
	}

	public IQueryable<IInstallableUnit> getMetadata(IProgressMonitor monitor) {
	 	... enhance here ...
	}
};
```

laeubi:
@ptziegler there is even `org.eclipse.equinox.p2.query.QueryUtil.compoundQueryable(IQueryable<T>, IQueryable<T>)`

merks:
@ptziegler 

I’m impressed by your resourcefulness!  🏆

ptziegler:
The usage of reflection has also been removed. That should be it, from my side.

ptziegler:
The [new](https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-962/5/maven-warnings/new/type.-1386543368/type.-1386543368/source.d5e9e1a6-371d-40f3-9247-f338aa9170c1/#13441) warning is unrelated to this change.
```
[INFO] --- tycho-p2-extras:4.0.5-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.pde.unittest.junit ---
[WARNING] Parameter 'comparator' is read-only, must not be used in configuration
```

laeubi (approved):
Looks good now, many thanks for your efforts in improving/fixing this.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,215 @@
...
+		}
+		try (InputStream is = getArtifactInputStream(artifactModel)) {
+			is.transferTo(destination);
+			return Status.OK_STATUS;
+		} catch (Exception e) {
+			return Status.error(e.getLocalizedMessage(), e);
+		}
+	}
+
+	private InputStream getArtifactInputStream(Object artifactModel) throws Exception {
```
HannesWell, line 123: This method would probably be simple, if in the caller to destination `OutputStream` would just be passed as an argument and if in this method the artifact would be written directly to the output stream.
For example in the last case, this would avoid creating an `InputStream` where the content is first written to an ByteArrayOutputStream.
Or in case there is a `featureJar` `File` you could just use `Files.copy(featureJar.toPath, destition)` to copy the file content to the output stream.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,215 @@
...
+
+	@Override
+	public IStatus getArtifacts(IArtifactRequest[] requests, IProgressMonitor monitor) {
+		MultiStatus multiStatus = new MultiStatus(getClass(), IStatus.INFO, "Perform Artifact Requests"); //$NON-NLS-1$
+		SubMonitor subMonitor = SubMonitor.convert(monitor, requests.length);
+		boolean isOk = true;
+		for (IArtifactRequest request : requests) {
+			request.perform(this, subMonitor.split(1));
+			multiStatus.add(request.getResult());
+			isOk &= request.getResult().isOK();
```
HannesWell, line 205: `MultiStatus` already accumulates the severity of all children as its own severity.
So you can just check below if `multiStatus.isOK()` in the end or just return the `multiStatus` as it is without the need collect its OK-status by yourself.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,215 @@
...
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.WriteResource;
+
+/**
+ * In-Memory representation of a artifact repository based on a non-IU target
+ * location. This repository is used during the planner resolution of an IU
+ * target location to supply artifacts from other (non-IU) locations.
+ */
+@SuppressWarnings("restriction")
+public class VirtualArtifactRepository extends AbstractArtifactRepository {
```
HannesWell, line 67: Have you checked if extending `SimpleArtifactRepository` would save some of the relatively simple implementations?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
```diff
@@ -327,9 +377,58 @@ public void testExternalModelManagerPreferences() throws Exception {
...
+		for (File jarFile : features.listFiles()) {
+			if (jarFile.isDirectory()) {
+				continue;
+			}
+			String jarFileName = jarFile.getName();
+			String jarFileBaseName = jarFileName.substring(0, jarFileName.length() - ".jar".length());
+
+			File outDir = new File(features, jarFileBaseName);
+			outDir.mkdir();
+			outDir.deleteOnExit();
```
HannesWell, line 397: I don't think this will be able to delete entire directories with their content.
If you need a temparary directory you can use JUnit's TemporaryFolder, which is reliably cleared depending on if it is a `@Rule` or `@ClassRule`.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
```diff
@@ -327,9 +377,58 @@ public void testExternalModelManagerPreferences() throws Exception {
...
+					File outFile = new File(outDir, jarEntry.getName());
+					outFile.deleteOnExit();
+					if (outFile.isDirectory()) {
+						outFile.mkdir();
+						return;
+					}
+					try (InputStream is = jar.getInputStream(jarEntry)) {
+						try (OutputStream os = new FileOutputStream(outFile)) {
+							is.transferTo(os);
+						}
```
HannesWell, line 410:
```suggestion
						Files.copy(is, outFile.toPath())
```
With that it is maybe more suitable to use `Path` as type of outFile.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
```diff
@@ -327,9 +377,58 @@ public void testExternalModelManagerPreferences() throws Exception {
...
+					} catch (IOException e) {
+						e.printStackTrace();
+						fail(e.getMessage());
```
HannesWell, line 413: Doesn't fail provide an overload, where you can pass an exception?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
```diff
@@ -327,9 +377,58 @@ public void testExternalModelManagerPreferences() throws Exception {
...
+				continue;
+			}
+			String jarFileName = jarFile.getName();
+			String jarFileBaseName = jarFileName.substring(0, jarFileName.length() - ".jar".length());
+
+			File outDir = new File(features, jarFileBaseName);
+			outDir.mkdir();
+			outDir.deleteOnExit();
+
+			try (JarFile jar = new JarFile(jarFile)) {
```
HannesWell, line 399: This entire block just extract the jar into a temporary directory, doesn't it?
I'm relatively sure, that there are (probably multiple) implementations of this functionality in some File-Utility classes in PDE (core) that can be reused here.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1078,11 +1078,24 @@ private void resolveWithPlanner(ITargetDefinition target, IProgressMonitor monit
...
 
-		ProvisioningContext context = new ProvisioningContext(getAgent());
+		List<IArtifactRepository> extraArtifactRepositories = new ArrayList<>();
+		List<IMetadataRepository> extraMetadataRepositories = new ArrayList<>();
+		getAdditionalProvisionIUs(target, extraArtifactRepositories, extraMetadataRepositories);
+		ProvisioningContext context = new ProvisioningContext(getAgent()) {
+			@Override
+			public IQueryable<IArtifactRepository> getArtifactRepositories(IProgressMonitor monitor) {
+				return QueryUtil.compoundQueryable(super.getArtifactRepositories(monitor),
+						(query, ignore) -> query.perform(extraArtifactRepositories.iterator()));
```
HannesWell, line 1088:
```suggestion
						QueryUtil.compoundQueryable(extraArtifactRepositories));
```
Can't we use QueryUtil.compoundQueryable() for a list of artifact repositories, like it is done for metadata-repositories?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,215 @@
...
+	private static final String NAME = "Non-IU Artifact Repository @ "; //$NON-NLS-1$
+	private static final String DESCRIPTION = """
+			In-Memory repository created for a single Non-IU repository, used
+			during the planner resolution of a real IU repository.
+			"""; // $NON-NLS-1$ //$NON-NLS-1$
```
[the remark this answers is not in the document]
ptziegler: I'm almost certain that's a JDT bug when using text blocks. When using only a single `// $NON-NLS-1$`, I still get the "Non-externalized string" warning. I'll check it again with the latest nightly-build tomorrow and create a separate issue, if necessary.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
ptziegler: It only does in [JUnit5](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#fail-java.lang.Throwable-), not in [JUnit4](https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#fail(java.lang.String)).

The catch-block is only necessary because of the lambda-statement. Perhaps I can rewrite this part to use a simple for-each loop... Otherwise it's probably better to simply wrap the IOException in an unchecked exception. Just so that the stack trace is preserved, in case this operation _does_ fail.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
laeubi: @HannesWell I think `SimpleArtifactRepository` is not really suitable as it adds to much  additional things here to be reusable, instead one might think as a follow up on this to make `AbstractArtifactRepository` implement more methods for convenience...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
laeubi: I think this is a common pattern in eclipse/equinox because the multi status can contain further messages. Also the Multistatus usually contains a message that indicates the error / action that failed so it could eb confusing to get a status that is okay but has a message that indicates an error...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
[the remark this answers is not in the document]
ptziegler: and done~

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1553,8 +1566,9 @@ private Collection<URI> getMetadataRepositories(ITargetDefinition target) throws
 		return result;
 	}
 
-	private List<IInstallableUnit> getAdditionalProvisionIUs(ITargetDefinition target) throws CoreException {
-		List<IInstallableUnit> result = new ArrayList<>();
+	private void getAdditionalProvisionIUs(ITargetDefinition target,
```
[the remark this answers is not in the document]
ptziegler: done~

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
ptziegler: It's sadly not that simple :/
Even with the temporary folder, I then still need to copy the update site to the test folder. In the end, I simply test against the update site in `modified-jdt-features.zip`, where the features are already unpacked.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
ptziegler: Obsolete, since the entire block is gone.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/IUBundleContainerTests.java
ptziegler: Also obsolete

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
HannesWell: Ok, that's fine, thanks for checking.
Adding more defaults sounds good.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,210 @@
...
+			}
+			try (InputStream is = location.toURL().openStream()) {
+				is.transferTo(destination);
+			}
+		} else if (artifactModel instanceof IFeatureModel featureModel) {
+			String installLocation = featureModel.getInstallLocation();
+			if (installLocation != null) {
+				File featureJar = new File(installLocation);
+				if (featureJar.isFile()) {
+					try (InputStream is = new FileInputStream(featureJar)) {
```
HannesWell, line 135: `Files.copy(featureJar.toPath(), destination)` would be simpler here.

HannesWell (commented):
Thank you for the updates.
I agree with all answers and just have one small suggestion to further simplify the copy.

With that being applied this is a perfectly fine change and also ready from my side.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,215 @@
...
+			ByteArrayOutputStream featureContent = new ByteArrayOutputStream();
+			try (PrintWriter writer = new PrintWriter(featureContent)) {
+				feature.write("", writer); //$NON-NLS-1$
+			}
+			// Generate in-memory feature jar (with only the feature.xml)
+			try (Jar jar = new Jar(feature.getId() + '_' + feature.getVersion())) {
+				jar.putResource("feature.xml", new WriteResource() { //$NON-NLS-1$
+					@Override
+					public void write(OutputStream out) throws Exception {
+						out.write(featureContent.toByteArray());
```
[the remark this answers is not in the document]
HannesWell: Good that you noticed this👍🏽

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
HannesWell: No Problem. 😄

laeubi:
I just wanted to note that the issue with two new maven warnings **can be ignored** if build/test is otherwise fine! I'm try to get rid of the warnings but its not that easy at least one of them should be fixed with this:
- https://github.com/eclipse-tycho/tycho/pull/3117

but will need probably 1 hour unless it show up.

ptziegler:
Hm... I checked the generated feature.jar and now it's empty again...

image

There's clearly something about the BND that I'm doing wrong... **argh**

ptziegler:
> > There's clearly something about the BND tool that I'm doing wrong... argh
> 
> Can't you just use the standard java `Jar/ZipOutputstream` or `Jar/ZipFile` instead of BND's `Jar`?

That's what I'm really tempted to do at this point. Debugging suggests the output-stream is never closed, but given that the BND class is 100% undocumented, I can't say whether that's by intent or what to do about it.

HannesWell:
> > > There's clearly something about the BND tool that I'm doing wrong... argh
> > 
> > Can't you just use the standard java `Jar/ZipOutputstream` or `Jar/ZipFile` instead of BND's `Jar`?
> 
> That's what I'm really tempted to do at this point. Debugging suggests the output-stream is never closed, but given that the BND class is 100% undocumented, I can't say whether that's by intent or what to do about it.

Please go for it. In general using standard Java instead of custom solutions should be preferred if the solutions are equivalent.

In general at the point when transfering the feature model to the destination OutputStrram what is the intended output? Should it just be the `feature.xml` or should it be in a jar?

HannesWell:
> In general at the point when transfering the feature model to the destination OutputStrram what is the intended output? Should it just be the feature.xml or should it be in a jar?

Looking at the code again, it is probably indeed a jar necessary.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,202 @@
...
+					return;
+				}
+			}
+			IFeature feature = featureModel.getFeature();
+			// Generate in-memory feature jar (with only the feature.xml)
+			Manifest manifest = new Manifest();
+			manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0"); //$NON-NLS-1$
+			try (JarOutputStream jos = new JarOutputStream(destination, manifest)) {
+				jos.putNextEntry(new JarEntry("feature.xml")); //$NON-NLS-1$
+				feature.write("", new PrintWriter(jos, true)); //$NON-NLS-1$
```
laeubi, line 146: Maybe I'm a bit paranoid, but as PrintWirter is buffered I always would suggest to flush its content or include it inside a try/resource before closing the entry / surrounding stream.

ptziegler: Closing the PrintWriter/JarOutputStream is dangerous, because it also closes the "destination" stream. Which is why I've removed the outer try-with block.

But explicitly flushing the streams sounds safe enough.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,207 @@
...
+				File featureJar = new File(installLocation);
+				if (featureJar.isFile()) {
+					Files.copy(featureJar.toPath(), destination);
+					return;
+				}
+			}
+			IFeature feature = featureModel.getFeature();
+			// Generate in-memory feature jar (with only the feature.xml)
+			Manifest manifest = new Manifest();
+			manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0"); //$NON-NLS-1$
```
HannesWell, line 143: I wonder if PDE or P2 considers the MANIFEST.MF or a feature-jar at any time and therefore if this is necessary?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,207 @@
...
+
+			JarOutputStream jos = new JarOutputStream(destination, manifest);
+			jos.putNextEntry(new JarEntry("feature.xml")); //$NON-NLS-1$
+			PrintWriter printWriter = new PrintWriter(jos);
+			feature.write("", printWriter); //$NON-NLS-1$
+			printWriter.flush();
+			jos.closeEntry();
+
+			jos.flush();
```
HannesWell, line 152: Looking at the implementation of `flush()` I got the impression that it is not necessary to call `jos.flush()`.
Furthermore maybe calling `jos.finish()` will also implicitly close the only entry and thus `jos.closeEntry()` might be not necessary as well.  At least `putNextEntry()` implicitly closes a previously opened entry.
But I cannot tell with confidence and would need to be checked.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
ptziegler: PDE usually only ever creates a default Manifest for features, so probably not... Perhaps there's a requirement when the features are signed, but that's not something that applies here.
So I guess it can go?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
ptziegler: The call to `closeEntry()` is indeed redundant. In hindsight, I would also expect the stream to write any buffered data, when `finish()` is called. So I assume that's also not strictly necessary.

```
public void finish() throws IOException {
	ensureOpen();
	if (finished) {
		return;
	}
	if (current != null) {
		closeEntry();
	}
	// write central directory
	long off = written;
	for (XEntry xentry : xentries)
		writeCEN(xentry);
	writeEND(off, written - off);
	finished = true;
}
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -0,0 +1,199 @@
...
+			if (installLocation != null) {
+				File featureJar = new File(installLocation);
+				if (featureJar.isFile()) {
+					Files.copy(featureJar.toPath(), destination);
+					return;
+				}
+			}
+			IFeature feature = featureModel.getFeature();
+			// Generate in-memory feature jar (with only the feature.xml)
+			JarOutputStream jos = new JarOutputStream(destination);
```
laeubi, line 140: What happend to the try/with resources here?

ptziegler: Closing the JarOutputStream also closes the underlying `destination` stream. Given that this stream is managed by the caller, it mustn't be closed by this method.

laeubi: you are right, what a mess :-D

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
HannesWell: > PDE usually only ever creates a default Manifest for features, so probably not... Perhaps there's a requirement when the features are signed, but that's not something that applies here. So I guess it can go?

Signing is indeed the only use-case where a Manifest is necessary, I'm not aware of any other one.
So yes I think it can go.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
HannesWell: > The call to `closeEntry()` is indeed redundant. In hindsight, I would also expect the stream to write any buffered data, when `finish()` is called. So I assume `flush()` also not strictly necessary.

Yes, especially since the destination does not have to be flushed (at least here, if it is necessary it is the responsibility of the caller).

HannesWell (approved):
Looks good now.
Thanks again.

PR #964, opened 2023-12-02
Enable the Quality Gate

laeubi:
FYI @jukzi @vogella @akurtakov


HannesWell (approved):
Yes please.

jukzi:
lets try

vogella:
Hammer! Please also add this check to platform repos. Finally an automatic way to prevent warning creep

laeubi:
> Hammer! Please also add this check to platform repos. 

As usually from then on each and every inconvenience with failing builds due to such quality checks are blamed on the messenger (aka committer of the config change) its unlikely I'll roll this out to the large, but the change is simple enough for anyone that finds it important enough for their project they fell confident enough to apply it anytime.

PR #971, opened 2023-12-03
Fix handling of class types and empty array in component property types

laeubi:
Currently for a class used in component property types the property looks like MyClass.class where according to the spec it should be a string holding the class name, also empty arrays result in a property with no value in the xml while the spec says it should be omitted,

This is now fixed, empty properties are only removed and Types are resolved to their type name from now on.

See also https://github.com/osgi/osgi/issues/646


laeubi:
I build started: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/19/

Issue #972, opened 2023-12-03
OSGi DS version 1.5 not selectable in PDE DS Annotation preferences


HannesWell:
In the PDE DS Annotation processing preference the entry for spec version 1.4 is duplicated and version 1.5 is not selectable:

![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/44067969/76e840aa-939f-420b-8895-310a56426156)

PR #974, opened 2023-12-03
Fix wrong constant in drop down

laeubi:
Fix https://github.com/eclipse-pde/eclipse.pde/issues/972


HannesWell:
Thanks.
Btw. since spec version 1.5 is not supported in the IDE, I think it would be good to lable it with something like `1.5 (not supported in IDE)` to make the user aware of it.

laeubi:
There is already a warning showing up, and it is supported but not all features of it :-)

PR #976, opened 2023-12-03
Add support for make jar as part of bnd build and adjust converter

laeubi:


laeubi:
Maven warnings can currently not be prevented, we are working on it ...

PR #979, opened 2023-12-04
Add support for bnd classpath as a replacement for jars.extra.classpath

laeubi:


laeubi:
Only quality gate for maven failed.

PR #933, opened 2023-11-24
Additions to PR#927

laeubi:
Additions to the proposed solution presented here:
- https://github.com/eclipse-pde/eclipse.pde/pull/927

this currently adds the equinox-executable feature so the test can find the "launcher" artifacts.

Issue #967, opened 2023-12-02
120 P2Tests fail on master since I20231201-1800


iloveeclipse:
* https://download.eclipse.org/eclipse/downloads/drops4/I20231202-0500/testresults/html/org.eclipse.pde.build.tests_ep431I-unit-macM1-java17_macosx.cocoa.aarch64_17.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20231202-0500/testResults.php

jukzi:
@laeubi may that be caused by recent PDE changes?

iloveeclipse:
> no one really wants to maintain them and target an unmaintained/deprecated component we probably should simply delete them

Sorry, this is not true in both assumptions. The functionality is there, it is used and it is not deprecated, there are people interesting in maintaining it.

But.

Every time release starts, lot of PR's are merged and usually no one cares about broken tests in SDK because most are only interested in pushing their own code and don’t care about SDK as a whole. This is wrong attitude.

There were changes on the pde and p2 and platform. Many of them could break pde.

People who merged into pde and/or p2 or platform should **at least** check if their changes were not breaking existing tests. Everyone can see current SDK nightly build results. It's not a big deal to check them on the next day after merging PR and ask "hmm, where all the test fails are coming from? Could it be my PR?"

One can't simply expect someone (like me) will always come and cleanup the mess afterwards. Honestly, I'm tired hunting regressions all over SDK code base.

laeubi:
> Sorry, this is not true in both assumptions. The functionality is there, it is used and it is not deprecated, there are people interesting in maintaining it.

See https://github.com/eclipse-pde/eclipse.pde/blob/master/build/README.md

But.

Don't take me to serious in that regard, I even take some attempts to re enable the tests
- https://github.com/eclipse-pde/eclipse.pde/pull/933
- https://github.com/eclipse-pde/eclipse.pde/pull/927

but this is currently not my main focus, so if someone has an idea and likes to help out its always welcome...

> One can't simply expect someone (like me) will always come and cleanup the mess afterwards.

No one expects that from you, its great you seem to be dedicated to this, thats why I'm always pushing forward things are detected in place when the change happens, if it shows up sometimes after, let it be warnings, javadoc errors test problems, its actually all to late given the low resources split over a large codebase...

HannesWell:
[...]

PR #980, opened 2023-12-04
Run some pde.build tests with Tycho (second attempt)

laeubi:
This is a fresh attempt based on

- https://github.com/eclipse-pde/eclipse.pde/pull/927

with some progress made here
- https://github.com/eclipse-pde/eclipse.pde/pull/933

but that was a dead end I think instead one should try to add a `category.xml` with another execution that mirrors into the "deltapack folder" so if the test executes in `target/work` the goal would be to have a folder `target/deltapack/eclipse` that mirrors everything needed as the "deltapack".

### DIFF
--- a/build/org.eclipse.pde.build.tests/pom.xml
+++ b/build/org.eclipse.pde.build.tests/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.eclipse.pde</groupId>
+		<artifactId>eclipse.pde.build</artifactId>
+		<version>4.30.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>org.eclipse.pde.build.tests</artifactId>
+	<version>1.4.200-SNAPSHOT</version>
+	<packaging>eclipse-test-plugin</packaging>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>target-platform-configuration</artifactId>
+				<configuration>
+					<dependency-resolution>
+						<extraRequirements>
+							<requirement>
+								<!-- The product -->
+								<type>p2-installable-unit</type>
+								<id>org.eclipse.sdk.ide</id>
+								<versionRange>0.0.0</versionRange>
+							</requirement>
+						</extraRequirements>
+					</dependency-resolution>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-surefire-plugin</artifactId>
+				<configuration>
+					<product>org.eclipse.sdk.ide</product>
+					<testRuntime>p2Installed</testRuntime>
+					<testClass>org.eclipse.pde.build.tests.PDEBuildTestSuite</testClass>
+					<systemProperties>
+						<pde.build.includeP2>false</pde.build.includeP2>
+					</systemProperties>
+				</configuration>
+				<executions>
+					<execution>
+						<id>p2-tests</id>
+						<goals><goal>test</goal></goals>
+						<configuration>
+							<testClass>org.eclipse.pde.build.tests.P2TestSuite</testClass>
+							<systemProperties/>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+</project>
--- a/build/org.eclipse.pde.build.tests/pom.xml
+++ b/build/org.eclipse.pde.build.tests/pom.xml
@@ -5,10 +5,10 @@
 	<parent>
 		<groupId>org.eclipse.pde</groupId>
 		<artifactId>eclipse.pde.build</artifactId>
-		<version>4.30.0-SNAPSHOT</version>
+		<version>4.31.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>org.eclipse.pde.build.tests</artifactId>
-	<version>1.4.200-SNAPSHOT</version>
+	<version>1.4.300-SNAPSHOT</version>
 	<packaging>eclipse-test-plugin</packaging>
 
 	<build>
--- a/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
+++ b/build/org.eclipse.pde.build.tests/META-INF/MANIFEST.MF
@@ -14,7 +14,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.apache.ant,
  org.eclipse.debug.core,
  org.eclipse.equinox.p2.publisher;bundle-version="1.1.0",
- org.eclipse.equinox.p2.publisher.eclipse;bundle-version="1.0.0"
+ org.eclipse.equinox.p2.publisher.eclipse;bundle-version="1.0.0",
+ org.eclipse.equinox.p2.repository.tools
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Import-Package: org.eclipse.equinox.frameworkadmin;version="2.0.0",
  org.eclipse.equinox.internal.p2.artifact.repository,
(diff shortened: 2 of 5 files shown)


iloveeclipse:
`org.eclipse.pde.build.tests: Unqualified OSGi version 1.4.300.qualifier must match unqualified Maven version 1.4.200-SNAPSHOT`

laeubi:
> `org.eclipse.pde.build.tests: Unqualified OSGi version 1.4.300.qualifier must match unqualified Maven version 1.4.200-SNAPSHOT`

Seems it was too late yesterday, I fixed that now and also added an assumption to the launcher find method, this seems for me the most suitable solution because then we can merge this attempt already and add more work (e.g. enable more test) on top of it.

laeubi:
This seems to reproduce the test failures we see here
- https://github.com/eclipse-pde/eclipse.pde/issues/967

I therefore think this PR should be merged so we can see them in regular verification builds and can fix/act on them... in addition to that it adds 35 new test we will be aware if they would break.

@HannesWell @vik-chand any concerns?

build/org.eclipse.pde.build.tests/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+					</dependency-resolution>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-surefire-plugin</artifactId>
+				<configuration>
+					<product>org.eclipse.sdk.ide</product>
+					<testRuntime>p2Installed</testRuntime>
+					<testClass>org.eclipse.pde.build.tests.PDEBuildTestSuite</testClass>
```
HannesWell, line 38: Am I right, that this testClass is used for the default `test` goal execution of the `tycho-surefire-plugin` for `eclipse-test-plugins` and the overwritten value in the separate `p2-tests` execution refelects the second execution defined in this projects test.xml?

build/org.eclipse.pde.build.tests/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+						<pde.build.includeP2>false</pde.build.includeP2>
+					</systemProperties>
+				</configuration>
+				<executions>
+					<execution>
+						<id>p2-tests</id>
+						<goals><goal>test</goal></goals>
+						<configuration>
+							<testClass>org.eclipse.pde.build.tests.P2TestSuite</testClass>
+							<systemProperties/>
```
HannesWell, line 49:
```suggestion
```
Is this necessary? The effective pom view says this element is not configured in a parent pom.

HannesWell (approved):
> This seems to reproduce the test failures we see here
> 
>     * [120 P2Tests fail on master since I20231201-1800 #967](https://github.com/eclipse-pde/eclipse.pde/issues/967)
> 

Yes, I checked a few failures and they are identical.

> 
> I therefore think this PR should be merged so we can see them in regular verification builds and can fix/act on them... in addition to that it adds 35 new test we will be aware if they would break.
> 

It's not ideal but sounds like a suitable solution for the current situation. It will also help to verify a fix sooner.

Thank you @laeubi for making this possible so that we can detect regressions early.
Hopefully this prevent future regressions and will reduce the burden for @iloveeclipse to hunt down their causes.

build/org.eclipse.pde.build.tests/pom.xml
laeubi: This is to clear the inherited properties from the default execution/plugin configuration, one might can configure the default execution but lets dont complicate it too much as it already is :-)

build/org.eclipse.pde.build.tests/pom.xml
laeubi: Yes

iloveeclipse:
Thanks for enabling tests! Was a long outstanding PDE issue.

build/org.eclipse.pde.build.tests/pom.xml
HannesWell: > but lets dont complicate it too much as it already is :-)

That's fine :)
But depending on where the to clear property is defined and how this might not be sufficient and something like https://stackoverflow.com/a/33252763/14542697 needs to be applied.
I have not further checked where this is coming from exactly, but I'm sure you are aware of all that.
Just wanted to make sure, it is not just forgotten.

akurtakov:
I just saw aggregator build failing https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-1639/1/console :
```
22:23:57  org.apache.maven.ProjectCycleException: The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT'}' and 'Vertex{label='org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT'}' introduces to cycle in the graph org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT --> org.eclipse.pde:org.eclipse.pde.build.tests:1.4.300-SNAPSHOT --> eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT --> org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT
```
I'm not sure it's caused by this one here but looks suspiciously connected.

iloveeclipse:
> I just saw aggregator build failing https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-1639/1/console :
> 
> ```
> 22:23:57  org.apache.maven.ProjectCycleException: The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT'}' and 'Vertex{label='org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT'}' introduces to cycle in the graph org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT --> org.eclipse.pde:org.eclipse.pde.build.tests:1.4.300-SNAPSHOT --> eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT --> org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT
> ```
> 
> I'm not sure it's caused by this one here but looks suspiciously connected.

This looks lile caused by this PR.

Should we quickly revert it before next build?

akurtakov:
> > I just saw aggregator build failing https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-1639/1/console :
> > ```
> > 22:23:57  org.apache.maven.ProjectCycleException: The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT'}' and 'Vertex{label='org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT'}' introduces to cycle in the graph org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT --> org.eclipse.pde:org.eclipse.pde.build.tests:1.4.300-SNAPSHOT --> eclipse.platform.repository:eclipse.platform.repository:4.31.0-SNAPSHOT --> org.eclipse.sdk.feature:org.eclipse.sdk.tests:4.31.0-SNAPSHOT
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure it's caused by this one here but looks suspiciously connected.
> 
> This looks lile caused by this PR.
> 
> Should we quickly revert it before next build?

I'm 90% that this will break the build.

iloveeclipse:
So let revert it.

iloveeclipse:
> So let revert it.

I see no possibility to do this from the github UI?

If I see it right, we only need revert for https://github.com/eclipse-pde/eclipse.pde/commit/6e5138acab9998128905e533edf5d573b58196a3

iloveeclipse:
> > So let revert it.
> 
> I see no possibility to do this from the github UI?
> 
> If I see it right, we only need revert for [6e5138a](https://github.com/eclipse-pde/eclipse.pde/commit/6e5138acab9998128905e533edf5d573b58196a3)

=> https://github.com/eclipse-pde/eclipse.pde/pull/985

iloveeclipse:
@laeubi : it is unfortunate we've had to revert this PR, but it has shown a potential to enable PDE tests. 
If you have an idea how to make it without breaking aggregator build, it would be great.

laeubi:
I'll give it a try later on, basically I think one can use a profile that we only enable when verification builds run (like the `tck` profile we already have) so it is not executed in a full aggregator build.

HannesWell:
> I'll give it a try later on, basically I think one can use a profile that we only enable when verification builds run (like the `tck` profile we already have) so it is not executed in a full aggregator build.

Sounds good.
I also had a quick look back then and wondered why there is even a cycle?
`org.eclipse.pde.build.tests` is not a fragment to pde.build and should not be contained in any feature. Therefore I wonder why the sdk had a reference to it so that a cycle can be formed?

Another point is that it looked like the activated tests only use what have been build with the I-builds (see https://github.com/eclipse-pde/eclipse.pde/pull/984) and if I understand the config right, this is no surprise since the test-runtime is p2 provisioned?

PR #982, opened 2023-12-05
Enable License Check for PDE

laeubi:
FYI @akurtakov @merks @HannesWell

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -1,0 +1,23 @@
+name: License vetting status check
+
+on:
+  push:
+    branches: 
+      - 'master'
+      - 'tycho-*'
+  pull_request:
+    branches: 
+     - 'master'
+     - 'tycho-*'
+  issue_comment:
+    types: [created]
+
+jobs:
+  call-license-check:
+    permissions:
+      pull-requests: write
+    uses: eclipse/dash-licenses/.github/workflows/mavenLicenseCheck.yml@master
+    with:
+      projectId: technology.tycho
+    secrets:
+      gitlabAPIToken: ${{ secrets.PDE_GITLAB_API_TOKEN }}


jukzi:
that breaks all builds now
image
for example 
https://github.com/eclipse-pde/eclipse.pde/actions/runs/7106050788/job/19344629747?pr=984

iloveeclipse:
> that breaks all builds now

Yep... Please fix or revert.

HannesWell:
We ware await the addition of a GITLAB_API_TOKEN for this organization by the EF-infra team:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4037

With that we can request the license vetting of the offending projects.

laeubi:
> > that breaks all builds now
> 
> Yep... Please fix or revert.

It does **NOT** break the build because it is not even _part_ of the build. It adds a new check that currently fails because some unvetted artifacts have recently slipped in and **that the purpose of this check**, to fail so such things did not happen unnoticed.

So can we please calm down a bit instead and try to understand why things happen, if anyone is that concerned we can of course fall back to manually checks and creating CQs by hand.

iloveeclipse:
@laeubi : the problem I see is that adding actions that are **known to fail** distracts everyone for no obvious reason.
If the problem is known, why not wait till it is fixed and enable the action **after that**?

With every new action added that **fails by default** everyone  learns to **ignore** github action errors and **keep ignoring them** in the future. This is https://wiki.c2.com/?FixBrokenWindows in pure form.

So we would be happy to see more checks and we value your work, but new checks should be introduced in a way that makes **everyone** happy.

akurtakov:
In this case the choices are:
* not have this one and suffer from issues like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1611 until we do the next one
* have this one which couldn't be tested until it's in and once the secret is added by webmaster we have better status quo
* not run licensecheck at all and ship with unvetted deps

As 3) is not an option and we can not fix 1) without 2) what we have here is the best forward looking path.

akurtakov:
Work towards fixing the licensecheck is ongoing in https://github.com/eclipse-pde/eclipse.pde/pull/983 . I am not aware of any other way to have this fixed without suffering short term.

jukzi:
i don't want to read through all the history, can you please summarize why such a change has to be submitted? As far as i understand such checks can also be used with not-submitted PRs, as each PR uses the workflows it submits.

akurtakov:
@jukzi  https://docs.github.com/en/actions/using-workflows/about-workflows#about-workflows :
```
Workflows are defined by a **YAML file checked in ** to your repository and will run when triggered by an event in your repository, or they can be triggered manually, or at a defined schedule.
```
A PR is not **checked in** thus workflow in it will not be executed.

laeubi:
> the problem I see is that adding actions that are known to fail distracts everyone for no obvious reason.

Some action require elevated permissions and can not run as part of a PR, so there is not much of an option.

> With every new action added that fails by default everyone learns to ignore github action errors and keep ignoring them in the future. This is https://wiki.c2.com/?FixBrokenWindows in pure form.

No this is maybe ignorance in pure form, please don't blame that on making things _VISIBLE_, the window is already _BROKEN_ but no one has noticed it yet so please don't start this fruitless discussion again. There is no single line in the committer handbook that says one can ignore failures just because it annoyed anyone in the past.

> i don't want to read through all the history, can you please summarize why such a change has to be submitted?

Please read the committer handbook about IP due diligence, if you feel that we should enforce each and every contribution to be checked by a committer **manually** then "we" don't need it, if we want to benefit from automatic checks and automatic request "we" need it.

jukzi:
> A PR is not **checked in** thus workflow in it will not be executed.

looks like a PR is treated to be checked in, see example
image
you also can see that "call-license-check / check-licenses" in this PR workflow was executed before it was checked in

akurtakov:
Because it already is as can be seen at https://github.com/eclipse-pde/eclipse.pde/blob/master/.github/workflows/licensecheck.yml . If there is no such workflow file in the repo it would not be executed.

jukzi:
> If there is no such workflow file in the repo it would not be executed.

thats simply not rue - you can see that "call-license-check / check-licenses" in this(**982**) PR workflow was executed before it was checked in

image

the same was true when i added codeQL

akurtakov:
The requirement seems to be specific to (parts?) of this workflow https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2966

PR #985, opened 2023-12-05
Revert "Run some pde.build tests with Tycho"

iloveeclipse:
This reverts commit 6e5138acab9998128905e533edf5d573b58196a3 as it will introduce cyclic build error in the aggregator build.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1639 
See https://github.com/eclipse-pde/eclipse.pde/pull/980


iloveeclipse:
image

PR #987, opened 2023-12-06
Separate the license checks in a check and and review part

laeubi:
and fix project id...

the review part needs to be merged to master to be fully tested.

grafik

now it nicely integrates with the other checks.


laeubi:
The check part works, now see if the request for review works as well...

PR #990, opened 2023-12-07
Jenkinsfile: qualityGate DELTA

jukzi:
do not fail build if total number of warnings have been reduced

see https://github.com/eclipse-pde/eclipse.pde/pull/989#issuecomment-1843399351

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,7 +33,7 @@ pipeline {
 						apiAnalyzer-workspace/.metadata/*.log')
 					junit '**/target/surefire-reports/*.xml'
 					discoverGitReferenceBuild referenceJob: 'eclipse.pde/master'
-					recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()], qualityGates: [[threshold: 1, type: 'NEW', unstable: true]]
+					recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()], qualityGates: [[threshold: 1, type: 'DELTA', unstable: true]]
 				}
 			}
 		}


laeubi:
I think this is best reviewed by @akurtakov in general I think changes should not introduce new warnings (or use `@suppresswarning` if unavoidable) but if it helps we allow to introduce a new one if another is fixed instead I can't judge.

akurtakov:
Please don't close this PR for now - it may be something we just apply soon if needed.

laeubi:
By the way, have you noticed how all checks are green now (including licence check)?

jukzi:
> By the way, have you noticed how all checks are green now (including licence check)?

gj!

akurtakov:
Step by step, we are really going towards far better development process! Of course it has its growing pains but there is always a price.

HannesWell:
> > I think I gave a better fix at the linked PR which removes the need for the "workaround" field. This happened only because of the threshold gate pointing the new issue thus I believe we should keep things as they are right now and not accept this one.
> 
> I totally agree that that particular solution is better, however in general it would be good to just reduce warnings insted of complaining about moved warnings.

Of course the best would be to not introduce new warnings, but having one new if ten are fixed is also a win.
However will this setting also permit a one-to-one replacement? Then I would say it is usually not an advantage.

laeubi:
if one thinks the change is good even though it introduce a new warning there is always the option to merge it anyways for "edge-cases"...

akurtakov:
@HannesWell as PL please make a decision on this one and either merge or close. It's good to keep the PRs queue shorter :) .

HannesWell:
> @HannesWell as PL please make a decision on this one and either merge or close. It's good to keep the PRs queue shorter :) .

I agree with Christoph's opinion on this. In general it should be attempted to not introduce new warnings and I'm convinced it is usually possible. But yes, in edge-cases it is still possible to overrule a failing quality-gate and submit changes nevertheless.

laeubi:
I just wanted to make you aware that there is an open pull request to add an option to only fail the quality gate but do not change the overall build result here:
- https://github.com/jenkinsci/warnings-ng-plugin/pull/1649

so if anyone is interested please express your support there!

PR #992, opened 2023-12-07
Add bnd templates template to feature

laeubi:
For license checks


akurtakov:
Just wondering - as this is external artifact shouldn't it be imported rather than included?

akurtakov:
> > Just wondering - as this is external artifact shouldn't it be imported rather than included?
> 
> Maybe yes it was just a quick thing to test the review check :-)

Sure, I am just putting my remarks as quite often I forget such things and end up pushing them :)

PR #993, opened 2023-12-08
Stop filtering javadoc

akurtakov:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1604 there is no longer timestamp in javadoc thus there is no need to filter javadoc.


akurtakov:
I am starting an I-build now to verify there wouldn't be a surprising comparator issue.

Issue #996, opened 2023-12-09
NullPointerException when resolving target-definition with 4.30 and onwards


ptziegler:
We use Maven dependencies together with p2 update sites for the WindowBuilder project. Trying to resolve the target-definition currently fails with the following stack trace:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.osgi.service.resolver.BundleDescription.getSymbolicName()" because "bundleDescription" is null
	at org.eclipse.pde.internal.core.target.InstallableUnitGenerator.generateInstallableUnits(InstallableUnitGenerator.java:85)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:273)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.StreamSpliterators$WrappingSpliterator.forEachRemaining(StreamSpliterators.java:310)
	at java.base/java.util.stream.Streams$ConcatSpliterator.forEachRemaining(Streams.java:734)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.getAdditionalProvisionIUs(P2TargetUtils.java:1579)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.getAdditionalProvisionIUs(P2TargetUtils.java:1572)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.resolveWithPlanner(P2TargetUtils.java:1090)
	at org.eclipse.pde.internal.core.target.P2TargetUtils.synchronize(P2TargetUtils.java:832)
	at org.eclipse.pde.internal.core.target.TargetDefinition.resolve(TargetDefinition.java:401)
	at org.eclipse.pde.internal.ui.editor.targetdefinition.TargetEditor$TargetChangedListener$1.run(TargetEditor.java:651)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

With the following Maven dependency being the culprit:

```
<dependency>
    <groupId>org.burningwave</groupId>
    <artifactId>core</artifactId>
    <version>12.64.2</version>
</dependency>
[...]

PR #997, opened 2023-12-09
Skip target bundles with invalid manifest during IU generation #996

ptziegler:
A call to BundlesAction.createBundleDescription() returns null, if the manifest of the current bundle contains invalid headers (e.g. the same package is imported twice).

If left unchecked, this causes an unhandled NullPointerException a few lines further down. Add an explicit null, in order to simply skip this invalid bundle.


laeubi (approved):
sounds resonable

HannesWell (approved):
Looks good.
Thank your for this fix.

Issue #140, opened 2022-06-05
Treat 'Require-Capability: osgi.ee' as equivalent to 'Bundle-RequiredExecutionEnvironment'


HannesWell:
If a bundle declares a `osgi.ee` capability like `Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"` but no `Bundle-RequiredExecutionEnvironment` but no `Bundle-RequiredExecutionEnvironment` the Manifest-Editor creates a warning about a missing execution environment and shows an empty List of EEs.

But BREE is deprecated by the OSGi spec:

> The Bundle-RequiredExecutionEnvironment manifest header provides the same function as the [osgi.ee Namespace](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.namespaces.html#framework.namespaces.osgi.ee). It allows a bundle to depend on the execution environment. **This header is deprecated** but must be fully supported by a compliant framework. Bundles **should not mix these headers** but use either an osgi.ee requirement or this header.

http://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.bree

Therefore such an osgi.ee capability should be treated equivalently regarding warnings, internal processing (I have not yet checked if it is already considered, maybe implicitly by the resolver?) and visual representation regarding in the Editor.

We could even consider to create `osgi.ee` capabilities instead of BREEs if new ones are added via the editor.

This is similar to [Bug 324209](https://bugs.eclipse.org/bugs/show_bug.cgi?id=324209).

PR #1000, opened 2023-12-11
Consider Require-Capability when validating BREE

alshamams:
Presence of osgi.ee capability ‘Require-Capability’ should mitigate absence of BREE headers in the manifest. The current parser has no awareness of Require-Capability header. In future the parser should consider this header too while evaluating BREE header. 

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/140

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
@@ -102,6 +102,7 @@ public class BundleErrorReporter extends JarManifestErrorReporter {
 	private boolean fOsgiR4;
 	private IPluginModelBase fModel;
 	private Set<String> fProjectPackages;
+	private static final String osgiEE = "osgi.ee"; //$NON-NLS-1$
 
 	public BundleErrorReporter(IFile file) {
 		super(file);
@@ -553,6 +554,21 @@ public class BundleErrorReporter extends JarManifestErrorReporter {
 		}
 	}
 
+	private boolean isCompatibleOsgiEE() {
+		IHeader header = getHeader(Constants.REQUIRE_CAPABILITY);
+		if (header == null) {
+			return false;
+		}
+		String ee = header.getValue();
+		String[] parts = ee.split(";"); //$NON-NLS-1$
+		if (parts[0] == null) {
+			return false;
+		}
+		if (!parts[0].equalsIgnoreCase(osgiEE)) {
+			return false;
+		}
+		return true;
+	}
 	private void validateRequiredExecutionEnvironment() {
 		int sev = CompilerFlags.getFlag(fProject, CompilerFlags.P_INCOMPATIBLE_ENV);
 		if (sev == CompilerFlags.IGNORE) {
@@ -587,8 +603,7 @@ public class BundleErrorReporter extends JarManifestErrorReporter {
 							IPath currentPath = entry.getPath();
 							if (JavaRuntime.newDefaultJREContainerPath().matchingFirstSegments(currentPath) > 0) {
 								String eeId = JavaRuntime.getExecutionEnvironmentId(currentPath);
-								IHeader header = getHeader(Constants.REQUIRE_CAPABILITY);
-								if (eeId != null && header == null) {
+								if (eeId != null && !isCompatibleOsgiEE()) {
 									VirtualMarker marker = report(PDECoreMessages.BundleErrorReporter_noExecutionEnvironmentSet, 1, sev, PDEMarkerFactory.M_EXECUTION_ENVIRONMENT_NOT_SET, PDEMarkerFactory.CAT_EE);
 									addMarkerAttribute(marker, "ee_id", eeId); //$NON-NLS-1$
 									addMarkerAttribute(marker,PDEMarkerFactory.compilerKey,	CompilerFlags.P_INCOMPATIBLE_ENV);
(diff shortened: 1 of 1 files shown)


laeubi:
@alshamams congratulation for open the 1.000 PR on PDE :1st_place_medal: 

> Presence of osgi.ee capability ‘Require-Capability’ should mitigate absence of BREE headers in the manifest. 

So I understand it correctly that it just not complains if any `Require-Capability` is found?

> In future the parser should consider this header too while evaluating BREE header

Do you maybe want to try add support for this? e.g on Tych you can take a look at:

- https://github.com/eclipse-tycho/tycho/blob/0842d2667a2643650dbf46d547388cbbe508da4b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ManifestComparator.java#L179
- https://github.com/eclipse-tycho/tycho/blob/0842d2667a2643650dbf46d547388cbbe508da4b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ManifestComparator.java#L185
- https://github.com/eclipse-tycho/tycho/blob/0842d2667a2643650dbf46d547388cbbe508da4b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java#L472
- https://github.com/eclipse-tycho/tycho/blob/0842d2667a2643650dbf46d547388cbbe508da4b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/BndManifestProcessor.java#L143

where Tycho tries to decide the EE value from the capability and parse the headers so maybe that can be used as some kind of inspiration here?

As a bonus, it would be great if such EE would be shown at the Manifest-Editor overview page (maybe with the edit buttons disabled)!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
```diff
@@ -587,7 +587,8 @@ private void validateRequiredExecutionEnvironment() {
 							IPath currentPath = entry.getPath();
 							if (JavaRuntime.newDefaultJREContainerPath().matchingFirstSegments(currentPath) > 0) {
 								String eeId = JavaRuntime.getExecutionEnvironmentId(currentPath);
-								if (eeId != null) {
+								IHeader header = getHeader(Constants.REQUIRE_CAPABILITY);
```
HannesWell, line 590: This is here and below is probably not sufficient because a Plugin/Bundle can have various difference required capabilities, e.g. custom ones or for osgi service loder mediator.
So the raw value has to be parsed for the `osgi.ee` capability. Please see the doc I linked in the issue.

HannesWell (changes requested):
Thank you @alshamams for working on this, but I think the current implementation is not sufficient.
Unfortunately the `osgi.ee` capability is not as simple as the BREE header since it shares the header with possible other required capabilities.

> As a bonus, it would be great if such EE would be shown at the Manifest-Editor overview page (maybe with the edit buttons disabled)!

Yes. As mentioned in the linked issue, we could even consider to fully replace the BREE header by a ee-capability´.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
alshamams: My assumption was that osgi.ee is the only namespace defined for required capabilities. I have now changed the code to parse the header value and check for the presence of osgi.ee namespace. Kindly review and let me know.

alshamams:
> Thank you @alshamams for working on this, but I think the current implementation is not sufficient. Unfortunately the `osgi.ee` capability is not as simple as the BREE header since it shares the header with possible other required capabilities.
> 
> > As a bonus, it would be great if such EE would be shown at the Manifest-Editor overview page (maybe with the edit buttons disabled)!
> 
> Yes. As mentioned in the linked issue, we could even consider to fully replace the BREE header by a ee-capability´.

A follow up PR is on its way.👍

alshamams:
I can think of corner cases such as presence of both the headers with conflicting EE versions. If you think this is a valid concern, I plan to investigate this in a follow up PR.

laeubi:
> I can think of corner cases such as presence of both the headers with conflicting EE versions. If you think this is a valid concern, I plan to investigate this in a follow up PR.

It would of course be good to have a warning as one is only _should_ to specify one of them but not both see https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.bree

>  Bundles should not mix these headers but use either an osgi.ee requirement or this header. If both are used, both constraints must be met to resolve. 

but it is even allowed and there can't be a conflict.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BundleErrorReporter.java
```diff
@@ -553,6 +554,21 @@ private void validateBundleVersion() {
...
+		}
+		String ee = header.getValue();
+		String[] parts = ee.split(";"); //$NON-NLS-1$
+		if (parts[0] == null) {
+			return false;
+		}
+		if (!parts[0].equalsIgnoreCase(osgiEE)) {
+			return false;
+		}
+		return true;
```
HannesWell, line 570: This is not very robust, in general you probably would want to use `ManifestElement.parseHeader()` instead.
But as written in the general comment, this should be done at a more abstract level.

HannesWell (commented):
@alshamams I think in the current from this is not sufficient, because if osgi.ee headers are used, not only their existence should be checked, but they should be merged with any potentially existing BREE-headers and then treated similarly. Further below in the code, their version should be checked as well, just like for the BREE header.

Therefore I suggest to treat all ee-requirements (regardless if they originate from a BREE header or a osgi.ee requirement) the same and read them at a more abstract level.
I'm thinking about reading
`List<Requirement> requirements = bundleDescription.getRequirements(ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE);`.
Then the EE Ids have to be extracted again, but that is probably not that easy. I have started to look into this a bit locally, but didn't found a satisfying solution yet.

HannesWell:
> I'm thinking about
> `List<Requirement> requirements = bundleDescription.getRequirements(ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE);`.
> Then the EE Ids have to be extracted again, but that is probably not that easy. I have started to look into this a bit locally, but didn't found a satisfying solution yet.

@alshamams I have just added a new method that does exactly that via https://github.com/eclipse-pde/eclipse.pde/pull/1325.
It is named `ManifestUtilss.getRequiredExecutionEnvironments(Resource)`.
Can you please update this PR to use that method to read the required EEs of a `bundleDescription`?

HannesWell:
> > I'm thinking about
> > `List<Requirement> requirements = bundleDescription.getRequirements(ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE);`.
> > Then the EE Ids have to be extracted again, but that is probably not that easy. I have started to look into this a bit locally, but didn't found a satisfying solution yet.
> 
> @alshamams I have just added a new method that does exactly that via #1325. It is named `ManifestUtilss.getRequiredExecutionEnvironments(Resource)`. Can you please update this PR to use that method to read the required EEs of a `bundleDescription`?

@gireeshpunathil could you ping @alshamams if she wants to continue on this one?

laeubi:
As no feedback closing this for now.

PR #1005, opened 2023-12-13
Various cleanups

akurtakov:
* Use simplified lambda syntax
* Use Collection.isEmpty
* Fix broken @see tag
* Use pattern matching
* Use try-with-resources
* Use StandardCharsets
* Immediate return

Complains from sonarlint in my workspace

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src_ant/org/eclipse/pde/api/tools/internal/tasks/APIToolsAnalysisTask.java
+++ b/apitools/org.eclipse.pde.api.tools/src_ant/org/eclipse/pde/api/tools/internal/tasks/APIToolsAnalysisTask.java
@@ -175,9 +175,7 @@ public class APIToolsAnalysisTask extends CommonUtilsTask {
 		}
 		List<Entry<String, IApiProblem[]>> allEntries = new ArrayList<>();
 		allEntries.addAll(entrySet);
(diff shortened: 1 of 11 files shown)


apitools/org.eclipse.pde.api.tools/src_ant/org/eclipse/pde/api/tools/internal/tasks/APIToolsAnalysisTask.java
```diff
@@ -175,9 +175,7 @@ private Summary[] createAllSummaries(Map<String, IApiProblem[]> allProblems) {
 		}
 		List<Entry<String, IApiProblem[]>> allEntries = new ArrayList<>();
 		allEntries.addAll(entrySet);
-		Collections.sort(allEntries, (entry1, entry2) -> {
-			return entry1.getKey().compareTo(entry2.getKey());
-		});
+		Collections.sort(allEntries, (entry1, entry2) -> entry1.getKey().compareTo(entry2.getKey()));
```
HannesWell, line 178: In a follow-up a static comparator factory could be used here and at other occasions:

```suggestion
		Collections.sort(allEntries, Comparator.comparing(Entry::geKey));
```

HannesWell (commented):
Thank for the clean up.

Issue #401, opened 2022-11-14
When importing a package, don't use the micro version


laeubi:
Currently in PDE when a package is missing, the Quickfix suggest to import the package and then automatically adds the exact version of that package from the target. As micro version should not change anything, it should not be part of the import, so if the package is of version `1.2.3` , the import should have a lower bound of `1.2`

alshamams:
Hi @laeubi ,
I want to take this up. I was trying to reproduce the problem, but could not.

- The quickfix list for missing imports in the source does not include package import suggestion. Can you point me to the workflow of the quickfix that you are referring to?
- Adding an import package in the UI results in insertion of the package in the manifest but doesn't involve quick fix.( Shouldn't the same change be implemented here as well?) Or is this what you intend?

laeubi:
You can use an existing plugin, then remove one of the require bundles / packages and then go to the place where the type is missing, if you then open quickfix you should find suggestion to import a package (maybe a bit down the list even though IMO it should be the first item).

alshamams:
Hi, @laeubi 
I tried recreating the issue as per your suggestion using an existing plugin project creation template and removed a required bundle. I have added the screenshot of possible quick fix suggestions that came up, since I did not see an option to import the missing package in the list.

<img width="674" alt="image" src="https://github.com/eclipse-pde/eclipse.pde/assets/122432113/376ffcd4-c826-4535-83ea-b21dd10f7004">

laeubi:
The following is for `org.eclipse.equinox.frameworkadmin` bundle of P2 repository:

1. go to the manifest and remove `org.eclipse.osgi.util` package import
2. Now you should get an error marker in the Utils class 
![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/1331477/346ee968-7649-469b-b50f-c9cb4dc8a44f)
3. the Quickfix (STRG+1) should suggest to import the package again
![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/1331477/8c93ce29-e34d-4955-9354-4433c7c5fb67)

To see what is described here, one needs to change that version of the package export of bundle `org.eclipse.osgi` for example from `1.1.0` in `1.1.1` in wich case the import will be after apply the quick-fix `1.1.1` while ist should be still `1.1.0` as the micro version is not important.

PR #1007, opened 2023-12-14
Strip micro-version and add upper bound when adding package imports

alshamams:
This commit resets the micro-version to ‘0’ while importing packages as part of quick fix, as it is not relevant. 

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/401

### DIFF
--- a/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/ImportPackageObject.java
+++ b/ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/ImportPackageObject.java
@@ -32,9 +32,11 @@ public class ImportPackageObject extends PackageObject {
 	private static final long serialVersionUID = 1L;
 
 	private static String getVersion(ExportPackageDescription desc) {
-		String version = desc.getVersion().toString();
-		if (!version.equals(Version.emptyVersion.toString())) {
-			return desc.getVersion().toString();
+		Version version = desc.getVersion();
+		if (version != null && !Version.emptyVersion.equals(version)) {
+			return new VersionRange(
+					"[" + version.getMajor() + "." + version.getMinor() + "," + (version.getMajor() + 1) + "]") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$//$NON-NLS-4$
+							.toString();
 		}
 		return null;
 	}
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java
@@ -49,7 +49,6 @@ import org.eclipse.pde.internal.core.text.bundle.ExportPackageHeader;
 import org.eclipse.pde.internal.core.text.bundle.ExportPackageObject;
 import org.eclipse.pde.internal.core.text.bundle.ImportPackageHeader;
 import org.eclipse.pde.internal.core.text.bundle.ImportPackageObject;
-import org.eclipse.pde.internal.core.text.bundle.ManifestHeader;
 import org.eclipse.pde.internal.ui.PDEPlugin;
 import org.eclipse.pde.internal.ui.PDEPluginImages;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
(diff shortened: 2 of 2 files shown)


HannesWell (changes requested):
Hi @alshamams thank you for working on this.
While this fixes the linked issue in its most basic form I think this is a good opportunity to fix a few other flaws in this quick-fix.

In general it would be good to also specify the upper bound of the import version range and to move the logic to a more general place, i.e. to `ImportPackageObject.getVersion()`.
You can get such String using the following to assemble a suitable VersionRange from a `org.osgi.framework.Version`:
```
new VersionRange("[" + version.getMajor() + "." + version.getMinor() + "," + (version.getMajor() + 1) + "]").toString()
```
With that the version of the `ImportPackageObject` does not have to be set explicitly anymore.

Besides that, the quick-fix currently only adds a version if the Import-Package header already exists. This could be fixed by replacing.

```
if (header == null) {
	bundle.setHeader(Constants.IMPORT_PACKAGE, pkgId);
} else if (header instanceof ImportPackageHeader ipHeader) {
```
by
```
if (header == null) {
	header = bundle.getModel().getFactory().createHeader(Constants.IMPORT_PACKAGE, pkgId);
}
if (header instanceof ImportPackageHeader ipHeader) {
```

alshamams:
Thanks @HannesWell for the quick review.

> In general it would be good to also specify the upper bound of the import version range and to move the logic to a more general place

Thanks, this looks clean and elegant. However, modifying the getVersion API will have implications in regular import package flow. (Not the quick fix one, but while adding import packages from Dependencies tab). Is this what you intend ? Just wanted to make sure you are aware of this change before I push.

> Besides that, the quick-fix currently only adds a version if the Import-Package header already exists. This could be fixed by replacing.

I was aware of it, and was planning to do it in a follow-up PR, given that it is unrelated to the issue at hand. I think that I will make the change here itself.

HannesWell:
> > In general it would be good to also specify the upper bound of the import version range and to move the logic to a more general place
> 
> Thanks, this looks clean and elegant. However, modifying the getVersion API will have implications in regular import package flow. (Not the quick fix one, but while adding import packages from Dependencies tab). Is this what you intend ? Just wanted to make sure you are aware of this change before I push.

I have not yet checked every caller path of the code, but yes in general it is good practice to specifiy an upper bound if one specifies a lower to follow semantic versioning. If one really wants to have no upper-bound, it can still be removed later. But from my experience I currently add an upper bound in 90% of the time I use the quick-fix.

> 
> > Besides that, the quick-fix currently only adds a version if the Import-Package header already exists. This could be fixed by replacing.
> 
> I was aware of it, and was planning to do it in a follow-up PR, given that it is unrelated to the issue at hand. I think that I will make the change here itself.

Yes that's fine from my POV, this change is small enough at the moment and it is related enough to mix it.

alshamams:
Hi @HannesWell ,
I have addressed the changes, and squashed and rebased the commits while adding you as the co-author. Kindly let me know in case of any more edits.

ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/ImportPackageObject.java
```diff
@@ -32,9 +32,17 @@ public class ImportPackageObject extends PackageObject {
...
 	private static String getVersion(ExportPackageDescription desc) {
-		String version = desc.getVersion().toString();
-		if (!version.equals(Version.emptyVersion.toString())) {
-			return desc.getVersion().toString();
+		// String version = desc.getVersion().toString();
+		//
+		// if (!version.equals(Version.emptyVersion.toString())) {
+		// return desc.getVersion().toString();
+		// }
+		// return null;
```
HannesWell, line 40: Please delete commented out code.
```suggestion
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java
```diff
@@ -321,14 +321,18 @@ protected void modifyModel(IBaseModel model, IProgressMonitor monitor) throws Co
...
+						header = bundle.getModel().getFactory().createHeader(Constants.IMPORT_PACKAGE, pkgId);
+					}
+					if (header instanceof ImportPackageHeader ipHeader) {
 						int manifestVersion = BundlePluginBase.getBundleManifestVersion(bundle);
-						String versionAttr = (manifestVersion < 2) ? ICoreConstants.PACKAGE_SPECIFICATION_VERSION : Constants.VERSION_ATTRIBUTE;
-						ImportPackageObject impObject = new ImportPackageObject((ManifestHeader) header, (ExportPackageDescription) getChangeObject(), versionAttr);
+						String versionAttr = (manifestVersion < 2) ? ICoreConstants.PACKAGE_SPECIFICATION_VERSION
+								: Constants.VERSION_ATTRIBUTE;
+						ImportPackageObject impObject = new ImportPackageObject((ManifestHeader) header, desc,
+								versionAttr);
```
HannesWell, line 335: If you inline `manifestVersion` and replace header by `ipHeader` we would get some nicely formatted code.
```suggestion
						String versionAttr = (BundlePluginBase.getBundleManifestVersion(bundle) < 2)
								? ICoreConstants.PACKAGE_SPECIFICATION_VERSION
								: Constants.VERSION_ATTRIBUTE;
						ImportPackageObject impObject = new ImportPackageObject(ipHeader, desc, versionAttr);
```

HannesWell (approved):
@alshamams I just applied the suggestions locally and also added handling for the default empty version (`0.0.0`) in `ImportPackageObject` so that in that case not a version range like `[0.0.0,1.0.0)` is added. This case 

The great thing about specifying the upper bound in `ImportPackageObject.getVersion()` is, that it is now also added if one select and adds a package via the Editor directly or through the `add dependencies` link in the `Automated Management of Dependencies` section.
This is IMHO a great improvement for PDE respectively its users.

The current state looks very good to me and is for me ready for submission.
Are you fine with this as well?

HannesWell:
I'm merging this now.
Thank you again @alshamams.

PR #1008, opened 2023-12-14
Handle bundle folders in VirtualArtifactRepository.transferArtifact

merks:
The implementation currently assumes all bundles are jars, but bundles from the shared bundle pool may be unpacked as folders.

https://github.com/eclipse-pde/eclipse.pde/pull/962

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
@@ -14,6 +14,7 @@
 package org.eclipse.pde.internal.core.target;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -22,9 +23,11 @@ import java.net.URI;
 import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
+import java.util.zip.ZipOutputStream;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
@@ -34,6 +37,7 @@ import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.equinox.frameworkadmin.BundleInfo;
+import org.eclipse.equinox.internal.p2.core.helpers.FileUtils;
 import org.eclipse.equinox.internal.p2.metadata.expression.QueryResult;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.core.ProvisionException;
@@ -123,8 +127,17 @@ public class VirtualArtifactRepository extends AbstractArtifactRepository {
 			if (location == null) {
 				throw new FileNotFoundException(bundleInfo.getSymbolicName());
 			}
-			try (InputStream is = location.toURL().openStream()) {
-				is.transferTo(destination);
+
+			File bundleLocation = new File(location);
+			if (bundleLocation.isFile()) {
+				try (InputStream is = new FileInputStream(bundleLocation)) {
+					is.transferTo(destination);
+				}
+			} else {
+				ZipOutputStream zip = new ZipOutputStream(destination);
+				FileUtils.zip(zip, bundleLocation, Set.of(), FileUtils.createRootPathComputer(bundleLocation));
+				zip.finish();
+				zip.flush();
 			}
 		} else if (artifactModel instanceof IFeatureModel featureModel) {
 			String installLocation = featureModel.getInstallLocation();


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -123,8 +125,25 @@ private void transferArtifact(Object artifactModel, OutputStream destination) th
...
+
+			File bundleLocation = new File(location);
+			if (bundleLocation.isFile()) {
+				try (InputStream is = new FileInputStream(bundleLocation)) {
+					is.transferTo(destination);
+				}
+			} else {
+				File zipFile = null;
+				try {
+					zipFile = File.createTempFile(bundleLocation.getName(), ".jar", null); //$NON-NLS-1$
```
laeubi, line 137: Can we use a `ZipOutputStream` instead of creating a physical file to open a stream fro it and copy it to the output stream?

merks: Yes, that's better.  Done.

laeubi (approved):
Looks good!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -123,8 +127,17 @@ private void transferArtifact(Object artifactModel, OutputStream destination) th
...
 				throw new FileNotFoundException(bundleInfo.getSymbolicName());
 			}
-			try (InputStream is = location.toURL().openStream()) {
-				is.transferTo(destination);
+
+			File bundleLocation = new File(location);
+			if (bundleLocation.isFile()) {
+				try (InputStream is = new FileInputStream(bundleLocation)) {
+					is.transferTo(destination);
+				}
```
HannesWell, line 135:
```suggestion
				Files.copy(bundleLocation.toPath(), destination);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -123,8 +127,17 @@ private void transferArtifact(Object artifactModel, OutputStream destination) th
...
+			} else {
+				ZipOutputStream zip = new ZipOutputStream(destination);
+				FileUtils.zip(zip, bundleLocation, Set.of(), FileUtils.createRootPathComputer(bundleLocation));
+				zip.finish();
+				zip.flush();
```
HannesWell, line 140: I think this flush() is not necessary.
```suggestion
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
```diff
@@ -123,8 +127,17 @@ private void transferArtifact(Object artifactModel, OutputStream destination) th
...
-			try (InputStream is = location.toURL().openStream()) {
-				is.transferTo(destination);
+
+			File bundleLocation = new File(location);
+			if (bundleLocation.isFile()) {
+				try (InputStream is = new FileInputStream(bundleLocation)) {
+					is.transferTo(destination);
+				}
+			} else {
+				ZipOutputStream zip = new ZipOutputStream(destination);
```
HannesWell, line 137: A few lines downwards we have used a `JarOutputStream`.
Is there a reason for the difference? I wonder if just a ZipOutputStream would be sufficient down there?
Looking at `JarOutputStream` the main addition to `ZipOutputStream` seems to be that it adds a `META-INF/MANIFEST.MF` entry and has a special magic-number for zips.

HannesWell (commented):
Thank you Ed for this fix.
I just have a few minor remarks. Please see them below.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
HannesWell: A few lines downwards we have used a `JarOutputStream`.
Is there a reason for the difference? I wonder if just a `ZipOutputStream` would be sufficient down there too?
Looking at `JarOutputStream` the main addition to `ZipOutputStream` seems to be that it adds a `META-INF/MANIFEST.MF` entry and has a special magic-number for zips.

merks: It would probably better for them to be consistent.,

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
merks: When the above change is made to make things consistent, this can be removed.

Perhaps @ptziegler would like to have a closer look?

laeubi: > I think this flush() is not necessary.

Why do you think this and why would it harm to explicitly flush the content?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/VirtualArtifactRepository.java
laeubi: `JarOutputStream` looks like a good idea, even though I think this is just jared to be extracted right away anyways but consistency is probably key here :-)

laeubi:
I'm merging this now and create a follow-up PR shortly.

PR #1009, opened 2023-12-14
Touch bundles affected by the changed ecj version

iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1394 
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1659


iloveeclipse:
No **real** test fails, only quality gate warnings interpreted as errors.

PR #1010, opened 2023-12-15
Unification / cleanup / nio usage for VirtualArtifactRepository

laeubi:


laeubi:
Maven warnings are cause by this:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1659#issuecomment-1858129712

PR #1015, opened 2023-12-16
Add a requirement to the org.eclipse.osgi.compatibility.state fragment

laeubi:
PDE effectively requires the org.eclipse.osgi.compatibility.state fragment but has no way to declare this requirement in a provisioned install setup so it must be given manually or by a feature.

This adds a p2.inf to ensure it is always included when installing pde.core bundle.


HannesWell (approved):

laeubi:
Test failures and Maven warnings are known issues.

PR #1017, opened 2023-12-18
Use the skipTests property for the execution of pde-build-tests

laeubi:


build/org.eclipse.pde.build.tests/pom.xml
```diff
@@ -148,7 +150,7 @@
 							<testRuntime>p2Installed</testRuntime>
 							<work>${sdkWorkDir}</work>
 							<profileName>${sdkProfile}</profileName>
-							<skipTests>true</skipTests> <!-- this supress the default execution -->
+							<skipTests>true</skipTests> <!-- this supress the default execution regardless of user proeprty given -->
```
Mailaender, line 153:
```suggestion
							<skipTests>true</skipTests> <!-- this supresses the default execution regardless of user property given -->
```

Issue #305, opened 2022-08-23
Validation dialog (when started from a launch) should have link/button to edit the launch config


mickaelistria:
Often, when I do Run As > Eclipse Application, I get the validation coming in to tell me some requirements are not resolved. In the vast majority of cases, remediation involves getting to the launch config editor to add the required deps.
It would save a lot of clicks and some time and attention if the Validation dialog had a "Edit Launch Configuration" link or button to go directly to the usual remediation place.

Dinesh0723:
Shall we add link/button in Validation dialog that will directly open the current configuration of eclipse application.

Dinesh0723:
@mickaelistria, please confirm if I can go ahead with the proposal as shown in the above screenshot?

HannesWell:
Both points sounds good.

But I want to mention that with the possible automated inclusion of dependencies missing items are more unlikely and if they occure are often a sign of configuration/OSGi-Metadata problems.

Nevertheless adding the suggested link still can be helpful, so please go on.

gireeshpunathil:
a more generic question: are there tradeoffs between a hyperlink versus a regular button?

mickaelistria:
links are supposed to move you someplace else without changing much of the state of the application; buttons are usually more to perform an on the current state of the application without necessarily navigating someplace else.

Dinesh0723:
The Edit launch configuration link in validation dialog directly opens the **Edit launch configuration properties** dialog for the current configuration. Is this ok and shall I go for PR? Please refer the attached video for reference,

https://github.com/eclipse-pde/eclipse.pde/assets/130960675/c222ab24-7d9b-4e18-8999-8bcb6bfe1191

iloveeclipse:
This is nice, however it opens for me not the current launch config but **completely unrelated** launch config (probably last executed?) and that's is more confusing as helpful.

Using I20231220-0750, please fix.

Dinesh0723:
> This is nice, however it opens for me not the current launch config but **completely unrelated** launch config (probably last executed?) and that's is more confusing as helpful.
> 
> Using I20231220-0750, please fix.

I have tested it in Eclipse version 2024-03 (4.31) with build ID I20231220-0750, and it is working fine (opens the current launch configuration every time). Please refer to the attached video,

https://github.com/eclipse-pde/eclipse.pde/assets/130960675/93f74782-54fd-480d-b202-cccbd070852e

laeubi:
[...]

PR #1023, opened 2023-12-20
Validation dialog - Edit launch configuration link

Dinesh0723:
Introduced a new link in validation dialog to open the current launch configuration.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/305


mickaelistria:
Thank you very much, this is going to be very useful!

PR #1033, opened 2023-12-31
Gitignore test classes directory

Bananeweizen:
pde.ds.annotations has main and test sources, which have 2 separate bin directories. The second bin directory should be gitignored.

### DIFF
--- a/ds/org.eclipse.pde.ds.annotations/.gitignore
+++ b/ds/org.eclipse.pde.ds.annotations/.gitignore
@@ -1,1 +1,2 @@
 /bin
+/bin_test/


HannesWell (approved):
Thanks.

In general I wonder why the test-class `NameGeneratorTest` is not simply placed in the existing `o.e.pde.ds.annotations.tests`, @laeubi?

laeubi:
> In general I wonder why the test-class NameGeneratorTest is not simply placed in the existing o.e.pde.ds.annotations.tests, @laeubi?

Because it is a plain unit test and not a plugin test.

PR #1034, opened 2023-12-31
Target platform icon for "Load TP" job

Bananeweizen:
Register the typical target platform icon for the job family of the "Loading target platform" job. That leads to the icon being shown next to the progress bar of the progress view, similar to how the build has its own icon and all egit, m2e etc. jobs have their own icon.

grafik


ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target/LoadTargetDefinitionJob.java
```diff
@@ -185,4 +185,11 @@ private void resetPlatform(IProgressMonitor monitor) {
 		PDECore.getDefault().getFeatureModelManager().targetReloaded();
 	}
 
+	/**
+	 * @since 3.17
+	 */
+	public static Object getFamily() {
```
HannesWell, line 191: Instead of adding this new API I would prefer to move `JOB_FAMILITY_ID` as public constant to an internal class, e.g. `ICoreConstants`, and references it that here and int PDEPlugin.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
```diff
@@ -42,6 +43,7 @@
 import org.eclipse.ui.internal.views.log.ILogFileProvider;
 import org.eclipse.ui.internal.views.log.LogFilesManager;
 import org.eclipse.ui.plugin.AbstractUIPlugin;
+import org.eclipse.ui.progress.IProgressService;
```
HannesWell, line 46: I wonder why there is `org.eclipse.ui.progress.IProgressService` and `org.eclipse.e4.ui.progress.IProgressService` which looks very similar.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
```diff
@@ -214,6 +216,18 @@ public void start(BundleContext context) throws Exception {
...
+
+	protected void registerProgressIcon() {
+		if (!PlatformUI.isWorkbenchRunning()) {
+			return;
+		}
+		IProgressService service = PlatformUI.getWorkbench().getProgressService();
+		if (service == null) {
+			return;
+		}
+		service.registerIconForFamily(PDEPluginImages.DESC_TARGET_DEFINITION, LoadTargetDefinitionJob.getFamily());
```
HannesWell, line 230: Can't this be simply done in a static initializer in `LoadTargetDefinitionJob`?
This would also solve the problem where to place the job-family constant.

HannesWell (commented):
Thanks, in general a good idea.
Just a few remarks about how to implement this.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
Bananeweizen: The job lives in pde.core. The registration for the icon happens with the workbench progress service, which looks like a UI dependent class to me. I'm not sure if that service can be retrieved in the plain pde.core bundle without disturbing the workbench lifecycle. In egit we therefore even put that code into a service reacting to the application has started lifecycle.
Do you know better?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
laeubi: One is the "old" eclipse.ui and the other the "new" rcp E4 ... but I think the E4 one is not really completed and would need a declarative way (e4xmi) to add icons for job groups.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
laeubi: I think given the current structure, this is fine, an alternative might be to use an addon in the e4xmi but that really seems a bit overhead here...

laeubi (approved):

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPlugin.java
HannesWell: > Do you know better?

You are right and no, I don't know better in this case.

HannesWell (changes requested):
Since we can't add the icon in the Job itself please create constant for the job-family instead of adding a new API method `getFamily()`.
Besides that this change is fine.

laeubi:
@Bananeweizen can you apply the suggestion from @HannesWell so we can merge this?

PR #1035, opened 2024-01-02
Fixup IU location update handling

laeubi:
Currently the IU location is updated in-place that has several drawbacks:
- we need to synchronize a to not mess up things
- cached items in the object can getting stale and drip into new state
- equals and hasCode can become unstable over time

to overcome this the code is changed in a way that an update is performed on a copy of the current state and a new object is returned and used to perform the update.


merks (approved):
I have tested these changes as part of preparing this PR:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1693

The behavior is much improved and one can really automatically update the versions of the units when updating the location URI.  Thanks!!

Issue #1042, opened 2024-01-07
NPE in ApiBaseline on adding new JVM


iloveeclipse:
Using 4.31.0.I20240104-1800, I've added new JRE in Java settings and right after doing that got error below in the log.

```
org.eclipse.core.jobs
Error
Fri Jan 05 09:33:17 CET 2024
An internal error occurred during: "Performing API Analysis".

java.lang.NullPointerException: Cannot invoke "java.util.HashSet.add(Object)" because "this.fSystemPackageNames" is null
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.isSystemPackage(ApiBaseline.java:659)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.resolvePackage(ApiBaseline.java:555)
	at org.eclipse.pde.api.tools.internal.model.ApiType.resolveSuperType(ApiType.java:286)
	at org.eclipse.pde.api.tools.internal.model.ApiType.getSuperInterfaces(ApiType.java:254)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.collectAllInterfaces(ClassFileComparator.java:906)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.getInterfacesSet(ClassFileComparator.java:2048)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.checkSuperInterfaces(ClassFileComparator.java:298)
	at org.eclipse.pde.api.tools.internal.comparator.ClassFileComparator.getDelta(ClassFileComparator.java:1193)
	at org.eclipse.pde.api.tools.internal.provisional.comparator.ApiComparator$2.visit(ApiComparator.java:751)
	at org.eclipse.pde.api.tools.internal.model.ArchiveApiTypeContainer.accept(ArchiveApiTypeContainer.java:189)
	at org.eclipse.pde.api.tools.internal.provisional.comparator.ApiComparator.internalCompare(ApiComparator.java:635)
	at org.eclipse.pde.api.tools.internal.provisional.comparator.ApiComparator.compare(ApiComparator.java:284)
	at org.eclipse.pde.api.tools.internal.provisional.comparator.ApiComparator.compare(ApiComparator.java:307)
	at org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.checkCompatibility(BaseApiAnalyzer.java:1457)
[...]

PR #1043, opened 2024-01-07
NPE in ApiBaseline on adding new JVM

iloveeclipse:
An attempt to reduce the probability of errors while API baseline is being updated due the JVM change.

- Move rebindVM() to a job and let it wait for pending API analysis jobs
- changed most of internal maps/lists to concurrent versions
- made initialize() synchronized

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1042

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
@@ -28,21 +28,23 @@ import java.util.Dictionary;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Hashtable;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.IVMInstallChangedListener;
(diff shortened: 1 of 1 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
```diff
@@ -966,36 +947,37 @@ public void vmChanged(PropertyChangeEvent event) {
...
+				return;
+			}
+			fVMBinding = null;
+			IApiComponent[] components = getApiComponents();
+			HashSet<String> ees = new HashSet<>();
+			for (IApiComponent component2 : components) {
+				ees.addAll(Arrays.asList(component2.getExecutionEnvironments()));
+			}
+			resolveSystemLibrary(ees);
+		}).schedule();
```
laeubi, line 974: should one probably `join()` on the job here? Previously the caller could be sure that the method does not return before the rebind was done.

iloveeclipse: The only caller was a listener. For the listener on JRE changes it is irrelevant to wait on this code to complete.

The reason to use a job here was the possibility of a deadlocks here, after initialize() was made synchronized.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
```diff
@@ -177,6 +179,10 @@ public class ApiBaseline extends ApiElement implements IApiBaseline, IVMInstallC
 	public ApiBaseline(String name) {
 		super(null, IApiElement.BASELINE, name);
 		fComponentsProvidingPackageCache = new ConcurrentHashMap<>(8);
+		fSystemLibraryComponentList = new CopyOnWriteArrayList<>();
+		fComponentsById = new ConcurrentHashMap<>();
+		fAllComponentsById = new ConcurrentHashMap<>();
+		fComponentsByProjectNames = new ConcurrentHashMap<>();
```
HannesWell, line 185: You could already initialize these fields at their declaration. Would save the lines here and easier to determine what implementation is actually used.

iloveeclipse: I prefer to init in constructor, this saves you many jumps while debugging.

iloveeclipse:
There are no test fails, despite the "quality gate" reporting the opposite:

image

and the only "errors" are **unrelated** messages from maven:
```
[INFO] --- tycho-eclipse:4.0.5-SNAPSHOT:eclipse-build (generate-xmls) @ org.eclipse.pde.ds.tck ---
[ERROR] tck/OSGI-INF/impl-src/org/osgi/impl/bundle/component/annotations/HelloWorld10.java:27 Invalid component descriptor namespace: http://www.osgi.org/xmlns/scr/v1.0.0
[ERROR] tck/OSGI-INF/impl-src/org/osgi/impl/bundle/component/annotations/ServiceBundle.java:30 Property 'servicefactory' is ignored when non-default scope is specified.
[ERROR] tck/OSGI-INF/impl-src/org/osgi/impl/bundle/component/annotations/ServiceSingleton.java:30 Property 'servicefactory' is ignored when non-default scope is specified.
[INFO]
```

PR #1047, opened 2024-01-12
Provide only compatible JUnit 5 test engines in RemotePluginTestRunner

HeikoKlare:
Currently, it is possible that the `RemotePluginTestRunner` loads bundles with "incompatible" test engines. This means, the engine is version-incompatible with the `TestEngine` interface. In such a case, the execution will fail,  because the `ServiceLoader` tries to load this engine as a provider for the `TestEngine` service but fails because it cannot assign the engine to the `TestEngine` interface. One such case is the execution of tests with Tycho, which can use a different JUnit version than the one used in the JDT test bundles. 

This change ensures that bundles providing incompatible engines are not added to the class loader used for executing the test to ensure that service loading for the engines does not fail.

## How to Reproduce
An example for the error is the execution of session tests with JUnit 5, as proposed in eclipse-platform/eclipse.platform#1086. There, the session test `org.eclipse.core.tests.runtime.jobs.Bug_412138` is ported to JUnit 5, but fails because the `org.eclipse.tycho.surefire.osgibooter` is found as an engine provider and provides incompatible engine versions. See for example [this log](https://github.com/eclipse-platform/eclipse.platform/actions/runs/7492899271/job/20397430481):
```
java.util.ServiceConfigurationError: org.junit.platform.engine.TestEngine: org.junit.jupiter.engine.JupiterTestEngine not a subtype
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:593)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1244)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1273)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1309)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1393)
	at java.base/java.lang.Iterable.forEach(Iterable.java:74)
	at org.junit.platform.launcher.core.LauncherFactory.collectTestEngines(LauncherFactory.java:160)
```
With this fix applied, the test runs fine, as the the engine is not loaded from the `org.eclipse.tycho.surefire.osgibooter` bundle's class loader, but from the one of the test engine bundles.

### DIFF
--- a/ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
@@ -12,4 +12,5 @@ Export-Package: org.eclipse.pde.internal.junit.runtime;x-internal:=true
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-ActivationPolicy: lazy
 Import-Package: org.eclipse.ui.testing;resolution:=optional
+DynamicImport-Package: org.junit.platform.engine
 Automatic-Module-Name: org.eclipse.pde.junit.runtime
--- a/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
+++ b/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
@@ -14,9 +14,12 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.junit.runtime;
 
+import static java.util.stream.Collectors.toCollection;
+
 import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.List;
@@ -99,19 +102,34 @@ public class RemotePluginTestRunner extends RemoteTestRunner {
 
 	private static List<Bundle> findTestEngineBundles() {
 		BundleContext bundleContext = FrameworkUtil.getBundle(RemotePluginTestRunner.class).getBundleContext();
-		List<Bundle> engineBundles = new ArrayList<>();
-		for (Bundle bundle : bundleContext.getBundles()) {
-			try {
-				BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
-				Collection<String> listResources = bundleWiring.listResources("META-INF/services", "org.junit.platform.engine.TestEngine", BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$//$NON-NLS-2$
-				if (!listResources.isEmpty()) {
-					engineBundles.add(bundle);
-				}
-			} catch (Exception e) {
-				// check the next bundle
+		return Arrays.stream(bundleContext.getBundles()).filter(RemotePluginTestRunner::providesCompatibleTestEngine).collect(toCollection(ArrayList::new));
+	}
+
+	/**
+	 * Checks whether the bundle provides test engines.
+	 * Ensures that test engines that can be loaded from the bundle
+	 * are compatible with the TestEngine version in current scope.
+	 * Otherwise, the JUnit platform's call to the ServiceLoader for
+	 * retrieving available engines will fail.
+	 * Incompatibilities can happen, e.g., in Tycho builds, where
+	 * the org.eclipse.tycho.surefire.osgibooter bundle is found
+	 * that may provide a different JUnit platform version than the
+	 * one available via the Eclipse target platform.
+	 */
+	private static boolean providesCompatibleTestEngine(Bundle bundle) {
+		try {
+			BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
+			String testEngineClass = "org.junit.platform.engine.TestEngine"; //$NON-NLS-1$
+			Collection<String> engineProviders = bundleWiring.listResources("META-INF/services", testEngineClass, BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$
+			if (!engineProviders.isEmpty()) {
+				Class<?> thisTestEngine = Class.forName(testEngineClass);
+				Class<?> bundleTestEngine = bundle.loadClass(testEngineClass);
+				return thisTestEngine == bundleTestEngine;
 			}
+		} catch (Exception e) {
+			// skip this bundle
 		}
-		return engineBundles;
+		return false;
 	}
 
 	/**


HeikoKlare:
@akurtakov @HannesWell May I ask (one of) you to have a look at this? Having this merged would allow me to proceed working on modernizing (i.e., migrating to JUnit 5) platform's session tests: eclipse-platform/eclipse.platform#1086

HannesWell:
> @akurtakov @HannesWell May I ask (one of) you to have a look at this? Having this merged would allow me to proceed working on modernizing (i.e., migrating to JUnit 5) platform's session tests: [eclipse-platform/eclipse.platform#1086](https://github.com/eclipse-platform/eclipse.platform/pull/1086)

I'll have a look at it over the weekend.
The main point that I want to check is if the new dependency is trivial. The reason is that `org.eclipse.pde.junit.runtime` runs in JVM's launched from a dev-eclipse together with the bundles from the TP. Therefore they can be (much) older than those currently in use and (probably) not even contain the JUnit-5 bundles.
But I have to check all of that.

HeikoKlare:
Thanks, Hannes!
I see, that's an interesting point. I can also change the implementation to be more resilient and also work when JUnit 5 bundles are not available. Then the dependency can either be made optional or replaced by using fully qualified names instead of the `TestEngine` class.

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
```diff
@@ -99,19 +104,38 @@ private static ClassLoader createJUnit5PluginClassLoader(String testPluginName)
...
+	 * that may provide a different JUnit platform version than the
+	 * one available via the Eclipse target platform.
+	 */
+	private static boolean providesCompatibleTestEngine(Bundle bundle) {
+		try {
+			BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
+			Collection<String> engineProviderSpecifications = bundleWiring.listResources("META-INF/services", "org.junit.platform.engine.TestEngine", BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$//$NON-NLS-2$
+			if (!engineProviderSpecifications.isEmpty()) {
+				Bundle engineBundleInThisClassLoader = FrameworkUtil.getBundle(TestEngine.class);
+				Class<?> testEngineClassInCurrentLoader = getTestEngineClassInBundle(engineBundleInThisClassLoader);
```
HannesWell, line 127: This will probably just return the `TestEngine.class` class.
```suggestion
				Class<?> testEngineClassInCurrentLoader = TestEngine.class;
```

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
```diff
@@ -99,19 +104,38 @@ private static ClassLoader createJUnit5PluginClassLoader(String testPluginName)
...
-		for (Bundle bundle : bundleContext.getBundles()) {
-			try {
-				BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
-				Collection<String> listResources = bundleWiring.listResources("META-INF/services", "org.junit.platform.engine.TestEngine", BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$//$NON-NLS-2$
-				if (!listResources.isEmpty()) {
-					engineBundles.add(bundle);
-				}
-			} catch (Exception e) {
-				// check the next bundle
+		return Arrays.stream(bundleContext.getBundles()).filter(RemotePluginTestRunner::providesCompatibleTestEngine).collect(toList());
```
HannesWell, line 107: Using `Collectors.toCollection(ArrayList::new)` would avoid the need to change the caller
```suggestion
		return Arrays.stream(bundleContext.getBundles()).filter(RemotePluginTestRunner::providesCompatibleTestEngine).collect(toCollection(ArrayList::new));
```

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
```diff
@@ -99,19 +104,38 @@ private static ClassLoader createJUnit5PluginClassLoader(String testPluginName)
...
+	 */
+	private static boolean providesCompatibleTestEngine(Bundle bundle) {
+		try {
+			BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
+			Collection<String> engineProviderSpecifications = bundleWiring.listResources("META-INF/services", "org.junit.platform.engine.TestEngine", BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$//$NON-NLS-2$
+			if (!engineProviderSpecifications.isEmpty()) {
+				Bundle engineBundleInThisClassLoader = FrameworkUtil.getBundle(TestEngine.class);
+				Class<?> testEngineClassInCurrentLoader = getTestEngineClassInBundle(engineBundleInThisClassLoader);
+				Class<?> testEngineInEngineProviderBundle = getTestEngineClassInBundle(bundle);
+				return testEngineInEngineProviderBundle != null && testEngineClassInCurrentLoader.isAssignableFrom(testEngineInEngineProviderBundle);
```
HannesWell, line 129: In order to avoid a hard dependency on `org.junit.platform.engine` I suggest using the following code.

```suggestion
			String testEngineClass = "org.junit.platform.engine.TestEngine"; //$NON-NLS-1$
			Collection<String> engineProviders = bundleWiring.listResources("META-INF/services", testEngineClass, BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$
			if (!engineProviders.isEmpty()) {
				Class<?> thisTestEngine = Class.forName(testEngineClass);
				Class<?> bundleTestEngine = bundle.loadClass(testEngineClass);
				return bundleTestEngine != null && thisTestEngine.isAssignableFrom(bundleTestEngine);
			}
```

I also suggest to shorten the variable names a bit. While descriptive names are definitively important, too long names are IMHO badly readable as well.

ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
```diff
@@ -11,5 +11,6 @@ Require-Bundle: org.eclipse.jdt.junit.runtime;bundle-version="[3.4.0,4.0.0)",
 Export-Package: org.eclipse.pde.internal.junit.runtime;x-internal:=true
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-ActivationPolicy: lazy
-Import-Package: org.eclipse.ui.testing;resolution:=optional
+Import-Package: org.eclipse.ui.testing;resolution:=optional,
+ org.junit.platform.engine
```
HannesWell, line 15: The mandatory import should be avoided. In general we could also use an optional import, but since we don't need any class from that package at compile-time I suggest to use [`DynamicImport-Package`](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.dynamicimportpackage).
```suggestion
Import-Package: org.eclipse.ui.testing;resolution:=optional
DynamicImport-Package: org.junit.platform.engine
```

HannesWell (changes requested):
OK, I checked the JUnit Plugin test launchers and it looks like as expected that one can launch such test without Junit-5.
But depending on the selected JUnit version in the launch-config there is code to ensure the required bundles for that version are present:

https://github.com/eclipse-pde/eclipse.pde/blob/a53dc16dbd82f8194758710e24a79143a94e6773/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java#L535-L549

and

https://github.com/eclipse-pde/eclipse.pde/blob/a53dc16dbd82f8194758710e24a79143a94e6773/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java#L501-L505

So yes, the code should be prepared to handle a completely unavailable `TestEngine` class and should not add a mandatory requirement to its package.
I added suggestions below.

But in general I'm surprised this code is used directly in Tycho.

HeikoKlare:
@HannesWell Thanks for the in-depth review and your suggestions. I have incorporated the changes in f4e9f33 so that no strict dependency to JUnit 5 is required anymore.

> But in general I'm surprised this code is used directly in Tycho.

Just for clarificaiton: It is not used "directly" in Tycho, but rather "indirectly". One consumer is the session test framework, which starts a remote test run via the `RemotePluginTestRunner` from within a test case, and that test case is executed by Tycho. Maybe it would also be possible to "cleanup" the classloader in the session test framework when starting a remote test run, but I thought there may be other situations than a Tycho run in which incompatible engine versions are on the classpath, so it makes sense to make the runner capable of dealing with that situation.

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
HeikoKlare: Yes, that seems to be useless and can be simplified.

ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
HeikoKlare: Thanks for the pointer to the dynamic import option. It's perfectly reasonable here.

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
HeikoKlare: Incorporated in f4e9f3358757e62cc21c0e9e0be12a0baac96a31

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
```diff
@@ -99,19 +102,34 @@ private static ClassLoader createJUnit5PluginClassLoader(String testPluginName)
...
+	 */
+	private static boolean providesCompatibleTestEngine(Bundle bundle) {
+		try {
+			BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
+			String testEngineClass = "org.junit.platform.engine.TestEngine"; //$NON-NLS-1$
+			Collection<String> engineProviders = bundleWiring.listResources("META-INF/services", testEngineClass, BundleWiring.LISTRESOURCES_LOCAL); //$NON-NLS-1$
+			if (!engineProviders.isEmpty()) {
+				Class<?> thisTestEngine = Class.forName(testEngineClass);
+				Class<?> bundleTestEngine = bundle.loadClass(testEngineClass);
+				return thisTestEngine != null && bundleTestEngine != null && thisTestEngine.isAssignableFrom(bundleTestEngine);
```
HannesWell, line 127: It looks like `Class.forName()` never returns null.
Further I assume there are no class loading tricks where a different `TestEngine` class is a super class of this bundle's one? Therefore an equals (maybe even identity?) check should be sufficient, shouldn't it?
```suggestion
				return thisTestEngine.equals(bundleTestEngine);
```

HannesWell (commented):
> Maybe it would also be possible to "cleanup" the classloader in the session test framework when starting a remote test run, but I thought there may be other situations than a Tycho run in which incompatible engine versions are on the classpath, so it makes sense to make the runner capable of dealing with that situation.

Thanks for the clarification. I think it is ok to do it here. This should not harm normal situations.

In general the change now looks good, I just have one small point open.

ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/RemotePluginTestRunner.java
HeikoKlare: Yes, an identity comparison is sufficient here (fixed with 19e3dd5). I had two false assumptions:
* I accidentally expected `isAssignableFrom()` to return true for two instances of the same class loaded by different class loaders. Since that's the operation used by the `ServiceLoader` that fails when incompatible engine versions are used, I reused it here.
* I expected `Class.forName()` to return null if the engine is not on the classpath, but it will of course throw an exception. A scenario where I would expect this happens to is as follows: execute the runner without any JUnit 5 dependencies, but use the argument `-runasjunit5` and have some bundle specifying a service provider for "org.junit.platform.engine.TestEngine".

HannesWell: > A scenario where I would expect this happens to is as follows: execute the runner without any JUnit 5 dependencies, but use the argument -runasjunit5 and have some bundle specifying a service provider for \"org.junit.platform.engine.TestEngine\".

I assume each bundle providing a TestEngine requires that class to implement it. And therefore, if the OSGi metadata are correct, it should depend on the providing bundle/package.
So I don't expect that to happen, but we'll see what specialties occur. ^^

HeikoKlare: > So I don't expect that to happen

Me neither 🙂 I should have mentioned that the situation may only occur with invalid metadata (i.e., the declaration of a service that is actually not provided). And even for such a very unlikely, accidental case it might be good to fail in a more "useful" way than throwing an NPE.

And, of course, thanks for merging!

PR #1054, opened 2024-01-15
Import support for Auxiliary Plugins in PDE/BND setups

laeubi:
BND allows arbitrary plugins to be provided, to account for this there is a special mechanism for an OSGi classpath that is a package is exported with `<package>;bnd-plugins=true` and a weaving hook is used to add additional dynamic imports.

This is specifically required for the `biz.aQute.repository` bundle that provides e.g. maven repository support.

CQ created here:
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/12777


laeubi:
CQ accepted!

PR #1055, opened 2024-01-15
Add support for using classpath of plain bnd projects

laeubi:
Currently if one imports a blain bnd-project the classpath is not setup unless one have some other tooling installed (e.g. bndtools)

This now adds support for evaluate the bnd project in case it is a "plain" bnd workspace project and adds the required items to the classpath container if it was added to the project.

To effectively use this feature one would additionally need:
- [x] https://github.com/eclipse-pde/eclipse.pde/pull/1054
- [ ] A way to update the classpath (currently Project > Plugin Tools > Update Classpath...) does not consider such projects
- [x] Add an option to enable the Plugin Dependencies container, currently one needs to enable it by manually edit the  classpath and project files, maybe in combination with previous entry it could just be added if update classpath is choosen

### DIFF
--- a/ui/org.eclipse.pde.core/.settings/.api_filters
+++ b/ui/org.eclipse.pde.core/.settings/.api_filters
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <component id="org.eclipse.pde.core" version="2">
-    <resource path="src/org/eclipse/pde/core/ModelChangedEvent.java" type="org.eclipse.pde.core.ModelChangedEvent">
-        <filter comment="This is PDEs own implementation" id="576725006">
-            <message_arguments>
-                <message_argument value="IModelChangedEvent"/>
-                <message_argument value="ModelChangedEvent"/>
-            </message_arguments>
-        </filter>
-    </resource>
     <resource path="src/org/eclipse/pde/internal/core/project/BundleProjectService.java" type="org.eclipse.pde.internal.core.project.BundleProjectService">
         <filter comment="Platform Team allows use of bundle importers for PDE import from source repository" id="640712815">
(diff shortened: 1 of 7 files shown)


laeubi:
It seems there is a "Fix Project Setup..." if one has compile problems this is the place where PDE should plug in I think...

laeubi:
> It seems there is a "Fix Project Setup..." if one has compile problems this is the place where PDE should plug in I think...

This is done now but requires that the plugin nature is already added, it would be usefull to have a Configure > Add Plugin nature (or even a generic add nature)....

laeubi:
Support to update the classpath is a bit messy as this assumes we have a plugin model, that is then passed in a long chain down to the final update that does a lot more than just refresh the classpath container.

So this is a more general one on if/how we can get a `IPluginModelBase` from a plain bnd project at all.

merks:
> This is done now but requires that the plugin nature is already added, it would be usefull to have a Configure > Add Plugin nature (or even a generic add nature)....

When does Configure -> Convert to Plug-in Projects kick in?

laeubi:
> > This is done now but requires that the plugin nature is already added, it would be usefull to have a Configure > Add Plugin nature (or even a generic add nature)....
> 
> When does Configure -> Convert to Plug-in Projects kick in?

It does a lot more than just adding a nature (e.g. adds manifests, builder and so on) I not yet fully understand the details but in the past it has worked not very well and needed careful review.

Also I think adding/removing a nature is something that might be useful in general, at the moment one has to implement it for each and every plugin again...

laeubi:
I now plan to merge this as is as it actually works already if making the changes manual, after looking at the alternatives, having an action to add the classpath container might be the best, a nature is not required for this to work.

Issue #898, opened 2023-11-07
Migrate the e4 tools to PDE


vogella:
e4 tools should IMHO be part of of PDE. These tools include for example,

* The application model editor
* Templates for model fragments

They were originally migrated to platform because of "the e4 people are mainly active in platform" (quote from memory).

I suggest we migrate them to PDE so that the plug-in tools get complete.

laeubi:
Can you explain what PDE has to do with E4? I don't really see any relation here.

vogella:
The e4 tools provide the tools to develop e4 application and plug-ins, e.g., the wizards to create application model fragments and the model editor. Did the "e4" part confuses you? It used to be the project name which developed the tools. The spies were also part of the e4 tools and we migrated them a while ago

mickaelistria:
PDE is about tools for Eclipse plugin development. e4 tools are tools for e4 plugin development, maintenance and monitoring. So any tool that is not targetting end-user of the IDE is most likely better fit in PDE.

laeubi:
>  So any tool that is not targetting end-user of the IDE is most likely better fit in PDE.

I must disagree here. One can use PDE for plain OSGi devlopment (and I do this for years) while E4 is a way to develop the RCP Framework what is a UI Framework ...

So if we decide that E4/RCP is not part of platform, it might be better to have a dedicated E4/RCP project than to move responsibilities to PDE that is already low on contribution resources.

PDE **might** include some extensions to E4 (spies?) that are closely related to OSGi, but pushing the burden of E4 UI in general is a bit to much, or for the same arguments we should push all P2 UI > Platform UI because these are actually UIs used in the Eclipse IDE ...

vogella:
> One can use PDE for plain OSGi devlopment 

Yes, and for pure 3.x  RCP development

akurtakov:
Being part of PDE or Platform would not affect people working on it as it's pretty much the same group. The separation should be functional in order to make our build and release procedures more  straightforward. From https://eclipse.dev/pde/ :
The Plug-in Development Environment (PDE) provides tools to create, develop, test, debug, build and deploy Eclipse plug-ins, fragments, features, update sites and RCP products. 

^^ Should be the guiding force whether something belongs in PDE or not.

laeubi:
Well then we can actually merge everything to PDE that would simplify things a lot ;-)

[...]

PR #1060, opened 2024-01-18
Write latest master build to downloads.eclipse.org

akurtakov:
Using https://download.eclipse.org/pde/builds/master/ as destination so it can be tested more broadly.
Mainly motivated to have a permanent download space which wasn't there as discovered in
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4172

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,15 +16,16 @@ pipeline {
 		stage('Build') {
 			steps {
 				wrap([$class: 'Xvnc', useXauthority: true]) {
-					sh '''
+					sh """
 						mvn clean verify --batch-mode -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 						-Pbree-libs \
 						-Papi-check \
 						-Pjavadoc \
+						${env.BRANCH_NAME=='master' ? '-Peclipse-sign': ''} \
 						-Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true \
 						-Dtycho.debug.artifactcomparator \
 						-Dpde.docs.baselinemode=fail
-					'''
+					"""
 				}
 			}
 			post {
@@ -42,5 +43,18 @@ pipeline {
 				}
 			}
 		}
+		stage('Deploy') {
+			when {
+				branch 'master'
+			}
+            steps {
+                sshagent ( ['projects-storage.eclipse.org-bot-ssh']) {
+                     sh '''
+                        ssh genie.pde@projects-storage.eclipse.org "rm -rf /home/data/httpd/download.eclipse.org/pde/builds/master/*"
+                        scp -r repository/target/repository/* genie.pde@projects-storage.eclipse.org:/home/data/httpd/download.eclipse.org/pde/builds/master/
+                        '''
+                }
+            }
+        }
 	}
 }
--- a/e4tools/features/org.eclipse.e4.tools.persistence.feature/forceQualifierUpdate.txt
+++ b/e4tools/features/org.eclipse.e4.tools.persistence.feature/forceQualifierUpdate.txt
@@ -1,1 +1,2 @@
 # To force a version qualifier update add the bug here
+Touch to publish source feature.
--- a/repository/category.xml
+++ b/repository/category.xml
@@ -6,6 +6,10 @@
    <feature id="org.eclipse.pde.spies.source" version="0.0.0"/>
    <feature id="org.eclipse.pde.unittest.junit" version="0.0.0"/>
    <feature id="org.eclipse.pde.unittest.junit.source" version="0.0.0"/>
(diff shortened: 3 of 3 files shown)


akurtakov:
@HannesWell wdyt? It's not giving any immediate benefits but it's one more step towards decoupling it from Platform. 
Once this is in I plan on trying to add this repo in releng.aggregator and have the bundles be used from it instead of rebuilt during I-build.

HannesWell (commented):
> @HannesWell wdyt? It's not giving any immediate benefits but it's one more step towards decoupling it from Platform. 

Regarding this change, the binaries published should probably be jar-signed, shouldn't they?

> Once this is in I plan on trying to add this repo in releng.aggregator and have the pde bundles be used from it instead of rebuilt during I-build.

Please don't do that before the cycles are cut. As long as there are other plugins in the SDK that depend on PDE, we should not do that. Otherwise we probably make the situation worse than it is now.

Looking at the Dependencies of `pde.build` (which is the plugin on that actually every other plugin in PDE depends on) I see mainly the following plugins with PDE dependencies:
- org.eclipse.e4.tools
- org.eclipse.e4.tools.emf.editor3x
- org.eclipse.e4.tools.emf.ui

Plus ant.test.core/ui and jdt.apt.tests.
And having a quick look at that I would say that it is probably possible, but not trivial to change.

grafik

laeubi:
> I see mainly the following plugins with PDE dependencies:

See
- https://github.com/eclipse-pde/eclipse.pde/issues/898

Jenkinsfile
```diff
@@ -41,5 +41,16 @@ pipeline {
...
 			}
 		}
+		stage('Deploy') {
+			when {
+				branch 'master'
+			}
+            steps {
+                sshagent ( ['projects-storage.eclipse.org-bot-ssh']) {
+                    sh 'ssh genie.pde@projects-storage.eclipse.org rm -rf /home/data/httpd/download.eclipse.org/pde/builds/master/*'
+                    sh 'scp -r repository/target/repository/* genie.pde@projects-storage.eclipse.org:/home/data/httpd/download.eclipse.org/pde/builds/master/'
```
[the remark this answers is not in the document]
akurtakov: Applied.

akurtakov:
> Regarding this change, the binaries published should probably be jar-signed, shouldn't they?

Also done.

Jenkinsfile
```diff
@@ -21,6 +21,7 @@ pipeline {
 						-Pbree-libs \
 						-Papi-check \
 						-Pjavadoc \
+						-Peclipse-sign \
```
laeubi, line 24: I think this should only be done on `MASTER` branch!

laeubi:
> Please don't do that before the cycles are cut. As long as there are other plugins in the SDK that depend on PDE, we should not do that. Otherwise we probably make the situation worse than it is now.

See
- https://github.com/eclipse-pde/eclipse.pde/pull/1081

Jenkinsfile
HannesWell: If you change the multi-line string with leading/trailing three single-quotes to an interpolated GroovyString with three double-quotes you could probably something like.
```suggestion
						${env.BRANCH_NAME=='master' ? '-Peclipse-sign': ''} \
```

laeubi:
@akurtakov now the e4 tools are part of this repo we maybe want to include the features in the updatesite

Jenkinsfile
akurtakov: Done

akurtakov:
> @akurtakov now the e4 tools are part of this repo we maybe want to include the features in the updatesite

Should be done now.

HannesWell (approved):
Looks good for a first attempt.

But on the long run we should discuss and have a more sophisticated repo structure.

akurtakov:
> Looks good for a first attempt.
> 
> But on the long run we should discuss and have a more sophisticated repo structure.

Definetly, esp with regards of releases and etc. We are far from such state yet and every step that would make things easier to go to the next step while not causing issue for anything else should be taken.

laeubi (approved):
Looks good for me

laeubi:
I don't know yet why but now master build fails :-\

https://ci.eclipse.org/pde/job/eclipse.pde/job/master/554/

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/pull/1095

akurtakov:
It still fails even after the test fix is in and this is most likely caused by signing as it fails in the very same way as figured in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1673 .

akurtakov:
If I fail to figure what's going wrong with comparison in the next few hours I'll revert the signing part only.

PR #1062, opened 2024-01-21
Support finding a model by Resource instead of BundleDescription

laeubi:
BundleDescription belongs to the old resolver API instead we should use the OSGi Resource API.

This does the following:
- add a new method to find the model by Resource
- deprecate for removal the old find method
- use adapter for internal usage of BundleDescription to lookup the model
- clear references to the deprecated method

This will allow to adapt PDE in a more flexible way to other plugin models (e.g. BND / tycho / p2 / ... )

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
@@ -28,6 +28,7 @@ import org.eclipse.pde.internal.core.build.WorkspaceBuildModel;
 import org.eclipse.pde.internal.core.project.PDEProject;
 import org.eclipse.pde.internal.core.util.VersionUtil;
 import org.osgi.framework.Version;
+import org.osgi.resource.Resource;
 
 /**
  * The central access point for models representing plug-ins found in the workspace
@@ -118,13 +119,30 @@ public class PluginRegistry {
 	/**
 	 * Returns a plug-in model associated with the given bundle description
 	 *
-	 * @param desc the bundle description
+	 * @param desc
+	 *            the bundle description
 	 *
-	 * @return a plug-in model associated with the given bundle description or <code>null</code>
-	 * 			if none exists
+	 * @return a plug-in model associated with the given bundle description or
+	 *         <code>null</code> if none exists
+	 * @deprecated Instead use {@link #findModel(Resource)}
 	 */
+	@Deprecated(forRemoval = true)
 	public static IPluginModelBase findModel(BundleDescription desc) {
-		return PDECore.getDefault().getModelManager().findModel(desc);
+		return findModel((Resource) desc);
+	}
+
+	/**
+	 * Returns a plug-in model associated with the given resource
+	 *
+	 * @param resource
+	 *            the (OSGi) resource to find a model for
+	 *
+	 * @return a plug-in model associated with the given bundle description or
+	 *         <code>null</code> if none exists
+	 * @since 3.18
+	 */
+	public static IPluginModelBase findModel(Resource resource) {
+		return PDECore.getDefault().getModelManager().findModel(resource);
 	}
 
 	/**
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/TargetRepository.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/TargetRepository.java
@@ -19,6 +19,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -39,6 +41,7 @@ import java.util.jar.Manifest;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.osgi.service.resolver.BundleDescription;
 import org.eclipse.osgi.service.resolver.State;
 import org.eclipse.pde.core.plugin.IPluginLibrary;
@@ -50,6 +53,7 @@ import org.osgi.resource.Capability;
 import org.osgi.resource.Requirement;
 import org.osgi.resource.Resource;
 import org.osgi.service.repository.ContentNamespace;
+import org.osgi.service.repository.Repository;
 import org.osgi.service.repository.RepositoryContent;
 
 import aQute.bnd.osgi.Instruction;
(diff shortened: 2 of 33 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
```diff
@@ -117,13 +118,29 @@ public static IPluginModelBase findModel(IProject project) {
...
 	 * Returns a plug-in model associated with the given bundle description
 	 *
-	 * @param desc the bundle description
+	 * @param desc
+	 *            the bundle description
 	 *
-	 * @return a plug-in model associated with the given bundle description or <code>null</code>
-	 * 			if none exists
+	 * @return a plug-in model associated with the given bundle description or
+	 *         <code>null</code> if none exists
```
HannesWell, line 125: Please also add a removal notice in the Java-doc, this allows JDT migrate existing callers via clean-ups.
```suggestion
	 *         <code>null</code> if none exists
	 * @deprecated Instead use {@link #findModel(Resource)}
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/TargetRepository.java
```diff
@@ -303,6 +309,35 @@ public InputStream getContent() {
...
+				if (location != null) {
+					Path path = Path.of(location);
+					if (Files.isRegularFile(path)) {
+						return adapter.cast(path);
+					}
+				}
+			}
+			if (adapter == Repository.class) {
+				return adapter.cast(repository);
+			}
```
HannesWell, line 337: Are the adapters to File, Path and Repo are currently in use?

HannesWell (commented):
Thanks for this change. This is definitively something that needs to be done in order to migrate off the old equinox resolver.

I'm just not sure if the generality of replacing `BundleDescription` by `Resource` is required or if replacing it with `BundleRevision` would be sufficient.
At the moment only `BundleDescription` instances are stored in the `PluginModelManager` and when we migrate that (probably) to `BundleRevision` only `BundleRevision` instances will be stored there. Of course Adapters could also be more specific if `BundleRevision` is used.

I see some places where a caller has to do a cast since the passed variable cannot be entirely be replaced with a Resource (because e.g. `getSymbolicName()` is used).

laeubi:
> I'm just not sure if the generality of replacing `BundleDescription` by `Resource` is required or if replacing it with `BundleRevision` would be sufficient.

 `BundleRevision` for example will again require a state where a bundle object can be acquired. I think the PDE way to bind itself to a resolved OSGi state is simply wrong, Tycho does not need this and still can work without it.

So `Resource` is much more generic and not bound to the wiring API so can be used anywhere, e.g. a Target Platform is actually a set of `Resource`s and would immediately be usable without any special handling on a pure metadata base. So the less specific we are the more generic we can craft methods so they can be used in any context.

> I see some places where a caller has to do a cast since the passed variable cannot be entirely be replaced with a Resource (because e.g. getSymbolicName() is used).

The BSN can be derived from a Resource by `aQute.bnd.osgi.resource.ResourceUtils.getIdentity(Resource)` but I don't want to change to much.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/TargetRepository.java
laeubi: Not yet but they are handy for further improvements.

HannesWell:
> > I'm just not sure if the generality of replacing `BundleDescription` by `Resource` is required or if replacing it with `BundleRevision` would be sufficient.
> 
> `BundleRevision` for example will again require a state where a bundle object can be acquired. I think the PDE way to bind itself to a resolved OSGi state is simply wrong, Tycho does not need this and still can work without it.

Please elaborate why you think it is simply wrong? Looking at Tycho I still see the `DependencyComputer` and the `EquinoxResolver` using the Wiring API.

> 
> So `Resource` is much more generic and not bound to the wiring API so can be used anywhere, e.g. a Target Platform is actually a set of `Resource`s and would immediately be usable without any special handling on a pure metadata base. So the less specific we are the more generic we can craft methods so they can be used in any context.
>

I understand and support the desire be as generic as possible, but in this case I don't think it is sufficient to just use `Resource` in PDE. While the TP is indeed a set of Resources at some point we always need to know how/by what requirements are satisfied, e.g. for the PDE-ClasspathContainer, the Dependencies View or Product launches, and for that I don't see another way than the Wiring API.
I also tried to rewrite PDE's DependencyManager only using the Resources API, but I cannot see how this feasible without `BundleWiring`/`BundleWire`.

laeubi:
> Please elaborate why you think it is simply wrong? 

Because you don't need a resolved state with bundles to compute the required information, it is already there in the meta-data, currently this requires PDE to download *all* bundles of a target platform even if only a few are used to resolve the projects classpath. Also this currently hinders PDE from using more than one target platform, e.g. you must always have one big-fat target that can satisfy all your workspace bundles, using different sets of bundles in the same workspace is not possible often without much efforts.

> looking at Tycho I still see the DependencyComputer and the EquinoxResolver using the Wiring API.

But this is only used to compute the access restriction rules and is even there to be replaced by OSGi Resource Model:
- https://github.com/eclipse-tycho/tycho/issues/1243

> While the TP is indeed a set of Resources at some point we always need to know how/by what requirements are satisfied, e.g. for the PDE-ClasspathContainer, the Dependencies View or Product launches, and for that I don't see another way than the Wiring API.

See [Resolver Service Specification ](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html) and especially [58.3.2 Resolving](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html#d0e50395)   and [ResolveContext.getWirings](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html#org.osgi.service.resolver.ResolveContext.getWirings--) that all works on pure Resource API without heavyweight Bundle Objects.

HannesWell (approved):
> > While the TP is indeed a set of Resources at some point we always need to know how/by what requirements are satisfied, e.g. for the PDE-ClasspathContainer, the Dependencies View or Product launches, and for that I don't see another way than the Wiring API.
> 
> See [Resolver Service Specification ](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html) and especially [58.3.2 Resolving](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html#d0e50395) and [ResolveContext.getWirings](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html#org.osgi.service.resolver.ResolveContext.getWirings--) that all works on pure Resource API without heavyweight Bundle Objects.

Indeed, you are right.
I forgot that there is a `Wiring` abstraction in the resources package on top of `BundleWiring`. With that for example the mention migration of `DependencyManager` is possible.
With that I'm fine with this change.

I have also created https://github.com/eclipse-pde/eclipse.pde/issues/1069, to track all the efforts and discuss general solutions for the migration off the equinox resolver API.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/bnd/TargetRepository.java
HannesWell: Then please let's add this as soon as it is used.
In case it isn't used an unlucky person will have a hard time in the future to find out if this is used or not.

PR #1066, opened 2024-01-22
A noextend API class that implements noimplement interface isn't a leak

merks:
https://github.com/eclipse-pde/eclipse.pde/discussions/1064


merks:
@laeubi 

This fixes the problem you noticed:

image

I.e., no more bogus warning.

PR #1075, opened 2024-01-26
Add an option to the product editor to flag it contain a JRE

laeubi:
Currently if one wants to include a JRE it is needed to specify a concrete EE for each of the linux/win/mac what is quite cumbersome and could easily forgotten to update.

Instead this adds a new flag "includeJRE" in the product that enables this for all os at once and always uses the default vm (what is the one from the target platform or user chosen default).

This will also make it easier for Tools like Tycho that already investigate the 'includeLaunchers' flag.

grafik

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
@@ -2120,6 +2120,7 @@ ProductInfoSection_new=New...
 ProductInfoSection_app=Application:
 ProductInfoSection_appTooltip=The application extension identifier
 ProductInfoSection_launchers=The product includes native launcher artifacts
+ProductInfoSection_jre=The product includes a JRE
 SplashSection_title=Location
 SplashSection_desc=The splash screen appears when the product launches. Specify the plug-in in which the splash screen is located.
 SplashSection_plugin=Plug-in:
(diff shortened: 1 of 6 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -2120,6 +2120,7 @@ ProductInfoSection_new=New...
 ProductInfoSection_app=Application:
 ProductInfoSection_appTooltip=The application extension identifier
 ProductInfoSection_launchers=The product includes native launcher artifacts
+ProductInfoSection_jre=The product includes a JRE
```
HannesWell, line 2123: A more active label would IMO be more suitable. I was always wondering why the launcher label is so passive.
```suggestion
ProductInfoSection_launchers=Include native launcher artifacts
ProductInfoSection_jre=Include a JRE
```

HannesWell (commented):
In general this looks good.

But the `.api_filters` are unrelated, aren't they? Please add them in a separate PR, if necessary.

laeubi:
> But the `.api_filters` are unrelated, aren't they? Please add them in a separate PR, if necessary.

See https://github.com/eclipse-pde/eclipse.pde/pull/1077

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
laeubi: I think it is maybe to make it look a bit nicer in the UI and sounds a bit more friendly, also both flags are only relevant in case one exports the product or for tools like Tycho / pdebuild so they are more like a tag.

PR #1077, opened 2024-01-27
Fail the build on API warnings

laeubi:


HannesWell (approved):
Looks good. Thanks for this.
Having those warnings strictly enforced is great!

Issue #1073, opened 2024-01-25
API tooling reports 300 false positive errors on SWT


iloveeclipse:
The issue is known since years, but so far only was visible on my workspaces, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=532282.

Now it seem to appear for more users, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/995.

I believe long time ago I've seen extra handling of SWT project in PDE, because of the "special" SWT nature (separation of code between common part & OS specific fragments).

I assume something in this "special" PDE code for SWT is buggy.

Would be nice if someone could look at that.

iloveeclipse:
I had a chance to debug it a bit, here brain dump before weekend.

The comparator walks first over API baseline in `ApiComparator.internalCompare()`, collecting types into `typeRootBaseLineNames` set. What is not in the set is considered as "new" in the workspace and if the types have no "since" tags, that is reported as errors. That's what we get here - comparator does not "see" anything in the baseline!!!

It looks like `ApiDescription.findNode(IElementDescriptor, boolean)` can't find anything for given SWT type names because `ApiDescription.fPackageMap` is empty and so the call to `ApiDescription.resolveAnnotations(IElementDescriptor)` returns `null` in `ApiComparator.internalCompare()` - because of that, `visibility` of the type is set to zero and the code doesn't got to the line where the visited baseline type is added to the `typeRootBaseLineNames`.

Because of that, while the comparator walks over the second `typeRootContainers2` set of classes in the workspace, it can't find the type in the "known API types set" and so assumes all the API types in the workspace are added and with that, "since" tags are required.

The question to answer now is: why `ApiDescription.fPackageMap` is empty for baseline SWT types?

laeubi:
Maybe related to 
- https://github.com/eclipse-pde/eclipse.pde/issues/801

iloveeclipse:
> Maybe related to ...

What wonders me is: why some users are able to get API tools working for SWT, and some not - and note, I personally see the problem with SWT **randomly since years**. Only recently it became permanent to me - but not for others.

iloveeclipse:
Also note: https://github.com/eclipse-pde/eclipse.pde/issues/801 is about project container types from the workspace, here we miss package names for baseline (bundle) container types. The types from workspace **are** found but falsely recognized as "added" API types (because they were not found in the baseline).

laeubi:
[...]

PR #1079, opened 2024-01-27
Restore the AnyValue(String) constructor to support filter wildcards

merks:
https://github.com/eclipse-pde/eclipse.pde/issues/1073


gireeshpunathil (approved):

laeubi (approved):
Obviously no test has ever covered this, so I think this can go in as soon as build finished without compile errors, Jenkins might bail out about a maven warning I'm currently looking into.

PR #1080, opened 2024-01-27
Capture comparator errors and compile logs

laeubi:
... currently no baseline compare errors so have to wait for the problem to reoccur.


laeubi:
Master shows compare error so lets see if we can capture the data: https://ci.eclipse.org/pde/job/eclipse.pde/job/master/547/maven-warnings/source.42a6c304-3397-440d-a176-7e38c73001e8/#1720

PR #1081, opened 2024-01-28
Migrate the e4 tools to PDE

laeubi:
Please carefully review if this is the desired state, I'll now start to integrate this into the build ...

Fix https://github.com/eclipse-pde/eclipse.pde/issues/898


laeubi:
Build works now quality gate errors are expected and should be handled afterwards.

HannesWell (commented):
Thanks for working on this.
In general this looks good, but I suggest to move the e4-tools projects into a folder named `e4-tools` not just `tools`.
We already have `apitools`, so we should be specific here too.

Furthermore the Oomph setup has to be adjusted, but that can also be done subsequently.

laeubi:
> I suggest to move the e4-tools projects into a folder named `e4-tools` not just `tools`.
> We already have `apitools`, so we should be specific here too.

I actually don't want to complicate things too much, the files where already moved to ui then moved there and so on.

If we have merged this, we can of course think about moving thing again around but for now I think we should complete this asap to not run into merge conflicts.

iloveeclipse:
> If we have merged this, we can of course think about moving thing again

I would prefer we would move it only once

> around but for now I think we should complete this asap to not run into merge conflicts.

There is no rush, as long as few active committers can have a consensus how a merge should happen.

HannesWell:
> > If we have merged this, we can of course think about moving thing again
> 
> I would prefer we would move it only once

Me too.
Does everyone participating in this PR currently is fine with `e4-tools`?

@laeubi if you have to now time to move it I can adjust this PR for you quickly.

iloveeclipse:
And yes, looking on the new structure I agree with Hannes, "tools" is non saying, while "e4-tools" is much better.
To be consistent with "apitools" may be better to have "e4tools" (no dash)?
Beside this, history seem to be OK after merge.

HannesWell:
> To be consistent with "apitools" may be better to have "e4tools" (no dash)?

Thought about this as well and personally would have liked it more with a dash. But since renaming `apitools` just for that is not really worth the effort I'm also fine with `e4tools` to have it consistent.

merks:
I’m fine with or without the dash.

HannesWell:
@laeubi I have it almost ready locall, with an itermediate commit just after the merge to move it to `e4tools` (without dash).
If you are fine, I can force push it to this branch.

HannesWell:
In order to speed this up, I took the liberty to update this PR with the suggested move after the merge and adapted to following commits accordingly.

laeubi:
> I would prefer we would move it only once

That's the point it is **NOT** moved see:

https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/tools

I just imported all commits from the repository as is so don't want to add additional "noise" for the inital import.

laeubi:
So can we **PLEASE** delay any "make it look nicer" to follow up PRs?

Merging repositories is already a daunting task so adding addition "cleanup" as part of the merge is not really anything beneficial beside it is required for the build to succeed.

HannesWell:
> I just imported all commits from the repository as is so don't want to add additional "noise" for the inital import.

Importing these commits is already 'noise' and moving them around is another 'noise'. Therefore I would have done that (closely) together in one 'noise'.

> Merging repositories is already a daunting task so adding addition "cleanup" as part of the merge is not really anything beneficial beside it is required for the build to succeed.

Well that's why I tried to help you with that extra work, but if you think it is so much better to do that separately and adjust the build twice you can restore your previous state and it can be moved later.

laeubi:
> but if you think it is so much better to do that separately and adjust the build twice

I won't do it at all for a simple reason

1. we previously always have choose the name of the original repo that was "tools" 
2. we don't have e3tools either so e4tools is missleading, also the ui folder already contains some "e4 tools" so this makes it not clearer or better
3. we now have two times interrupting the history, e.g. look at [this file](https://github.com/eclipse-pde/eclipse.pde/commits/3d0cbb83b870d8e155926f905c414b6bd35845dd/e4tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java) it show:

Renamed from tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java
[(Browse History)](https://github.com/eclipse-pde/eclipse.pde/commits/13b9e5b3e2cab3904e08eb900365855297d076bf/tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java?browsing_rename_history=true&new_path=e4tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java&original_branch=3d0cbb83b870d8e155926f905c414b6bd35845dd)

if you then follow it it shows you the previous move from Lars:
Renamed from bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java
[(Browse History)](https://github.com/eclipse-pde/eclipse.pde/commits/e25b69309c18275a8001849136f35119e38ad338/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java?browsing_rename_history=true&new_path=e4tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java&original_branch=3d0cbb83b870d8e155926f905c414b6bd35845dd)

and now you see the "real" history, so for me renaming the folder does not bring anything useful in terms of git history either...

HannesWell:
> * we previously always have choose the name of the original repo that was "tools"

But that was for another repository, which does not have other 'tools'.

> 
> * we don't have e3tools either so e4tools is missleading, also the ui folder already contains some "e4 tools" so this makes it not clearer or better

And we don't have e2 and e1, so I don't understand why it is a prerequisite?
In general I agree that the overall structure is not ideal, but has probably grown historically, but I don't see a big need to change the existing structure (which affects existing PRs and local changes). In the end the path within the repo does not really matter.
But if we add new files to the repo, we can and should try how to fit that best into the existing structure (existing PRs local changes have to be adapted any way). And since all the added projects are named `org.eclipse.e4.tools*` and are therefore different from the existing projects, they should IMHO reside in a specific location.

> 
> * we now have two times interrupting the history, e.g. look at [this file](https://github.com/eclipse-pde/eclipse.pde/commits/3d0cbb83b870d8e155926f905c414b6bd35845dd/e4tools/bundles/org.eclipse.e4.tools.compat/src/org/eclipse/e4/tools/compat/internal/CopyAction.java) it show:

That's right, but I don't see why this is a problem as long as within git-blame per line is preserved (which a move does).

HannesWell:
But as we have discussed privately in the end the location is not a show stopper and all of use can do more important things.
@iloveeclipse and @merks if you are fine with the current location in `e4tools` I'll continue with the merge to not delay this further.

iloveeclipse:
> if you are fine with the current location in `e4tools` I'll continue with the merge to not delay this further.

Yes, I'm fine.

HannesWell:
Merged it now. Thank you @laeubi for this work.

mickaelistria:
Thank you very much for this refactoring!

PR #1085, opened 2024-01-28
Adapt PDE's Oomph setupt to e4.tools merge

HannesWell:
Follow-up of https://github.com/eclipse-pde/eclipse.pde/pull/1081.

@merks can you tell why the e4.tools project are not imported in the WS? I actually assumed that the targlet task should do that without adjustments.

### DIFF
--- a/releng/org.eclipse.pde.setup/PDE.setup
+++ b/releng/org.eclipse.pde.setup/PDE.setup
@@ -177,7 +177,7 @@
             project="org.eclipse.pde"/>
         <operand
             xsi:type="workingsets:ExclusionPredicate"
-            excludedWorkingSet="//'pde.workingsets'/@workingSets[name='PDE%20API%20Tools'] //'pde.workingsets'/@workingSets[name='PDE%20Tests'] //'pde.workingsets'/@workingSets[name='PDE%20Build']"/>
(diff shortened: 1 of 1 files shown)


merks:
Did you delete the other projects (with the same names) from the other repository.  Obviously for any project name x there can be only one x in the workspace.  Also, often when you delete a project something will create a new .project file in the old location.   I'd suggest just committing this change and tomorrow morning I can help fix any outstanding issues...

merks:
Here's what I do for cases like this.  I change to show projects as top level elements and delete them all, without deleting them in the underlying file system (clone).   Do a pull all the clones.  The resolve the targlet again.  The files are imported:

image

And this working set is empty:

image

So all looks fine.

HannesWell:
> Did you delete the other projects (with the same names) from the other repository. Obviously for any project name x there can be only one x in the workspace.

Thanks Ed for testing, but now I have PDE in a dedicated workspace with only P2. But I'll figure out what's wrong at my computer.



> The files are imported:
> 
> image

Its good that they are imported, but they should be added to a new `E4 Tools` working set.

merks:
The picture was from **before** the pde setup change was merged.

PR #1087, opened 2024-01-29
Improve PDE.setup

merks:


merks:
It looks like this:

image

merks:
New picture without pde.build working set:

image

Issue #1088, opened 2024-01-29
Deadlock in API tooling after removing projects from workspace


iloveeclipse:
Using build I20240128-1800. After updating workspace on latest master (where e4 projects were removed) API tools caused a deadlock - after pull, while build was running, I've selected one of the "removed" e4 tools projects and that immediately freeze the UI.

Looks like the deadlock happens because the **disposed** API component (from removed project) is used in two threads - one that runs API checks in a Job and another one that disposes the same component as a result of project removal.

The synchronization introduced in 1ac36e3b3ab471423454932740e7ca1466fd6415 is one part of the problem.
Another one is the synchronization on `org.eclipse.pde.api.tools.internal.model.Component.fApiDescription` field in `Component.dispose()`. In general, all synchronization in PDE is pretty inconsistent and only works by occasion.

I believe the **main issue** is however that `WorkspaceDeltaProcessor.processJavaElementDeltas()` calls  to `ApiBaselineManager.disposeWorkspaceBaseline()` **while** API analysis is running.
There was an attempt to mitigate that by cancelling ApiAnalysisJob jobs in `ApiBaselineManager.disposeWorkspaceBaseline()` but this is obviously not enough.

I think we should check if we can run this baseline disposal as a job and make it conflicting with API analysis jobs, so they can't run in parallel.

I will try to provide a PR.

[Full thread dump](https://github.com/eclipse-pde/eclipse.pde/files/14081441/pde_api_tools_deadlock.txt)

Here is the relevant part:

```
Found one Java-level deadlock:
=============================
"main":
  waiting to lock monitor 0x00007ffdb8140890 (object 0x00000004061b8e48, a java.lang.Object),
  which is held by "Worker-137: Updating API analysis markers on org.eclipse.e4.ui.workbench.renderers.swt"

"Worker-137: Updating API analysis markers on org.eclipse.e4.ui.workbench.renderers.swt":
  waiting to lock monitor 0x00007ffd181f31d0 (object 0x00000004a8f5ad58, a org.eclipse.pde.api.tools.internal.model.ProjectComponent),
  which is held by "Worker-300: Performing API Analysis"

"Worker-300: Performing API Analysis":
  waiting to lock monitor 0x00007ffdb8140890 (object 0x00000004061b8e48, a java.lang.Object),
  which is held by "Worker-137: Updating API analysis markers on org.eclipse.e4.ui.workbench.renderers.swt"

Java stack information for the threads listed above:
===================================================
[...]

PR #1089, opened 2024-01-29
Deadlock in API tooling after removing projects from workspace

iloveeclipse:
Makes workspace baseline disposal task to job and makes sure it doesn't run in parallel with API analysis jobs.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1088

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiBaselineManager.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiBaselineManager.java
@@ -40,15 +40,22 @@ import org.eclipse.core.resources.ISaveContext;
 import org.eclipse.core.resources.ISaveParticipant;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@laeubi : quality gate is finding 2 maven warnings that weren't introduced by this PR.
So I assume I have to ignore that as usually with all maven warnings, or is this a problem that need to be reported somewhere?

laeubi:
I have not yet find out why baseline errors are reported here, it seems always one or two items but I can't reproduce it locally.

If you have an idea what can causing this (it is always the same project and always the same file) that would be helpfull but for this PR I think we can't fix it see:

- https://github.com/eclipse-pde/eclipse.pde/pull/1080

beside that PDE should normally show zero maven warnings.

iloveeclipse:
> If you have an idea what can causing this

Sorry, no idea about Maven / tycho in general and for the particular warnings in particular. I guess some maven/tycho specific parameters in Jenkinsfile that makes Jenkins execution differ from the official SDK build, but honestly no idea which one.

laeubi:
> Sorry, no idea about Maven / tycho in general and for the particular warnings in particular.

The warning basically says that baseline compile result does not match verification compile result I have created
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921

with compiler logs and baseline versus compile results.

PR #1090, opened 2024-01-29
fix "The value of the local variable is not used"

jukzi:


jukzi:
@laeubi the maven plugin again fails on unrelated warnings :-(

laeubi:
> @laeubi the maven plugin again fails on unrelated warnings :-(

Its not the maven plugin its a compiler bug:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921

Issue #1084, opened 2024-01-28
Remove special treatment of org.eclipse.swt in API-Tools


HannesWell:
At the moment API-tools have special treatment for the `org.eclipse.swt` project, because within the IDE the java class reside within the o.e.swt (host/main) bundle, while in the baseline the classes reside within the platform specific fragments:

https://github.com/eclipse-pde/eclipse.pde/blob/b95c6e569079d9726ec249ebccc38ce9ac81cdd1/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/util/Util.java#L308

https://github.com/eclipse-platform/eclipse.platform.swt/pull/1008 this could change and we can probably remove the special handling for swt in api-tools.

See also https://github.com/eclipse-pde/eclipse.pde/issues/1073#issuecomment-1912506323 (although this missed some parts).

PR #1094, opened 2024-01-31
Remove special treatment of org.eclipse.swt in API-Tools

HannesWell:
Now that https://github.com/eclipse-platform/eclipse.platform.swt/pull/1008 is submitted, I think we can remove the special treatment of SWT and its fragments, since they should simply be compared against the baseline as they are.
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1084


akurtakov:
This reminds me of dropping SWT special treatment in Tycho.  It's so nice to see SWT being less and less of an exception. Thanks @HannesWell !

laeubi:
@akurtakov As far as I know Tycho has long time ago dropped special treatment of SWT, this is now handled by p2.inf + filters and was even improved thanks to @merks adding new testcases, also note that p2.inf customization is not special to swt,its even used in JDT / PDE / Equinox.

PR #1096, opened 2024-01-31
Make the generation of source feature configurable and skip e4.tools

laeubi:
PDE has an option to generate source features but the e4 tools seem not to be published with source features so disable it to not get a warning about additional artifacts compared to baseline.

### DIFF
--- a/e4tools/features/org.eclipse.e4.tools.persistence.feature/build.properties
+++ b/e4tools/features/org.eclipse.e4.tools.persistence.feature/build.properties
@@ -1,2 +1,4 @@
 bin.includes = feature.xml,\
                feature.properties
+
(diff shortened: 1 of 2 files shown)


akurtakov:
In https://github.com/eclipse-pde/eclipse.pde/pull/1060 I touch the persistence feature and the source feature is uploaded https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1060/6/artifact/repository/target/repository/features/ . It should have the same effect for I-builds p2 repo too. This is better path from consistency POV. I'll push the patch in question now.

laeubi:
Ok then this is obsolete and actually the warning was correct.

PR #1099, opened 2024-01-31
Touch o.e.pde.spy.model

akurtakov:
After https://github.com/eclipse-pde/eclipse.pde/pull/1060 got merged the build failed due to signing. This is similar issue like eclipse-platform/eclipse.platform.releng.aggregator#1673 so this is a simple check whether a "touch" would be enough.


akurtakov:
Merging as failure happens only for master (as signing is not enabled for PRs) so results will be seen on master build.

laeubi:
@akurtakov if you like to test you can of course enable signing for a PR as well, if you are commiter of a product those changes will still be picked up!

akurtakov:
> @akurtakov if you like to test you can of course enable signing for a PR as well, if you are commiter of a product those changes will still be picked up!

Thanks, I totally forgot about that.

PR #1104, opened 2024-02-01
Do not submit: Enable signing

akurtakov:
Test for failing pde.build.tests if signing is enabled.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ pipeline {
 						-Pbree-libs \
 						-Papi-check \
 						-Pjavadoc \
-						${env.BRANCH_NAME=='master' ? '-Peclipse-sign': ''} \
+						-Peclipse-sign \
 						-Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true \
 						-Dtycho.debug.artifactcomparator \
 						-Dpde.docs.baselinemode=fail


laeubi:
@akurtakov maybe after signing we need to update the p2 metadata? Or can we try to disable signing for the test project itself somehow by disable the signing plugin ther explicitly?

akurtakov:
The log puzzles me "Installation of IUs [org.eclipse.pde.build.tests, org.eclipse.tycho.surefire.osgibooter, org.eclipse.tycho.surefire.junit4] failed: Call to p2 director application failed:The installable unit org.eclipse.pde.build.tests has not been found."
Disabling the signing is not what I look for as this is the same/similar root issue to the aggregator verification build.

laeubi:
We try to install the test runtime with P2 and this means that for some reason the "project repository" seem not contain the item anymore, if you run locally you can see this artifact in `eclipse.pde/build/org.eclipse.pde.build.tests/target/local-artifacts.properties` as well as in the p2content / p2artifacts ... somewhere here the signing seems to "change" something and now it can't be found anymore (e.g. it might use a different file name for the signed items).

akurtakov:
Ah, so this has been long standing issue and I-build is not showing it only because it skips tests.

laeubi:
I haven't looked into the details of how this "signing" work, but for example with source features it is required to afterwards update the p2 info as well.

Issue #1107, opened 2024-02-02
Parallel API Analysis: IllegalStateException: The bundle belongs to another state


jukzi:
```
java.lang.IllegalStateException: The bundle belongs to another state: org.eclipse.help_3.10.200.v20231106-1240
	at org.eclipse.osgi.internal.resolver.StateImpl.basicAddBundle(StateImpl.java:734)
	at org.eclipse.osgi.internal.resolver.StateImpl.addBundle(StateImpl.java:199)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.addApiComponents(ApiBaseline.java:414)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.restoreFrom(ApiBaseline.java:800)
	at org.eclipse.pde.api.tools.internal.ApiBaselineManager.loadBaselineInfos(ApiBaselineManager.java:249)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.loadBaselineInfos(ApiBaseline.java:772)
	at org.eclipse.pde.api.tools.internal.model.ApiBaseline.getApiComponent(ApiBaseline.java:731)
	at org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.analyzeComponent(BaseApiAnalyzer.java:239)
	at org.eclipse.pde.api.tools.internal.builder.ApiAnalysisBuilder.buildAll(ApiAnalysisBuilder.java:917)
	at org.eclipse.pde.api.tools.internal.builder.ApiAnalysisBuilder.work(ApiAnalysisBuilder.java:402)
	at org.eclipse.pde.api.tools.internal.builder.ApiAnalysisBuilder$ApiAnalysisJob.run(ApiAnalysisBuilder.java:554)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
eclipse.buildId=4.31.0.I20240131-1800
java.version=21.0.2

@iloveeclipse  i mark this as regression, as there is currently no workaround anymore to disable parallel execution

laeubi:
Maybe better make the job / analysis more robust? Running it not as a job has major performance impact

iloveeclipse:
> Maybe better make the job / analysis more robust?

Sure, I was only mentioning possibility to configure it in a more convenient way.

Please be aware, PDE code is a nightmare of synchronization problems, it is like a can of worms...

iloveeclipse:
[...]

PR #1109, opened 2024-02-02
Revert "remove RunAPIAnalysisBuilderAsJob option from perferences UI"

jukzi:
Reverts eclipse-pde/eclipse.pde#568

reason: Parallel API Analysis not stable, see

https://github.com/eclipse-pde/eclipse.pde/issues/1107
https://github.com/eclipse-platform/eclipse.platform/issues/200


iloveeclipse:
Change in PDEUIMessages is missing.

gireeshpunathil:
If parallel analysis is not stable, why are we introducing this option to the user again, instead of disallowing parallel analysis itself?

iloveeclipse:
Because it is hard to make it stable and user should be able to choose what way should Eclipse work - slow and stable or fast and unstable.

laeubi:
> slow and stable or fast and unstable.

We might rename the option into "fail fast mode" :-D

gireeshpunathil:
> Because it is hard to make it stable and user should be able to choose what way should Eclipse work - slow and stable or fast and unstable.

along with the revert, are we warning the user around the checkbox saying: `on: fast and unstable`, `off: slow but stable` ? if not, this revert alone looks to me like misguiding the user!

iloveeclipse:
There is not much to warn, every code has bugs, so should every preference has a warning?

laeubi:
> There is not much to warn, every code has bugs, so should every preference have a warning?

It should warn if there are unforeseen consequence. e.g as A user I might need to know what will happen if I choose the option e.g. what does it means to run it as a job and if it is faster, why I should NOT choose that option ...

Given that here
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/898

it was even asked for explaining the whole story of JUnit4/5 in a single checkbox by the same audience of people it seems fair to address @gireeshpunathil concerns here in the same way...

jukzi:
i vote for: disable parallel by default, add the UI option to enable again.
Reasoning: the few people who ever will find time to enable the option will know that they enabled something non-standard. it feels like common sense that something that is meant to better but not enabled by default will come with some drawback.

merks:
BTW, I provisioned a new IDE today with the latest of everything, and it had an API error that went away with a clean build; very annoying.

gireeshpunathil:
> There is not much to warn, every code has bugs, so should every preference have a warning?

no, but if I know a feature has a bug in it, then I would definitely warn, if bringing it back and present to the user.

iloveeclipse:
> i vote for: disable parallel by default, add the UI option to enable again.

That would be my prefered solution.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -368,6 +368,7 @@ MainPreferencePage_ShowTargetStatus=S&how current target platform in status bar
 MainPreferencePage_updateStale=&Update stale manifest files prior to launching
 MainPreferencePage_WorkspacePluginsOverrideTarget=Workspace p&lug-ins override target platform plug-ins with the same id
 MainPreferencePage_DisableAPIAnalysisBuilder=&Disable API analysis builder
+MainPreferencePage_RunAPIAnalysisBuilderAsJob=&Run API analysis parallel to the build job
```
laeubi, line 371:
```suggestion
MainPreferencePage_RunAPIAnalysisBuilderAsJob=&Run API analysis in the background (experimental)
```

jukzi:
> it blocks all user operation

that could be solved like in JDT, where the build jobs checks for interrupt/cancel

gireeshpunathil (approved):

PR #1114, opened 2024-02-03
P2Utils.createFeatureIUs should guard against missing map entries

merks:
When features include os-specific fragments, those included bundles may not be present in the map of plug-ins.  In general it's good to avoid an NPE if/when a map entry is missing.


laeubi (approved):
Thanks for spotting this!

PR #1115, opened 2024-02-03
Add support to contribute IClasspathContributor as an OSGi service

laeubi:
Currently IClasspathContributor can only be contributed by an extension point, this is often not very flexible and one can't use OSGi DI techniques like declarative services.

This adds support for contribute IClasspathContributor as OSGi services as well.


laeubi:
For failure see
- https://github.com/eclipse-pde/eclipse.pde/issues/1119

PR #1120, opened 2024-02-03
Test for bug 1119

laeubi:


laeubi:
The build script looks like this, the question is more why should there be 4 entries and what entry is wrong (and why only on linux):
```

<project name="Compile feature" default="main">
	<target name="main">
		<parallel threadsPerProcessor='3'>
			<ant antfile="build.xml" dir="plugins/A" target="build.jars"/>
		</parallel>
		<parallel threadsPerProcessor='3'>
			<ant antfile="build.xml" dir="plugins/C" target="build.jars"/>
			<ant antfile="build.xml" dir="plugins/B" target="build.jars"/>
		</parallel>
		<parallel threadsPerProcessor='3'>
			<ant antfile="build.xml" dir="plugins/D" target="build.jars"/>
		</parallel>
		<parallel threadsPerProcessor='3'>
			<ant antfile="build.xml" dir="plugins/E" target="build.jars"/>
		</parallel>
		<parallel threadsPerProcessor='3'>
			<ant antfile="build.xml" dir="plugins/F" target="build.jars"/>
		</parallel>
	</target>
</project>
```

PR #1121, opened 2024-02-04
Handle return value 24 the same as 23

merks:
https://github.com/eclipse-equinox/p2/issues/166


merks:
I've tested this and this is indeed the place where the framework must restart the launched process based on the exit code which can now be 24 instead of only 23 to indicate a restart.

PR #1123, opened 2024-02-04
Revert "Handle return value 24 the same as 23"

merks:
Reverts eclipse-pde/eclipse.pde#1121

The changes is https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622 avoid this problem.


HannesWell:
@merks since you are about to touch the code again, could you please use the existing `IApplication` constants for the remaining 23 as indicated in https://github.com/eclipse-equinox/equinox/pull/498#issuecomment-1925690359?

merks:
Apparently I'm stupid and this is the result:

https://github.com/eclipse-pde/eclipse.pde/pull/1124

PR #1124, opened 2024-02-04
Revert "Handle return value 24 the same as 23"

merks:
This reverts commit e41959b261bcf067243f292639bf2811072b901d.


HannesWell (approved):
Thank you!

Issue #1059, opened 2024-01-18
[target] Saving .target xml file ignores existing file's line endings


haubi:
So we do have some .target files committed to our Git repo for ages, and we use the Eclipse IDE on both Linux and Windows. 

Recently, I've added Maven artifacts to that .target files, using Eclipse IDE on Linux.

Now, when developers on Windows just perform "Apply" in the Target Platform Preferences dialog, even without actually changing anything there, they see changes in the line endings on all the .target files that reference Maven artifacts.

My **first** thought was for the difference with Maven artifacts:
The result is PR https://github.com/eclipse-m2e/m2e-core/pull/1649, to avoid writing the .target file when the content is identical.

My **second** thought was for the line endings, and this is what I figured out:
It actually is the System.lineSeparator() used, and I fail to find any code in this area to eventually reuse the existing file's line separator, or to look at Eclipse or Git properties (e.g. core.autocrlf) for the preferred line separator.
So even if the .target file is written because of real changes, the line separator may change for the whole file, depending on the OS where Eclipse was running on to create that file before.

After all, it really is confusing and annoying to developers, when they see local changes to files they don't touch at all, and leads to additional effort to revert and/or ignore these changes.

Any thought for how to work around this problem until a proper fix is available?

For the records: The stack in **Eclipse 2023-09** for where the first line separator added to the .target file content is:
```
ToXMLStream.startDocumentInternal() line: 188	
ToXMLStream(SerializerBase).startDocument() line: 1076	
DOM2TO.parse() line: 91	
TransformerImpl.transformIdentity(Source, SerializationHandler) line: 718	
TransformerImpl.transform(Source, SerializationHandler, String) line: 780	
TransformerImpl.transform(Source, Result) line: 395	
TargetDefinitionPersistenceHelper.persistXML(ITargetDefinition, OutputStream) line: 131	
TargetDefinition.write(OutputStream) line: 825	
WorkspaceFileTargetHandle.save(ITargetDefinition) line: 81	
TargetPlatformService.saveTargetDefinition(ITargetDefinition) line: 303	
TargetPlatformPreferencePage.performOk() line: 952	
TargetPlatformPreferencePage(PreferencePage).performApply() line: 457	
TargetPlatformPreferencePage(PreferencePage).lambda$1(SelectionEvent) line: 289	
0x00007f9c88f9c6c0.accept(Object) line: not available	
SelectionListener$1.widgetSelected(SelectionEvent) line: 84	
[...]

PR #1127, opened 2024-02-05
TargetDefinition: compare target locations along serialized value

haubi:
Implementations of ITargetLocation may have an equals() implementation that does not compare all their serialized attributes. In particular, this is true for MavenTargetLocation, causing any check for redundancy upon the 'Apply' button in the Target Platform Preferences Editor to fail, causing each time:
- the .target file to be rewritten
- the Target Platform State to be re-evaluated

Instead, we compare along the serialized value of ITargetLocation, ignoring any particular order of target locations as well.

This originates in the discussion at https://github.com/eclipse-m2e/m2e-core/pull/1649#issuecomment-1926831214

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.ParserConfigurationException;
@@ -879,7 +880,7 @@ public class TargetDefinition implements ITargetDefinition {
 				Objects.equals(getVMArguments(), definition.getVMArguments()) && //
 				Objects.equals(getJREContainer(), definition.getJREContainer()) && //
 				Arrays.equals(getIncluded(), definition.getIncluded()) && //
-				Arrays.equals(getTargetLocations(), definition.getTargetLocations()) && //
+				unorderedArraysEqualsSerialized(getTargetLocations(), definition.getTargetLocations()) && //
 				Arrays.equals(getImplicitDependencies(), definition.getImplicitDependencies());
 	}
 
@@ -899,8 +900,22 @@ public class TargetDefinition implements ITargetDefinition {
 				isArgsNullOrEqual(getVMArguments(), definition.getVMArguments()) && //
 				Objects.equals(getJREContainer(), definition.getJREContainer()) && //
 				Arrays.equals(getIncluded(), definition.getIncluded()) && //
-				Arrays.equals(getTargetLocations(), definition.getTargetLocations()) && //
+				unorderedArraysEqualsSerialized(getTargetLocations(), definition.getTargetLocations()) && //
 				Arrays.equals(getImplicitDependencies(), definition.getImplicitDependencies());
+	}
+
+	private static boolean unorderedArraysEqualsSerialized(ITargetLocation[] one, ITargetLocation[] two) {
+		if (one == two) {
+			return true;
+		}
+		if (one == null || two == null || one.length != two.length) {
+			return false;
+		}
+		Set<Object> oneXml = Arrays.stream(one).filter(Objects::nonNull)
+				.map(l -> Objects.requireNonNullElse(l.serialize(), l)).collect(Collectors.toSet());
+		Set<Object> twoXml = Arrays.stream(two).filter(Objects::nonNull)
+				.map(l -> Objects.requireNonNullElse(l.serialize(), l)).collect(Collectors.toSet());
+		return oneXml.equals(twoXml);
 	}
 
 	private boolean isArgsNullOrEqual(String args1, String args2) {


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
```diff
@@ -899,10 +900,21 @@ public boolean isContentEquivalent(ITargetDefinition definition) {
...
 	}
 
+	private boolean unorderedArrayEqualsSerialized(ITargetLocation[] one, ITargetLocation[] two) {
+		if (one.length != two.length) {
+			return false;
+		}
+		Set<String> oneXml = new HashSet<>(
+				Stream.of(one).filter(l -> l != null).map(ITargetLocation::serialize).toList());
+		Set<String> twoXml = new HashSet<>(
+				Stream.of(two).filter(l -> l != null).map(ITargetLocation::serialize).toList());
```
laeubi, line 914: What do you think about clean up the xml e.g. replace all \r \n\t by empty string?

Or even parse it with a dom parser and then process it on the dom level, e.g. order of attributes should not matter also.

haubi: Uhm, that feels like overengineering:
AFAICS, a difference in the serialized string without a difference in the real content only happens if someone really opened the editor, performing some changes _and reverting_ them. In this case I don't think writing the file again or reloading target platform state really is a problem.

And the newline thingy is a different problem anyway: https://github.com/eclipse-pde/eclipse.pde/issues/1059

haubi: > What do you think about clean up the xml e.g. replace all \r \n\t by empty string?

Thinking more about this, also having https://github.com/eclipse-pde/eclipse.pde/issues/1059 in mind, brings me to the idea that serialize() probably isn't the best approach to write into XML files or even create XML strings, because it doesn't allow to use some specific (XML) Writer doing whatever formatting is wanted (regarding line breaks, indentation, ...).
Heck, IMHO the data object shouldn't be aware at all of whether the result should be XML/JSON/Binary/Whatever, nor File/String/Stream/Whatever, and any particular formatting, no?

laeubi: Yes but as it is API we can't easily change that... so we can only do our best here I also planned to rework the API but its not top priority at them moment as it works so far.

haubi:
For the remaining failing tests I believe these are independent of my commit, no?

laeubi:
> For the remaining failing tests I believe these are independent of my commit, no?

It looks like this yes...

laeubi (approved):
Looks good to me, rebased to restart the build

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
```diff
@@ -899,10 +901,24 @@ public boolean isContentEquivalent(ITargetDefinition definition) {
...
+	private static boolean unorderedArraysEqualsSerialized(ITargetLocation[] one, ITargetLocation[] two) {
+		if (one == two) {
+			return true;
+		}
+		if (one == null || two == null || one.length != two.length) {
+			return false;
+		}
+		Set<Object> oneXml = new HashSet<>(Stream.of(one).filter(l -> l != null)
+				.map(l -> Objects.requireNonNullElse(l.serialize(), l)).collect(Collectors.toSet()));
+		Set<Object> twoXml = new HashSet<>(Stream.of(two).filter(l -> l != null)
```
HannesWell, line 917: Here and above the pipeline could be 'simplified'
```suggestion
		Set<Object> twoXml = new HashSet<>(Arrays.stream(two).filter(Objects::nonNull)
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
```diff
@@ -899,10 +901,24 @@ public boolean isContentEquivalent(ITargetDefinition definition) {
...
+				.map(l -> Objects.requireNonNullElse(l.serialize(), l)).collect(Collectors.toSet()));
+		return oneXml.size() == twoXml.size() && oneXml.containsAll(twoXml);
```
HannesWell, line 919: Isn't this the same as just equals?
```suggestion
		return oneXml.equals(twoXml);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
haubi: Yep, and this exactly is one of the reasons why I don't call myself an experienced Java programmer yet, thanks!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
```diff
@@ -899,10 +900,24 @@ public boolean isContentEquivalent(ITargetDefinition definition) {
...
 	}
 
+	private static boolean unorderedArraysEqualsSerialized(ITargetLocation[] one, ITargetLocation[] two) {
+		if (one == two) {
+			return true;
+		}
+		if (one == null || two == null || one.length != two.length) {
+			return false;
+		}
+		Set<Object> oneXml = new HashSet<>(Arrays.stream(one).filter(Objects::nonNull)
```
laeubi, line 914: Why wrap the set into another set? Do we need a specific type?

if yes use `toSet(HashSet::new)`

laeubi (approved):
Looks ok to me.

HannesWell (approved):
Looks good to me as well. Thanks!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
HannesWell: Yep, I also learned a lot in such reviews and still learn new things. 😃
Thanks for applying.

Issue #1129, opened 2024-02-06
StatusLineManager warning on saving bundle manifest


iloveeclipse:
With I20240205-1800
Created new plugin, opened MANIFEST.MF, checked "This plugin is a singleton" checkbox, saved editor via Ctrl+S => warning reported in the log.
```
java.lang.IllegalStateException: beginTask should only be called once per instance. At least call done() before further invocations
	at org.eclipse.jface.action.StatusLineManager$1.beginTask(StatusLineManager.java:161)
	at org.eclipse.core.runtime.ProgressMonitorWrapper.beginTask(ProgressMonitorWrapper.java:53)
	at org.eclipse.ui.internal.dialogs.EventLoopProgressMonitor.beginTask(EventLoopProgressMonitor.java:64)
	at org.eclipse.core.runtime.ProgressMonitorWrapper.beginTask(ProgressMonitorWrapper.java:53)
	at org.eclipse.ui.internal.dialogs.EventLoopProgressMonitor.beginTask(EventLoopProgressMonitor.java:64)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:557)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:512)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2439)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2473)
	at org.eclipse.ui.actions.WorkspaceModifyOperation.run(WorkspaceModifyOperation.java:131)
	at org.eclipse.ui.internal.editors.text.WorkspaceOperationRunner.run(WorkspaceOperationRunner.java:76)
	at org.eclipse.ui.internal.editors.text.WorkspaceOperationRunner.run(WorkspaceOperationRunner.java:66)
	at org.eclipse.ui.editors.text.TextFileDocumentProvider.executeOperation(TextFileDocumentProvider.java:461)
	at org.eclipse.ui.editors.text.TextFileDocumentProvider.saveDocument(TextFileDocumentProvider.java:783)
	at org.eclipse.ui.editors.text.ForwardingDocumentProvider.saveDocument(ForwardingDocumentProvider.java:141)
	at org.eclipse.pde.internal.ui.editor.context.InputContext.doSave(InputContext.java:213)
	at org.eclipse.pde.internal.ui.editor.context.InputContextManager.save(InputContextManager.java:82)
[...]

PR #1130, opened 2024-02-06
InputContextManager.save(): split monitor #1129

jukzi:
https://github.com/eclipse-pde/eclipse.pde/issues/1129

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
@@ -15,6 +15,7 @@ package org.eclipse.pde.internal.ui.editor.context;
 
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jukzi:
@iloveeclipse please test if that solves your issue.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
```diff
@@ -77,9 +79,11 @@ public void dispose() {
 	 * Saves dirty contexts.
 	 */
 	public void save(IProgressMonitor monitor) {
-		for (InputContext context : inputContexts.values()) {
+		Collection<InputContext> values = inputContexts.values();
+		SubMonitor subMon = SubMonitor.convert(monitor, values.size());
+		for (InputContext context : values) {
 			if (context.mustSave())
-				context.doSave(monitor);
+				context.doSave(subMon.newChild(1));
```
laeubi, line 86:
```suggestion
				context.doSave(subMon.split(1));
```

jukzi:
@iloveeclipse ping

iloveeclipse:
> ping

I can't reproduce that, so I can't say if the patch makes any difference.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/context/InputContextManager.java
jukzi: split() would check for cancel. i don't know if that is safe at this place and don't think it's needed - never had a lag there that i would like to cancel.

jukzi:
> I can't reproduce that, so I can't say if the patch makes any difference.

nevertheless that code looks wrong, i suggest to submit.

iloveeclipse:
> nevertheless that code looks wrong, i suggest to submit.

If you think so, please. I don't want to look into monitor code if I'm not forced to.

PR #1133, opened 2024-02-06
Add message value to API-tools annotations

HannesWell:
This allows users of the annotations to provide contextual information to clients about why the annotations is applied.
This is common for users of the javadoc-annotations, but at the moment not possible when using the java-annotations.

Later PDE's API-Tools could also provide that information to developers of clients.

Instead of
```
/**
 * The doc.
 * @noextend This class is not intended to be sub-classed by clients.
 */
public abstract class MyClass {
```
one can use 
```
/**
 * The doc.
 */
@NoExtend("This class is not intended to be sub-classed by clients.")
public abstract class MyClass {
```

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.annotations/src/org/eclipse/pde/api/tools/annotations/NoExtend.java
+++ b/apitools/org.eclipse.pde.api.tools.annotations/src/org/eclipse/pde/api/tools/annotations/NoExtend.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013 IBM Corporation and others.
+ * Copyright (c) 2013, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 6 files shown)


laeubi (approved):
Looks good!

laeubi:
@HannesWell just one thing to note as we almost ever have the same "text" maybe the default should not be empty (even though I'm not sure if default text will showup in the javadoc?)

HannesWell:
> @HannesWell just one thing to note as we almost ever have the same "text" maybe the default should not be empty (even though I'm not sure if default text will showup in the javadoc?)
> 

That's a good suggestion.
I'll check later how this will render in Javadoc.

HannesWell:
> @HannesWell just one thing to note as we almost ever have the same "text" maybe the default should not be empty (even though I'm not sure if default text will showup in the javadoc?)

Added a default value for all annotations based on the texts commonly used throughout the Eclipse SDK.
In case one of them is not suitable we can still improve the default later, but I would like to submit this as soon as the build is green so that it can be used in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1039 after tonight's I-build.

I also checked if default values/messages are displayed in the javadoc and at least the Eclipse built-in Javadoc view does not render the `default` of the value element. But if one clicks on the annotation the annotation-class documentation is shown that provides similar information and the annotation itself is already relatively clear.
Nevertheless API tools could also show the default message in its warnings if desired.

PR #1135, opened 2024-02-07
Touch bundles affected by compiler implementation change

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1781


iloveeclipse:
Ignoring warnings reported, they are not caused by this build anyway.

PR #1137, opened 2024-02-08
Remove references to Platform#getPlatformAdmin

laeubi:
PDE is currently the only consumer of the Platform#getPlatformAdmin method that uses the discouraged PlatformAdmin requiring the compatibility fragment.

To allow platform to move away from PlatformAdmin altogether, we can simply have a dedicated method in the PDE space that fetches the PlatformAdmin service.


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PluginModelManager.java
```diff
@@ -559,14 +556,6 @@ private void initializeTable(IProgressMonitor monitor) {
 			return;
 		}
 
-		// Check if PlatformAdmin service is available (Bug 413450)
-		PlatformAdmin pAdmin = Platform.getPlatformAdmin();
```
HannesWell, line 563: Have you considered to keep this as an early check?
I assume something else will fail later anyways if the admin is absent?
But now that you have moved the error message to the PlatformAdmin-'Getter' I think that's save.

HannesWell (approved):
Looks good, I was thinking about something like this too. Thanks!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PluginModelManager.java
laeubi: I have moved it to the getter because effectively this is/was the only location where a null check was performed, all other locations instantly dereference the value so will lead to a nasty Nullpointer exception and then you hopefully somewhere find a message in the log whats really going on...

laeubi:
Finally ...

PR #1140, opened 2024-02-08
uniformly use 4GB Heap on Workspace

jukzi:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1612


merks:
Has no build impact.

PR #1145, opened 2024-02-10
Remove the org.eclipse.pde.build from the pde feature

laeubi:
Currently the `org.eclipse.pde.build` is directly included in the PDE feature but pde-build is already a requirement and will be included automatically. Beside that it prevents to update `org.eclipse.pde.build` independent from PDE.

With this PR
- https://github.com/eclipse-pde/eclipse.pde/pull/1137

I noticed that the test still run against the i-build version and not the local compiled version. The reason is that the pde feature (that is installed as part of the test runtime) has a strict requirement to the `org.eclipse.pde.build`and therefore can't be updated.

As `org.eclipse.pde.build` is a requirement for pde.core it will be included anyways, also we currently don't really [maintain pde build anymore](https://github.com/eclipse-pde/eclipse.pde/tree/master/build) therefore it seems reasonable to handle it like any other dependency an not explicitly list it in the feature.

### DIFF
--- a/features/org.eclipse.pde-feature/feature.xml
+++ b/features/org.eclipse.pde-feature/feature.xml
@@ -26,15 +26,11 @@
       <import plugin="org.osgi.service.component.annotations"/>
       <import plugin="org.osgi.service.metatype.annotations"/>
       <import plugin="org.bndtools.templates.template"/>
-      <import plugin="biz.aQute.repository" />
+      <import plugin="biz.aQute.repository"/>
    </requires>
 
    <plugin
          id="org.eclipse.pde"
-         version="0.0.0"/>
-
-   <plugin
-         id="org.eclipse.pde.build"
          version="0.0.0"/>
 
    <plugin


akurtakov:
Why not simplifying the feature significantly? E.g. remove everything the pde.ui already requires and other transitive deps I haven't thought about now.

laeubi:
> Why not simplifying the feature significantly? E.g. remove everything the pde.ui already requires and other transitive deps I haven't thought about now.

I just try to be as small steps as possible :-D

merks (commented):
Given previous discussions about "deep updates"

https://github.com/eclipse-equinox/p2/issues/173

I would be concerned that the pde.build bundle will not update until some other bundle's version constraint requires a newer version.  Before going down this path, we'd better be really 100% sure that updates will actually update such things, especially if this will become a new trend.  I have the suspicion is that it will not do so and that this is a very dark path.

laeubi:
> I would be concerned that the pde.build bundle will not update until some other bundle's version constraint requires a newer version

This depends a bit, currently one can only update whole PDE feature or nothing, now it could be updated independently and given that no one _needs_ a newer version why install one automatically?

merks:
> This depends a bit, currently one can only update whole PDE feature or nothing, now it could be updated independently and given that no one _needs_ a newer version why install one automatically?

This line of reasoning concerns me **a lot**.  It's likely that pde.build will never update (even again).  The fact that it can be updated  (independently) doesn't mean it will be updated.  I will assert, until proven otherwise, that it will not updated; I've provided the link to the issue about that very issue.   If we plan to go farther down this path, and my assertion is correct, then only the UI plugin will ever update. 

We really must make decisions understanding **the full impact** of those decisions.  The question "if it doesn't need an update why update it" is not the right question.   If we fix a problem in something (pde.build), or changing anything for any reason, such that there is a new version, but an update doesn't install it, does anyone need it?  What's the point of changing things at all if no one needs it and what's the point of changing things if updates don't update them. 

I hope this doesn't become some kind of circularity.  We really need to focus on facts here.  We must not make changes because some internal build infrastructure thing is simplified yet the whole update story becomes distorted or broken.

laeubi:
@merks you can try it out yourself, updating any PDE plugin alone is not possible within an IDE, you always must update the full feature, and in that case I would expect that if anything is fixed in `pde.build` it has a new version and consumers will require that version at the minimum, how else would the fix become effective?

merks:
I must run now.  I will do some experiments.  In the meantime, think carefully about your statement "the consumers will require that version at the minimum", because right now the feature is the consumer and the installation consumes that feature.  If you break that link, then nothing may require the new version and the new version may not be installed, even though it has enhancement and bug fixes that maybe nothing strictly requires but certainly the users will want.

Instead of hypothetical discussions, actually demonstrate that transitively required bundle will update when the feature than transitively depends on it is update.  I will try that when I have time, but given you are the one making the change, it's actually your obligation to demonstrate that it works as you assert...

HannesWell:
> Given previous discussions about "deep updates"
> 
> [eclipse-equinox/p2#173](https://github.com/eclipse-equinox/p2/issues/173)
> 
> I would be concerned that the pde.build bundle will not update until some other bundle's version constraint requires a newer version. Before going down this path, we'd better be really 100% sure that updates will actually update such things, especially if this will become a new trend. I have the suspicion is that it will not do so and that this is a very dark path.

While I understand the desire to solve the mentioned build/test problem, I also share Ed's concern and agree with him that this has to be solved first.
I also understand the concerns about deep updates ('never touch a running system', or at least don't touch it unnecessarily), but on the other hand it is also a good idea to keep dependencies up to date (nowadays, but I think it was always true) to get bug- and security-fixes. And just lifting the lower bound of a dependency to get a fix is not the right thing to do since the dependencies declared in a bundle should follow semantic-versioning and should be done based on the API requirements. Its the task of the provisioning tool to handle things like 'patch' updates. But I think that should to be discussed in https://github.com/eclipse-equinox/p2/issues/173.

If we can't introduce a deep update mode, we could alternatively think about introducing a less strict mode for included bundles in feature where only the lower bound is the exact current version (including the qualifier) and the upper-bound is more flexible. Of course the question is how flexible: Only the same major, the same minor or the same micro/service version (the latter should already solve the build problem, shouldn't it?).
This could be also achieved using imports in a feature, where one specifies the exact current version as lower bound and then selects the desired version match rule.
In order to make this more convenient to handle we could think about introducing some kind of macro that resolves to the 'imported' bundle's current version. Then we could define:
```
   <requires>
      <import plugin="org.eclipse.pde.builod" version="@bundle_version" match="perfect"/>
   </requires>
```
Can anybody tell if there is already a common style for such macros in Eclipse? I.e. do they use @ or % or ${} as special symbols?

laeubi:
I really don't want to introduce any macro magic just to solve the problem that projects don't manage their dependencies correctly.

Just let me repeat, **it is currently impossible to update** `org.eclipse.pde.build` alone.
So every bugfix will require a new release of `org.eclipse.pde` (feature) what will include any other PDE dependencies that **need to be updating their lower bounds**, this has nothing to do with semantic versioning, that will still work.

Otherwise we will release a bugfix that will not require the bugfix in any of the using bundles what is simply wrong, or we want to only release a bugfix of  `org.eclipse.pde.build` that the must be installed independently anyways!

HannesWell:
> Just let me repeat, **it is currently impossible to update** `org.eclipse.pde.build` alone.

That's right, but this PR doesn't solve this too. That's a matter of https://github.com/eclipse-equinox/p2/issues/173.
And I haven't seen a button in my Eclipse that says 'Update only this Plugin', it is currently not even possible to install a single Plugin/Bundle.

> Otherwise we will release a bugfix that will not require the bugfix in any of the using bundles what is simply wrong, or we want to only release a bugfix of `org.eclipse.pde.build` (without anything else!) that then must be installed independently anyways!

Realistically I don't think a standalone release of `pde.build` is gonna happen. But even if we would provide patch releases for single bundle it could be any PDE bundle, in fact it could be any bundle in the Eclipse SDK. And if we follow that logic, eventually we would have to throw away features at all. And I haven't understood why the pde.build bundle is special compared to others.

I understand your desire to get https://github.com/eclipse-pde/eclipse.pde/pull/1137 done, but this change is IMHO the wrong solution, at least in the current form. I haven't looked into the details, but can't you make the test picking up the feature from the reactor?

laeubi:
> That's right, but this PR doesn't solve this too. 

It does because now you can update it independently (if wanted)

> That's a matter of https://github.com/eclipse-equinox/p2/issues/173.

That a complete different topic, even with that implemented it wont be possible to update, also this ticket is about transitive dependencies, of any kind not direct dependencies of the same project.

> Realistically I don't think a standalone release of pde.build is gonna happen.

I never claimed that nor recommend nor wanted that, this was brought up by @merks and you that something won'T get updated if there are updates (wherever they should come from if not PDE I don't know).

> And if we follow that logic, eventually we would have to throw away features at all

Exactly, and thats what @akurtakov mentioned, we should actually probably always only need `pde.ui` an nothing else as everything is transitive dependent from that ;-)

> I understand your desire to get https://github.com/eclipse-pde/eclipse.pde/pull/1137 done, but this change is IMHO the wrong solution, at least in the current form. 

That's completely independent from this one as it was just used as an example why this is bad...

> I haven't looked into the details, but can't you make the test picking up the feature from the reactor?

It could but then it won't be updatable because the sdk requires PDE with a exact range as well ;-)

PR #1147, opened 2024-02-10
Move the features from e4tools to the global features folder

laeubi:
PDE has all its features under <projectroot>/features, only the imported e4tools where organized differently.

This moves the e4tools features to the usual features folder.


HannesWell (approved):
Makes sense. Thanks.

PR #1149, opened 2024-02-11
Use a dedicated test project that allows updating pde feature

laeubi:
Currently we use the sdk product to create the test runtime, this has the drawback that we can not use the current build pde code but always using the latest from the ibuilds,

This uses an own copy of the sdk product to setup the necessary
test runtime so we can customize this as required.


HannesWell:
This is a duplicate of https://github.com/eclipse-pde/eclipse.pde/pull/1151, isn't it?

laeubi:
Ni I just splitted some parts fom it there

Issue #1119, opened 2024-02-03
org.eclipse.pde.build.internal.tests.ScriptGenerationTests fails on github linux only


laeubi:
See for example this run: https://github.com/eclipse-pde/eclipse.pde/runs/21185950650

```
expected:<4> but was:<5>
java.lang.AssertionError: expected:<4> but was:<5>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.eclipse.pde.build.internal.tests.ScriptGenerationTests.testBug238177(ScriptGenerationTests.java:873)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
[...]

PR #1150, opened 2024-02-11
Disable ScriptGenerationTests#testBug238177

laeubi:
The test fails on some linux systems for unknown reasons.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1119

### DIFF
--- a/build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/ScriptGenerationTests.java
+++ b/build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/ScriptGenerationTests.java
@@ -827,6 +827,7 @@ public class ScriptGenerationTests extends PDETestCase {
 	}
 
 	@Test
+	@Ignore("this test fails on some linux versions for unkown reasons")
 	public void testBug238177() throws Exception {
 		IFolder buildFolder = newTest("238177");
 		IFolder a = Utils.createFolder(buildFolder, "plugins/A");


build/org.eclipse.pde.build.tests/src/org/eclipse/pde/build/internal/tests/ScriptGenerationTests.java
```diff
@@ -827,6 +827,7 @@ public void testBug248767_212467() throws Exception {
 	}
 
 	@Test
+	@Ignore("this test fails on some linux versions for unkown reasons")
```
HannesWell, line 830: Instead of disabling it for all platforms it would be less invasive to just disable it for Linux using Assumptions.

laeubi: As said anyone that wants to invest time in pde-build and thinks this test is important can do so.

iloveeclipse: That's OK for Linux, but why should it be disabled on other platforms?

HannesWell: > Instead of disabling it for all platforms it would be less invasive to just disable it for Linux using Assumptions.

See https://github.com/eclipse-pde/eclipse.pde/pull/1152.

Issue #312, opened 2022-08-25
APIToolsJavadocCompletionProposalComputer: Resource was not properly disposed


jukzi:
(after pressing ctrl+shift in a javadoc @see link)
```
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:713)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:341)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:290)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:268)
	at org.eclipse.pde.api.tools.ui.internal.completion.APIToolsJavadocCompletionProposalComputer.computeJavadocProposals(APIToolsJavadocCompletionProposalComputer.java:141)
	at org.eclipse.pde.api.tools.ui.internal.completion.APIToolsJavadocCompletionProposalComputer.computeCompletionProposals(APIToolsJavadocCompletionProposalComputer.java:115)
	at org.eclipse.jdt.internal.ui.text.java.CompletionProposalComputerDescriptor.computeCompletionProposals(CompletionProposalComputerDescriptor.java:348)
	at org.eclipse.jdt.internal.ui.text.java.CompletionProposalCategory.computeCompletionProposals(CompletionProposalCategory.java:340)
	at org.eclipse.jdt.internal.ui.text.java.ContentAssistProcessor.collectProposals(ContentAssistProcessor.java:333)
	at org.eclipse.jdt.internal.ui.text.java.ContentAssistProcessor.computeCompletionProposals(ContentAssistProcessor.java:289)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$10(AsyncCompletionProposalPopup.java:378)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$9(AsyncCompletionProposalPopup.java:377)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
[...]

PR #1156, opened 2024-02-15
APIToolsJavadocCompletionProposalComputer: MT hardening #312

jukzi:
collectExistingTags() is running in background while sessionEnded() is running in UI Thread.

https://github.com/eclipse-pde/eclipse.pde/issues/312

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
+++ b/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
@@ -15,10 +15,10 @@ package org.eclipse.pde.api.tools.ui.internal.completion;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.core.CompletionContext;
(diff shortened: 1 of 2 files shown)


apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
```diff
@@ -71,10 +71,10 @@
  */
 public class APIToolsJavadocCompletionProposalComputer implements IJavaCompletionProposalComputer {
 
-	private String fErrorMessage = null;
-	private Image fImageHandle = null;
-	private ASTParser fParser = null;
-	HashMap<String, Boolean> fExistingTags = null;
+	private volatile String fErrorMessage = null;
+	private volatile Image fImageHandle = null;
```
laeubi, line 75: If you want to MT harden this, it has to be an atomic reference, volatile is not enough, same for the other ones ...

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
```diff
@@ -377,11 +378,11 @@ public String getErrorMessage() {
 	public void sessionEnded() {
 		if (fImageHandle != null) {
 			fImageHandle.dispose();
+			fImageHandle = null;
```
laeubi, line 381: This is not multithreading save, fImageHandle must be extracted into a local variable if it can really be modified / accessed from multiple threads

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
jukzi: i harden it, it does not promise to make it totally mt safe. That would need much more work, as currently in the extreme case the sessionEnded() can even be called before collectExistingTags() because it is not waited on the backgroundthread after a timeout. However i have not observes such cases in real life. (logs an exception that fParser is null)

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/completion/APIToolsJavadocCompletionProposalComputer.java
jukzi: s.o.

Issue #329, opened 2022-09-06
ConcurrentModificationException in BundleSourcePage / Bundle.fDocumentHeaders


jukzi:
Bundle.fDocumentHeaders is not a concurrent datastructure and access not synchronized 

```
java.util.ConcurrentModificationException
	at java.base/java.util.TreeMap$PrivateEntryIterator.nextEntry(TreeMap.java:1208)
	at java.base/java.util.TreeMap$ValueIterator.next(TreeMap.java:1253)
	at org.eclipse.pde.internal.ui.editor.plugin.BundleSourcePage.getRangeElement(BundleSourcePage.java:285)
	at org.eclipse.pde.internal.ui.editor.PDESourcePage.handleSelectionChangedSourcePage(PDESourcePage.java:516)
	at org.eclipse.pde.internal.ui.editor.plugin.BundleSourcePage.handleSelectionChangedSourcePage(BundleSourcePage.java:651)
	at org.eclipse.pde.internal.ui.editor.PDESourcePage$PDESourcePageChangedListener.selectionChanged(PDESourcePage.java:92)
	at org.eclipse.jface.text.TextViewer.firePostSelectionChanged(TextViewer.java:2714)
	at org.eclipse.jface.text.TextViewer.firePostSelectionChanged(TextViewer.java:2644)
	at org.eclipse.jface.text.TextViewer$2.run(TextViewer.java:2623)
	at org.eclipse.swt.widgets.Display.runTimer(Display.java:4158)
	at org.eclipse.swt.widgets.Display.messageProc(Display.java:3262)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3630)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:644)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
[...]

PR #1157, opened 2024-02-15
HeaderMap: replace with a concurrent #329

jukzi:
to avoid ConcurrentModificationException
https://github.com/eclipse-pde/eclipse.pde/issues/329


laeubi:
Do you haven ensured that the contract is still satisfied (case insensitive keys):

https://github.com/eclipse-pde/eclipse.pde/blob/72fa789951b9e14d8392abc184797c04496ea493/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/HeaderMap.java#L21-L22

jukzi:
> Maybe this is more a sign that the Map is accesed outside the SWT Thread, what seems wrong for a document?

The issue points to Bundle.fDocumentHeaders, which has plenty callers without guarantee to be accessed only by SWT

PR #1162, opened 2024-02-20
fix some javadoc

jukzi:


jukzi:
some of the `: error:` logged in https://ci.eclipse.org/pde/job/eclipse.pde/job/master/601/consoleText

Issue #1069, opened 2024-01-23
Migrate PDE off the antiquated Equinox resolver API


HannesWell:
As said in [Bug 570198](https://bugs.eclipse.org/bugs/show_bug.cgi?id=570198) and [Bug 575374](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575374), "the org.eclipse.osgi.service.resolver [API] is antiquated and has not been used by the Equinox Framework since Kepler release." and PDE should be migrated to modern alternatives, ideally standard OSGi APIs.

This is intended as an umbrella issue to track all efforts in the area as well as to discuss the general approaches to solve this great effort.

The API in o.e.osgi.service.resolver is used for different areas within PDE.

- **`VersionRange`**: can be replaced with osgi.VersionRange (which it already extends). Used in PDE APIs so alternatives have to be added and existing methods have to be deprecated for removal.
- **`State`**: For the resolver State used in `PDEState` probably the OSGi resolver service and/or the implementations in the `org.eclipse.osgi.container` can serve as an replacement
- **`BundleDescription`**/Plugin dependencies: The initial proposal was to replace `BundleDescription` by `BundleRevision`. But in https://github.com/eclipse-pde/eclipse.pde/pull/1062#issuecomment-1905449203 @laeubi suggested to use the even more generic osgi Resouce API instead, which would allow even lighter Bundle/Plug-in representations

  > you don't need a resolved state with bundles to compute the required information, it is already there in the meta-data, currently this requires PDE to download _all_ bundles of a target platform even if only a few are used to resolve the projects classpath. Also this currently hinders PDE from using more than one target platform, e.g. you must always have one big-fat target that can satisfy all your workspace bundles, using different sets of bundles in the same workspace is not possible often without much efforts.
  
  And in combination with the suggested usage of the [`ResolveContext.getWirings()`](https://docs.osgi.org/specification/osgi.core/7.0.0/service.resolver.html#org.osgi.service.resolver.ResolveContext.getWirings--) most of the information from the `BundleRevision` are also available for a osgi Resource , just in a slightly more abstract form. I agree, that it is probably possible to use mostly the Resources API.
  But maybe a migration is easier in two phases, first from `BundleDescription` to `BundleRevision` and the to `Resource`. The second step is then probably simpler, since a `BundleRevision` is already a Resource.

[...]

PR #1163, opened 2024-02-21
Rework PluginRegistry API and introduce VersionMatchRule enum

HannesWell:
Modernize the PluginRegistry API that contains the antiquated `org.eclipse.osgi.service.resolver.VersionRange` by providing alternatives using the `org.osgi.framework.VersionRange` and deprecating the existing methods for removal.

In conjunction with that also mark `PluginRegistry.PluginFilter` as deprecated for removal since it is effectively a `Predicate<IPluginModelBase>` and just use the latter in the new methods.

This is currently an early draft to discuss the overall direction. Since the intention is to focus on the API callers of the now deprecated methods are not replaced in this PR. I'll do that in a follow-up.

About the API I'm also contemplating to further modernize the methods in `PluginRegistry` that now return a List to return a Stream instead. Then we could also save the filter arguments since callers can just filter the returned stream if desired.

Furthermore I would also like to replace the integer and string constants in `IMatchRules` by a corresponding Enum `VersionMatchRules` (or similar) that also provides method to obtain a predicate for a reference version that implements the literals matching rule or to obtain a corresponding `VersionRange` from a reference version (which ever is needed).

Part of https://github.com/eclipse-pde/eclipse.pde/issues/1069

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2012 IBM Corporation and others.
+ * Copyright (c) 2007, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,24 +10,26 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Hannes Wellmann - Rework PluginRegistry API to modernize it and replace Equinox resolver's VersionRange
  *******************************************************************************/
 package org.eclipse.pde.core.plugin;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.osgi.service.resolver.BundleDescription;
-import org.eclipse.osgi.service.resolver.VersionRange;
 import org.eclipse.pde.core.build.IBuildModel;
 import org.eclipse.pde.internal.core.PDECore;
 import org.eclipse.pde.internal.core.build.WorkspaceBuildModel;
 import org.eclipse.pde.internal.core.project.PDEProject;
 import org.eclipse.pde.internal.core.util.VersionUtil;
 import org.osgi.framework.Version;
+import org.osgi.framework.VersionRange;
 import org.osgi.resource.Resource;
 
 /**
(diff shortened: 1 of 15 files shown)


HannesWell:
This PR now focus only on the rework of the `PluginRegistry` API and has a adjustments of the pde.project description API moved out into https://github.com/eclipse-pde/eclipse.pde/pull/1340.

In the meantime I made greater changes to the `PluginRegistry.findModel(s)()` methods and introduced a `VersionMatchRule` enum. `VersionMatchRule` represents the rules currently defined in `IMatchRules` as int constants and handled in various places in the PDE code base in a type-safe manner and also provides the represented logic in one place.

The rework is not yet completely finished, but the overall direction is settled and I would like to gather feedback already.
On the long run the goal is also to replace `IMatchRules` respectively the usage of int literals by the new `VersionMatchRule` enum.

ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
```diff
@@ -45,15 +46,24 @@
...
+	 *
+	 * @see PluginRegistry#findModels(String, String, VersionMatchRule)
+	 * @see PluginRegistry#findModels(String, VersionRange)
 	 * @since 3.6
+	 * @deprecated Instead use {@link Predicate} and filter the stream returned
+	 *             by
+	 *             {@link PluginRegistry#findModels(String, String, VersionMatchRule)}
+	 *             or {@link PluginRegistry#findModels(String, VersionRange)}
 	 */
+	@Deprecated(forRemoval = true, since = "3.19 (removal in 2026-09 or later)")
```
HannesWell, line 66: Btw. what do you think about mentioning the earliest removal date here?
I think it would help consumers to understand better when the actual removal will happen earliest.

ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java
```diff
@@ -264,61 +278,85 @@ public static IPluginModelBase[] getExternalModels() {
...
+	 *            must fulfill with respect to the specified reference version
+	 *            in order to be selected. May not be null.
 	 *
 	 * @return a matching models, possibly an empty collection
+	 * @since 3.19
+	 */
+	public static Stream<IPluginModelBase> findModels(String id, String version, VersionMatchRule matchRule) {
+		Version reference = version != null ? Version.valueOf(version) : null;
+		return selectModels(id, version != null ? p -> matchRule.matches(VersionUtil.getVersion(p), reference) : null);
+	}
```
HannesWell, line 336: Instead of having one overload of `findModel(s)` that accepts a `VersionRange` and one that accepts a version String plus a `VersionMatchRule` we could also just provide one method with a `VersionRange`.
With the new `VersionMatchRule` one would then use
```
findModel("foo.bar", EQUIVALENT.rangeFor(version))
```
instead of 
```
findModel("foo.bar", versionStr, EQUIVALENT)
```
Of course the level of convenience is also influenced from if one has a `Version` object or a version `String`.
Maybe this justifies to keep both so in case one has a Version object the former solution can be used and if one has a version `String` the latter approach is nicer.

I'm torn on this question.

HannesWell: Because there is more than just one difference as described above, I think it is fine to preserve this structure.

HannesWell:
In general I think this rework is now ready, although there might be a few more minor adjustments necessary.

So anybody interested in this change, please share your opinion as soon as possible.
Otherwise I plan to complete this by the end of this week.

I have also evaluated the possibility to return `Optional<IPluginModelBase>` in the new overloads of `PluginRegistry.findModel()` that return `null` if nothing was found. But this creates an odd asymmetry with respect to the existing overloads. I find it more important to keep the API consistent here. In the other overload that was added recently, we also continued to return `null`: https://github.com/eclipse-pde/eclipse.pde/pull/1062

PR #1165, opened 2024-02-22
Migrate https://wiki.eclipse.org/Execution_Environments

jonahgraham:
Quick pass to migate https://wiki.eclipse.org/Execution_Environments to PDE repo.

It was pointed out to me that https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/Simultaneous_Release_Requirements.md#execution-environment referenced this page.

Not sure if this is worth moving or if it should just be discarded due to many out of date parts.

@vogella you have been doing a bunch of migration, your thoughts are solicited here.


vogella:
I think it is fine to convert as it is, if necessary this allows us to rework the information later.

Thanks @jonahgraham

PR #1168, opened 2024-02-25
Add bnd repository view

laeubi:
CQ, Description and cleanup will follow shortly.

Requires
- [x] Dependencies https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1811
- [x] CQ https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/13813


laeubi:
Build works now CQ is opened.

HannesWell:
Can't we use the original view from bnd as an ordinary dependency or is there a plan to use the original bnd-repository view at some time in the future? The amount of copied code is quite high.

laeubi:
> Can't we use the original view from bnd as an ordinary dependency or is there a plan to use the original bnd-repository view at some time in the future? The amount of copied code is quite high.

No, I used as much as is available as "reusable" components [here](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1811) already but the rest is just intermingled with bndtools and there is not much spin at bndtools to change it there.

iloveeclipse:
@laeubi : I see now compilation errors in PDE in my workspace.

Some dependencies are additionally required now, but there is no hint which exactly. 
Which bundles provide packages missing below:
- org.bndtools.api
- org.bndtools.versioncontrol.ignores.manager.api
- org.osgi.namespace.*

laeubi:
@iloveeclipse you run the setup actions this should discover new dependencies automatically, they where added here
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1811

merks:
I ran into this after a pull and yes, the setup actions fix all the problems as expected.   But I don't think @iloveeclipse uses the Oomph setups, so he might need to import some projects and probably needs to re-resolve the target platform.

laeubi:
> Yes, to be more precise: which p2 repositories contain the new requirements?

I-Build should contain them maybe already or you can try using the bnd repo from here https://github.com/bndtools/bnd?tab=readme-ov-file#using-the-latest-development-snapshot-build-of-bndbndtools but in general there is no specialized one.

PDE also produces a repository but I'm not sure about the URL (@akurtakov ?)

merks:
Not all requirements come from p2 repositories.  You can see these new things all come directly from Maven:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/0795d76a8b50cd88f322c19a262c0d3bdda6f4e5

That being said,  Orbit's generator infrastructure looks at the Platform's *.target and incorporates such dependencies so that they are also available from Orbit, so currently already here:

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/nightly/latest

iloveeclipse:
OK, thanks, I've fixed setup now: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1763

PR #1172, opened 2024-02-27
Execution environments

vogella:


vogella:
@HannesWell I don't know if we still use execution environments, I have the dim memory that we stopped doing this at some point but maybe this memory is wrong. So I did a few minor tweaks here to the content @jonahgraham migrated.

Issue #769, opened 2023-09-25
Creating a new product extension via the product configuration file does not update the plug-in  editor content


vogella:
If I create a product extension via the product configuration file this does not update the plug-in editor content

Steps to reproduce:

1.) Start with empty workspace
2.) Create plug-in without extensions -> no plugin.xml
3.) Open the MANIFEST.MF
4.) Created product and press the "New" button

![training_product28](https://github.com/eclipse-pde/eclipse.pde/assets/139910/55b46705-7829-4a90-8f24-300522dad032)

Add entries like the following:

![training_product30](https://github.com/eclipse-pde/eclipse.pde/assets/139910/3cac6c2d-fc0b-46ab-9197-7568355865aa)

5.) Press Finish and double-click on plugin.xml and check the "Extensions" tab -> empty
6.) Close editor and re-open -> "Extensions" filled

PR #1176, opened 2024-02-28
New product extension issue #769

SudhindraKKulkarni:
Extension tab of the plugin editor content does not get updated when new product is created via the product configuration file#769

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/769

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/PluginInputContextManager.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/PluginInputContextManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2000, 2015 IBM Corporation and others.
+ *  Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/PluginInputContextManager.java
```diff
@@ -53,16 +59,32 @@ public IModel getPluginModel() {
...
+
 	@Override
-	protected void structureChanged(IFile file, boolean added) {
-		// If a plugin.xml file has been added to the project the editor should update
-		if (added && ICoreConstants.PLUGIN_FILENAME_DESCRIPTOR.equalsIgnoreCase(file.getName())) {
-			IProject project = getCommonProject();
-			if (project == null || project.equals(file.getProject())) {
-				monitorFile(file);
-			}
+	public void resourceChanged(IResourceChangeEvent event) {
```
laeubi, line 64: is it really desired to replace the pld code?

SudhindraKKulkarni: The creation of a new product using the product configuration file creates the plugin.xml file.Due to this the "resourceChanged" api of InputContextManager gets called, but however the monitoredFiles list does not contain the "plugin.xml".Due to this the "structureChanged" api of PluginInputContextManager will never be called. Hence it is redundant and can be removed.
The "structureChanged" api of PluginInputContextManager was checking if plugin.xml was added to the project and if so was adding the plugin.xml to the monitoredFiles list.
This operation is now done in the "resourceChanged" api of the PluginInputContextManager so that the "structureChanged" api of the InputContextManager gets called and the plug-in editor content is updated with the new product extension.

iloveeclipse:
@SudhindraKKulkarni : do you plan to work on this fix?

SudhindraKKulkarni:
> @SudhindraKKulkarni : do you plan to work on this fix?

I have handled the review comments and committed the changes as part of the new PR https://github.com/eclipse-pde/eclipse.pde/pull/1230

Issue #1185, opened 2024-03-06
"build.properties does not exist" warning on no-PDE projects


jukzi:
Happens in platform workspace and is also asked often on Stackoverflow
![image](https://github.com/eclipse-pde/eclipse.pde/assets/51790620/35c702f9-8016-4f3e-aa03-0509cedd70aa)
It is produced by ManifestConsistencyChecker  - no idea why it runs on non-PDE projects, Anybody an idea? otherwise i would just add a doublecheck that it does nothing when the project has no plugin nature.

merks:
It's strange and annoying because it seems like an intermittent problem.  Have you been able to reproduce it in a debug launch?  I failed in that effort last I tried.

jukzi:
> Have you been able to reproduce it in a debug launch?

I was not able to reproduce it in debug or manual set up projects. Best thing i can offer is to double check and may be log an IllegalStateexception to get a stacktrace how it happens. it is only supposed to run when there is a 
```
		<buildCommand>
			<name>org.eclipse.pde.ManifestBuilder</name>
			<arguments>
			</arguments>
		</buildCommand>
	<natures>
		<nature>org.eclipse.pde.PluginNature</nature>
	</natures>
```
in the `.project` and since normally that seems to hold the error is never removed even if the workspace rebuilds (because such project are not build)

(see `projectToBuild.hasNature(PDE.PLUGIN_NATURE)` in org.eclipse.pde.internal.ui.preferences.PDECompilersConfigurationBlock.doFullBuild().new Job() {...}.run(IProgressMonitor)

laeubi:
> Happens in platform workspace and is also asked often on Stackoverflow

Is there a link / reproducer? I'm not sure I really understand when it happens, e.g if the builder is configured for the project or has this projects no builder and no nature at all?

tivervac:
Just want to confirm that I experienced the same in our company's workspace. Sadly, I have no reproducer.

I got it on the `.JETEmitters` project, which only has the `javabuilder` build command and the java nature. Other projects which have false positives even have _no_ builders and _no_ natures.

merks:
Yes, I keep getting this problem too, on many projects, but it kind of comes and goes.  Yesterday my Oomph environment has that, but a clean build make it go away.

jubax:
I seem to have the same problem. After I installed a new default Java runtime I suddenly got

[...]

PR #1186, opened 2024-03-06
Do not create "build.properties does not exist" marker on non-PDE #1185

jukzi:
but log a IllegalStateException() to find out how that happens.

https://github.com/eclipse-pde/eclipse.pde/issues/1185

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ManifestConsistencyChecker.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ManifestConsistencyChecker.java
@@ -31,6 +31,7 @@ import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.SubMonitor;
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ManifestConsistencyChecker.java
```diff
@@ -372,13 +375,21 @@ private void validateBuildPropertiesExists(IProject project) {
 			}
 			// if build.properties doesn't exist and build problems != IGNORE, create a marker on the project bug 172451
 			try {
+				if (!project.hasNature(PDE.PLUGIN_NATURE)) {
+					// hod did this happen?
```
merks, line 379: I assume this is temporary so the hod versus how is not so important.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ManifestConsistencyChecker.java
```diff
@@ -364,6 +366,7 @@ private void validateBuildProperties(IProgressMonitor monitor) {
 
 	// Will place a marker on the project if the build.properties does not exist
 	private void validateBuildPropertiesExists(IProject project) {
+
```
merks, line 369: Why add a blank line?

merks (approved):
It's fine.

gireeshpunathil (approved):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ManifestConsistencyChecker.java
jukzi: thanks for mentioning this typos

PR #1188, opened 2024-03-07
Use soft references to cache image data in Plug-in Image Browser

### LINKED ISSUE #1179: OutOfMemoryError in Plug-in Image Browser
This problem frequently shows up in my workspace, whenever I try to look at all icons that are available in my workspace.

Because we keep our documentation together with the source code, we have roughly 4000 images in our workspace, a lot of them being large screenshots. I think that when populating the cache, those images eat up all of the available memory.

```
java.lang.OutOfMemoryError: Java heap space
	at org.eclipse.swt.internal.image.PNGFileFormat.setPixelData(PNGFileFormat.java:242)
	at org.eclipse.swt.internal.image.PNGFileFormat.setImageDataValues(PNGFileFormat.java:307)
	at org.eclipse.swt.internal.image.PNGFileFormat.readNonInterlacedImage(PNGFileFormat.java:492)
	at org.eclipse.swt.internal.image.PNGFileFormat.readPixelData(PNGFileFormat.java:321)
	at org.eclipse.swt.internal.image.PNGFileFormat.readNextChunk(PNGFileFormat.java:139)
	at org.eclipse.swt.internal.image.PNGFileFormat.loadFromByteStream(PNGFileFormat.java:80)
	at org.eclipse.swt.internal.image.FileFormat.loadFromStream(FileFormat.java:58)
	at org.eclipse.swt.internal.image.FileFormat.load(FileFormat.java:89)
	at org.eclipse.swt.graphics.ImageLoader.load(ImageLoader.java:154)
	at org.eclipse.swt.graphics.ImageDataLoader.load(ImageDataLoader.java:25)
	at org.eclipse.swt.graphics.ImageData.<init>(ImageData.java:334)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.AbstractRepository.createImageData(AbstractRepository.java:87)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.WorkspaceRepository.lambda$0(WorkspaceRepository.java:83)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.WorkspaceRepository$$Lambda$2661/0x0000023187401490.visit(Unknown Source)
	at org.eclipse.core.internal.resources.Resource.lambda$0(Resource.java:122)
	at org.eclipse.core.internal.resources.Resource$$Lambda$487/0x00000231862fa848.visitElement(Unknown Source)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:85)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:90)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:90)
	at org.eclipse.core.internal.watson.ElementTreeIterator.doIteration(ElementTreeIterator.java:90)
	at org.eclipse.core.internal.watson.ElementTreeIterator.iterate(ElementTreeIterator.java:135)
	at org.eclipse.core.internal.resources.Resource.accept(Resource.java:131)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.WorkspaceRepository.populateCache(WorkspaceRepository.java:69)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.AbstractRepository.run(AbstractRepository.java:66)
	at org.eclipse.pde.internal.ui.views.imagebrowser.repositories.WorkspaceRepository.run(WorkspaceRepository.java:117)
```

ptziegler:
The WorkspaceRepository class used by this view initializes the image data for all images that are found in the workspace. This approach doesn't scale very well and may cause OutOfMemory errors on large projects.

By using those soft references in combination with a lazy initialization, we can reduce the overall footprint on low-memory systems, without losing most of the benefits gained by caching.

~~Note that this approach doesn't work for images retrieved from jar files, as the image data can't be reloaded once the archive has been closed again which is why hard references are mandatory here.~~

One noteworthy difference to the old implementation that invalid/broken images are no longer skipped. Instead, the default image data of the ImageDescriptor (also used by e.g. the MissingImageDescriptor) is returned.

This change also resolves a suppressed ClassCastException in the searchDirectory() method of the AbstractRepository class, when trying to cast a java.io.File to an org.eclipse.core.resources.IFile.

Resolves #1179

### DIFF
--- a/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.ui; singleton:=true
-Bundle-Version: 3.15.100.qualifier
+Bundle-Version: 3.15.200.qualifier
 Bundle-Activator: org.eclipse.pde.internal.ui.PDEPlugin
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageElement.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageElement.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2012 Christian Pontesegger and others.
+ *  Copyright (c) 2012, 2024 Christian Pontesegger and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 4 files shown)


ptziegler:
This should -in principle- drastically reduce the change of an OOM error. And at first glance, the view is populated just fine using the lazy-loading mechanism.

image

laeubi:
> Note that this approach doesn't work for images retrieved from jar files, as the image data can't be reloaded once the archive has been closed again which is why hard references are mandatory here.

Thanks this looks good, just one note, even though you can't read from the same stream, it looks like the filename is known and therefore one can reopen the jar/zip again, do you think this is worth to consider?

laeubi:
I wonder if it would be good to use Bundle.findEntry instead, this should allow to get an URL to the content of the bundle without loading anything and then one can later load it by URL

ptziegler:
> I wonder if it would be good to use Bundle.findEntry instead, this should allow to get an URL to the content of the bundle without loading anything and then one can later load it by URL

While it would probably work -in principle-  I'm a little bit concerned that there might be unexpected side effects with this approach. The current implementation works directly on the file system. So there are no assumptions about whether the bundle is in the workspace or even whether it has been installed by the IDE.

> even though you can't read from the same stream, it looks like the filename is known and therefore one can reopen the jar/zip again, do you think this is worth to consider?

I was a little hesitant at first, but I think that makes sense. At the very least, the lazy-loading should be the same, regardless of whether the image is inside a jar.

laeubi:
> While it would probably work -in principle- I'm a little bit concerned that there might be unexpected side effects with this approach. The current implementation works directly on the file system. So there are no assumptions about whether the bundle is in the workspace or even whether it has been installed by the IDE.

I thought it shows the images of installed bundles but then maybe I'm wrong :-)

Issue #1189, opened 2024-03-09
Product Editor "The product editor includes native" clipped text


Phillipus:
Eclipse SDK
Version: 2024-03 (4.31)
Build id: I20240229-0520

Tested on Windows and Mac.

Should say "The product editor includes native launcher artifacts"

<img width="706" alt="Screenshot 2024-03-09 at 13 05 57" src="https://github.com/eclipse-pde/eclipse.pde/assets/600504/b6786470-7e77-4a96-9d47-5c904ec6dbfc">

Phillipus:
Maybe this did it?

laeubi:
It shows fine under linux.

Phillipus:
https://github.com/eclipse-pde/eclipse.pde/pull/1190

Phillipus:
> It shows fine under linux.

When making changes to the UI it really is a good idea to check on Windows, Mac and Linux.

laeubi:
> > It shows fine under linux.
> 
> When making changes to the UI it really is a good idea to check on Windows, Mac and Linux.

Feel free to contact me and offer additional hardware for testing ;-)

Beside that it looks like a bug in SWT, a label should always show with its full size regardless of filling given by grid layout.

Phillipus:
> Feel free to contact me and offer additional hardware for testing ;-)

No I won't do that because I live off of PayPal donations to support my product, and that's my only income. From that I have to buy Macs, Windows machines and other infrastructure because I care about my users. Failing that you could ask others to test.

> Beside that it looks like a bug in SWT, a label should always show with its full size regardless of filling given by grid layout.

I really don't think so. 

There's a PR that fixes it.

laeubi:
> I really don't think so.

Why? As said filling the label might fix it visually but should not be necessary.

> There's a PR that fixes it. I spent over an hour of my time on this, so thanks to me for that. You're welcome.

Thank you very much for looking into it and proposing a fix.

Phillipus:
Take it or leave it, I'm out.

merks:
It definitely doesn't show fine on Windows without the PR and the PR definitely fixes the problem, on Windows. 

Thanks @Phillipus  ❤️ 

Of course we'd all like multiple hardware, but virtual boxes work too, though it's super hard to get one for macos.  Even then, it's a lot of extra work to try each and every thing out of multiple OSes so mostly we assume that it works the same everywhere, which is not a great assumption but is true more often than not....

PR #1190, opened 2024-03-09
Fix clipped text in Product Editor

Phillipus:
- See https://github.com/eclipse-pde/eclipse.pde/issues/1189


laeubi (approved):
Looks good to me.

merks (approved):
I tested locally before and after.  This definitely fixes the problem on Windows.

Issue #146, opened 2022-06-08
Show of packages / bundles that are already imported.


laeubi:
One really annoying feature is that PDE hides bundles/packages if they are already imported.

This leads to strange behavior, e.g. I can't import a package if it is (transitively) imported and even worse the search is simply empty.

It would be good if PDE at least shows the package (maybe with a hint: already imported by...)

lathapatil:
I can make it look like below but I still need a hint on styling the text (grey out) inside the dialog  because the underlying  class responsible for the packages list is FilteredList.java which intern not using SWT label but a custom one.

https://github.com/eclipse-pde/eclipse.pde/assets/130035155/b2f2512f-0b12-484b-9a92-fd259329c560

Empty package image used for this may look inappropriate ? then it can be reverted to package image.

laeubi:
@lathapatil thanks for working on this, this looks fine to me, just a space between version and the text would be good.
Also it would be nice to have a checkbox (enabled by default) that reads "Show already imported packages"

lathapatil:
> @lathapatil thanks for working on this, this looks fine to me, just a space between version and the text would be good. Also it would be nice to have a checkbox (enabled by default) that reads "Show already imported packages"

Thanks for the quick feedback .

does it make sense to disable 'ADD' button during multi-selection if one of the selection is already imported ?

If this is not ok.. we can enable 'ADD' button even though one of the selection is already imported and can handle the addition of packages at backend

laeubi:
I think the button should only be disabled if ALL packages are already imported.

> Should this behavior be applied for "Required plug-ins" and "Exported Packages" as well?

Yes that would be good.

PR #1195, opened 2024-03-13
Packages/bundles that are already imported/added are shown

lathapatil:
Packages/bundles that are already imported/added are shown followed by [already imported/added]

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/146

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDELabelProvider.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDELabelProvider.java
@@ -16,6 +16,7 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.ui;
 
+import java.util.HashMap;
 import java.util.Locale;
 
 import org.eclipse.core.resources.IProject;
@@ -86,6 +87,7 @@ import org.eclipse.pde.internal.core.text.bundle.ExportPackageObject;
 import org.eclipse.pde.internal.core.text.bundle.ImportPackageObject;
 import org.eclipse.pde.internal.core.text.bundle.PackageObject;
 import org.eclipse.pde.internal.core.util.VersionUtil;
+import org.eclipse.pde.internal.ui.dialogs.PluginSelectionDialog;
 import org.eclipse.pde.internal.ui.elements.NamedElement;
 import org.eclipse.pde.internal.ui.util.SWTUtil;
 import org.eclipse.pde.internal.ui.util.SharedLabelProvider;
@@ -96,6 +98,11 @@ import org.osgi.resource.Resource;
 
 public class PDELabelProvider extends SharedLabelProvider {
 	private static final String SYSTEM_BUNDLE = "system.bundle"; //$NON-NLS-1$
+	private IPluginModelBase currentModel;
+
+	public IPluginModelBase getCurrentPluginModel() {
+		return currentModel;
+	}
 
 	public PDELabelProvider() {
 	}
@@ -203,6 +210,13 @@ public class PDELabelProvider extends SharedLabelProvider {
 		}
 		if (pluginBase.getModel() != null && !pluginBase.getModel().isInSync())
 			text += " " + PDEUIMessages.PluginModelManager_outOfSync; //$NON-NLS-1$
+
+		HashMap<String, Boolean> existingImports = PluginSelectionDialog.getExistingImports(currentModel, false);
+		if (existingImports.get(pluginBase.getId()) != null && existingImports.get(pluginBase.getId())) {
+			text += " " + PDEUIMessages.PluginModelManager_alreadyAddedViaReexport; //$NON-NLS-1$
+		} else if (existingImports.get(pluginBase.getId()) != null && !existingImports.get(pluginBase.getId())) {
+			text += " " + PDEUIMessages.PluginModelManager_alreadyAdded; //$NON-NLS-1$
+		}
 		return text;
 	}
 
@@ -952,4 +966,7 @@ public class PDELabelProvider extends SharedLabelProvider {
 		return versionRange;
 	}
 
+	public void setCurrentModel(IPluginModelBase currentModel) {
+		this.currentModel = currentModel;
+	}
 }
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/PluginSelectionDialog.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/PluginSelectionDialog.java
@@ -18,7 +18,6 @@ package org.eclipse.pde.internal.ui.dialogs;
 
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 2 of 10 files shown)


gireeshpunathil:
@lathapatil - did change set from #1194 carry over into this PR?

lathapatil:
> @lathapatil - did change set from #1194 carry over into this PR?

No, I am trying to figure out what went wrong. Looks like #1194 has not reached here along with its branch !

gireeshpunathil:
it is probably because you branched out for this PR from the previous branch, so the tip of the branch got carried over.

nothing to worry: I will help you fix that. if you are using command line then do this:
 - `git rebase -I HEAD~2`
 - this opens up the commit editor in vim editor
 - you will see 2 entries for the two commits.
 - remove the line that has `214071f` in the commit hash
 - save and close
 - git push -f <origin branch> < this pr branch>`

lathapatil:
@laeubi Could you please test the PR if it meets the expectations and possibly review the code ?

laeubi:
@lathapatil do I need to consider something special?

I fetched this PR then started an embedded workbench and tried the following:
grafik

but package `aQute.bnd.build` is not shown.

I also think it would be good to have a checkbox (enabled by default) "Show already imported packages", at best the user choice is persisted to survive a restart.

lathapatil:
> I fetched this PR then started an embedded workbench and tried the following:

Checkout PR , in PR branch run Eclipse application and create a plugin project , and check manifest file for import packages , require bundles, export packages

This is what I have tried and the things are reflecting.

What do you mean by embedded workbench ?

lathapatil:
> I also think it would be good to have a checkbox (enabled by default) "Show already imported packages"

This is already part of the PR. But I observed there is some delay in loading all packages

lathapatil:
Please find attached video for working of the feature from the branch https://github.com/lathapatil/eclipse.pde/tree/Issues/146_Show_imported_exported_required_bundles


https://github.com/eclipse-pde/eclipse.pde/assets/130035155/87bb4aac-1215-4e8c-a329-aa936ad4b27a

lathapatil:
@laeubi could you please try testing and reviewing this ?
@gireeshpunathil Can you please try to test this .. if you are also not able to see the changes

gireeshpunathil:
@lathapatil - sure, will do. also, could you pls rebase your changes? as it was quite some time, now there is a conflict that has emerged.

lathapatil:
> @lathapatil - sure, will do. also, could you pls rebase your changes? as it was quite some time, now there is a conflict that has emerged.

Now it is rebased with conflicts resolved. Could you please verify the behavior

gireeshpunathil:
@lathapatil - thanks, I tested this and found two minor issues. not sure if you want to fix it in this PR or another one:
 - when I select plugins or packages that are already in and not already in together, then the `add` button is ON. puzzled on what would be the right behaviour!
 
Screenshot 2024-10-22 at 5 53 49 PM

 - export packages section is showing empty, even when I have one. your video shows it right, however. so I am wondering what would be the difference between the environment of yours and mine!

Screenshot 2024-10-22 at 6 00 23 PM

gireeshpunathil:
@lathapatil - I tested the new changes, and it works fine, as you have showed in the video. thanks!

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/PluginSelectionDialog.java
```diff
@@ -153,8 +159,8 @@ private static IPluginModelBase[] getElements(boolean includeFragments) {
 		return PluginRegistry.getActiveModels(includeFragments);
 	}
 
-	public static HashSet<String> getExistingImports(IPluginModelBase model, boolean includeImportPkg) {
-		HashSet<String> existingImports = new HashSet<>();
+	public static HashMap<String, Boolean> getExistingImports(IPluginModelBase model, boolean includeImportPkg) {
```
HannesWell, line 162: Please use only collection interfaces (as abstract as possible) if possible here and in all other changed code. Using a specific implementation is usually unnecessarily restrictive.
```suggestion
	public static Map<String, Boolean> getExistingImports(IPluginModelBase model, boolean includeImportPkg) {
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDELabelProvider.java
```diff
@@ -203,6 +210,13 @@ public String getObjectText(IPluginBase pluginBase) {
 		}
 		if (pluginBase.getModel() != null && !pluginBase.getModel().isInSync())
 			text += " " + PDEUIMessages.PluginModelManager_outOfSync; //$NON-NLS-1$
+
+		HashMap<String, Boolean> existingImports = PluginSelectionDialog.getExistingImports(currentModel, false);
```
HannesWell, line 214: Instead of setting `IPluginModelBase currentModel` via a setter and saving it, could you just use the following?
```
if (pluginBase.getModel() instanceof IPluginModelBase pluginModel) {
  ...
```
Because the less state we have the better. I also don't know if `currentModel` can be null at this point in any usage scenario and if the called `PluginSelectionDialog.getExistingImports` method would handle that well.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/PluginSelectionDialog.java
```diff
@@ -298,4 +305,22 @@ protected void createButtonsForButtonBar(Composite parent) {
...
+		super.updateButtonsEnableState(status);
+		Button okButton = getOkButton();
+		StructuredSelection currentSelection = super.getSelectedItems();
+		HashMap<String, Boolean> existingImports = PluginSelectionDialog
+				.getExistingImports(PDEPlugin.getDefault().getLabelProvider().getCurrentPluginModel(), false);
+		if (!currentSelection.isEmpty())
+			okButton.setEnabled(false);
+		for (Object selection : currentSelection) {
+			if (selection instanceof IPluginModelBase
+					&& !(existingImports.keySet().contains(((IPluginModelBase) selection).getPluginBase().getId()))) {
```
HannesWell, line 319: Please use pattern matching for instanceof in all changed code and if applicable.
```suggestion
			if (selection instanceof IPluginModelBase pluginModel && !(existingImports.keySet().contains(pluginModel).getPluginBase().getId()))) {
```
Please also reformat the line, not sure if it fits in one line now.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/dialogs/PluginSelectionDialog.java
```diff
@@ -153,8 +159,8 @@ private static IPluginModelBase[] getElements(boolean includeFragments) {
 		return PluginRegistry.getActiveModels(includeFragments);
 	}
 
-	public static HashSet<String> getExistingImports(IPluginModelBase model, boolean includeImportPkg) {
-		HashSet<String> existingImports = new HashSet<>();
+	public static HashMap<String, Boolean> getExistingImports(IPluginModelBase model, boolean includeImportPkg) {
+		HashMap<String, Boolean> existingImports = new HashMap<>();
```
HannesWell, line 163: Maybe I oversaw something, but I can only see the key-set of the map being used and values seem not to be considered at all. If that's correct, can this `Map` stay a `Set`?

HannesWell (commented):
Thanks for working on this.
I havn't done a technical/functional review but have a few style remarks below.

PR #1196, opened 2024-03-13
Multifix resolution for removal of bundle/package from Quick Fix  #685

### LINKED ISSUE #685: quick-fix for removing require-bundle: fault-lines do not match resolution batch
steps to recreate:
 - create a simple plugin project
 - add a few non-existent bundles under require-bundle header in the manifest
 - right click on the error(s) and select quick-fix
 - select `remove bundle from the list` option
 - select all the applicable lines from the below list and finish
 
observe only the first item in the list as removed. Expectation is to remove all the selected bundles.

lathapatil:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/685

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/RemoveExportPackageResolution.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/RemoveExportPackageResolution.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2005, 2020 IBM Corporation and others.
+ *  Copyright (c) 2005, 2024 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
+ *	   Latha Patil (ETAS GmbH) - Issue #685
  *******************************************************************************/
 package org.eclipse.pde.internal.ui.correction;
 
@@ -35,8 +36,19 @@ public class RemoveExportPackageResolution extends AbstractManifestMarkerResolut
 		String markerPackage = marker.getAttribute("packageName", (String) null); //$NON-NLS-1$
 		Bundle bundle = (Bundle) model.getBundle();
 		ExportPackageHeader header = (ExportPackageHeader) bundle.getManifestHeader(Constants.EXPORT_PACKAGE);
-		if (header != null)
+		if (header != null) {
+			// Issue 685 - If `markerPackage` can be null, it might indicate
+			// that the marker is related to the removal of RequireBundle or
+			// others. If it is not null, then markerPackage might represent the
+			// imported package or others, and the ExportPackageHeader(header)
+			// does not include it.
+			if (markerPackage == null || !header.hasPackage(markerPackage)) {
+				MultiFixResolution multiFixResolution = new MultiFixResolution(marker, null);
+				multiFixResolution.run(marker);
+			} else {
 			header.removePackage(markerPackage);
+			}
+		}
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


lathapatil:
> @lathapatil - when I tested your patch I did not see any difference unfortunately! it may be that the bundle references are removed rom the `Require-Bundle` header actually, but not reflected in the manifest editor.
> 
> I remember there is a way to sync the manifest content in the view and the in-memory copy; but forgot which one. can you pls check if there is one such thing?

It is reflecting in manifest file as well when I test from my workspace / branch. How do I reproduce this?

gireeshpunathil:
- add two new, bogus bundles against Require-Bundle: 
```
 Require-Bundle: org.eclipse.ui,
 foo,
 bar
```
 - select quick fix
 - select both the errors
 - apply
 
I see only one getting removed (the pre-existing behaviour).

HannesWell:
> @lathapatil - I don't want to block this PR just because I am unable to see the change - may be I am doing some silly mistake. So I request @HannesWell to have a look and advise.

I'll have a look at it as soon as possible, but I'm quite busy this week, so I probably won't have the time before the weekend.

lathapatil:
> @lathapatil - I don't want to block this PR just because I am unable to see the change - may be I am doing some silly mistake. So I request @HannesWell to have a look and advise.

@gireeshpunathil I've observed the same behavior you mentioned when unresolved bundles are added at the end, as in the example you provided. However, when we add them in between, it seems to work fine. I'll need to debug this case further. I'll keep you updated on my findings.

lathapatil:
@gireeshpunathil When I debugged this case, I found that [while saving the removal of the first marker](https://github.com/eclipse-platform/eclipse.platform.ui/blob/dd5068c243f37be4ab941620170978a2a0858931/bundles/org.eclipse.core.filebuffers/src/org/eclipse/core/internal/filebuffers/ResourceTextFileBuffer.java#L353) into the manifest file, the other markers were not updated properly; [instead, they were deleted.](https://github.com/eclipse-platform/eclipse.platform.ui/blob/dd5068c243f37be4ab941620170978a2a0858931/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractMarkerAnnotationModel.java#L644)

Working case - 
Require-Bundle: bcpg;bundle-version="1.77.0",dummy1,dummy2,
 bcprov.source;bundle-version="1.77.0"
 
 Non-working case -
Require-Bundle: bcpg;bundle-version="1.77.0",
 dummy1,dummy2,
 bcprov.source;bundle-version="1.77.0"

After removal of the first bundle manifest file is being saved and when the file commit happens [updateMarkers ](https://github.com/eclipse-platform/eclipse.platform.ui/blob/dd5068c243f37be4ab941620170978a2a0858931/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractMarkerAnnotationModel.java#L614) method of AbstractMarkerAnnotationModel is called and here I see the difference with [annotationMap ](https://github.com/eclipse-platform/eclipse.platform.ui/blob/dd5068c243f37be4ab941620170978a2a0858931/bundles/org.eclipse.ui.editors/src/org/eclipse/ui/texteditor/AbstractMarkerAnnotationModel.java#L618) being populated in working case and empty in non-working case hence markers are not updated properly.

But I do not have deeper knowledge of these annotation models. It might take more time for me to check on this.. If you already know what's happening here you can help me out !

gireeshpunathil:
@lathapatil - it is possible that we (not your changes, but the original design itself) are missing to `connect` all the markers that are relevant, that appear before the current line being altered. 

for example:
 - line x is being quick-fixed, markers for lines x+ are scanned and updated: working case
 - line x is being quick-fixed, markers for lines x- are not scanned so not updated: failing case
 
so looks like you are in the right direction, focus on how the markers are selected for update - if we are moving forward in the document for searching of markers, we need to do it in both directions - at least in this case.

does it make sense at all?

PR #1199, opened 2024-03-15
[APITools] Various minor and local code clean-ups and simplifications

HannesWell:


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/FilterStore.java
```diff
@@ -58,22 +60,24 @@ public class FilterStore implements IApiFilterStore {
...
+	 * Constant representing the path of the default path of the .api_filters file.
 	 */
-	static final String SETTINGS_FOLDER = ".settings"; //$NON-NLS-1$
+	static final IPath API_FILTERS_XML_PATH = IPath.fromOSString(".settings") //$NON-NLS-1$
+			.append(IApiCoreConstants.API_FILTERS_XML_NAME);
 	/**
 	 * The mapping of filters for this store.
 	 */
-	protected HashMap<String, Set<IApiProblemFilter>> fFilterMap = null;
+	protected Map<String, Set<IApiProblemFilter>> fFilterMap = null;
```
HannesWell, line 70: @laeubi it looks like changing the type of this field breaks the `tycho-apitools-plugin:verify` verify mojo.
I saw executions that fail with
```
Caused by: java.lang.NoSuchFieldError: fFilterMap
    at org.eclipse.tycho.apitools.ApiAnalysis$3.initializeApiFilters (ApiAnalysis.java:470)
    at org.eclipse.pde.api.tools.internal.FilterStore.isFiltered (FilterStore.java:181)
```
Maybe I understand something wrong, but does Tycho really use internals of API-Tools and at the same time loads the classes from a current I-builds repo?

HannesWell: Created https://github.com/eclipse-pde/eclipse.pde/pull/1203 to restore the field of the internal class for now.
But on the long run there should be a more robust solution.

laeubi: > Maybe I understand something wrong, but does Tycho really use internals of API-Tools and at the same time loads the classes from a current I-builds repo?

Tycho loads the api-tools bundles and then runs an embedded framework where it calls the code like a regular bundle, as API Tools have literally no API it must call the (exported) internal API. As it compiles against the latest release there is some chance of incompatibility if one mixes a newer/older API tools Tycho is not build against.

In this case it extends `FilterStore` in `org.eclipse.tycho.apitools.ApiAnalysis.getApiFilterStore(BundleComponent)` to intercept the loading as otherwhise there is no way to influence where the store loads the file from, you can provide a constructor that allows to supply a file (or path) then we can use that one from the next release on.

PR #1200, opened 2024-03-16
Fix UseScanManager.isValidArchive() and UseReportConverter

HannesWell:
In UseScanManager.isValidArchive() the jar/zip-file was closed too early making it impossible to read from it.

And in UseReportConverter.getXML() the default look up looked not at the wrong file when computing the filtered-counts.

Since especially the former seems to be broken for some time I wonder if API use-scans are a  used feature?
Does anybody know if the is used for the Eclipse SDK or SimRel? With that I could verify if this really is fixed (at the moment this is just an anticipated fix from looking at the code). Maybe @merks or @iloveeclipse?

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/search/UseReportConverter.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/search/UseReportConverter.java
@@ -1680,7 +1680,7 @@ public class UseReportConverter extends HTMLConvertor {
 		}
 		// try looking in the default 'xml' directory as a raw report root
 		// might have been specified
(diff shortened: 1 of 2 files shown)


merks:
I imagine/expect they are only used if it can be used via Tycho.

laeubi:
Sometimes these things are made available as an Ant Task and then called by scripts. Beside that Tychon "only" calls the `BaseApiAnalyzer` so whatever is analyzed there will be used.

HannesWell:
The code (probably) broken is only called by the preference-page. My remark was meant if API use-scans are used in general and if anybody knows a place where some are produced so that I could use them as sample or if I had to create them by myself.

HannesWell:
Let's submit this, although I cannot test this at the moment, it should not make it worse and I expect to improve the situation.

PR #1202, opened 2024-03-17
Fix escaping of line-break in api.tools/build.properties

HannesWell:
Forgotten in https://github.com/eclipse-pde/eclipse.pde/pull/1116.


laeubi (approved):
Looks good

Issue #1205, opened 2024-03-18
Two ClasspathResolutionTest failures since  I20240317-1800


iloveeclipse:
We have two new ClasspathResolutionTest failures since https://download.eclipse.org/eclipse/downloads/drops4/I20240317-1800/

See for example https://download.eclipse.org/eclipse/downloads/drops4/I20240317-1800/testresults/html/org.eclipse.pde.ui.tests_ep432I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

Most likely regression from ab28042c7d6616e422203cf4d731fea9a923b062 that added new dependencies.

```
org.eclipse.pde.core.tests.internal.classpath.ClasspathResolutionTest	testImportSystemPackageDoesntAddExtraBundleJava8	Failure	Expecting empty but was: ["org.osgi.annotation.versioning_1.1.2.202109301733.jar", "org.osgi.annotation.bundle_2.0.0.202202082230.jar", "org.osgi.service.component.annotations_1.5.1.202212101352.jar", "org.osgi.service.metatype.annotations_1.4.1.202109301733.jar"]

java.lang.AssertionError:
Expecting empty but was: ["org.osgi.annotation.versioning_1.1.2.202109301733.jar",
"org.osgi.annotation.bundle_2.0.0.202202082230.jar",
"org.osgi.service.component.annotations_1.5.1.202212101352.jar",
"org.osgi.service.metatype.annotations_1.4.1.202109301733.jar"]
at org.eclipse.pde.core.tests.internal.classpath.ClasspathResolutionTest.testImportSystemPackageDoesntAddExtraBundleJava8(ClasspathResolutionTest.java:132)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:276)
at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:276)
at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:276)
```

```
[...]

PR #1207, opened 2024-03-19
Add org.eclipse.pde.feature dependency to pull additional requirements

laeubi:
Currently the tests are missing some indirect requirements defined in the feature on their test, this adds the org.eclipse.pde.feature to make sure all declared items are pulled in.

See 
- https://github.com/eclipse-pde/eclipse.pde/issues/1205


HannesWell (approved):
Looks good and should fix the I-build test-failure. 👍🏽

PR #1212, opened 2024-03-25
Exclude upper-bound in version ranges when adding Package-imports

HannesWell:
I did that wrong in my proposal in https://github.com/eclipse-pde/eclipse.pde/pull/1007#pullrequestreview-1782918647.


HannesWell:
I also fixed wrong version ranges in PDE bundles, probably introduced by this bug.
But there are more in the SDK, although many are much older and not caused by this bug.

Issue #802, opened 2023-10-18
PDE uses a different stack then the regular stack for its views


vogella:
Last week I gave an introduction to a group of new PDE users and the main complaint they had was that we are using two stacks in case PDE perspective is used.

![image](https://github.com/eclipse-pde/eclipse.pde/assets/139910/ca9b4219-291f-45a6-88e1-c09b395e0705)

To reproduce, reset your PDE perspective and open the Console.

Still very annoying. Git Tooling seems to have no issue using the "regular" bottom stack. 

IIRC this was reported in the past but I think in Bugzilla.

Dinesh0723:
@vogella , refer the attached image,
![2024-04-11_13h59_43](https://github.com/eclipse-pde/eclipse.pde/assets/130960675/67c70e76-9204-465e-8539-5470d7ca5d56).

Are there any other views that need to be opened in the same stack?

vogella:
@Dinesh0723 did you fix something? What if you reset the perspective?

Dinesh0723:
> @Dinesh0723 did you fix something? What if you reset the perspective?

Yeah, I have added a placeholder for respective views in the above screenshot.

If we do reset perspective, it will show only the default views (Problems view, Target Platform State view) in bottom stack..

vogella:
Looks good to me. Thanks for the fix. Issue can be closed in case your fix is merged.

Dinesh0723:
Sure. Thank you.

vogella:
Thank you for fixing

HannesWell:
> Looks good to me. Thanks for the fix. Issue can be closed in case your fix is merged.

@Dinesh0723 can you please link your PR here?

vogella:
Fixed via https://review.gerrithub.io/c/eclipse-egit/egit/+/1245418

PR #1229, opened 2024-04-15
PDE uses a different stack then the regular stack for its views.

Dinesh0723:
PDE should support to open Console, Git Repositories and Git Staging views in a same stack.

Fixes:https://github.com/eclipse-pde/eclipse.pde/issues/802

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPerspective.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEPerspective.java
@@ -11,6 +11,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Mickael Istria (Red Hat Inc.) - 482905 Project Explorer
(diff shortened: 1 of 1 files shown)


vogella:
Sorry for overseeing the PR. Are you still around? 

The Console line is right and should go in. org.eclipse.ui.console contributes no perspectiveExtension at all, so in a perspective without a placeholder the compat layer creates a brand new stack for it. Adding the placeholder is precisely what JDT does (JavaPerspectiveFactory.java:56, same constant, same bottom folder). And org.eclipse.ui.console is already a required bundle in ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF:116, so no manifest change is needed.

The Egit lines should be removed, I fix this directly in EGit.

vogella:
Egit Gerrit https://review.gerrithub.io/c/eclipse-egit/egit/+/1245418

vogella:
Fixed via https://review.gerrithub.io/c/eclipse-egit/egit/+/1245418

Issue #1232, opened 2024-04-15
Expensive operations running on initialization of tabs


fedejeanne:
Some tabs still rely on the method `initializeFrom` to perform the heavy lifting. 

This sometimes leads to unnecessary (sometimes even expensive) calculations, since not all tabs are ever going to be **activated** (_i.e._ selected).

These are the ones I've identified so far:

* ~`TestTab` (belongs to _Plugin **Test**_ configuration)~ --> Not possible, see [comment](https://github.com/eclipse-pde/eclipse.pde/issues/1232#issuecomment-2061200283)
* [ ] `PluginsTab` (in the _Plugin_ and _Plugin **Test**_ configurations)  --> #1233

fedejeanne:
BTW these are some samplings showing the initialization times of `TestTab` and `PluginsTab` (see https://github.com/vi-eclipse/Eclipse-PDE/issues/2)

![Image](https://github.com/vi-eclipse/Eclipse-PDE/assets/2205684/b786ac81-d58d-4093-91c5-c3250a3c8dfd)

Snapshot (same as https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1214): [open-run-configurations-dialog.zip](https://github.com/eclipse-jdt/eclipse.jdt.ui/files/11937105/open-run-configurations-dialog.zip)

fedejeanne:
#1233 was merged and then reverted (#1251). Since the new fix (#1252) has been lying around for more than a year, I'd say this issue is not going to be solved.

Issue #1227, opened 2024-04-14
PDE errors on all bundles that use system packages since I20240412-1800


iloveeclipse:
Since https://download.eclipse.org/eclipse/downloads/drops4/I20240412-1800/ we have PDE errors reported on **all** bundles that make use of system packages, like on jface:

```
Description	Resource	Path	Location	Type
Neither an available bundle nor the associated JRE export package 'javax.xml.parsers'	MANIFEST.MF	/org.eclipse.jface/META-INF	line 42	Plug-in Problem
Neither an available bundle nor the associated JRE export package 'org.w3c.dom'	MANIFEST.MF	/org.eclipse.jface/META-INF	line 44	Plug-in Problem
Neither an available bundle nor the associated JRE export package 'org.xml.sax'	MANIFEST.MF	/org.eclipse.jface/META-INF	line 45	Plug-in Problem
```

![image](https://github.com/eclipse-pde/eclipse.pde/assets/964108/04a304d1-7262-4f1c-8cfa-44266b9ae6ba)

This is a regression from https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/426, reverting it fixes the issue.

TL;DR: at some place in PDE broken 23 EE is picked up as compatible with Java 17, and it since it has zero system packages, we have trouble in resolving system package dependencies in PDE.

Looks like adding "broken" `JavaSE-23 EE` profile (it doesn't have proper compliance set in `org.eclipse.jdt.internal.launching.environments.EnvironmentsManager.getExecutionEnvironmentCompliance(IExecutionEnvironment)`) confuses PDE. I haven't found exact place yet, but it is most likely due the fact that EE's are sorted by compliance and since  `JavaSE-23 EE` has `1.3` compliance it seem to be picked before any other "Java 17 compatible" EE and because of `org.eclipse.pde.internal.build.site.PDEState.querySystemPackages(IExecutionEnvironment, Properties)` code no system packages can be found if Java 23 JVM is not "installed" as JVM in Eclipse.

This is most likely also the reason for 3 new PDE test failures in ClasspathResolutionTest that needs proper resolved system packages, see
https://download.eclipse.org/eclipse/downloads/drops4/I20240412-1800/testresults/html/org.eclipse.pde.ui.tests_ep432I-unit-win32-java17_win32.win32.x86_64_17.html

iloveeclipse:
[...]

PR #946, opened 2023-11-29
Recreate #674: Postpone long-running operations to the activation of the "Tracing" tab.

fedejeanne:
Recreate #674 which was reverted in #925.

Needs:
* [x] https://github.com/eclipse-platform/eclipse.platform/pull/860

Fixes to https://github.com/eclipse-pde/eclipse.pde/issues/679

PR #1233, opened 2024-04-15
Move long running operations to the activation of the PluginsTab

fedejeanne:
Move the long running operations in the `PluginsTab` from `initializeFrom` to `activated`. Only run these operations the first time the tab is activated in order to:
a. Only run them when the tab is actually selected
b. Not run them again if the tab is deselected and selected again (without leaving the current launch configuration)

## How to test
* Create 2 **new** launch configurations from inside the _Run Configurations_ dialog: one _Eclipse Application_ (1) and one _JUnit-Plugin Test_ (2)
* Select the `Plugins` tab in one of them and make some changes (save them)
* Move to the other configuration --> The `Plugins` tab should be selected and it should look fine. The activation should **not** be skipped _i.e._ it should **not** hit the `return;` in line 167
* Move back to the first configuration --> The `Plugins` tab should be selected and contain all changes you applied before. The activation should **not** be skipped
* Select another tab
* Select the `Plugins` tab again --> The activation should be skipped _i.e._ hit should hit the `return;` in line 167

image

* [x] I tested my changes thoroughly

Contributes to https://github.com/eclipse-pde/eclipse.pde/issues/1232

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
@@ -59,6 +59,7 @@ public class PluginsTab extends AbstractLauncherTab {
 	private Combo fDefaultAutoStart;
 	private Spinner fDefaultStartLevel;
 	private final Listener fListener;
+	private boolean fActivated;
 
 	private static final int DEFAULT_SELECTION = 0;
 	private static final int PLUGIN_SELECTION = 1;
@@ -156,6 +157,16 @@ public class PluginsTab extends AbstractLauncherTab {
 
 	@Override
 	public void initializeFrom(ILaunchConfiguration configuration) {
+		// Long-running initialization happens on first activation of this tab
+	}
+
+	@Override
+	public void activated(ILaunchConfigurationWorkingCopy configuration) {
+		if (fActivated) {
+			// Since this method can be expensive, only activate this tab once.
+			return;
+		}
+
 		try {
 			int index = DEFAULT_SELECTION;
 			if (configuration.getAttribute(IPDELauncherConstants.USE_CUSTOM_FEATURES, false)) {
@@ -171,6 +182,9 @@ public class PluginsTab extends AbstractLauncherTab {
 			fDefaultAutoStart.setText(Boolean.toString(auto));
 			int level = configuration.getAttribute(IPDELauncherConstants.DEFAULT_START_LEVEL, 4);
 			fDefaultStartLevel.setSelection(level);
+
+			// If everything ran smoothly, this tab is activated
+			fActivated = true;
 		} catch (CoreException e) {
 			PDEPlugin.log(e);
 		}
@@ -188,7 +202,9 @@ public class PluginsTab extends AbstractLauncherTab {
 		int index = fSelectionCombo.getSelectionIndex();
 		configuration.setAttribute(IPDELauncherConstants.USE_DEFAULT, index == DEFAULT_SELECTION);
 		configuration.setAttribute(IPDELauncherConstants.USE_CUSTOM_FEATURES, index == FEATURE_SELECTION);
-		fBlock.performApply(configuration);
+		if (fActivated) {
+			fBlock.performApply(configuration);
+		}
 		// clear default values for auto-start and start-level if default
 		String autoText = fDefaultAutoStart.getText();
 		if (Boolean.toString(false).equals(autoText)) {
@@ -217,16 +233,9 @@ public class PluginsTab extends AbstractLauncherTab {
 		return fImage;
 	}
 
-	/**
-	 * Validates the tab.  If the feature option is chosen, and the workspace is not correctly set up,
-	 * the error message is set.
-	 *
-	 * @see org.eclipse.pde.ui.launcher.AbstractLauncherTab#validateTab()
-	 */
 	@Override
 	public void validateTab() {
-		String errorMessage = null;
-		setErrorMessage(errorMessage);
+		setErrorMessage(null);
 	}
 
 	@Override
--- a/org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
+++ b/org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.doc.user; singleton:=true
-Bundle-Version: 3.15.100.qualifier
+Bundle-Version: 3.15.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.help;bundle-version="[3.2.0,4.0.0)"
--- a/org.eclipse.pde.doc.user/pom.xml
+++ b/org.eclipse.pde.doc.user/pom.xml
@@ -17,7 +17,7 @@
     <version>4.32.0-SNAPSHOT</version>
   </parent>
   <artifactId>org.eclipse.pde.doc.user</artifactId>
-  <version>3.15.100-SNAPSHOT</version>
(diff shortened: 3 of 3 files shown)


fedejeanne:
Test failures seem to be unrelated and they fail also locally (I tested on my machine) even before applying this PR. See #1227 

The license check seems unrelated too, though I don't know why it happens.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
```diff
@@ -156,6 +157,16 @@ public void createControl(Composite parent) {
 
 	@Override
 	public void initializeFrom(ILaunchConfiguration configuration) {
+		// Long-running operations happen during the first activation of the tab
```
HannesWell, line 160:
```suggestion
		// Long-running initialization happens on first activation of this tab
```

laeubi:
I just remember that there was already an attempt in the past and there where also some glitches so one should be aware of that.

The build failure effectively says that you changed the public API documentation and therefore need to touch the docs bundle.

laeubi:
By the way if the expensive operation can be identified, it could be wrapped in a `BusyIndicator.call(...)` to have better UI feedback.

fedejeanne:
> Thank you for this.
> 
> I have tested this and in general this works well

👍 

> I just noticed that when I switch between the configs with the `Plug-ins` tab activate and don't save, the activation is never skipped. Is that intended?

Yes, it's intended: I can't skip the activation when first switching (or switching back) to a launch config because the tab is instantiated again and everything needs to be populated again. Skipping is only possible when switching back and forth between tabs in the same launch config because the config (and all its tab) remain in memory.

> I just remember that there was already an attempt in the past and there where also some glitches so one should be aware of that.

@laeubi you mean https://github.com/eclipse-pde/eclipse.pde/pull/946 ? The glitch was that the `activated` method of the default tab (or better said: "the first selected tab") of a launch configuration was not being called. I checked this now and it's being called :-)

fedejeanne:
I also inlined a call in `validateTab` and removed its JavaDoc since it was inconsistent.

I hope I addressed everything, @laeubi and @HannesWell ?

fedejeanne:
I found an error when creating a new Eclipse launch configuration in the dialog:

```java
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.Button.getSelection()" because "this.fIncludeOptionalButton" is null
	at org.eclipse.pde.internal.ui.launcher.AbstractPluginBlock.performApply(AbstractPluginBlock.java:951)
	at org.eclipse.pde.internal.ui.launcher.BlockAdapter.performApply(BlockAdapter.java:84)
	at org.eclipse.pde.ui.launcher.PluginsTab.performApply(PluginsTab.java:210)
	at org.eclipse.debug.ui.AbstractLaunchConfigurationTabGroup.performApply(AbstractLaunchConfigurationTabGroup.java:106)
	at org.eclipse.pde.ui.launcher.EclipseLauncherTabGroup.performApply(EclipseLauncherTabGroup.java:46)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupWrapper.performApply(LaunchConfigurationTabGroupWrapper.java:205)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupViewer.handleApplyPressed(LaunchConfigurationTabGroupViewer.java:1521)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.handleLaunchConfigurationSelectionChanged(LaunchConfigurationsDialog.java:1045)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.lambda$1(LaunchConfigurationsDialog.java:607)
	at org.eclipse.jface.viewers.StructuredViewer$3.run(StructuredViewer.java:820)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.viewers.StructuredViewer.firePostSelectionChanged(StructuredViewer.java:817)
	at org.eclipse.jface.viewers.ColumnViewer.firePostSelectionChanged(ColumnViewer.java:1065)
	at org.eclipse.jface.viewers.StructuredViewer.setSelection(StructuredViewer.java:1665)
	at org.eclipse.jface.viewers.TreeViewer.setSelection(TreeViewer.java:1091)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationView.handleConfigurationAdded(LaunchConfigurationView.java:330)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationView.lambda$0(LaunchConfigurationView.java:305)
...
```

I'm setting this PR to be a draft so I can further look into it tomorrow. I'll let you know when it's ready to be reviewed again

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
```diff
@@ -943,6 +947,12 @@ private int countChecked(Object[] elements) {
 	}
 
 	public void performApply(ILaunchConfigurationWorkingCopy config) {
+		if (!fControlCreated) {
+			// This method is also being called before the proper UI elements
+			// are created, which might cause a NPE
+			return;
+		}
```
fedejeanne, line 954: FYI this one is to prevent the NPE mentioned in [my comment](https://github.com/eclipse-pde/eclipse.pde/pull/1233#issuecomment-2061789215)

fedejeanne:
Ready for review. The NPE was because of a premature invocation of `AbstractPluginBlock.performApply`. Skipping that invocation until the block has been created (_i.e._ until **after** `createControl` has been called) fixes the issue and has no side-effects.

FTR:
* [x] I tested my changes thoroughly :-)

ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
```diff
@@ -188,7 +202,9 @@ public void performApply(ILaunchConfigurationWorkingCopy configuration) {
 		int index = fSelectionCombo.getSelectionIndex();
 		configuration.setAttribute(IPDELauncherConstants.USE_DEFAULT, index == DEFAULT_SELECTION);
 		configuration.setAttribute(IPDELauncherConstants.USE_CUSTOM_FEATURES, index == FEATURE_SELECTION);
-		fBlock.performApply(configuration);
+		if (fActivated) {
```
HannesWell, line 205: > Ready for review. The NPE was because of a premature invocation of `AbstractPluginBlock.performApply`. Skipping that invocation until the block has been created (_i.e._ until **after** `createControl` has been called) fixes the issue and has no side-effects.

I think preventing the block's performApply method here already is better since we don't need to maintain the `isActivated` state in another place too.

HannesWell (commented):
Thanks for the update.
While trying out the change I noticed that the activated method is already called in a 'busy-curser' context, so we don't have to start another nested one and save the extra code(complexity). Btw. in this case I would have just passed `null` as Display to reuse the current one.

To prevent the NPE I also think we should use a more local check.
I pushed my suggestions as another commit to your branch. Please incorporate it as you find it suitable.

If you agree with the suggestions, I'm fine with this change.

fedejeanne:
@HannesWell thank you for the suggestion, I like the changes!

Ready to merge on my part 👍  

I assume in this case 2 commits are OK? If not, let me know and I can squash them and add you as co-author :-)

HannesWell:
Great.
Please just squash them. No need to persist the discussion in the git history. :)

fedejeanne:
Squashed 👍

fedejeanne:
Failing test is unrelated and already documented here: https://github.com/eclipse-pde/eclipse.pde/issues/554

HannesWell (approved):
Thank you @fedejeanne. Nice contribution.
Activating a Eclipse launch config becomes faster and faster with your contributions. :)

iloveeclipse:
I assume this change caused new regression,  please check https://github.com/eclipse-pde/eclipse.pde/issues/1250.

PR #1237, opened 2024-04-15
[APItools] Leverage ExecutionEnvironmentDescription(Map) constructor

HannesWell:
Leverage the new ExecutionEnvironmentDescription(Map) constructor introduced in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/282


HannesWell:
The last fix would be obsolete with https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/432.

PR #1238, opened 2024-04-16
Touching bundles after compiler change

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1979


iloveeclipse:
Test fails are known and on master, see https://github.com/eclipse-pde/eclipse.pde/issues/1227

Issue #1231, opened 2024-04-15
Error validating SDK launch due missing JRE 1.1 support


iloveeclipse:
After removing support for JRE 1.1 (https://github.com/eclipse-equinox/equinox/pull/572 & https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/424) there is a validation error shown on startup of SDK from SDK.

PDE should update it's logic to validate launches.

![image](https://github.com/eclipse-pde/eclipse.pde/assets/964108/12c40bf7-0935-4d99-b232-2c013f0a3923)

akurtakov:
@merks As this one comes from Orbit I would say that the fix should be there, no?

iloveeclipse:
PDE allows to use **any** libraries in the platform, and PDE can be used to debug older platforms.
So the fix must be in PDE / SDK.

merks:
There's nothing wrong with such a requirement.  Even the latest Java 21 provides those system capabilities:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/35e929d7-cba0-4101-8669-8e7aa71357ff)

So if PDE says those capabilities are missing, it's not looking at the right thing or not looking in the right way.

akurtakov:
Acked about capability being provided and thus PDE fix is due.
https://github.com/eclipse-equinox/equinox/issues/571#issuecomment-2046055636 makes me believe that we should try to not ship bundles with "non-standard" EEs.

merks:
FYI, for history, you can see this issue:

https://github.com/eclipse-orbit/orbit/issues/52

This legacy bundle is a drop-in replacement for this:

https://repo1.maven.org/maven2/xml-apis/xml-apis-ext/1.3.04/

That thing provides no source.  If it did, I would have just wrapped in with m2e/tycho BND support which would produce this manifest:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/7b59aae7-903b-4dc7-86a5-6d88ea7839e6)

So standard or not, BND produces that and org.eclipse.osgi.storage.Storage.calculateEECapabilities(Version) computes that, so I tried to reproduce something that includes actual sources and that follows that same expect result.

HannesWell:
> PDE allows to use **any** libraries in the platform, and PDE can be used to debug older platforms. So the fix must be in PDE / SDK.

[...]

PR #1239, opened 2024-04-16
Rely on system-packages of Equinox JRE-1.1 profile

HannesWell:
With https://github.com/eclipse-equinox/equinox/pull/596 the JRE-1.1 profile embedded in Equinox was updated to have the expected system-packages.
Therefore the hard-coded ones in PDE can be removed.

See also https://github.com/eclipse-pde/eclipse.pde/issues/1231#issuecomment-2057561119.


akurtakov:
I haven't looked into "why" but since this commit we get https://ci.eclipse.org/pde/job/eclipse.pde/job/master/643/maven-warnings/new/source.ac496625-1c6e-4e5d-9b31-fdbd8eeb12b5/#8143 in master builds.

HannesWell:
> I haven't looked into "why" but since this commit we get https://ci.eclipse.org/pde/job/eclipse.pde/job/master/643/maven-warnings/new/source.ac496625-1c6e-4e5d-9b31-fdbd8eeb12b5/#8143 in master builds.

I tested this before submitting this in a debugged IDE and at least there it worked.
I don't know the setup of the DS TCK exactly, could it be the case that there the Equinox system bundle from the latest releases and the latest I-build of PDE or even the one build is mixed in the runtime?
IIRC at least Tycho's API tools plugin did something like that, but @laeubi probably knows it better than I do.

laeubi:
The warnings will only go away after the next release when Tycho has picked up the Equinox change.

PR #1242, opened 2024-04-18
Allow launch prototypes in JUnit Plug-in Test (generic Test view)

akurtakov:
All the tabs are already hooked for prototype support (shared with other launch configs that support prototypes) thus only the final enable and show the configuration tab is needed.


akurtakov:
Unrelated test failure.

PR #1247, opened 2024-04-22
Clean caches upon deletion of a Target Definition #1246

### LINKED ISSUE #1246: TargetDefinition leak
While investigating [a potential memory leak in RCPTT](https://github.com/eclipse/org.eclipse.rcptt/issues/71) I've discovered that 
`org.eclipse.pde.internal.core.target.TargetPlatformService.deleteTarget(ITargetHandle)` does not clean up `org.eclipse.pde.internal.core.TargetPlatformHelper.fgCachedTargetDefinitionMap`.

Moreover I could not find an API that performs cleanup of both `TargetPlatformService` and `TargetPlatformHelper`, so the workaround ought to be managing both explicitly on every deletion.

Suggested fix: make `TargetPlatformService.deleteTarget()` cleanup the `TargetPlatformHelper` cache.

basilevs:
TargetPlatformHelper and TargetPlatformService keep separate caches for target definition, resulting in a leak of TargetDefinition objects.

The leak happens when TargetDefinition.resolve() caches data in TargetPlatformHelper.

Fixes #1246

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetPlatformService.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetPlatformService.java
@@ -74,6 +74,7 @@ import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.core.PDECore;
 import org.eclipse.pde.internal.core.PDEPreferencesManager;
 import org.eclipse.pde.internal.core.TargetDefinitionManager;
+import org.eclipse.pde.internal.core.TargetPlatformHelper;
 import org.osgi.service.prefs.BackingStoreException;
 
 /**
@@ -156,6 +157,7 @@ public class TargetPlatformService implements ITargetPlatformService {
 			fExtTargetHandles.remove(((ExternalFileTargetHandle) handle).getLocation());
 		}
 		((AbstractTargetHandle) handle).delete();
+		TargetPlatformHelper.getTargetDefinitionMap().remove(handle);
 	}
 
 	@Override
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/LocalTargetDefinitionTests.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/LocalTargetDefinitionTests.java
@@ -766,6 +766,22 @@ public class LocalTargetDefinitionTests extends AbstractTargetTest {
 	}
 
 	/**
+	 * Test for https://github.com/eclipse-pde/eclipse.pde/issues/1246
+	 */
+	@Test
+	public void testDeleteCleansCaches() throws Exception {
+		ITargetDefinition definition = getNewTarget();
+		try {
+			assertFalse(TargetPlatformHelper.getTargetDefinitionMap().containsKey(definition.getHandle()));
+			definition.resolve(null);
+			assertTrue(TargetPlatformHelper.getTargetDefinitionMap().containsKey(definition.getHandle()));
+		} finally {
+			getTargetService().deleteTarget(definition.getHandle());
+			assertFalse(TargetPlatformHelper.getTargetDefinitionMap().containsKey(definition.getHandle()));
+		}
+	}
+
+	/**
 	 * Tests that a single (lower) version of a bundle can be included in the
 	 * target platform.
 	 */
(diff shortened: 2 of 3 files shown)


HannesWell (commented):
Thank you @basilevs for this fix.
In general it looks reasonable to me, but after a quick check of the callers I think we have to check in detail if `TargetPlatformPreferencePage.handleRemove/performOk()` needs adjustments to ensure a target is not prematurely removed from the map.
I can can help with that this evening.

HannesWell (approved):
> In general it looks reasonable to me, but after a quick check of the callers I think we have to check in detail if `TargetPlatformPreferencePage.handleRemove/performOk()` needs adjustments to ensure a target is not prematurely removed from the map.

I checked the preference-page in detail and the call to `TargetPlatformHelper.getTargetDefinitionMap().remove()` in `TargetPlatformPreferencePage.handleRemove` can be removed since it will be called in /`performOk()` with the removed targets when `TargetPlatformService.deleteTarget()` is called.

I applied that already so that this change can be submitted as soon as the build completes.

Thank you for this contribution.

basilevs:
> since it will be called in /`performOk()` with the removed targets when `TargetPlatformService.deleteTarget()` is called.

Not for Cancel?

HannesWell:
> > since it will be called in /`performOk()` with the removed targets when `TargetPlatformService.deleteTarget()` is called.
> 
> Not for Cancel?

Yes if the dialog is canceled `deleteTarget()` is not called but also the target is not deleted. Since `handleRemove()` only adds a target to the list of targets to be removed but does not yet delete them. So deleting it from the other cache at that point in time was actually too early since as discovered, the target is not deleted if Cancel is pressed.
Therefore moving the removal from the TargetPlatformHelper-cache to the time when the deletion is really performed actually also fixes that 'bug'.

HannesWell:
Btw. the new test fails in I-builds:
https://download.eclipse.org/eclipse/downloads/drops4/I20240518-1800/testresults/html/org.eclipse.pde.ui.tests_ep432I-unit-win32-java17_win32.win32.x86_64_17.html

HannesWell:
> Btw. the new test fails in I-builds: https://download.eclipse.org/eclipse/downloads/drops4/I20240518-1800/testresults/html/org.eclipse.pde.ui.tests_ep432I-unit-win32-java17_win32.win32.x86_64_17.html

For some of the test-classes the test cases are not found and zero tests are executed ...

```
Running AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.LocalTargetDefinitionTests
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s -- in AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.LocalTargetDefinitionTests
Running AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.WorkspaceTargetDefinitionTests
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s -- in AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.WorkspaceTargetDefinitionTests
```
I'll investigate that.

HannesWell:
> For some of the test-classes the test cases are not found and zero tests are executed ...
> 
> ```
> Running AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.LocalTargetDefinitionTests
> Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s -- in AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.LocalTargetDefinitionTests
> Running AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.WorkspaceTargetDefinitionTests
> Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s -- in AllPDETests org.eclipse.pde.ui.tests.target.AllTargetTests org.eclipse.pde.ui.tests.target.WorkspaceTargetDefinitionTests
> ```
> 
> I'll investigate that.

Looks like I disabled those tests in https://github.com/eclipse-pde/eclipse.pde/pull/745#issuecomment-1722879441 since they have to be reworked to be able to run in I-builds.
https://github.com/eclipse-pde/eclipse.pde/blob/e41da1385c56b9a2ac0808b584ae7f77809eabe1/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/LocalTargetDefinitionTests.java#L75-L77

I'll have a look at it.

PR #1248, opened 2024-04-23
Combine definition and declaration in ApiFilterStoreTests

akurtakov:
Minimal readability improvement found while looking into unstable test.


akurtakov:
@HannesWell As I have you here, what do you think of the MacOS gha failure? Java 8 no longer available on macos?

HannesWell:
> @HannesWell As I have you here, what do you think of the MacOS gha failure? Java 8 no longer available on macos?

That's new to me. Is this really permanent or maybe just a temporary issue?

akurtakov:
Failed on rerun here. We will see on master..

PR #1249, opened 2024-04-24
Publish a product IU into the PDE launch profile

laeubi:
Currently PDE published already root-feature what is already a great win. But in a usual product install there is also one unit for the product and because that is missing it can't be updated in a launched product from the IDE.

This adds a very bare unit to the profile by just publish the product id/version/name, but already allows that P2 pick up the unit when an update is requested.


merks:
I assume you mean IU and not UI...

PR #1253, opened 2024-04-24
Add an option to ignore references

tivervac:
Note: there might be another way to do this, as I found a few half-wired up places. However,  I couldn't find one while investigating my use case, so I added somewhat simple code that implements my use case.

I'm trying to proxy an entire target platform through an artifact repository, i.e., Nexus. As such, I've defined the software sites in my target platform to point to proxies of my own server. Sadly, it seems like p2 software sites are not necessarily self-contained. Namely, they can contain references to other repositories.

In our case, this causes us to load 122 external repos. Strikingly, most of these come from _disabled_ references. As such I've added some code that would skip over these, let me know if that isn't allowed for some reason.
Using only the enabled references, I still have 5 hits, much less, much faster, but not good enough. As such, I included a new backwards-compatible Software Site option that includes (or excludes) the traversal of references. In essence, this forces your software sites to be self-contained.

This ticket inspired another small ticket in [p2](https://github.com/eclipse-equinox/p2/pull/506).

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/IUBundleContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/IUBundleContainer.java
@@ -111,6 +111,12 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	 */
 	public static final int INCLUDE_CONFIGURE_PHASE = 1 << 3;
 
+
+	/**
+	 * Whether this container should follow repository references.
+	 */
+	public static final int FOLLOW_REPOSITORY_REFERENCES = 1 << 4;
+
 	/**
 	 * IU identifiers.
 	 */
@@ -152,10 +158,20 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	/**
 	 * Constructs a installable unit bundle container for the specified units.
 	 *
-	 * @param ids IU identifiers
-	 * @param versions IU versions
-	 * @param repositories metadata repositories used to search for IU's or <code>null</code> for default set
-	 * @param resolutionFlags bitmask of flags to control IU resolution, possible flags are {@link IUBundleContainer#INCLUDE_ALL_ENVIRONMENTS}, {@link IUBundleContainer#INCLUDE_REQUIRED}, {@link IUBundleContainer#INCLUDE_SOURCE}, {@link IUBundleContainer#INCLUDE_CONFIGURE_PHASE}
+	 * @param ids
+	 *            IU identifiers
+	 * @param versions
+	 *            IU versions
+	 * @param repositories
+	 *            metadata repositories used to search for IU's or
+	 *            <code>null</code> for default set
+	 * @param resolutionFlags
+	 *            bitmask of flags to control IU resolution, possible flags are
+	 *            {@link IUBundleContainer#INCLUDE_ALL_ENVIRONMENTS},
+	 *            {@link IUBundleContainer#INCLUDE_REQUIRED},
+	 *            {@link IUBundleContainer#INCLUDE_SOURCE},
+	 *            {@link IUBundleContainer#INCLUDE_CONFIGURE_PHASE},
+	 *            {@link IUBundleContainer#FOLLOW_REPOSITORY_REFERENCES}
 	 */
 	IUBundleContainer(String[] ids, Version[] versions, URI[] repositories, int resolutionFlags) {
 		fIds = ids;
@@ -368,7 +384,8 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	public synchronized IUBundleContainer update(Set<String> toUpdate, IProgressMonitor monitor) throws CoreException {
 		SubMonitor progress = SubMonitor.convert(monitor, 100);
 		URI[] updateRepos = fRepos == null ? null : fRepos.clone();
-		IQueryable<IInstallableUnit> source = P2TargetUtils.getQueryableMetadata(updateRepos, progress.split(30));
+		IQueryable<IInstallableUnit> source = P2TargetUtils.getQueryableMetadata(updateRepos, IsFollowRepositoryReferences(),
+				progress.split(30));
 		boolean updated = false;
 		String[] updateIDs = fIds.clone();
 		Version[] updateVersions = fVersions.clone();
@@ -550,7 +567,7 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	public boolean getIncludeAllRequired() {
 		// if this container has not been associated with a container, return its own value
 		if (fSynchronizer == null) {
-			return (fFlags & INCLUDE_REQUIRED) == INCLUDE_REQUIRED;
+			return (fFlags & INCLUDE_REQUIRED) != 0;
 		}
 		return fSynchronizer.getIncludeAllRequired();
 	}
@@ -566,7 +583,7 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	public boolean getIncludeAllEnvironments() {
 		// if this container has not been associated with a container, return its own value
 		if (fSynchronizer == null) {
-			return (fFlags & INCLUDE_ALL_ENVIRONMENTS) == INCLUDE_ALL_ENVIRONMENTS;
+			return (fFlags & INCLUDE_ALL_ENVIRONMENTS) != 0;
 		}
 		return fSynchronizer.getIncludeAllEnvironments();
 	}
@@ -580,7 +597,7 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	public boolean getIncludeSource() {
 		// if this container has not been associated with a container, return its own value
 		if (fSynchronizer == null) {
-			return (fFlags & INCLUDE_SOURCE) == INCLUDE_SOURCE;
+			return (fFlags & INCLUDE_SOURCE) != 0;
 		}
 		return fSynchronizer.getIncludeSource();
 	}
@@ -593,9 +610,23 @@ public class IUBundleContainer extends AbstractBundleContainer {
 	public boolean getIncludeConfigurePhase() {
 		// if this container has not been associated with a container, return its own value
 		if (fSynchronizer == null) {
-			return (fFlags & INCLUDE_CONFIGURE_PHASE) == INCLUDE_CONFIGURE_PHASE;
+			return (fFlags & INCLUDE_CONFIGURE_PHASE) != 0;
 		}
 		return fSynchronizer.getIncludeConfigurePhase();
+	}
+
+	/**
+	 * Returns whether or not repository references should be followed
+	 *
+	 * @return whether or not repository references should be followed
+	 */
+	public boolean IsFollowRepositoryReferences() {
+		// if this container has not been associated with a container, return
+		// its own value
+		if (fSynchronizer == null) {
+			return (fFlags & FOLLOW_REPOSITORY_REFERENCES) != 0;
+		}
+		return fSynchronizer.isFollowRepositoryReferences();
 	}
 
 	/**
@@ -663,6 +694,7 @@ public class IUBundleContainer extends AbstractBundleContainer {
 		fSynchronizer.setIncludeAllEnvironments((fFlags & INCLUDE_ALL_ENVIRONMENTS) == INCLUDE_ALL_ENVIRONMENTS);
 		fSynchronizer.setIncludeSource((fFlags & INCLUDE_SOURCE) == INCLUDE_SOURCE);
 		fSynchronizer.setIncludeConfigurePhase((fFlags & INCLUDE_CONFIGURE_PHASE) == INCLUDE_CONFIGURE_PHASE);
+		fSynchronizer.setFollowRepositoryReferences((fFlags & FOLLOW_REPOSITORY_REFERENCES) == FOLLOW_REPOSITORY_REFERENCES);
 	}
 
 	@Override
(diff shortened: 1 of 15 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinitionPersistenceHelper.java
```diff
@@ -87,6 +87,7 @@ public class TargetDefinitionPersistenceHelper {
 	static final String ATTR_INCLUDE_ALL_PLATFORMS = "includeAllPlatforms"; //$NON-NLS-1$
 	static final String ATTR_INCLUDE_SOURCE = "includeSource"; //$NON-NLS-1$
 	static final String ATTR_INCLUDE_CONFIGURE_PHASE = "includeConfigurePhase"; //$NON-NLS-1$
+	static final String ATTR_INCLUDE_REFERENCES = "includeReferences"; //$NON-NLS-1$
```
laeubi, line 90: In P2 director application this option is named `followReferences` maybe we should for consitency name this here the same
```suggestion
	static final String ATTR_INCLUDE_REFERENCES = "followReferences"; //$NON-NLS-1$
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -703,6 +714,26 @@ public boolean getIncludeConfigurePhase() {
...
+		fIncludeReferences = value;
+	}
+
+
+	/**
+	 * Return whether or not repository references should be considered
+	 * 
+	 * @return whether or not repository references should be considered
+	 */
+	public boolean getIncludeReferences() {
```
laeubi, line 733:
```suggestion
	public boolean isIncludeReferences() {
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1046,7 +1079,8 @@ private static void addReferences(IMetadataRepository repository, List<IMetadata
 		Collection<IRepositoryReference> references = repository.getReferences();
 		SubMonitor subMonitor = SubMonitor.convert(monitor, references.size() * 2);
 		for (IRepositoryReference reference : references) {
-			if (reference.getType() == IRepository.TYPE_METADATA && seen.add(reference)) {
+			if (reference.getType() == IRepository.TYPE_METADATA && reference.getOptions() == IRepository.ENABLED
```
laeubi, line 1082: Good catch! According to the API docs `IRepositoryReference#getOptions` is a bitfield (even though it only supports two values at the moment).

That said, the (internal) usage of bitfields is quite cumbersome maybe we should add a (default implemented) method `isEnabled()` to the interface that query the bitfield and then can be used in a more semantic way.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1090,7 +1124,7 @@ private void resolveWithPlanner(ITargetDefinition target, IProfile profile, IPro
 		List<IArtifactRepository> extraArtifactRepositories = new ArrayList<>();
 		List<IMetadataRepository> extraMetadataRepositories = new ArrayList<>();
 		addAdditionalProvisionIUs(target, extraArtifactRepositories, extraMetadataRepositories);
-		ProvisioningContext context = new ProvisioningContext(getAgent()) {
+		ProvisioningContext context = new ProvisioningContext(getAgent(), getIncludeReferences()) {
```
laeubi, line 1127: As mentioned on the P2 PR I don't think we need an additional constructor parameter here but can just set the property like below.

laeubi (commented):
@tivervac thank you for the PR that looks like a useful improvement, I added some comments. Also note that a corresponding change would be required in Tycho to support the new attribute.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/IUBundleContainer.java
```diff
@@ -598,6 +615,20 @@ public boolean getIncludeConfigurePhase() {
...
+	/**
+	 * Returns whether or not repository references should be considered
+	 *
+	 * @return whether or not repository references should be considered
+	 */
+	public boolean getIncludeReferences() {
+		// if this container has not been associated with a container, return
+		// its own value
+		if (fSynchronizer == null) {
+			return (fFlags & INCLUDE_REFERENCES) == INCLUDE_REFERENCES;
```
laeubi, line 627: I think this is the usual pattern and avoids using the constant twice.

```suggestion
			return (fFlags & INCLUDE_REFERENCES) != 0;
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinitionPersistenceHelper.java
HannesWell: I would even go one step further and name it `followRepositoryReferences` otherwise the chances are high that it is confused with something like inclusion of transitive dependencies or alike.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/IUBundleContainer.java
```diff
@@ -684,6 +716,7 @@ public String serialize() {
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_ALL_PLATFORMS, Boolean.toString(getIncludeAllEnvironments()));
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_SOURCE, Boolean.toString(getIncludeSource()));
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_CONFIGURE_PHASE, Boolean.toString(getIncludeConfigurePhase()));
+		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_REFERENCES, Boolean.toString(getIncludeReferences()));
```
HannesWell, line 719: Since this option is for a relatively special use-case (I assume most users are happy with the default true) I think the attribute should only be written when its value is not the default (i.e. if it is false).
Otherwise all existing target definitions are changed as soon as they get touched after this change has landed. And since the list of options is already lengthy, I think it should not be made even longer.

HannesWell:
> Using only the enabled references, I still have 5 hits, much less, much faster, but not good enough.

That's a good catch. Disabled references should not be considered.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinitionPersistenceHelper.java
laeubi: I'm not a fan of exaggerated naming schemes. People should not "guess" things and anyone that is uncertain can simply use the UI.

tivervac:
@laeubi The build here seems to fail. Do you know what's going on? Also, what's the difference between `continuous-integration/jenkins/pr-head` and `Jenkins`?

laeubi:
That looks strange an I have never seen this error before in PDE ... but the Github UI suggest that there are recent changes in the master branch, can you try to rebase your changes an current master if it solves the issue?

Beside that "continuous-integration/jenkins/pr-head" is the actual Jenkinsjob and "Jenkins " is the github action integration status check on Github level.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinitionPersistenceHelper.java
HannesWell: > People should not "guess" things

Absolutely, That's why I suggested a more clear name. Of course finding a suitable name is not easy, since it should be as short as possible but also expressive and clear. Yes it's longer, but that does not automatically makes it exaggerated, its always a compromise and one has to find a balance.

> and anyone that is uncertain can simply use the UI.

At least for me the (long-term) goal is actually to require the UI less and since target-definition just like code is read more often than written it should be clear what's intended.

Maybe somebody else also has an opinion?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/IUBundleContainer.java
```diff
@@ -684,6 +716,18 @@ public String serialize() {
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_ALL_PLATFORMS, Boolean.toString(getIncludeAllEnvironments()));
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_SOURCE, Boolean.toString(getIncludeSource()));
 		containerElement.setAttribute(TargetDefinitionPersistenceHelper.ATTR_INCLUDE_CONFIGURE_PHASE, Boolean.toString(getIncludeConfigurePhase()));
+		// Include references was only added quite late, defaults to true and
+		// most users will never edit it.
+		// As such, for stability, conciseness, and readability, we specifically
+		// don't serialize its default state.
+		boolean includeReferences = getIncludeReferences();
+		if (includeReferences) {
```
HannesWell, line 724:
```suggestion
		boolean includeReferences = getIncludeReferences();
		if (includeReferences) { // only serialize the a non-default value
```
The comment can be shorter here. The intend should be clear.

HannesWell:
> @laeubi The build here seems to fail. Do you know what's going on? Also, what's the difference between `continuous-integration/jenkins/pr-head` and `Jenkins`?

Looking at the failing tests it seems like new attribute somehow shows up in existing test-cases altough it should not be persisted if it's value is the default.
I suggest you debug the failing cases in your IDE as JUnit Plug-in test and check what's going on.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinitionPersistenceHelper.java
merks: I prefer the longer name that makes it 100% clear that we are talking about repository references. I’ll bet 99% of the users have no clue what that even is so more descriptive is important.

laeubi: `followReferences` is the parameter name of the director and no one ever was "confused" and I'll bett 99% of the people don't even know what repository reference are either... so for these one needs a good UI with a descriptive help if one really is concerned about these 99% of people because longer names won't help them in any ways. We also don't use the term "reference" anywhere else so there is really not any way to be unclear what is meant here.

merks: We’re all entitled to opinions and personal preferences.  😜 They need not be the same as someone else’s.  I won’t complain about the decision even if it is different from the one I would make.  Don’t worry, be happy.

HannesWell: PDE's and P2's users are not identical also with regard to their experience.

And what's so bad about the extra word?
It's just written if set to false and therefore won't be visible most of the time anyway.

laeubi: This location type is the "P2 Location" it is all about P2, so if it won't be visible most of the time anyway whats wrong with having it consistent with P2 terminology? 😜

laeubi:
@tivervac I think the tests needs adjustments now because they supply a new completion e.g. here:

> There should not be any proposals at index 61. Following proposals found: [includeReferences]

So I think it fair enough to simply make the test assume that new attribute now (what ever it will be named).

laeubi (approved):
Looks good for me!

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -1017,15 +1049,16 @@ static IQueryable<IInstallableUnit> getQueryableMetadata(URI[] repos, IProgressM
 		int repoCount = repos.length;
 		SubMonitor subMonitor = SubMonitor.convert(monitor, repoCount * 2);
 
-		Set<IRepositoryReference> seen = new HashSet<>();
```
HannesWell, line 1020: Maybe I have missed it, but what's the reason for the seen set to be re-created for each repository in a container?
If multiple repos referenced the same third repo it can now end up multiple times in the resulting list, can't it?

tivervac: I was simply trying to clarify that seen is not used in this function at all by inlining it. However, I didn't notice the once vs every iteration difference. I'll have to revert that one line as this changes the behavior

tivervac: See https://github.com/eclipse-pde/eclipse.pde/pull/1257

PR #1256, opened 2024-04-26
Correctly update location attributes in Target Editor

### LINKED ISSUE #1255: Modifying the software site "Included Software" settings in the Target Editor doesn't do anything
The TP serializer is "smart". It doesn't serialize the entire document every time. Instead, it tries to do a minimal update. For these settings, it does so in `TargetDefinition.updateIUContainerElements()`. This code compares the old nodes with the new ones, and if any changes were made to the `repository` or the `Installable Units`, this is reflected. However, it doesn't look at the `location` itself.

See the self-explanatory video below. If you're wondering, saving the file does not solve the issue.

[Screencast from 2024-04-26 11-10-59.webm](https://github.com/eclipse-pde/eclipse.pde/assets/3389524/398e54d8-b97d-43dd-8f6d-e5bf11536bd7)

tivervac:
Fixes #1255


laeubi (approved):
It looks fine for me and a changed URL should of course be respected.

HannesWell:
I'll have a look at this tomorrow.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
```diff
@@ -1321,10 +1319,16 @@ private void updateIUContainerElements(Element containersElement, List<Element>
...
 					TargetDefinitionDocumentTools.addChildWithIndent(containersElement, movedContainer);
+					updatedContainers.remove(container);
 				}
 			}
 		}
 
+		// Use a mock comparator, we only want to use the
+		// "elementAttributesComparator" also used in updateElements
+		TargetDefinitionDocumentTools.updateElements(containersElement, oldContainers, updatedContainers,
+				(Element e1, Element e2) -> 0);
```
HannesWell, line 1330:
```suggestion
				(e1, e2) -> 0);
```

HannesWell (changes requested):
I have tried out your change and it updates the location's attribute. But at the same time it write the entire location element and all its nested children in one line.
I have not found out why this happens, I guess that modifying the XML element does not preserve existing white-space, but that has to be fixed before this can be submitted.

ui/org.eclipse.pde.genericeditor.extension.tests/src/org/eclipse/pde/genericeditor/extension/tests/Bug531602FormattingTests.java
```diff
@@ -13,13 +13,13 @@
  *******************************************************************************/
 package org.eclipse.pde.genericeditor.extension.tests;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
```
tivervac, line 22:
```suggestion
import static org.junit.Assert.assertEquals;

import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.net.URI;
import java.nio.charset.StandardCharsets;
```

tivervac:
> ... it write the entire location element and all its nested children in one line. I have not found out why this happens, I guess that modifying the XML element does not preserve existing white-space, but that has to be fixed before this can be submitted.

@HannesWell I did notice this when creating this patch, see my comment above for more details. As mentioned, if fixing this is _required_ then sadly this ticket will not land at all as I do not plan to fix the existing bug.

HannesWell:
> > ... it write the entire location element and all its nested children in one line. I have not found out why this happens, I guess that modifying the XML element does not preserve existing white-space, but that has to be fixed before this can be submitted.
> 
> @HannesWell I did notice this when creating this patch, see my comment above for more details. As mentioned, if fixing this is _required_ then sadly this ticket will not land at all as I do not plan to fix the existing bug.

Sorry, it looks like it was already a bit late yesterday.
But this is an unfortunate situation. Trading one flaw for the other is not really a satisfying solution.
Some would probably be happy that the UI now works and don't care about the formatting, but at least for me with my usage schema (having a target-file in a SCM, the new situation would be a bit worse: Now I would have to restore the old formatting after I have changed the location through the UI, which would probably take me much time than just editing the attributes in the Source-tab (both would have to be done manually). And I guess there are others that would be even less happy with the new situation.
Therefore I'm not inclined towards the new state. Sorry for that.

Ideally both would be fixed and I would be fine to keep this open until you or somebody else finds the time/motivation to fix the other issues.

laeubi:
I now choose a different approach here:
- https://github.com/eclipse-pde/eclipse.pde/pull/1546

and simply added a new method that synchronize attribute values between two elements. with that I could not notice any negative impact on formatting.

Sadly this shows that some features are "baked" into PDE and other location types can not benefit from such "optimization" and there is always a chance for a surprise.

PR #1257, opened 2024-05-02
Fix regression in P2TargetUtils

tivervac:
Introduced in #1253


HannesWell:
Thank you @tivervac for this follow-up.

PR #1258, opened 2024-05-03
[DS] Use simple class name for single-element property annotations

HannesWell:
Using the full name of the AST element representing a single-element property annotations to derive the property name, makes the generated annotation depend on the way a ComponentPropertyType annotation is applied. For example with fully-qualified name or with the outer-class if it's a inner class.
Instead the simple name of the resolved annotation class should be used as stated in the OSGi compendium spec in '112.8.2.1 Component Property Mapping' [1]:
```
However, if the component property type is a single-element annotation, see 9.7.3 in [7] The Java Language Specification,
Java SE 8 Edition, then the property name for the value method is derived from the name of the component property type
rather than the name of the method.

In this case, the simple name of the component property type, that is, the name of the class without any package name or outer
class name, if the component property type is an inner class, must be converted to the property name as follows:
```

Without this the generated component XML for the following class contains the wrong property names, instead of `event.topics` and `my.comp` the names are `org.osgi.service.event.propertytypes.event.topics` and `my.comp.my.type`.

```
package test.ds.component.props;

@Component(service = EventHandler.class)
@org.osgi.service.event.propertytypes.EventTopics(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE)
@MyComp.MyType(Map.class)
public class MyComp implements EventHandler {

	@ComponentPropertyType
	@Target(ElementType.TYPE)
	public @interface MyType {
		Class<?>[] value();
	}

	@Override
	public void handleEvent(Event event) {
	}
}
```

[1] - https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.component.html#service.component-component.property.mapping

### DIFF
--- a/ds/org.eclipse.pde.ds.annotations/src/org/eclipse/pde/ds/internal/annotations/AnnotationVisitor.java
+++ b/ds/org.eclipse.pde.ds.annotations/src/org/eclipse/pde/ds/internal/annotations/AnnotationVisitor.java
@@ -1328,8 +1328,9 @@ public class AnnotationVisitor extends ASTVisitor {
 
 	private void collectComponentPropertyTypes(IDSDocumentFactory dsFactory,
 			LinkedHashMap<String, IDSProperty> newPropMap, Annotation propertyType) {
-		String fqdn = propertyType.getTypeName().getFullyQualifiedName();
 		ITypeBinding propertyTypeBinding = propertyType.resolveTypeBinding();
+		String simpleName = propertyTypeBinding.getName();
+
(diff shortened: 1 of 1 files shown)


laeubi:
@HannesWell can you make sure we track this as a TCK in the OSGi repository?

HannesWell:
> @HannesWell can you make sure we track this as a TCK in the OSGi repository?

I'll try to prepare a PR in the OSGi repository for such test-case. But since it will take some time any-ways until that is available I'll submit this now already.

PR #1259, opened 2024-05-06
Inline variable and remove JavaDoc in PluginsTab

fedejeanne:
The JavaDoc is outdates and the variable is a null value

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
@@ -217,16 +217,9 @@ public class PluginsTab extends AbstractLauncherTab {
 		return fImage;
 	}
 
-	/**
-	 * Validates the tab.  If the feature option is chosen, and the workspace is not correctly set up,
-	 * the error message is set.
-	 *
-	 * @see org.eclipse.pde.ui.launcher.AbstractLauncherTab#validateTab()
-	 */
 	@Override
 	public void validateTab() {
-		String errorMessage = null;
-		setErrorMessage(errorMessage);
+		setErrorMessage(null);
 	}
 
 	@Override


ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/PluginsTab.java
```diff
@@ -217,16 +217,9 @@ public Image getImage() {
 		return fImage;
 	}
 
-	/**
-	 * Validates the tab.  If the feature option is chosen, and the workspace is not correctly set up,
-	 * the error message is set.
-	 *
-	 * @see org.eclipse.pde.ui.launcher.AbstractLauncherTab#validateTab()
-	 */
```
gireeshpunathil, line 225: instead of removing the doc, could you modify it to reflect what the public API is doing?

fedejeanne: the behavior was changed back in 2010 (https://github.com/eclipse-pde/eclipse.pde/commit/b787978d3202383ca77c1ad4c37a36b5863997f5), so better not.

gireeshpunathil: I agree, so I am not advocating to retain the doc as is. Instead, modify the doc to reflect the new behaviour. (premise: when API is modified, doc is adjusted too to match the new behaviour)

fedejeanne: A JavaDoc here would only be outdated the next time someone decides to change the implementation to actually do something. I think the default one is ok:

image

HannesWell (approved):
I think just removing the javadoc is fine in this case. Thanks for this.

akurtakov:
Please keep an eye on failing builds and don't push things if the build fails. In this case it causes the I-build comparator issue. Tools get better with time but we still need committers to keep an eye for the cases where things are not handled properly.

HannesWell:
Sorry for the trouble, my bad. I was careless assuming this PR is trivial.
Thanks for the fix!

PR #1268, opened 2024-05-15
Only schedule a build if manifest builder is enabled for project

laeubi:
Currently MinimalState schedules a build for every project that has any PDE nature, this can lead to projects scanned that have no ManifestBuilder enabled at all and creating unrelated warnings.

This now checks if the builder is enabled for the project before scheduling a build. Also the check for possible invalid invocations is moved up so it is triggered even if a build.properties exits.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1185


HannesWell (approved):
This looks good to me.
When implementing this I assumed that the builder is actually only executed if it is also configured in the project but that seems wrong.

Thank you for your investigation and fix.

Issue #1265, opened 2024-05-14
[win] NewFeatureProjectTestCase randomly failed


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20240513-1800/testresults/html/org.eclipse.pde.ui.tests_ep432I-unit-win32-java17_win32.win32.x86_64_17.html

```
Failed to delete P/com.junitTest.feature

java.lang.AssertionError: Failed to delete P/com.junitTest.feature
at org.eclipse.pde.ui.tests.PDETestCase.tearDown(PDETestCase.java:112)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

```
java.lang.AssertionError: Testing for existing java nature...
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.pde.ui.tests.wizards.NewFeatureProjectTestCase.testJavaFeature(NewFeatureProjectTestCase.java:129)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

can not reproduce locally, no recent commit to PDE

Project.hasNature(String) is already known to sometimes return wrong result - see https://github.com/eclipse-pde/eclipse.pde/issues/1185#issuecomment-2053715406 , https://github.com/eclipse-platform/eclipse.platform/pull/1312 /

during NewFeatureProjectTestCase org.eclipse.core.internal.resources.ProjectDescription.natures is accessed in multiple threads:

```
Thread [main] (Suspended (modification of field natures in ProjectDescription))	
	owns: CreateFeatureProjectOperation  (id=239)	
	owns: RunnableLock  (id=240)	
	ProjectDescription.<init>() line: 125	
	Project.create(IProjectDescription, int, IProgressMonitor) line: 302	
	Project.create(IProjectDescription, IProgressMonitor) line: 291	
	Project.create(IProgressMonitor) line: 286	
	CoreUtility.createProject(IProject, IPath, IProgressMonitor) line: 96	
	CreateFeatureProjectOperation(AbstractCreateFeatureOperation).createProject(IProgressMonitor) line: 104	
	CreateFeatureProjectOperation(AbstractCreateFeatureOperation).createFeature(IProgressMonitor) line: 84	
	CreateFeatureProjectOperation(AbstractCreateFeatureOperation).execute(IProgressMonitor) line: 72	
[...]

PR #1269, opened 2024-05-15
Retry delete after a little delay / refresh in PDETestCase

laeubi:
Currently a failure to delete a project fails the test randomly on windows. This can be caused for example when things happen too fast and a file is still locked.

This first try the delete and if it fails wait one second, sync the project and if it still exits retry the delete.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1265

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/PDETestCase.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/PDETestCase.java
@@ -13,7 +13,6 @@
  *******************************************************************************/
 package org.eclipse.pde.ui.tests;
 
-import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeTrue;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


laeubi:
@HeikoKlare maybe yes, I just found that some instabilities are often only a timing issue, it would of course be great to have a JUnit5 Rule / Extension / ... that would encapsulate all these parts, for example we often have the demand to wait for Autobuild, disable autobuild, wait for jobs to finish and so on. @akurtakov might know if there is a good place to collect such things. The [osgi-test](https://github.com/osgi/osgi-test) project even goes a little further and even records what is done and undo the actions afterwards, I think something like this for a "Workspace" that has options to import / create projects would be nice so they automatically get cleaned up like for bundles / services in OSGi test.

HeikoKlare:
Great, thank you for that input! I will keep that in mind when proceeding with the JUnit 5 migration in the platform. The existing rules are rather limited in terms of reusability (as being placed in specific test plug-ins), but having reusable (JUnit 5) extensions for workspace-setup/cleanup-related tasks will make sense anyway.

PR #1270, opened 2024-05-15
[PDE-Build] Always use the OSGi profiles provided by the runtime bundle

HannesWell:
The code in PDE(-build) usually only matches respectively is adapted to content of the profiles provided by the org.eclipse.osgi bundle of the same release and not necessarily of previous or even future releases. For example with https://github.com/eclipse-pde/eclipse.pde/pull/1239 PDE-build now relies on the system.packages provided by the JRE-1.1 profile in o.e.osgi since this release. Of course these profiles are missing in previous versions of o.e.osgi. This causes errors during PDE's product export if an earlier version of o.e.osgi is in the target-platform because the system.packages for the JRE-1.1 are then null/empty.


HannesWell:
API-tools failures in Windows/Mac GH workflows seem unrelated.

Issue #995, opened 2023-12-08
Workspace Product Export does not respect 'Eclipse-BundleShape: dir' Manifest-Header anymore


HannesWell:
When exporting a product from the workspace using PDE's product export the `Eclipse-BundleShape: dir` Manifest headers are not considered and plug-ins containing it are therefore not extracted.

To replicate just import the following simple product definition and export the `my.product`:
[my.example.product.zip](https://github.com/eclipse-pde/eclipse.pde/files/13614567/my.example.product.zip)

This has the consequence that the exported product cannot even be launched since for example `org.eclipse.equinox.launcher.*` is not extracted and the launcher cannot find its companion shared library.
Thus this is a sever regression for those that use the product export (like us), probably from
- https://github.com/eclipse-pde/eclipse.pde/pull/770
or its follow-up
 - https://github.com/eclipse-pde/eclipse.pde/pull/886

I can take core of this later (probably at the weekend).

Rainer7000:
Is there any progress with this bug, or is it paused until   https://github.com/eclipse-pde/eclipse.pde/issues/862  is resolved ?
I'm asking because after 2023-09 and 2023-12 this is the third time in a row that we are unable to update to the current version of eclipse because of a unresolved but already known bug. In this case the short term solution could be rather trivial (directory-unpack-flag gets ignored somewhere) and the complete rework of the PDE Export Wizard could lead to a very long development phase where the export-functionality is still broken in the eclipse-distribution.

laeubi:
If this is crucial to your business and likes to speed up the development you can [contact me](https://xn--lubisoft-0za.gmbh/en/#opensource) to enter a contract about this specific issue or if you need dedicated enterprise support.

Of course a PR to fix the problem would also be welcome as it seems you have developers at hand that are familiar with Eclipse RCP, see the [contribution guides here](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md) and if Eclipse is an important Technology for you you maybe even want to join the [Eclipse IDE Workingroup](https://www.eclipse.org/membership/#tab-membership) that also regularly funds development in critical areas or [consider a sponsoring](https://www.eclipse.org/sponsor/ide/) to improve Eclipse IDE in general.

Rainer7000:
Unfortunately we are a rather small company and I'm the only Java-Developer, so our resources are very limited.
Although we decided to dedicate 100% of our java-developer-resources to the issue for quite sime time ;)
[...]

PR #1271, opened 2024-05-15
[PDE-Build] Copy Eclipse-BundleShape and other headers in wrapped states

HannesWell:
For wrapped OSGi States obtained from pde.core/ui the Manifest 'Eclipse-BundleShape' entries, besides others, were not copied into the bundle's user-object Properties. This had the consequence that in ShapeAdvisor.getUnpackClause() the value of that entry could never be considered.
https://github.com/eclipse-pde/eclipse.pde/blob/c2d4b0c101bbcb0697edd9b8edf2a1aeb8e71b16/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/ShapeAdvisor.java#L110-L113

The method only used to return the expected boolean value because of corresponding unpack-attributes in 'plugin' entries of feature.xml files. But since the removal of that otherwise unused attribute in https://github.com/eclipse-pde/eclipse.pde/pull/770 this makeshift was gone and ShapeAdvisor.getUnpackClause() always return false leading to all bundles in an exported product being in jar-shape, even if the 'Eclipse-BundleShape' entry in the MANIFEST.MF told something different.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/995

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java
@@ -104,6 +104,10 @@ public class PDEState implements IPDEBuildConstants, IBuildPropertiesConstants {
 		addedBundle = new ArrayList<>();
 		unqualifiedBundles = new ArrayList<>();
 		//forceQualifiers();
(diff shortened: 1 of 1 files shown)


HannesWell:
API-tools failures in Windows GH workflows seem unrelated.

PR #1274, opened 2024-05-19
Enable most LocalTargetDefinitionTests in verification builds

HannesWell:
And fix LocalTargetDefinitionTests.testDeleteCleansCaches failing in I-builds.

Follow-up on https://github.com/eclipse-pde/eclipse.pde/pull/1247#issuecomment-2119162437.


HannesWell:
For now I have disabled all cases in LocalTargetDefinitionTests that fail in verification builds and only fixed LocalTargetDefinitionTests.testDeleteCleansCaches, which fails in I-builds https://github.com/eclipse-pde/eclipse.pde/pull/1247#issuecomment-2119160053.
The trick is to add some content to the target, if it is empty a target is not cached upon resolution.

Issue #1278, opened 2024-05-30
Wrong property fired in Product#setIncludeJre


Phillipus:
The property `P_INCLUDE_LAUNCHERS` is fired here:

https://github.com/eclipse-pde/eclipse.pde/blob/96484560b85cf472802badf93bd4555a3e881641/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/product/Product.java#L899-L906

It should be `P_INCLUDE_JRE`

Phillipus:
PR https://github.com/eclipse-pde/eclipse.pde/pull/1279

PR #1279, opened 2024-05-30
Fire the correct property in Product#setIncludeJre

Phillipus:
The property should be P_INCLUDE_JRE not P_INCLUDE_LAUNCHERS

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1278


laeubi (approved):

HannesWell (approved):
Thank you for this fix.

Approve as PL/PMC.

HannesWell:
Actually we already have RC2 promotion today with the I-build from yesterday. So unless there is a RC2a for the SDK, this will fix will not be in Eclipse 2024-06.
But I think that's not a blocker that alone justifies a RC2a.

merks:
I'm not sure we should be committing changes before master is open again....

akurtakov:
@HannesWell For the future please don't push changes at that state of the cycle.

PR #1282, opened 2024-06-02
Shortcut to Target Platform preferences in editor

### LINKED ISSUE #1281: Unload target definition / Fallback to "Running Platform" or preference shortcut
I find myself working with different target platform definitions quite often and noticed some possible improvements.

When working inside the target definition editor it is nice to have the button **Set as Active Target Platform** / **Reload Target Platform** but I also need to *unload* the  target platform quite often. When switching to another target platform I just navigate to to it and open it and then click on **Set as Active Target Platform** but when I want to go back to the clean state of the *Running Platform* target definition I have to manually open the [preferences](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Fguide%2Ftools%2Fpreference_pages%2Ftarget_platform.htm) and activate it there.

It would be great if the target definition editor would feature a **Unload Target Platform** button for when this very target definition is loaded.
The other option which would maybe fit even more is to have a shortcut inside the target definition editor as well as the **Target Platform State** view that opens the preferences right at the *Plug-in Development -> Target Platform* page.
This would also be beneficial for switching between target definitions as following the shortcut would be simpler than navigating to the file, opening it and then setting it as active target platform.

thahnen:
This helps when a user wants to switch between target definitions without the need of either navigating to another target definition file or the preference page listing all Target Platforms.

Resolves #1281

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
@@ -70,6 +70,7 @@ import org.eclipse.pde.internal.ui.IHelpContextIds;
 import org.eclipse.pde.internal.ui.PDEPlugin;
 import org.eclipse.pde.internal.ui.PDEPluginImages;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
+import org.eclipse.pde.internal.ui.preferences.TargetPlatformPreferencePage;
 import org.eclipse.pde.internal.ui.shared.target.ITargetChangedListener;
 import org.eclipse.pde.internal.ui.shared.target.TargetContentsGroup;
 import org.eclipse.pde.internal.ui.shared.target.TargetLocationsGroup;
@@ -87,6 +88,7 @@ import org.eclipse.ui.IURIEditorInput;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.dialogs.PreferencesUtil;
 import org.eclipse.ui.dialogs.SaveAsDialog;
 import org.eclipse.ui.forms.AbstractFormPart;
 import org.eclipse.ui.forms.HyperlinkGroup;
@@ -406,8 +408,19 @@ public class TargetEditor extends FormEditor {
 		export.setToolTipText("Export"); //$NON-NLS-1$
 		export.setImageDescriptor(PDEPluginImages.DESC_EXPORT_TARGET_TOOL);
 
+		Action openTPPreference = new Action("openTPPreference") { //$NON-NLS-1$
+			@Override
+			public void run() {
+				PreferencesUtil.createPreferenceDialogOn(getSite().getShell(), TargetPlatformPreferencePage.PAGE_ID,
+						null, null).open();
+			}
+		};
+		openTPPreference.setToolTipText(PDEUIMessages.AbstractTargetPage_openPreferences);
+		openTPPreference.setImageDescriptor(PDEPluginImages.DESC_SETTINGS_OBJ);
+
 		form.getToolBarManager().add(setAsTarget);
 		form.getToolBarManager().add(export);
+		form.getToolBarManager().add(openTPPreference);
 		form.getToolBarManager().add(help);
 		form.updateToolBar();
 
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
@@ -488,6 +488,7 @@ SchemaAttributeDetails_use=Use:
 AbstractPluginBlock_counter={0} out of {1} selected
 AbstractTargetPage_setTarget=Set as Active Target Platform
 AbstractTargetPage_reloadTarget=Reload Target Platform
+AbstractTargetPage_openPreferences=Open the Target Platform preferences
 ###### Launchers #######################################
 MainTab_name = &Main
 MainTab_AttributeLabel_LauncherPDEVersion=PDE version
--- a/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.ui; singleton:=true
-Bundle-Version: 3.15.200.qualifier
+Bundle-Version: 3.15.300.qualifier
 Bundle-Activator: org.eclipse.pde.internal.ui.PDEPlugin
 Bundle-Vendor: %provider-name
 Bundle-Localization: plugin
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -39,6 +39,8 @@ public class PDEUIMessages extends NLS {
 
 	public static String AbstractTargetPage_reloadTarget;
 
+	public static String AbstractTargetPage_openPreferences;
+
 	public static String AddActivationHeaderResolution_label;
 
 	public static String AddAutomaticModuleResolution_desc;


thahnen:
This failing build doesn't seem related to my changes if I've read it correctly. Maybe it is due to the current failing master?

merks:
Yes, this doesn't look related to your changes but rather looks like existing problems. 😞

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -488,6 +488,7 @@ SchemaAttributeDetails_use=Use:
 AbstractPluginBlock_counter={0} out of {1} selected
 AbstractTargetPage_setTarget=Set as Active Target Platform
 AbstractTargetPage_reloadTarget=Reload Target Platform
+AbstractTargetPage_showAll=Show all Target Platforms
```
HannesWell, line 491: Although it is true that the opened preference page shows all target-definitions, what the action is actually doing is opening the `Target Platform` preference page.
Therefore I think the tool-tip should be accordingly:
```suggestion
AbstractTargetPage_showAll=Open 'Target Platform' preference
```
If you want to also hint that it shows all targets, you could append something like `(lists all available targets)`

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
```diff
@@ -406,8 +408,19 @@ public void run() {
 		export.setToolTipText("Export"); //$NON-NLS-1$
 		export.setImageDescriptor(PDEPluginImages.DESC_EXPORT_TARGET_TOOL);
 
+		Action showAll = new Action("showAll") { //$NON-NLS-1$
```
HannesWell, line 411:
```suggestion
		Action openTPPreference = new Action("openTPPreference") { //$NON-NLS-1$
```

HannesWell (changes requested):
Thank you @thahnen for this PR.

Adding a toolbar-button to open the Target Platform preference page looks like a convenient feature. Since this Editor is already not that simple, we should be careful when adding new capabilities. But the proposed addition adds a quite general feature in a not too prominent location so I think it is acceptable.

But as explained below I think the tool-tip and correspondingly the ids/variables should say that the button opens the `Target Platform` preference instead of `show all targets`.

Another point is distinction between a Action/Button and a link:
An action usually changes the model (in this case the target) or does something with it, for example export it. A link on the other hand, as the name somehow implies links to another part of the IDE respectively leads the user to it.
So far the theory, but `Set as Active Target Platform` is currently a link (probably to make it easier to read), while it should actually be a button and `Help` is a button while it should be a link.
So we already have a mixture in the editor and since making this a link would lead to another text in the tool-bar, I'm fine to keep this a button/action.

Could you please also add a screenshot of the changed Editor UI to this PR, in order to allow others better to understand the change without fetching this change?

thahnen:
Hi @HannesWell,

thanks a lot for the review. I'm going to tackle the proposed changes any minute now. For now, here is a screenshot of what the Target Definition Editor will look like. I've re-used the icon as it seems fit: It displays two options and only one is selected. I've looked through all other icons but didn't find something even more fitting and as this is the settings / preference icon it made the most sense to me. I'm also super bad at creating icons, otherwise I might have come up with a *new* one.

I've placed it after the *Export* action as it made the most sense to me. The **Set as Active Target Platform** / **Reload Target Platform** ~action~ link as well as the **Export** action both apply to this very opened target definition, the new **Open 'Target Platform' preferences** action applies to target definitions in general while the **Help** action is even more general.

eclipse pde-I1281-1

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -488,6 +488,7 @@ SchemaAttributeDetails_use=Use:
 AbstractPluginBlock_counter={0} out of {1} selected
 AbstractTargetPage_setTarget=Set as Active Target Platform
 AbstractTargetPage_reloadTarget=Reload Target Platform
+AbstractTargetPage_openPreferences=Open 'Target Platform' preference (lists all available target)
```
merks, line 491: I have a feeling that ' needs to be specified as ''.  The wording seems awkward.  No doubt @HannesWell will have a constructive suggestion.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
```diff
@@ -406,8 +408,19 @@ public void run() {
 		export.setToolTipText("Export"); //$NON-NLS-1$
 		export.setImageDescriptor(PDEPluginImages.DESC_EXPORT_TARGET_TOOL);
 
+		Action openTPPreference = new Action("openTPPreference") { //$NON-NLS-1$
+			@Override
+			public void run() {
+				String pageId = TargetPlatformPreferencePage.PAGE_ID;
```
merks, line 414: It's a nit, but this assignment seems pointless for a a constant used only once.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
laeubi, line 491:
```suggestion
AbstractTargetPage_openPreferences=Show all Target Platforms known to the IDE
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
thahnen: You're right, copy-pasted that but having that assignment doesn't make sense.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
thahnen: You're right, I'll change that to what @laeubi suggested.

thahnen: Accepted

HannesWell:
Thank you for the update.

> I've looked through all other icons but didn't find something even more fitting and as this is the settings / preference icon it made the most sense to me. I'm also super bad at creating icons, otherwise I might have come up with a _new_ one.
> 
> I've placed it after the _Export_ action as it made the most sense to me. The **Set as Active Target Platform** / **Reload Target Platform** ~action~ link as well as the **Export** action both apply to this very opened target definition, the new **Open 'Target Platform' preferences** action applies to target definitions in general while the **Help** action is even more general.

That's all perfectly fine. I think it is often even better to reuse existing icons so that it is easier to recognize the buttons general purpose.

This PR looks quite good to me, but since we now had different suggestions for the tool-tip text we should first get consensus to avoid changing it back and forth (finding good names/text is often the hardest part).
As explained in https://github.com/eclipse-pde/eclipse.pde/pull/1282#discussion_r1623414366 I think the tool-tip should say what the button is actually doing, which is opening the `Target Platform` preference page. It can often be helpful to mention that this also provides a list of available targets, but since one gets more than just a list, the current text does not look ideal to me.
Of course my initial suggestion can be enhanced. Maybe I have a better suggestion tomorrow and of course other suggestions are welcome.

We are currently in a quite-period after the development cycle of the 2024-06 release anyway and have to wait until the master is prepared and opened for 24-09.

> I have a feeling that ' needs to be specified as ''. The wording seems awkward. No doubt @HannesWell will have a constructive suggestion.

@merks I tested it and `'` works as expected. And checking the doc of the NLS class, single-quotes are only special in that sense that between them the variable substitution does not happen. So `NLS.bind("Some '{0}' text", "sepcial")` does not become `Some 'special' text` but just `Some '{0}' text`. To get the former one has to use double single-quites: `NLS.bind("Some ''{0}'' text", "sepcial")`.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -488,6 +488,7 @@ SchemaAttributeDetails_use=Use:
 AbstractPluginBlock_counter={0} out of {1} selected
 AbstractTargetPage_setTarget=Set as Active Target Platform
 AbstractTargetPage_reloadTarget=Reload Target Platform
+AbstractTargetPage_openPreferences=Show all Target Platforms known to the IDE
```
HannesWell, line 491: @merks, @laeubi what do you think about
```suggestion
Open 'Target Platform' preference page to manage all targets known to the IDE.
```
or just `Open 'Target Platform' preference page to manage all known targets.` or a bit more _technical_ `Open 'Target Platform' preference page to manage all available targets.`?

I think this should cover both aspects, mentioning that the button opens the preference page and that this page lists all known targets.

laeubi: I think we should not use `'` in the text there is no need for it.

"all available targets" don't sound right for me, because there could be numerous targets that we can't show here e.g. I don't think targets are listed if one D&D a file into the IDE (but I haven't specifically tested that).

merks: I would try to make the sentence grammatically complete with the word 'the' added:
```
Open the 'Target Platform' preference page to manage all known targets
```
Note that there is precedent for including the quotes

image

laeubi: > Note that there is precedent for including the quotes

At this place it even looks more strange (to me) because they are already highlighted by a link, but at least one is already at the preferences and there is a direct connection to `'thing'` we don't have here, so maybe even don't focus at the "page" aspect at all and simply name it 

`Open the target platform preference`

thahnen: I'm fine with the latter one as it is basically saying what is done by the action. The implication that the preference contains all known target definitions is given once the preference page is shown.

HannesWell: The intention of the single-quotes was to emphazie that `Target Platform` is the title of the preference. If the tool-tip would support formatting like here in GitHub I would suggest a text like:
Open the `Target Platform` preference page...

So the single-quotes were the fall-back. But I'm also fine with removing them, I would just write Target Platform in capitalized like it is done in the page title.
Regarding if it is all known/available/... targets: That page definitively lists all target files in any open project, I don't know if it shows targets dropped to the IDE, but also don't know if that's even possible. We could just remove that part or just be vague and say `all targets`.

Based on that my suggestions are either
`Open the Target Platform preference`
or
`Open the Target Platform preference to manage all targets.`

merks: Given it's Window -> **Preferences**, and the title of the dialog is **Preferences** is think the word preference should be plural or probably better would be preference page like here:

image

laeubi: I would prefer not to include "page"  or any "manage XYZ" because maybe in the future we want to show multiple pages and "PreferencePage" is really a technical term, to open the Preferences we also have "Window > Preferences" and not "Window > Workspace Preferences Dialog"

That's why I think even `Open preferences` is enough as it is clear from the context...

merks: I did show reference to existing terminology that words it exactly this way.  There is much to be said for consistency...

I would wonder if Open preference would open the preferences for the editor.

So maybe is okay and still short enough?
```
Open the Target Platform preferences
```

Let's hope we don't discuss colors next, then we'll never find agreement.  😝

laeubi: > I did show reference to existing terminology that words it exactly this way. There is much to be said for consistency...

We can also adjust the references, there is so much things in platform that are far from optimal, if we always try to be "consistent" we would never be able to improve anything. e.g there are a lot of Preference pages that do not react well to size changes, should we (for consistency) make all pages react poorly to size changes?

Beside that, as mentioned this is *inside* the preferences dialog, while here we are inside an Editor and the other Icons simply read `Export` (not `Open Export Dialog (Wizard)`) and "`Help` (not `Open Eclipse Help (Dialog)`) so if we want to be really consistent the new icon should simply read `Preferences` ...

merks: Export is a verb.   👅   Preferences... is appropriate for a menu, not for a tool tip...

Yes, consistently bad is not good.  But I don't think what I showed is bad; it's proper well-worded English.

I suggest doing a search `preference page` in `*.properties` and finding all consistently similar examples:

image

If we determine they are all consistently bad, we should consistently fix them all, rather than do something different.  I don't think being inconsistent is ever good.

In any case, I have said **more than enough**.  Hannes, as team lead can decide whatever he deems best and I have provided sufficient input to the process.  👼

HannesWell: > I like the simplistic approach of `Open the Target Platform preferences` the best.

Agree. Lets continue with that. I think that's a good compromise.

HannesWell:
@thahnen please adjust the tool-tip to the agreed text. This I would be fine with this PR and we can submit this.

When you push an update, please make sure you rebased on top of the current master.

The latest build will probably fail due to missing version bumps in the manifest of `org.eclipse.pde.ui`.
If that's the case, please increase the service version segment by +100 in
https://github.com/eclipse-pde/eclipse.pde/blob/219444193e62086bd46148caa507a4b07c626071/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF#L5

to `Bundle-Version: 3.15.300.qualifier`.

thahnen:
Hey @HannesWell, you already rebased the current master changes on my fork, so working with that I adjusted based on the agreed text.

HannesWell (approved):
Thank you. The change looks good.

> The latest build will probably fail due to missing version bumps in the manifest of `org.eclipse.pde.ui`. If that's the case, please increase the service version segment by +100 in
> 
> https://github.com/eclipse-pde/eclipse.pde/blob/219444193e62086bd46148caa507a4b07c626071/ui/org.eclipse.pde.ui/META-INF/MANIFEST.MF#L5
> 
> to `Bundle-Version: 3.15.300.qualifier`.

But to fix the build failure you have to apply the change mentioned above.
We usually do this in a separate commit with message `Version bump for 4.33 stream`, ideally before the actual change so that each commit is 'buildable'.

thahnen:
@HannesWell my bad, updated the bundle version.

HannesWell:
> @HannesWell my bad, updated the bundle version.

Thank you.
Could you please change the message from that commit where you update the version to `Version bump for 4.33 stream` and rebase your change on the current master and force push to your branch so that the merge commit vanishes now already?

merks:
@thahnen 

Thank you for your perseverance.   🏅

thahnen:
Ah damn missed the merge commit (thanks for nothing GH webpage), sorry for that thought I'd selected rebasing. Interactive rebasing is my nemesis, I've resetted to the non-fork master and applied both commits via patches 😮‍💨

HannesWell (approved):
> Thank you for your perseverance. 🏅

I can only second that, discussing names always takes longer than one actually likes.

Thanks again for this PR, the change now looks perfectly fine.

The current build failure seem unrelated, I'm investigating its cause, but probably nothing you can do about.

thahnen:
Alright, thanks for all the input and making this possible. [Heading off now](https://imgflip.com/i/8t4vxc)!

HannesWell:
> Alright, thanks for all the input and making this possible. [Heading off now](https://i.imgflip.com/8t4vxc.jpg)!

That's indeed a relive. Until we meet again :D

The build failure is somewhat strange and says that two different versions of org.eclipse.help.source.feature seem to be required?
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:4.0.8:test (pde-build-tests) on project org.eclipse.pde.build.tests:
Installation of IUs [org.eclipse.pde.build.tests, org.eclipse.tycho.surefire.osgibooter, org.eclipse.tycho.surefire.junit4, org.eclipse.pde.feature.group, org.eclipse.sdk.test.ide] failed:
Call to p2 director application failed:
Cannot complete the install because of a conflicting dependency.:
[Software being installed: Eclipse SDK 4.33.0.v20240604-1451 (org.eclipse.sdk.test.ide 4.33.0.v20240604-1451);
Only one of the following can be installed at once: :
[Eclipse Help System Developer Resources 2.3.1900.v20240606-1800 (org.eclipse.help.source.feature.jar 2.3.1900.v20240606-1800);
Eclipse Help System Developer Resources 2.3.1900.v20240606-1620 (org.eclipse.help.source.feature.jar 2.3.1900.v20240606-1620)];
Cannot satisfy dependency::
[From: Eclipse Help System Developer Resources 2.3.1900.v20240606-1620 (org.eclipse.help.source.feature.group 2.3.1900.v20240606-1620);
To: org.eclipse.equinox.p2.iu; org.eclipse.help.source.feature.jar [2.3.1900.v20240606-1620,2.3.1900.v20240606-1620], filter=(org.eclipse.update.install.features=true)];
Cannot satisfy dependency::
[From: Eclipse Help System Developer Resources 2.3.1900.v20240606-1800 (org.eclipse.help.source.feature.group 2.3.1900.v20240606-1800);
To: org.eclipse.equinox.p2.iu; org.eclipse.help.source.feature.jar [2.3.1900.v20240606-1800,2.3.1900.v20240606-1800], filter=(org.eclipse.update.install.features=true)];
Cannot satisfy dependency::
[From: Eclipse Platform Developer Resources 4.33.0.v20240606-1942 (org.eclipse.platform.source.feature.group 4.33.0.v20240606-1942);
To: org.eclipse.equinox.p2.iu; org.eclipse.help.source.feature.group [2.3.1900.v20240606-1620,2.3.1900.v20240606-1620]];
Cannot satisfy dependency::
[From: Eclipse SDK 4.33.0.v20240604-1451 (org.eclipse.sdk.test.ide 4.33.0.v20240604-1451);
To: org.eclipse.equinox.p2.iu; org.eclipse.help.source.feature.group [2.3.1900.v20240606-1800,2.3.1900.v20240606-1800]];
Cannot satisfy dependency::
[From: Eclipse SDK 4.33.0.v20240604-1451 (org.eclipse.sdk.test.ide 4.33.0.v20240604-1451); To: org.eclipse.equinox.p2.iu; org.eclipse.platform.source.feature.group [4.33.0.v20240606-1942,4.33.0.v20240606-1942]]].
```

Maybe this a consequence of the I-builds that yesterday run relatively short one after the another?
I'll await tonight's I-build and run this PRs build again.

HannesWell:
> Maybe this a consequence of the I-builds that yesterday run relatively short one after the another?

Great, after restarting the builds succeeded after waiting another night. 
Submitting.

Issue #1178, opened 2024-03-01
Add a new "Upgrade" button to EE


laeubi:
Currently PDE only support adding and removing EEs:

![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/1331477/5c829f5e-117c-4cb8-bae9-8039275d228d)

but a more common thing is to *upgrade* to a new EE, currently this requires:

1. remove the old EE
2. add the new EE
3. press the "Update Classpath settings"

There should be a new "Upgrade" button that allows to select the new EE, removes all old ones and updates the classpath in one go.

As part of this, this dialog should then be sorted by EE in descending order (highest EE first):
![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/1331477/edfe3ec0-5503-474e-a598-78002271b632)

akurtakov:
Even simpler from end user POV could be double click on the EE and be able to edit directly.

laeubi:
Yes but "discover" this might be problematic, so there should always be a "visible way" (aka button/link) to do this as well

lathapatil:
> As part of this, this dialog should then be sorted by EE in descending order (highest EE first):

@laeubi Should the contents of the dialog be sorted as below ?

[JavaSE-21, JavaSE-20, JavaSE-19, JavaSE-18, JavaSE-17, JavaSE-16, JavaSE-15, JavaSE-14, JavaSE-13, JavaSE-12, JavaSE-11, JavaSE-10, JavaSE-9, JavaSE-1.8, JavaSE-1.7, JavaSE-1.6, J2SE-1.5, J2SE-1.4, J2SE-1.3-NO-SOUND, J2SE-1.3, OSGi/Minimum-1.2, J2SE-1.2, OSGi/Minimum-1.1, JRE-1.1, CDC-1.1/Foundation-1.1, OSGi/Minimum-1.0, CDC-1.0/Foundation-1.0]

laeubi:
@lathapatil yes that's correct, so "highest" java should be on top.

lathapatil:
@laeubi 

There is a confusion in implementing this in the specified order (Highest Java to lowest). Which Approach is preferred ?

Problem :  Even if we sort the list in the desired order and set the elements to dialog, the parent class (FilteredList) is sorting it back to its original form.

Approach 1: 
[...]

PR #1285, opened 2024-06-03
Add a new "Modify" button to Execution Environment section

lathapatil:
Fixes : #1178

A new "Modify" button that allows to select the new EE, removes all old ones and updates the classpath in one go.

As part of this, this dialog should then be sorted by EE in descending order (highest EE first):

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
@@ -12,6 +12,7 @@
  *     IBM Corporation - initial API and implementation
  *     Lars Vogel <Lars.Vogel@vogella.com> - Bug 487988
  *     Martin Karpisek <martin.karpisek@gmail.com> - Bug 351356
+ *     Latha Patil (ETAS GmbH) - Issue 1178 Add 'Change' button to change a Plugin's required EE and classpath
  *******************************************************************************/
 package org.eclipse.pde.internal.ui.editor.plugin;
 
@@ -50,6 +51,7 @@ import org.eclipse.pde.internal.core.ibundle.IManifestHeader;
 import org.eclipse.pde.internal.core.natures.PluginProject;
 import org.eclipse.pde.internal.core.text.bundle.ExecutionEnvironment;
 import org.eclipse.pde.internal.core.text.bundle.RequiredExecutionEnvironmentHeader;
+import org.eclipse.pde.internal.core.util.VMUtil;
 import org.eclipse.pde.internal.ui.IHelpContextIds;
 import org.eclipse.pde.internal.ui.PDEPlugin;
 import org.eclipse.pde.internal.ui.PDEPluginImages;
@@ -71,6 +73,7 @@ import org.eclipse.swt.widgets.Table;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.actions.ActionFactory;
 import org.eclipse.ui.dialogs.ElementListSelectionDialog;
+import org.eclipse.ui.dialogs.FilteredList;
 import org.eclipse.ui.forms.events.IHyperlinkListener;
 import org.eclipse.ui.forms.widgets.FormToolkit;
 import org.eclipse.ui.forms.widgets.Hyperlink;
@@ -114,6 +117,7 @@ public class ExecutionEnvironmentSection extends TableSection {
 
 	public ExecutionEnvironmentSection(PDEFormPage page, Composite parent) {
 		super(page, parent, Section.DESCRIPTION, new String[] { //
+				PDEUIMessages.RequiredExecutionEnvironmentSection_modify,
 				PDEUIMessages.RequiredExecutionEnvironmentSection_add,
 				PDEUIMessages.RequiredExecutionEnvironmentSection_remove });
 		createClient(getSection(), page.getEditor().getToolkit());
@@ -160,21 +164,11 @@ public class ExecutionEnvironmentSection extends TableSection {
 		gd.horizontalSpan = 2;
 		link.setLayoutData(gd);
 
-		final IProject project = getPage().getPDEEditor().getCommonProject();
+		IProject project = getPage().getPDEEditor().getCommonProject();
 		if (project != null && PluginProject.isJavaProject(project)) {
 			link = toolkit.createHyperlink(container, PDEUIMessages.ExecutionEnvironmentSection_updateClasspath, SWT.NONE);
 			link.addHyperlinkListener(IHyperlinkListener.linkActivatedAdapter(e -> {
-				try {
-					getPage().getEditor().doSave(null);
-					IPluginModelBase model = PluginRegistry.findModel(project);
-					if (model != null) {
-						ClasspathComputer.setClasspath(project, model);
-						if (PDEPlugin.getWorkspace().isAutoBuilding()) {
-							doFullBuild(project);
-						}
-					}
-				} catch (CoreException e1) {
-				}
+				updateClasspathSettings(project);
 			}));
 			gd = new GridData();
 			gd.horizontalSpan = 2;
@@ -192,6 +186,19 @@ public class ExecutionEnvironmentSection extends TableSection {
 		section.setClient(container);
 	}
 
+	private void updateClasspathSettings(IProject project) {
+		try {
+			getPage().getEditor().doSave(null);
+			IPluginModelBase model = PluginRegistry.findModel(project);
+			if (model != null) {
+				ClasspathComputer.setClasspath(project, model);
+				if (PDEPlugin.getWorkspace().isAutoBuilding()) {
+					doFullBuild(project);
+				}
+			}
+		} catch (CoreException e1) {
+		}
+	}
 	@Override
 	public void dispose() {
 		IBundleModel model = getBundleModel();
@@ -208,8 +215,9 @@ public class ExecutionEnvironmentSection extends TableSection {
 	@Override
 	protected void buttonSelected(int index) {
 		switch (index) {
-			case 0 -> handleAdd();
-			case 1 -> handleRemove();
+			case 0 -> handleModify();
+			case 1 -> handleAdd();
+			case 2 -> handleRemove();
 		}
 	}
 
@@ -244,8 +252,9 @@ public class ExecutionEnvironmentSection extends TableSection {
 	private void updateButtons() {
 		Table table = fEETable.getTable();
 		TablePart tablePart = getTablePart();
-		tablePart.setButtonEnabled(0, isEditable());
-		tablePart.setButtonEnabled(1, isEditable() && table.getSelection().length > 0);
+		tablePart.setButtonEnabled(0, isEditable() && table.getItemCount() == 1);
+		tablePart.setButtonEnabled(1, isEditable());
+		tablePart.setButtonEnabled(2, isEditable() && table.getSelection().length > 0);
 	}
 
 	private void handleRemove() {
@@ -260,17 +269,53 @@ public class ExecutionEnvironmentSection extends TableSection {
 	}
 
 	private void handleAdd() {
-		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(), new EELabelProvider());
+		Object[] selection = showEESelectionDialog(true);
+		if (selection != null) {
+			addExecutionEnvironments(selection);
+		}
+	}
+
+	private void handleModify() {
+		RequiredExecutionEnvironmentHeader header = getHeader();
+		Object[] selection = showEESelectionDialog(false);
+		if (selection != null) {
+
+			List<String> existingEEs = header.getEnvironments();
+			// Remove existing EE (Upgrade is allowed only if a single EE is
+			// present in the header)
+			getHeader().removeExecutionEnvironment(existingEEs.get(0));
+
+			// add selected Java EE to the header
+			addExecutionEnvironments(selection);
+
+			IProject project = getPage().getPDEEditor().getCommonProject();
+			updateClasspathSettings(project);
+		}
+
+	}
+
+	private Object[] showEESelectionDialog(boolean allowMultiSelection) {
+		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(),
+				new EELabelProvider()) {
+			@Override
+			protected FilteredList createFilteredList(Composite parent) {
+				FilteredList filteredList = super.createFilteredList(parent);
+				filteredList.setComparator(VMUtil.ASCENDING_EE_JAVA_VERSION.reversed());
+				return filteredList;
+			}
+		};
 		dialog.setElements(getEnvironments());
 		dialog.setAllowDuplicates(false);
-		dialog.setMultipleSelection(true);
+		dialog.setMultipleSelection(allowMultiSelection);
 		dialog.setTitle(PDEUIMessages.RequiredExecutionEnvironmentSection_dialog_title);
 		dialog.setMessage(PDEUIMessages.RequiredExecutionEnvironmentSection_dialogMessage);
 		dialog.create();
-		PlatformUI.getWorkbench().getHelpSystem().setHelp(dialog.getShell(), IHelpContextIds.EXECUTION_ENVIRONMENT_SELECTION);
+		PlatformUI.getWorkbench().getHelpSystem().setHelp(dialog.getShell(),
+				IHelpContextIds.EXECUTION_ENVIRONMENT_SELECTION);
 		if (dialog.open() == Window.OK) {
-			addExecutionEnvironments(dialog.getResult());
+			return dialog.getResult();
 		}
+		return null;
 	}
 
 	@SuppressWarnings("deprecation")
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
@@ -2283,6 +2283,7 @@ GetNonExternalizedStringsAction_allExternalizedMessage=All strings in manifest f
 RequiredExecutionEnvironmentSection_title=Execution Environments
 RequiredExecutionEnvironmentSection_add=Add...
 RequiredExecutionEnvironmentSection_remove=Remove
+RequiredExecutionEnvironmentSection_modify=Modify...
 RequiredExecutionEnvironmentSection_fragmentDesc=Specify the minimum execution environments required to run this fragment.
 RequiredExecutionEnvironmentSection_pluginDesc=Specify the minimum execution environments required to run this plug-in.
 RequiredExecutionEnvironmentSection_dialog_title=Execution Environments
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -2901,6 +2901,8 @@ public class PDEUIMessages extends NLS {
 
 	public static String EditorUtilities_missingIcoNote;
 
+	public static String RequiredExecutionEnvironmentSection_modify;
+
 	public static String RequiredExecutionEnvironmentSection_add;
 
 	public static String RequiredExecutionEnvironmentSection_remove;


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -324,6 +350,28 @@ private void handleAdd() {
...
+			ExecutionEnvironment[] addedExecutionEnvironments = header.getEnvironments();
+			// Remove existing EE ( Upgrade is allowed only when single Java EE
+			// is present in the header)
+			getHeader().removeExecutionEnvironment(addedExecutionEnvironments[0]);
+
+			SortedMap<String, IExecutionEnvironment> executionEnvironmentMap = new TreeMap<>(new VersionComparator());
+			Arrays.asList(executionEnvironments).forEach(executionEnvironment -> executionEnvironmentMap
+					.put(executionEnvironment.getId(), executionEnvironment));
+			// add highest Java EE to the header
+			header.addExecutionEnvironment(executionEnvironmentMap.get(executionEnvironmentMap.firstKey()));
```
HannesWell, line 368: I don't think it isn't necessary to sort all EEs since IIRC the list of EEs `JavaRuntime.getExecutionEnvironmentsManager().getExecutionEnvironments()` is already sorted in ascending order and you can therefore just use the last element.

And in case it would be sorted, I would use
`Comparator.comparing(ee -> new org.osgi.framework.Version(getVersionString(ee)))`
to create a comparator without the need to implement a version comparison logic again.

HannesWell, line 368: The current implementation of the `Upgrade` button just upgrades to the latest version supported by PDE/JDT.
While I really like the upgrade in one step (compared to add with subsequent remove and CP update), I think we should present a dialog to select the upgrade target, similar to the dialog that is presented when one clicks add.

At the moment clicking Upgrade moves a bundle to `JavaSE-23` that will be released in September. I assume that's not what most users will target in the meantime. And in general one does not always want to upgrade to the latest but for example from one LTS like 17 to a later one like Java-21. Without the ability to select the target this new features would often not be useable.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -127,7 +133,12 @@ public Object[] getElements(Object inputElement) {
 	}
 
 	public ExecutionEnvironmentSection(PDEFormPage page, Composite parent) {
-		super(page, parent, Section.DESCRIPTION, new String[] {PDEUIMessages.RequiredExecutionEnvironmentSection_add, PDEUIMessages.RequiredExecutionEnvironmentSection_remove, PDEUIMessages.RequiredExecutionEnvironmentSection_up, PDEUIMessages.RequiredExecutionEnvironmentSection_down});
+		super(page, parent, Section.DESCRIPTION,
+				new String[] { PDEUIMessages.RequiredExecutionEnvironmentSection_add,
+						PDEUIMessages.RequiredExecutionEnvironmentSection_upgrade,
+						PDEUIMessages.RequiredExecutionEnvironmentSection_remove,
+						PDEUIMessages.RequiredExecutionEnvironmentSection_up,
+						PDEUIMessages.RequiredExecutionEnvironmentSection_down });
```
HannesWell, line 141: Since `Upgrade` does more than Add/Remove/Up/Down I think it should be either first or last to have the existing buttons grouped.
```suggestion
				new String[] {  PDEUIMessages.RequiredExecutionEnvironmentSection_upgrade,
						PDEUIMessages.RequiredExecutionEnvironmentSection_add,
						PDEUIMessages.RequiredExecutionEnvironmentSection_remove,
						PDEUIMessages.RequiredExecutionEnvironmentSection_up,
						PDEUIMessages.RequiredExecutionEnvironmentSection_down });
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -270,16 +285,19 @@ public void run() {
 	}
 
 	private void updateButtons() {
+		String javaExecutionEnvironmentMatcher = "(?i)\\b(jre|javase|j2se).*\\b"; //$NON-NLS-1$
```
HannesWell, line 288: To improve the performance you should in general consider to compile the regex into a static final `java.util.regex.Pattern`.
But I wonder if a simple `text.startswith("jre") || text.startswith("javase") || text.startswith("j2se")` would be simpler to read (and faster).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -311,7 +329,15 @@ private void handleRemove() {
 	}
 
 	private void handleAdd() {
-		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(), new EELabelProvider());
+		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(),
+				new EELabelProvider()) {
+			@Override
+			protected FilteredList createFilteredList(Composite parent) {
+				FilteredList filteredList = super.createFilteredList(parent);
+				filteredList.setComparator(new VersionComparator());
```
HannesWell, line 337: As mentioned below the EEs are probably already sorted in the EE-Manager. So just reversing the list might be sufficient.

HannesWell (commented):
Thank you @lathapatil for this contribution.
In general I like the proposed functionality a lot, but I think it needs some fine-tuning. Please see the comments below.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
laeubi: I won't rely on the order of some method we don't control.

laeubi: I agree that the user should be asked for what is to be upgraded, I remember we specifically discussed this in the issue itself, this will also remove any need to sort the list as the dialog itself is sorted already.

lathapatil: > And in case it would be sorted, I would use
> `Comparator.comparing(ee -> new org.osgi.framework.Version(getVersionString(ee)))`
> to create a comparator without the need to implement a version comparison logic again.

@HannesWell I only have option to set Comparator to `org.eclipse.ui.dialogs.FilteredList` and the sorting is done in `TwoArrayQuickSorter` .

I get the versions in ascending order using `Comparator.comparing(ee -> new org.osgi.framework.Version(getVersionString(ee)))` but the requirement is to get it in descending order .

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
lathapatil: @HannesWell  Yes. They are sorted based on string comparison and not versions. 
If I am not wrong the list is sorted in natural order in `setElements` method of `FilteredList` class (unless we set a comparator explicitly ) even if the list is revered before setting it to dialog.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
lathapatil: @HannesWell Please review the latest code

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -283,17 +300,62 @@ private void handleRemove() {
...
+				// add selected Java EE to the header
+				addExecutionEnvironments(dialog.getResult());
+
+				final IProject project = getPage().getPDEEditor().getCommonProject();
+				updateClasspathSettings(project);
+			}
+		}
+	}
+
+	private ElementListSelectionDialog showDialog(boolean isUpgrade) {
```
HannesWell, line 331: Let's be more descriptive about what this change is:
```suggestion
	private ElementListSelectionDialog showDialog(boolean allowMultiSelection) {
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -283,17 +300,62 @@ private void handleRemove() {
...
+				return ((String) ee).replaceAll(versionMatcher, "$1"); //$NON-NLS-1$
+			}
+		};
 		dialog.setElements(getEnvironments());
 		dialog.setAllowDuplicates(false);
-		dialog.setMultipleSelection(true);
+		if (isUpgrade)
+			dialog.setMultipleSelection(false);
+		else
+			dialog.setMultipleSelection(true);
```
HannesWell, line 352: In general you should always use braces even if if-else blocks only contain a single statement.
But in this case this could be even simpler:
```suggestion
		dialog.setMultipleSelection(allowMultiSelection);
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -283,17 +300,62 @@ private void handleRemove() {
...
+		}
+	}
+
+	private ElementListSelectionDialog showDialog(boolean isUpgrade) {
+		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(),
+				new EELabelProvider()) {
+			@Override
+			protected FilteredList createFilteredList(Composite parent) {
+				FilteredList filteredList = super.createFilteredList(parent);
+				filteredList.setComparator(Comparator.comparing(ee -> new VersionComparator(getVersionString(ee))));
```
HannesWell, line 337: When https://github.com/eclipse-pde/eclipse.pde/pull/1324 is completed, this can just be
```suggestion
				filteredList.setComparator(VMUtil.ASCENDING_EE_JAVA_VERSION.reversed());
```

HannesWell (commented):
@lathapatil sorry for the delayed reply.
In https://github.com/eclipse-pde/eclipse.pde/pull/1324 I'm about to add a `Comparator` to compare EEs based on their java version named `VMUtil.ASCENDING_EE_JAVA_VERSION`. My intention was also that you can re-use it for this PR and we don't have to re-invent the logic at different places.
I didn't responded yet since I wanted to have the mentioned PR completed first, but unfortunately it takes longer than expected. But it should complete in the next days.
I'll let you know.

lathapatil:
> since I wanted to have the mentioned PR completed first

Sure , I will wait for the other PR . Thanks for the update.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
+		if (dialog.open() == Window.OK) {
+			addExecutionEnvironments(dialog.getResult());
+		}
+	}
+
+	private void handleUpgrade() {
+		RequiredExecutionEnvironmentHeader header = getHeader();
+		IExecutionEnvironment[] executionEnvironments = JavaRuntime.getExecutionEnvironmentsManager()
+				.getExecutionEnvironments();
+		if (executionEnvironments.length > 0) {
```
HannesWell, line 287: I cannot think of a situation where the if-condition is not true.
```suggestion
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
+
+				List<String> addedExecutionEnvironments = header.getEnvironments();
+				// Remove existing EE ( Upgrade is allowed only when single Java
+				// EE is present in the header)
+				getHeader().removeExecutionEnvironment(addedExecutionEnvironments.get(0));
+
+				// add selected Java EE to the header
+				addExecutionEnvironments(dialog.getResult());
+
+				final IProject project = getPage().getPDEEditor().getCommonProject();
```
HannesWell, line 299:
```suggestion
				IProject project = getPage().getPDEEditor().getCommonProject();
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
+		IExecutionEnvironment[] executionEnvironments = JavaRuntime.getExecutionEnvironmentsManager()
+				.getExecutionEnvironments();
+		if (executionEnvironments.length > 0) {
+			ElementListSelectionDialog dialog = showDialog(false);
+			if (dialog.open() == Window.OK) {
+
+				List<String> addedExecutionEnvironments = header.getEnvironments();
+				// Remove existing EE ( Upgrade is allowed only when single Java
+				// EE is present in the header)
+				getHeader().removeExecutionEnvironment(addedExecutionEnvironments.get(0));
```
HannesWell, line 294: The variable name could be a bit more precise and the comment can be updated that all single EEs are upgradable.
```suggestion
			List<String> existingEEs = header.getEnvironments();
			// Remove existing EE (Upgrade is allowed only if a single EE is
			// present in the header)
			getHeader().removeExecutionEnvironment(existingEEs.get(0));
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
 		dialog.setTitle(PDEUIMessages.RequiredExecutionEnvironmentSection_dialog_title);
 		dialog.setMessage(PDEUIMessages.RequiredExecutionEnvironmentSection_dialogMessage);
 		dialog.create();
-		PlatformUI.getWorkbench().getHelpSystem().setHelp(dialog.getShell(), IHelpContextIds.EXECUTION_ENVIRONMENT_SELECTION);
-		if (dialog.open() == Window.OK) {
-			addExecutionEnvironments(dialog.getResult());
-		}
+		PlatformUI.getWorkbench().getHelpSystem().setHelp(dialog.getShell(),
+				IHelpContextIds.EXECUTION_ENVIRONMENT_SELECTION);
+		return dialog;
```
HannesWell, line 323: Handling cancellations and alike here avoids having the logic in all callers.
```suggestion
		if (dialog.open() == Window.OK) {
			return dialog.getResult();
		}
		return null;
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
+				// add selected Java EE to the header
+				addExecutionEnvironments(dialog.getResult());
+
+				final IProject project = getPage().getPDEEditor().getCommonProject();
+				updateClasspathSettings(project);
+			}
+		}
+	}
+
+	private ElementListSelectionDialog showDialog(boolean allowMultiSelection) {
```
HannesWell, line 305: The method name should be a bit more precise.
```suggestion
	private Object[] showEESelectionDialog(boolean allowMultiSelection) {
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -262,17 +274,53 @@ private void handleRemove() {
...
+		}
+	}
+
+	private void handleUpgrade() {
+		RequiredExecutionEnvironmentHeader header = getHeader();
+		IExecutionEnvironment[] executionEnvironments = JavaRuntime.getExecutionEnvironmentsManager()
+				.getExecutionEnvironments();
+		if (executionEnvironments.length > 0) {
+			ElementListSelectionDialog dialog = showDialog(false);
+			if (dialog.open() == Window.OK) {
```
HannesWell, line 289:
```suggestion
			Object[] selection = showEESelectionDialog(false);
			if (selection != null) {
```

HannesWell (changes requested):
Thanks @lathapatil for the update, I think we are on the finishing stretch with this one and I'm looking forward to use it. This will be very handy.
I just have a few minor remarks for further enhancements.

The only general point I have is the label of the Button. Since one can also downgrade the EE, I wonder if we would better label the button with `Change` instead of `Upgrade`. What do you think?

Since the change to sort the EEs in descending order is already valuable by itself, if you want to, you can also move that change into a second commit of this PR.

The commit message should be updated and the header should only be a single line. For example:
```
Add 'Upgrade' button to change a Plugin's required EE and classpath

Changing the required execution environment through that button changes
the corresponding header in the MANIFEST.MF as well as the JRE_CONTAINER
listed in the '.classpath' file as well as the compiler settings in the JDT
preferences in one action.
```

lathapatil:
> The only general point I have is the label of the Button. Since one can also downgrade the EE, I wonder if we would better label the button with `Change` instead of `Upgrade`. What do you think?

+1 for `Change`

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -246,8 +254,12 @@ public void run() {
 	private void updateButtons() {
 		Table table = fEETable.getTable();
 		TablePart tablePart = getTablePart();
-		tablePart.setButtonEnabled(0, isEditable());
-		tablePart.setButtonEnabled(1, isEditable() && table.getSelection().length > 0);
+		tablePart.setButtonEnabled(0,
+				isEditable() && table.getItemCount() == 1 && (table.getItem(0).getText().startsWith("JRE") //$NON-NLS-1$
+						|| table.getItem(0).getText().startsWith("JavaSE") //$NON-NLS-1$
+						|| table.getItem(0).getText().startsWith("J2SE"))); //$NON-NLS-1$
```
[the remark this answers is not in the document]
lathapatil: I think Its better to have below condition for `Change` button

`tablePart.setButtonEnabled(0, isEditable() && table.getSelectionCount() == 1);`

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -113,7 +116,8 @@ public void dispose() {
 	}
 
 	public ExecutionEnvironmentSection(PDEFormPage page, Composite parent) {
-		super(page, parent, Section.DESCRIPTION, new String[] { //
+		super(page, parent, Section.DESCRIPTION, new String[] {
```
HannesWell, line 119: Nitpick: Please preserve the comment to prevent changes by future formatting.
```suggestion
		super(page, parent, Section.DESCRIPTION, new String[] { //
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
```diff
@@ -260,17 +269,53 @@ private void handleRemove() {
 	}
 
 	private void handleAdd() {
-		ElementListSelectionDialog dialog = new ElementListSelectionDialog(PDEPlugin.getActiveWorkbenchShell(), new EELabelProvider());
+		Object[] selection = showEESelectionDialog(false);
```
HannesWell, line 272: Shouldn't the argument be `true`?
```suggestion
		Object[] selection = showEESelectionDialog(true);
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -2283,6 +2283,7 @@ GetNonExternalizedStringsAction_allExternalizedMessage=All strings in manifest f
 RequiredExecutionEnvironmentSection_title=Execution Environments
 RequiredExecutionEnvironmentSection_add=Add...
 RequiredExecutionEnvironmentSection_remove=Remove
+RequiredExecutionEnvironmentSection_change=Change
```
HannesWell, line 2286: Like for the add-button we should also append three dots to indicate this opens a dialog.
```suggestion
RequiredExecutionEnvironmentSection_change=Change...
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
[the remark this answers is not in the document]
HannesWell: That would make the implementation more difficult. At the moment you can just replace the first (and only) EE `existingEEs.get(0)` in `handleChange()`. Furthermore then the question would arise when to update the Classpath? Only if the EE corresponding to the java version configured in the classpath/compiler settings?
Because the vast majority of Plugin/Bundle projects has only one EE I suggest to keep this button for that simple case and let those that have complex EE configs handle everything on their own.

HannesWell (commented):
@lathapatil thanks for the update. This looks very good now.
Please the three comments for the two little things that should be adjusted.

Regarding the label of the button: As an alternative to `Change...` we could also use `Modify...`?
I believe modify sounds a bit more technical, but I have no real preference and just wanted to give you another idea.

And if you want to, you could also add a tool-tip that explains that the Change/Modify button also updates the Java Compiler and classpath configuration.

vogella:
Thanks. Cool stuff

HannesWell:
> Thanks. Cool stuff 

Absolutely.
Btw. @lathapatil to tell more users about this new cool stuff can you please provide a new and noteworthy entry about this in https://github.com/eclipse-platform/www.eclipse.org-eclipse ?

HannesWell:
I showed this new button at OCX/EclipseCon in my presentation about the latest enhancements in PDE and many people liked this new feature very much. 😃🚀

Issue #1289, opened 2024-06-05
NPE in ReferenceAnalyzer


jukzi:
during Github verification build:
```
Caused by: java.lang.NullPointerException: Cannot load from object array because "this.this$0.fIndexedDetectors" is null
    at org.eclipse.pde.api.tools.internal.builder.ReferenceAnalyzer$Visitor.visit (ReferenceAnalyzer.java:105)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.doVisit (ProjectTypeContainer.java:102)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.doVisit (ProjectTypeContainer.java:121)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.doVisit (ProjectTypeContainer.java:121)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.doVisit (ProjectTypeContainer.java:121)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.doVisit (ProjectTypeContainer.java:121)
    at org.eclipse.pde.api.tools.internal.model.ProjectTypeContainer.accept (ProjectTypeContainer.java:62)
    at org.eclipse.pde.api.tools.internal.model.AbstractApiTypeContainer.accept (AbstractApiTypeContainer.java:60)
    at org.eclipse.pde.api.tools.internal.model.Component.accept (Component.java:66)
    at org.eclipse.pde.api.tools.internal.builder.ReferenceAnalyzer.extractReferences (ReferenceAnalyzer.java:210)
    at org.eclipse.pde.api.tools.internal.builder.ReferenceAnalyzer.analyze (ReferenceAnalyzer.java:239)
    at org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.checkApiUsage (BaseApiAnalyzer.java:1248)
    at org.eclipse.pde.api.tools.internal.builder.BaseApiAnalyzer.analyzeComponent (BaseApiAnalyzer.java:288)
    at org.eclipse.tycho.apitools.ApiAnalysis.call (ApiAnalysis.java:175)
    at jdk.internal.reflect.DirectMethodHandleAccessor.invoke (DirectMethodHandleAccessor.java:103)
    at java.lang.reflect.Method.invoke (Method.java:580)
    at org.eclipse.tycho.osgi.framework.EclipseFramework.execute (EclipseFramework.java:141)
[...]

PR #1290, opened 2024-06-05
fix NPE for ReferenceAnalyzer.fIndexedDetectors null #1289

jukzi:
https://github.com/eclipse-pde/eclipse.pde/issues/1289

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/ReferenceAnalyzer.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/ReferenceAnalyzer.java
@@ -146,7 +146,7 @@ public class ReferenceAnalyzer {
 	 * are interested in. Provides a fast way to hand references off to
 	 * interested problem detectors.
 	 */
-	IApiProblemDetector[][] fIndexedDetectors;
+	private IApiProblemDetector[][] fIndexedDetectors;
 
 	/**
 	 * Indexes the problem detectors by the reference kinds they are interested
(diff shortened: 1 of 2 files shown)


gireeshpunathil (approved):

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/ReferenceAnalyzer.java
```diff
@@ -230,10 +230,14 @@ void extractReferences(IApiTypeContainer scope, IProgressMonitor monitor) throws
 	 */
 	public IApiProblem[] analyze(IApiComponent component, IApiTypeContainer scope, IProgressMonitor monitor) throws CoreException {
 		SubMonitor localMonitor = SubMonitor.convert(monitor, 4);
-		// build problem detectors
-		IApiProblemDetector[] detectors = buildProblemDetectors(component, ProblemDetectorBuilder.K_ALL, localMonitor.split(1));
-		// analyze
 		try {
+			// build problem detectors
+			IApiProblemDetector[] detectors = buildProblemDetectors(component, ProblemDetectorBuilder.K_ALL,
```
HannesWell, line 235: Is there a reason why you moved this into the try-catch-block?

jukzi: sure: the final block cleans up a field that is initialized somewhen within  "buildProblemDetectors". Any exception after initializing the field would leave a leak.

Issue #1283, opened 2024-06-03
API error on SWT "execution environments have been changed since"


iloveeclipse:
I've set my baseline to 4.32RC2a build and see now this API error on SWT:

`The minor version should be incremented in version 3.126.0, since execution environments have been changed since version 3.126.0`

I'm not aware about any execution environment change in SWT, so why do we have that error now?
Is something broken again with API descriptions, like we had it in 4.31 (se https://github.com/eclipse-pde/eclipse.pde/issues/1187)?

merks:
The API baseline I find to be notoriously hard to update.   If you to to the preference, double click it and refresh it, does that help?  If not, restart, and do that again maybe helps?

iloveeclipse:
Does it mean, you can't reproduce? I see it on both my Linux/Windows workspaces, so I assume this should be reproducible.

merks:
I’m driving around so not tried yet. So it was just a thought.

merks:
Yes, I have that error too:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/2e96f43e-3c7f-4035-8f18-302eaa58efaf)

The thing doesn't have a BREE though and never did.   Last release cycle it was version bumped without a reason...

merks:
Debugging, here is the problem:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/c16cde23-4946-4758-ad28-b16265c2f947)

The API reference indicates there is no EE at all, but from the bundle description the current project computes that it has EE  Java-17 even though that value is not present in the MANIFEST.MF.

While building the description, the entry is in the map, but not in the file itself:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/55af3de5-3e17-4b71-af22-40f244bec7a2)

And finally, we see here that PDE is adding the EE property, computed from the JRE associated with the project, to the manifest properties as if it were present in the loaded resource:

![image](https://github.com/eclipse-pde/eclipse.pde/assets/208716/df8a3fde-614e-477c-a762-b1005d65eafb)

It's not entirely clear to me how to fix this problem.  For validation would need to should be able to distinguish between a computed/implicit default value versus an explicitly-present value...  I don't think it makes so much sense that the baseline ought to record some value not actually present and one that depends on the workspace's JRE.  Note in particular that if one associates the JavaSE-17 EE with a Java 21 JRE/JDK, the computed value changes:

[...]

PR #1292, opened 2024-06-06
Track whether the manifest of a bundle component has an explicit EE

merks:
https://github.com/eclipse-pde/eclipse.pde/issues/1283


merks:
@iloveeclipse 

I've verified that this fixed the problem where the baseline version and the workspace version as the same and now both are validated such that both indicate they have no BREE and hence to BREE change:

image

Also verified that if I were to add an explicit BREE, this would be detected as a change to the BREE:

image

merks:
@iloveeclipse 

I'll give some more time to review, but I'd like to commit this today so that I can update my IDE to fix the problem tomorrow....

iloveeclipse (approved):
Looks good for me and I also can confirm that the API error on Linux on SWT is gone with this patch.

iloveeclipse:
> I'll give some more time to review, but I'd like to commit this today so that I can update my IDE to fix the problem tomorrow....

Sure. I've checked the patch on Linux, sounds reasonable, but I'm not that deep in this PDE code. I would say we should merge to get rid of the current (false positive) error on SWT and if the patch could be improved, it should be improved later.

PR #1293, opened 2024-06-07
Test removing org.eclipse.rcp.configuration.feature.group

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2109

### DIFF
--- a/build/org.eclipse.pde.build.tests/pom.xml
+++ b/build/org.eclipse.pde.build.tests/pom.xml
@@ -80,7 +80,6 @@
 										<iu><id>org.eclipse.platform.feature.group</id></iu>
 										<iu><id>org.eclipse.jdt.feature.group</id></iu>
 										<iu><id>org.eclipse.jdt.source.feature.group</id></iu>
-										<iu><id>org.eclipse.rcp.configuration.feature.group</id></iu>
 										<iu><id>org.eclipse.equinox.executable</id></iu>
 										<iu><id>org.eclipse.e4.rcp.feature.group</id></iu>
 										<iu><id>org.eclipse.e4.rcp.source.feature.group</id></iu>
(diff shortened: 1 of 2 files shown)


merks:
@laeubi 

Surprisingly it seems all the tests pass!

@HannesWell 

The failures look like build/infrastructure problems, not test failures...

laeubi:
Just be aware that removing it from `test.xml` will not be validated in a PR but at least for maven it looks like it works...

merks:
I see.  Sort of but not really.  I see it referenced here:

image

So I assumed that tests just use it.  

What else uses this and when/where?  

This suggested there is no way to know without changing it and hoping for the best.   😨

laeubi:
It is used by the I-Build: https://wiki.eclipse.org/Platform-releng/Eclipse_Test_Framework

the referenced one in code might or might not be *taht* `test.xml` as there are many of them...

HannesWell (approved):
Let's try if this works in I-builds.
If it fails, reverting it is overall simpler than thinking about it much longer.

merks:
@iloveeclipse @akurtakov  @MohananRahul 

FYI, this change as some chance to break the I-Build via the PDE Build Test so I thought I'd give you a heads up...

akurtakov:
Please start extra I build to test it.

merks:
I've started https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/13/

merks:
The build finished and I think it launched these tests:

image

But my knowledge of how this part works is poor.  😢

merks:
If I'm reading the tea leaves correctly, only this JDT test is failing:

image

So am I safe to assume this change caused neither build nor test regressions?  Then I can proceed to remove what has no become an unused feature...

iloveeclipse:
Yes, if you mean these tests: https://download.eclipse.org/eclipse/downloads/drops4/I20240611-0150/testResults.php I don't see "unexpected" new failures.

merks:
@iloveeclipse 

Thanks for confirming!   That's very good news!  🎉

PR #1296, opened 2024-06-12
ResourceApiTypeRoot: use IFile.readAllBytes()

jukzi:


HannesWell (commented):
Are these preference changes intended?
Apart from that, this looks good to me. 👍🏽

jukzi:
> Are these preference changes intended?

removed.

PR #1300, opened 2024-06-13
Open edit launch dialog using correct mode

kysmith-csg:
When a launch fails validation, the user is given a link to Edit Launch Configuration. Previously, this link always opened the Edit Run Configurations dialog in Run mode, even if the user initially attempted to Debug. This change ensures that the edit dialog is opening using the original mode.

### DIFF
--- a/features/org.eclipse.pde.unittest.junit-feature/feature.xml
+++ b/features/org.eclipse.pde.unittest.junit-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.unittest.junit"
       label="%featureName"
-      version="1.0.800.qualifier"
+      version="1.0.900.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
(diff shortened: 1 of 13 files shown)


HannesWell (changes requested):
Thanks for this PR, and sorry for the delayed reply.
The goal is good, but I think the implementation should be enhanced.
Instead of having a global static field in `LauncherUtils` that can be set by everyone and has to be remembered to be set, I think it would be better just pass through the `launchMode` from a launch.

While checking if that's possible I already made the necessary changes and pushed them as separate commit to your (rebased branch). If you agree with them, please incorporate them into your commit.

kysmith-csg:
@HannesWell 

>  please incorporate them into your commit

What does this mean exactly? Do you want me to force push them? You could also just squash-merge this pull request to accept it and it would only make a single commit to the main repo.

HannesWell (approved):
> @HannesWell
> 
> > please incorporate them into your commit
> 
> What does this mean exactly? Do you want me to force push them?

Yes. I wanted to give you the possibility to comment on the change in case you don't think it's fine.
But from your comment I think you are fine with it and I just squashed it. And added a missing version bump.

> You could also just squash-merge this pull request to accept it and it would only make a single commit to the main repo.

Yes that would work, but github's squash-and-merge leads to a generic committer mail, which I don't think is good since it hides the person that actually submitted a change.

But this is now fine. Thanks for this contribution.

PR #1313, opened 2024-06-25
Bypass API analysis for classes from a test classpath entry

merks:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1298


laeubi (approved):
Looks good!

HannesWell:
In a (private) chat with Ed I understood this as a 'last' resort solution but since we got https://github.com/eclipse-platform/eclipse.platform.swt/pull/1305 this is not necessary anymore to fix the underlying issue.
So should we still keep this? If yes, maybe we should move the check to a higher level and let the analyzer just skip test source folders. Then it does not have to be checked again and again when creating problems (which are not always errors).

merks:
I didn’t merge it. I’m not sure we really need this given swt solved the problem without this.

laeubi:
>  I’m not sure we really need this given swt solved the problem without this.

Well SWT not really "solved" this, they installed a workaround (making thing package private) wile this works in this case I still believe test code should not be API checked, if it can happen on a higher level why not.

Issue #1310, opened 2024-06-18
Execute ApiApplication failed


laeubi:
This was shown in a Github Action and shows a place where a component is disposed while still be used:

Disposal Stacktrace:
```
 Caused by: java.lang.RuntimeException: Component was disposed here
    at org.eclipse.pde.api.tools.internal.model.BundleComponent.dispose (BundleComponent.java:203)
    at org.eclipse.pde.api.tools.internal.model.ProjectComponent.dispose (ProjectComponent.java:178)
    at org.eclipse.pde.api.tools.internal.model.ApiBaseline.doDispose (ApiBaseline.java:837)
    at org.eclipse.pde.api.tools.internal.model.WorkspaceBaseline.dispose (WorkspaceBaseline.java:56)
    at org.eclipse.pde.api.tools.internal.ApiBaselineManager.lambda$2 (ApiBaselineManager.java:639)
    at org.eclipse.pde.api.tools.internal.ApiBaselineManager.disposeWorkspaceBaseline (ApiBaselineManager.java:662)
    at org.eclipse.pde.api.tools.internal.WorkspaceDeltaProcessor.processJavaElementDeltas (WorkspaceDeltaProcessor.java:80)
    at org.eclipse.pde.api.tools.internal.WorkspaceDeltaProcessor.elementChanged (WorkspaceDeltaProcessor.java:51)
    at org.eclipse.jdt.internal.core.DeltaProcessor$3.run (DeltaProcessor.java:1751)
    at org.eclipse.core.runtime.SafeRunner.run (SafeRunner.java:47)
    at org.eclipse.jdt.internal.core.DeltaProcessor.notifyListeners (DeltaProcessor.java:1739)
    at org.eclipse.jdt.internal.core.DeltaProcessor.firePostChangeDelta (DeltaProcessor.java:1572)
    at org.eclipse.jdt.internal.core.DeltaProcessor.fire (DeltaProcessor.java:1548)
    at org.eclipse.jdt.internal.core.DeltaProcessor.notifyAndFire (DeltaProcessor.java:2259)
    at org.eclipse.jdt.internal.core.DeltaProcessor.resourceChanged (DeltaProcessor.java:2151)
    at org.eclipse.jdt.internal.core.DeltaProcessingState.resourceChanged (DeltaProcessingState.java:501)
    at org.eclipse.core.internal.events.NotificationManager$1.run (NotificationManager.java:321)
[...]

PR #1316, opened 2024-06-29
Show threads involved in disposing of components

laeubi:
Currently we show where the component was disposed but not by what thread. As this can be interesting for debugging the cause, this information is now added.

Just i minor improvement regarding https://github.com/eclipse-pde/eclipse.pde/issues/1310 to get more information about the involved parties.


HannesWell (approved):

iloveeclipse:
Can it be merged?

PR #1321, opened 2024-07-04
Remove rarely usable Drag&Drop of EEs in Manifest-Editor Overview page

HannesWell:
Drag&drop support was introduced via [Bug 192890](https://bugs.eclipse.org/192890), but only supports reordering of required Execution-Environments in the same Plug-in project.
Because in the vast majority of cases Plug-ins only have one EE defined that feature is only rarely usable. And if a Plug-in project really defines multiple BREEs the Manifest Editor already provides an alternative with the 'Up' and 'Down' buttons next to the table.

grafik

While being only rarely usable support D&D support for required EEs makes it more difficult to support OSGi EE requirements in https://github.com/eclipse-pde/eclipse.pde/pull/1318.
Therefore I propose to remove it.

@laeubi, @merks or anbody else, do you have any objections?

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExecutionEnvironmentSection.java
@@ -39,7 +39,6 @@ import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.LabelProvider;
 import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.viewers.TableViewer;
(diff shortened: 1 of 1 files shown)


merks:
There are many places where drag and drop would be nice for reordering but this  seems at the very bottom of the list of useful places. 👍

HannesWell:
> There are many places where drag and drop would be nice for reordering but this seems at the very bottom of the list of useful places. 👍

Absolutely.
Let's give other some time to react, but if there are no objections until tomorrow evening, I'll submit this.

laeubi:
I don't even think the order matters at all, so maybe we can even remove the buttons...

vogella:
+1 for removing DND here.

HannesWell:
> I don't even think the order matters at all, so maybe we can even remove the buttons...

See https://github.com/eclipse-pde/eclipse.pde/pull/1324.

PR #1324, opened 2024-07-06
Always sort BREE header values in ascending Java version order

HannesWell:
Remove the ability to manually move required execution environments in the Manifest Editor. Instead always sort the values in ascending Java version order.

From a technical perspective the order of the values of the `Bundle-RequiredExecutionEnvironment` header is irrelevant, similar to for example the `Import-Package` header. Removing the buttons to reorder the values simplifies the UI.

Values of existing Plug-in projects are displayed in sorted order regardless of if they are actually sorted in the `MANIFEST.MF` or not. On the next write all values will be written in sorted order.

With this change applied the EE section now looks like this:

grafik


HannesWell:
@laeubi, @merks what do you think about this? Any objections?

As a side note: I just created https://github.com/eclipse-equinox/equinox/pull/655 as an attempt to make the order of some old Java version as expected. But since it only affects old versions I don't consider it a prerequisite.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/VMUtil.java
```diff
@@ -96,12 +101,31 @@ public static String getVMInstallName(IExecutionEnvironment ee) throws CoreExcep
...
+				case "J2SE-1.2" -> 1.2; //$NON-NLS-1$
+				case "J2SE-1.3" -> 1.3; //$NON-NLS-1$
+				case "J2SE-1.4" -> 1.4; //$NON-NLS-1$
+				default -> {
+					Properties properties = ee.getProfileProperties();
+					Object target = properties != null //
+							? properties.get(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM)
+							: null;
+					yield target instanceof String version ? Double.parseDouble(version) : 0;
+				}
```
HannesWell, line 123: Just for clarification: This part could probably be simplified to only the default case if https://github.com/eclipse-equinox/equinox/pull/655 were available.

HannesWell:
I'm going to submit this now to unblock #1285, but still hope that https://github.com/eclipse-equinox/equinox/pull/655 can be leveraged later.

PR #1328, opened 2024-07-10
Updating Execution_Environments.md to more recent examples

vogella:
Removes examples for JRE which do not exist anymore and other very specific cases which also do not exit more, also remove old BREE listing and add one one which are still relevant these days.


vogella:
Merging this, the document was so outdated that any potential reader must have left the document very fast. Hopeful the shorter and more recent text will find a few readers in the future.

PR #1336, opened 2024-07-13
Simplify and clean-up API baseline initialization

HannesWell:
Also remove unused method ApiBaseline.getVisibleDependentComponents().

@merks you often looked into the API-baseline tooling. Do you want to review this?

I assume `ApiBaselineManager.createWorkspaceBaseline()` used to be so complicated because it grew historically.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiBaselineManager.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ApiBaselineManager.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


merks:
I've only tried to make the complicated stuff work when it's not working, not try to understand why it's so complicated...

HannesWell:
> I've only tried to make the complicated stuff work when it's not working, not try to understand why it's so complicated...

Understand. Then I assume you are fine with this.

All tests pass, although I see that the GH workflows are not running...

merks:
It sure bypasses a lot of stuff, but if you've tested it and the test pass, it definitely gets rid of a lot of complexity!!

HannesWell:
> It sure bypasses a lot of stuff, but if you've tested it and the test pass, it definitely gets rid of a lot of complexity!!

In my local test (but I could only test a simple case) this worked well and all tests pass.

PR #1337, opened 2024-07-13
[Build] Adapt license-check to its move to own eclipse-dash organization

HannesWell:
See https://github.com/eclipse-dash/dash-licenses/issues/302


HannesWell:
License check passed. Submitting.

HannesWell:
@laeubi can you tell what the advantage of having the license check and request for licensing vetting split in different workflows?
From my perspective it just makes maintenance harder since more than one file handles license checking.

PR #1338, opened 2024-07-13
[Build] Handle license check and vetting request in one workflow

HannesWell:
This simplifies maintenance since the license handling is not distributed over multiple workflows.

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -1,6 +1,12 @@
-name: Request license-review
+# This workflow will check for Maven projects if the licenses of all (transitive) dependencies are vetted.
+
+name: License status
 
 on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
   issue_comment:
     types: [created]
 
@@ -12,4 +18,4 @@ jobs:
     with:
       projectId: eclipse.pde
     secrets:
-      gitlabAPIToken: ${{ secrets.PDE_GITLAB_API_TOKEN }}
+      gitlabAPIToken: ${{ github.event_name == 'issue_comment' && secrets.PDE_GITLAB_API_TOKEN || '' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,6 @@ jobs:
   check-merge-commits:
     if: github.event_name == 'pull_request'
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/checkMergeCommits.yml@master
-  check-dash-licenses:
-    uses: eclipse-dash/dash-licenses/.github/workflows/mavenLicenseCheck.yml@master
-    with:
-      projectId: eclipse.pde
   build:
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/mavenBuild.yml@master
 


laeubi (changes requested):
@HannesWell actually this was intended, because the *check* itself requires less permissions, I assume this does not change that often anyways to become an issue.

laeubi:
This seem to complicate the handling a lot just to save a duplicate line that almost never changes... 

Also I actually like it to be shown as part of the "Continuous Integration" workflow instead of a separate one.

.github/workflows/licensecheck.yml
```diff
@@ -1,6 +1,12 @@
-name: Request license-review
+# This workflow will check for Maven projects if the licenses of all (transitive) dependencies are vetted.
+
+name: License vetting status check
```
laeubi, line 3: This is now misleading as it also handles the previous `Request license-review`

HannesWell:
> This seem to complicate the handling a lot just to save a duplicate line that almost never changes...
>
Using one [ternary operator](
https://docs.github.com/en/actions/learn-github-actions/expressions#operators) does not seem to be a lot more complicated to me (tbh the GH workflow syntax is unknown also to me, but that's their decision).

On the other hand having the license handling distributed over multiple files also makes it more complicated to understand, even if does not change that often. Which weights more is a matter of preference.

> Also I actually like it to be shown as part of the "Continuous Integration" workflow instead of a separate one.

I don't think this is a big issue. The license check is a 'meta' check any ways not directly related to building a change.
And in the end in the vast majority of cases I just have a look at the check summary at the bottom of a PR and there all checks are listed in just one flat list. How it is present in the `Checks` tab of the PR or even in the `Actions` section of this repo is at least by myself not often looked at.

I understand the technical concerns about too elevated permission, but not those about the presentation.
Furthermore everyone else that I know of who is using this workflow is using it from a single file.

> 
> Actually it would even be better if they where two different workflow files for the different purposes...


Feel free to provide PRs to https://github.com/eclipse-dash/dash-licenses, but it's not that easy to achieve that without a major rework because many is shared in both cases. And in the end I don't think it is worth the effort.

.github/workflows/licensecheck.yml
HannesWell: Then lets just name it `License status`.

laeubi:
> I don't think this is a big issue. The license check is a 'meta' check any ways not directly related to building a change.

For me this is an important part, having the same workflow referenced in two places is something I way less often need to think about than looking at the actual checks.

So I still don't see any benefits here beside that now the check runs with elevated peremissions (PR write) and it feels confusing that the workflow performs two different tasks (one for checking license and one for requesting a review).

It would actually be better if the check itself would be split into two dedicated ones, one for checking the license and one for requesting a review, then there won't be any "duplication".

.github/workflows/licensecheck.yml
laeubi: This is still wrong, requesting a review is not a status, also I suspect that now the check will maybe become green if requesting succeeds... I still see no benefit in mixing these two concerns into one workflow run that makes it more complex to manage and understand and test.

e.g. previously if the check failed one could rerun it, what will happen now is unclear, will it repeat the check of licenses or will it repeat the request for license and all that kind of uncertainties.

laeubi:
> On the other hand having the license handling distributed over multiple files also makes it more complicated to understand, even if does not change that often.

We have two cases:

1. check license
2. request licensee review

so these are not "multiple places" it is exactly one place for one purpose... on the other hand now I need to understand that the shared workflow has two cases (check/request) and the workflow referenced also has different cases... so for me this still is not worth the hassle just because we have change the name once what most probably will never happen again.

HannesWell:
> > I don't think this is a big issue. The license check is a 'meta' check any ways not directly related to building a change.
> 
> For me this is an important part, having the same workflow referenced in two places is something I way less often need to think about than looking at the actual checks.

Looks like this is the root of our disagreement. For me its the other way round.

> So I still don't see any benefits here beside that now the check runs with elevated peremissions (PR write) and it feels confusing that the workflow performs two different tasks (one for checking license and one for requesting a review).

What the workflow does or in how many files it is distributed in the end is probably irrelevant for most contributors since they only care if it works.
But the elevated permissions are indeed a drawback, therefore I'm closing this.

PR #1340, opened 2024-07-14
Replace project APIs using antiquated Equinox resolver's VersionRange

HannesWell:
Provide alternatives that are using `org.osgi.framework.VersionRange` and deprecate the existing methods for removal.

This is extracted from https://github.com/eclipse-pde/eclipse.pde/pull/1163 handling only the `simple` cases.

Part of https://github.com/eclipse-pde/eclipse.pde/issues/1069.

### DIFF
--- a/org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
+++ b/org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.doc.user; singleton:=true
-Bundle-Version: 3.15.300.qualifier
+Bundle-Version: 3.16.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.help;bundle-version="[3.2.0,4.0.0)"
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/project/IHostDescription.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/project/IHostDescription.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010 IBM Corporation and others.
+ * Copyright (c) 2010, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,7 +13,7 @@
  *******************************************************************************/
 package org.eclipse.pde.core.project;
 
-import org.eclipse.osgi.service.resolver.VersionRange;
+import org.osgi.framework.VersionRange;
 
 /**
  * Describes a fragment host. Instances of this class can be created
@@ -30,14 +30,22 @@ public interface IHostDescription {
 	 *
 	 * @return symbolic name of the host
 	 */
-	public String getName();
+	String getName();
 
 	/**
 	 * Returns the version constraint of the host or <code>null</code>
 	 * if unspecified.
 	 *
 	 * @return version constraint or <code>null</code>
+	 * @since 3.19
 	 */
-	public VersionRange getVersionRange();
+	VersionRange getVersion();
+
+	/** @deprecated Instead use {@link #getVersion()} */
+	@Deprecated(forRemoval = true, since = "4.19")
+	default org.eclipse.osgi.service.resolver.VersionRange getVersionRange() {
+		VersionRange version = getVersion();
+		return version != null ? new org.eclipse.osgi.service.resolver.VersionRange(version.toString()) : null;
+	}
 
 }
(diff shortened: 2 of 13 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/core/project/IHostDescription.java
```diff
@@ -30,14 +30,25 @@ public interface IHostDescription {
...
 	/**
 	 * Returns the version constraint of the host or <code>null</code>
 	 * if unspecified.
 	 *
 	 * @return version constraint or <code>null</code>
+	 * @since 3.19
 	 */
-	public VersionRange getVersionRange();
+	VersionRange getVersion();
+	// TODO: alternatively name it getHostVersionRange() ?
```
HannesWell, line 43: Since only the return type changes from `org.eclipse.osgi.service.resolver.VersionRange` to `org.osgi.framework.VersionRange` to we have to find a new name for the method.
The OSGi spec only talks about the _version_ of a requirement (e.g. host, package or bundle) although technically it can be a version range. With that in mind I found naming the new method just `getVersion()` a good choice, especially since there is no need to reflect the exact return type in a method name. The method name should reflect the semantic of a method.
Alternatively we could name it for example `getHostVersionRange()`. But not only would this 'duplicate' the return type information from the method signature but also the information of the containing class.

The same applies for `IPackageImportDescription.getVersion()` and `IRequiredBundleDescription.getVersion()`.

HannesWell (commented):
@laeubi or @merks or anyone else, are you interested in reviewing this?

org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.doc.user; singleton:=true
-Bundle-Version: 3.15.300.qualifier
+Bundle-Version: 3.15.400.qualifier
```
HannesWell, line 5: Btw. does anybody know the policy when to bump the minor and when to bump the micro/service version of a doc bundle? Because it is usually not used from code (isn't it?) that bundle itself does not really have an API and versioning according to that not really possible, I think?

merks: At one point in the history is was 3.14.2200.qualifier suggesting 22 +100 increments.  I doubt there is ever a compelling reason to increment the minor version.

HannesWell: That's right. Just incrementing the service version forever does not look suitable.
We could just increment the minor version when new API is added or some is removed after the deprecation period. But this could mean that the minor version also raises relatively quickly since the doc is some kind of an aggregator. Again on the other hand it still wouldn't be higher than the aggregator features since they are incremented every time.

laeubi: One could think about just increment the minor version on each release and therefore reset the micro version to zero.

ui/org.eclipse.pde.core/src/org/eclipse/pde/core/project/IHostDescription.java
laeubi: What about `getRange()`? or we can folow the record semantics and name it `versionRange()` even though it is a bit unusual at the first place, I found this often convenient to have an interface and use a record as the implementation, then these interface methods are automatically implemented.

HannesWell: For me `getRange()` is not more clear than `getVersion()`.
The point about records is true and actually I like using records but since the implementations are not records we don't have an immediate benefit from that and would would just be a single method with a different style.
But I still find this idea interesting and created https://github.com/eclipse-pde/eclipse.pde/pull/1341 to explore it further.

In the meantime my suggestion is to continue with `getVersion()`. Since we have enough time to complete that in this release cycle we can immediately replace that method e.g. by `version()` or `versionRange()` in https://github.com/eclipse-pde/eclipse.pde/pull/1341.

org.eclipse.pde.doc.user/META-INF/MANIFEST.MF
HannesWell: That would be possible, but for this change I think it is sufficient if we agree to bump the minor version.

ruspl-afed: What if we align the version of PDE Doc bundle with the version of `org.eclipse.pde` feature? The argument could be: PDE Doc bundle content should be in sync with a content of main PDE feature.

HannesWell: Fair point.
Although this means that I had to bump the minor version of the o.e.pde feature.
I assumed the p2 version checker would suggest to bump a feature's minor version if a plugin's minor has changed. 🤔
But maybe I confused this.

HannesWell:
As a follow up I have created https://github.com/eclipse-pde/eclipse.pde/pull/1343 to replace the methods deprecated here in the PDE code base.

If objections are raised I plan to submit this tomorrow Tuesday (European) evening.

HannesWell:
Thanks to everyone involved.
The new deprecation  warnings will be fixed in https://github.com/eclipse-pde/eclipse.pde/pull/1343

PR #1345, opened 2024-07-21
Implement pde.project description interfaces using records

HannesWell:
The implementations were already immutable and using records avoids implementing some boilerplate code.

And remove duplicated java-doc in BundleProjectService identically copied from the implemented interface.

This is the internal part of https://github.com/eclipse-pde/eclipse.pde/pull/1341.


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/BundleProjectService.java
```diff
@@ -89,110 +89,94 @@ public static synchronized IBundleProjectService getDefault() {
...
+		}
+
+		@Override
+		public Version getVersion() {
+			return version();
+		}
+
+		@Override
+		public String[] getFriends() {
+			return friends.isEmpty() ? null : friends.toArray(String[]::new);
```
ruspl-afed, line 153: Is it possible to avoid returning `null` for array?

merks: That does seem better.

HannesWell: > Is it possible to avoid returning `null` for array?

Unfortunately not really because the doc of `IPackageExportDescription.getFriends()` states:
```
Returns the declared friends of this package or <code>null</code> if none.
```

But in the API rework in https://github.com/eclipse-pde/eclipse.pde/pull/1341, the replacement of that method returns an empty `(Sorted)Set` if an exported package has no friends.

In general this change is intended as clean-up without behavioral change and to keep the change in #1341 smaller. But I'm happy to discuss any enhancements of this API there.

HannesWell: I'm submitting this now, lets continue any API discussion in #1341.

PR #1349, opened 2024-07-25
Move api.tools.annotations to Java-1.8

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2197
Caused by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551


HannesWell:
There are also many test bundles that have a BREE below 1.8. Lets see if they have to be bumped too.

HannesWell:
For reference: The target was lowered in https://github.com/eclipse-pde/eclipse.pde/pull/913 to allow using these annotations (that are usually provided by the Eclipse runtime used for development IIRC) in applications targeting older java (and Eclipse releases).

iloveeclipse:
> There are also many test bundles that have a BREE below 1.8. Lets see if they have to be bumped too.

If the projects are compiled in Eclipse during tests, yes.

HannesWell:
> > There are also many test bundles that have a BREE below 1.8. Lets see if they have to be bumped too.
> 
> If the projects are compiled in Eclipse during tests, yes.

Yes. But I'm reluctant to just bump everything because I don't know what is verified in tests as well.

laeubi:
As far as I have understand the compiler will simply no longer support these so the build will fail anyways unless we configure an older compiler in the build.

HannesWell:
> As far as I have understand the compiler will simply no longer support these so the build will fail anyways unless we configure an older compiler in the build.

It depends. Maybe these test-projects don't contain sources, I haven't checked it yet.
But my goal is to make the build at at least compile again, test-failures can still be fixed later.
I don't have that much time this morning to do a deep check what's necessary.

But Jenkins and the GH workflow for Windows already succeeded, GH-workflow for Linux has one test-failure but that looks unrelated.
That's sufficient to submit it now even tough the Mac-workflow is still running.

laeubi:
@HannesWell as long as no new compiler is deployed (and the aggregator pom updated) still the old compiler is used in verification builds.

HannesWell:
> @HannesWell as long as no new compiler is deployed (and the aggregator pom updated) still the old compiler is used in verification builds.

Yes that could definitively reveal more older targets. But it looks like there are still test resources that have no sources, too. E.g. there are numerous resources for the api.tools tests that all have a `J2SE-1.4` BREE and only compares an `.api_description` file.

PR #1350, opened 2024-07-25
Move API-Tools test resources to JavaSE-1.8

HannesWell:
ECJ does not support Java versions below 1.8 anymore

Contributes to fixing
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2197

Caused by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551


HannesWell:
Looks like all API-Tools failures are fixed and only pde.build.tests failures remain.
Will handle the latter in another PR.

PR #1352, opened 2024-07-26
[PDE-Build] Use File to represent file-system paths instead of String

HannesWell:
And don't name variables that denote local files 'urls'.

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/AbstractScriptGenerator.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/AbstractScriptGenerator.java
@@ -35,6 +35,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/AbstractScriptGenerator.java
```diff
@@ -84,7 +85,7 @@ public abstract class AbstractScriptGenerator implements IXMLConstants, IPDEBuil
 	private static PDEUIStateWrapper pdeUIState;
 
 	/** Location of the plug-ins and fragments. */
-	protected String[] sitePaths;
+	protected List<File> sitePaths;
```
laeubi, line 88: Even though its already merged, why not even `use` Path instead of `File` ?

HannesWell: I thought about that as well, but it would have required more changes. Since my main goal was to get rid of the File-to-URL and URL-to-File conversion I didn't want to do that as well in this change.
But since this is all internal we can change to Path at our own discretion at any time.

PR #1354, opened 2024-07-26
[PDE-Build] Use Path instead of URL to represent local files

HannesWell:
This avoids unnecessary conversations that also uses deprecated methods like File.toURL()

### DIFF
--- a/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/BuildTimeFeature.java
+++ b/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/BuildTimeFeature.java
@@ -14,12 +14,9 @@
 package org.eclipse.pde.internal.build.site;
 
 import java.io.File;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
+import java.nio.file.Path;
 import java.util.ArrayList;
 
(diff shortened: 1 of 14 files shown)


build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/BuildTimeFeature.java
```diff
@@ -156,31 +153,28 @@ public BuildTimeSiteContentProvider getFeatureContentProvider() {
...
-				rootLocation = location.getPath();
 			}
-			int i = rootLocation.lastIndexOf(Constants.FEATURE_FILENAME_DESCRIPTOR);
-			if (i != -1)
-				rootLocation = rootLocation.substring(0, i);
+			location = location.toAbsolutePath();
+			if (location.endsWith(Constants.FEATURE_FILENAME_DESCRIPTOR)) {
+				location = location.getParent();
+			}
+			rootLocation = location.toString() + File.separator;
```
HannesWell, line 177: This is another reason why using Strings to model local file-system paths is a really bad idea, besides the forward-/backwards-slash topic: One always has to pay attention if trailing separators are present or not...

But since that method is called in a lot of places, I'll leave that for a follow-up PR.

PR #1359, opened 2024-07-30
[API-Tools] Use JDT's JRT-FileSystem cache

HannesWell:
@merks you implemented to method in https://github.com/eclipse-pde/eclipse.pde/pull/357, do you have any objection?

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ArchiveApiTypeContainer.java
@@ -14,21 +14,16 @@
 package org.eclipse.pde.api.tools.internal.model;
 
 import java.io.IOException;
-import java.net.URI;
 import java.nio.file.FileSystem;
-import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


merks:
If using internals isn't a problem then it's fine.   (Of course that's why internals are not really so internal that one can assume changing them breaks no one.)

HannesWell:
> If using internals isn't a problem then it's fine.   (Of course that's why internals are not really so internal that one can assume changing them breaks no one.)

We used `JRTUtil` in that meethod already before this change. 
And yes it's not ideal to use internals but I hope we are friends with JDT even though the api.tools plugin is not mentioned as such the manifest of the providing plugin.
And even if PDE is not notified in advance about a breaking change we will notice it latest in the next nightly I-Build.
Sharing the cache is IMO worth it.

merks:
Note that I think it shares the cache either way because that sharing is what the JRE is providing.

HannesWell:
> Note that I think it shares the cache either way because that sharing is what the JRE is providing.

I debuged through the creation of the JRT file-system with a JDK-17 and `jdk.internal.jimage.ImageReader` is indeed cached within the JDK already. I guess that's the heaviest thing? But still there are some more objects created without it. And as long as JDT also has a cache for the JRT-filesystems I think it is reasonable to reuse it instead of having our own. If caching not necessary at all we should remove it here and in JDT.

Then lets have this. Thank you for your explanations Ed.

Issue #1360, opened 2024-07-30
JUnit5SuiteExecutionTest fails with "Resource '/verification.tests.junit5' already exists." on Linux


HannesWell:
At the moment, only on Linux the initialization of the test-class `org.eclipse.pde.junit.runtime.tests.JUnit5SuiteExecutionTest` persistently fails with the following error:
```
Resource '/verification.tests.junit5' already exists.
org.eclipse.core.internal.resources.ResourceException(/verification.tests.junit5)[374]: java.lang.Exception: Resource '/verification.tests.junit5' already exists.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:343)
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:330)
	at org.eclipse.core.internal.resources.Project.assertCreateRequirements(Project.java:94)
	at org.eclipse.core.internal.resources.Project.create(Project.java:305)
	at org.eclipse.core.internal.resources.Project.create(Project.java:291)
	at org.eclipse.pde.ui.tests.util.ProjectUtils.importTestProject(ProjectUtils.java:180)
	at org.eclipse.pde.junit.runtime.tests.JUnitExecutionTest.setupProjects(JUnitExecutionTest.java:73)
	at org.eclipse.pde.junit.runtime.tests.JUnit5SuiteExecutionTest.setupProjects(JUnit5SuiteExecutionTest.java:40)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.RunBefores.invokeMethod(RunBefores.java:33)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
[...]

PR #1361, opened 2024-07-30
Prevent multiple execution of junit.runtime.tests and always clear workspace

HannesWell:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1360


HannesWell:
Looks like JUnit5SuiteExecutionTest also called `JUnitExecutionTest.setupProjects()` but didn't have the `CLEAR_WORKSPACE` TestRule specified. Furthermore both the `JUnit5SuiteExecutionTest` and the `JUnitExecutionTest` were executed twice in verification builds because `JUnitRuntimeTests` suite defined in the package was not declared as `testClass` in the `pom.xml` (only referenced in the `test.xml` for I-Builds)

HannesWell:
Great all tests pass again now.
As expected, the number of test execution has been reduced but only because the duplicated execution of  `JUnitExecutionTest` and `JUnit5SuiteExecutionTest` is now avoided. I verified that both are now executed exactly once.

PR #1363, opened 2024-08-01
Fix most warnings in e4-tools

HannesWell:


HannesWell:
Fixed issues 162 (mainly because resource warnings are not suppressed in the Maven compiler settings anymore), 120 warnings remaining.

Issue #1365, opened 2024-08-06
Unable to open Advanced Search Dialog in Bundle Repository view more than once


ptziegler:
To reproduce from a clean workspace:

- Open the Bundle Repositories view
- Open the Advanced Search Dialog
- Click through the individual tabs
- Close the Advanced Search Dialog
- Try to open the Advanced Search Dialog again

_Expectation:_

The dialog should open again, just like it did the first time.

_Actual Behavior:_

The dialog fails to open with the following error, caused by the image set for the tab item already being disposed:

```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4903)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:500)
	at org.eclipse.swt.widgets.Item.setImage(Item.java:180)
	at org.eclipse.swt.widgets.TabItem.setImage(TabItem.java:307)
	at org.eclipse.pde.bnd.ui.views.repository.AdvancedSearchDialog.createDialogArea(AdvancedSearchDialog.java:95)
	at org.eclipse.jface.dialogs.TitleAreaDialog.createContents(TitleAreaDialog.java:166)
	at org.eclipse.jface.window.Window.create(Window.java:431)
	at org.eclipse.jface.dialogs.Dialog.create(Dialog.java:1092)
	at org.eclipse.jface.window.Window.open(Window.java:788)
	at org.eclipse.pde.bnd.ui.views.repository.RepositoriesView$9.run(RepositoriesView.java:700)
	at org.eclipse.jface.action.Action.runWithEvent(Action.java:474)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:581)
	at org.eclipse.jface.action.ActionContributionItem.lambda$5(ActionContributionItem.java:454)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4326)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4124)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3712)
[...]

PR #1367, opened 2024-08-06
Fix: Unable to open Advanced Search Dialog more than once

ptziegler:
When opening this dialog, the tab items are decorated with images created by the individual search panels. When the dialog is closed, those images are disposed again.

Problem arises because the search panels don't create new images, but instead use shared images. Meaning when the dialog is opened a second time, those images have already been disposed, preventing the dialog from showing up.

Reason for this regression is the switch from the Icons class of the bndtools project to the PDE Resources class. Former has an explicit "isDisposed()" check, which would've created a new image on-the-fly, while latter doesn't. Given that the images are not owned by this dialog, the solution to this problem is to simply not have them be disposed by the dialog.

See https://github.com/eclipse-pde/eclipse.pde/issues/1365

### DIFF
--- a/ui/org.eclipse.pde.bnd.ui/src/org/eclipse/pde/bnd/ui/views/repository/AdvancedSearchDialog.java
+++ b/ui/org.eclipse.pde.bnd.ui/src/org/eclipse/pde/bnd/ui/views/repository/AdvancedSearchDialog.java
@@ -17,8 +17,6 @@ package org.eclipse.pde.bnd.ui.views.repository;
 
 import java.beans.PropertyChangeListener;
 import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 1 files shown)


ptziegler:
This is the [original code](https://github.com/bndtools/bnd/blob/035bab9ea70ab28a7d16cf08b89bc0507b5852b5/bndtools.core/src/org/bndtools/core/ui/icons/Icons.java#L163-L191) that was used to create those images. The "isDisposed()" check is what prevents this problem from showing up with the bndtools.:

```java
public static Image image(String name, String... decorators) {
	synchronized (images) {
		Key k = new Key(name, decorators);
		Image image = images.get(k);
		if (image != null && !image.isDisposed())
			return image;


		if (decorators.length == 0) {
			ImageDescriptor desc = desc(name, false);
			if (desc == null) {
				desc = desc(ICONS_MISSING_GIF);
			}
			Image baseImage = desc.createImage();
			images.put(k, baseImage);
			return baseImage;
		}


		Image baseImage = image(name);


		ImageDescriptor overlays[] = new ImageDescriptor[decorators.length];
		for (int i = 0; i < decorators.length; i++) {
			overlays[i] = desc(decorators[i], true);
		}
		DecorationOverlayIcon decoratedImage = new DecorationOverlayIcon(baseImage, overlays);
		Image result = decoratedImage.createImage();
		images.put(k, result);
		return result;
	}
}
```

HannesWell (approved):
That's a good catch and nice investigation. 👍🏽

Thank you.

PR #1373, opened 2024-08-09
Adjust lower bound for dependency to JDT Launching in PDE Build

ptziegler:
The PDEState class is using JavaRuntime.getProvidedVMPackages(...), which has only been added with version 3.22. But the Manifest still only requires 3.21. This causes issues when installing the latest PDE in the Eclipse 4.31, as the JDT bundles are not updated, which therefore leads to a NoSuchMethodError.

See https://github.com/eclipse-windowbuilder/windowbuilder/issues/865 for a real-life example of this problem.

Amends 00c9caf09d4a5f5be92cbb3e628ef64fa73440f6


merks:
You really deserve a prize.  🥇

Issue #1381, opened 2024-08-28
CSS Spy: NPE in WidgetTreeProvider


jukzi:
eclipse.buildId=4.33.0.I20240827-1800

```
org.eclipse.e4.core.di.InjectionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.e4.ui.css.core.dom.CSSStylableElement.getNativeWidget()" because the return value of "org.w3c.dom.NodeList.item(int)" is null
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:66)
	at org.eclipse.e4.core.internal.contexts.ContextObjectSupplier$ContextInjectionListener.update(ContextObjectSupplier.java:95)
	at org.eclipse.e4.core.internal.contexts.TrackableComputationExt.update(TrackableComputationExt.java:103)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.processScheduled(EclipseContext.java:358)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.set(EclipseContext.java:374)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.activate(EclipseContext.java:661)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.activateBranch(EclipseContext.java:670)
	at org.eclipse.e4.ui.internal.workbench.PartActivationHistory.activate(PartActivationHistory.java:60)
	at org.eclipse.e4.ui.internal.workbench.PartServiceImpl.activate(PartServiceImpl.java:767)
	at org.eclipse.e4.ui.internal.workbench.PartServiceImpl.activate(PartServiceImpl.java:686)
	at org.eclipse.e4.ui.internal.workbench.swt.AbstractPartRenderer.activate(AbstractPartRenderer.java:95)
	at org.eclipse.e4.ui.workbench.renderers.swt.ContributedPartRenderer.lambda$0(ContributedPartRenderer.java:63)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4326)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1198)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1183)
[...]

PR #1382, opened 2024-08-28
CSS Spy: fix NPE in WidgetTreeProvider #1381

jukzi:
for items that can not be styled with CSS

https://github.com/eclipse-pde/eclipse.pde/issues/1381

### DIFF
--- a/features/org.eclipse.pde.spies-feature/feature.xml
+++ b/features/org.eclipse.pde.spies-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.spies"
       label="%featureName"
-      version="1.0.400.qualifier"
+      version="1.0.500.qualifier"
       provider-name="%providerName"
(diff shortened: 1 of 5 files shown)


HannesWell (approved):
Thank you for this fix.
I could reproduce it and verify that this fixes the issue.

But currently the build fails because the `org.eclipse.pde.spies` feature also needs a version bump and the Jenkins builds also complains about differences in the org.eclipse.pde.doc.user bundle (not sure why, but if it persists after the feature version is bumped, I suggest to force a qualifier update there).

Although the change is fine and looks save I think we are too late for the upcoming release because the last scheduled I-build is about to happen tonight and I don't think this is worth the work of a re-spin:
https://calendar.google.com/calendar/u/0/embed?src=prfk26fdmpru1mptlb06p0jh4s@group.calendar.google.com

Or do you think is is super important?

jukzi:
this is not urgent

Issue #1386, opened 2024-09-04
API error on SWT on a 4.34 master


iloveeclipse:
I see API error on **SWT** if using https://download.eclipse.org/eclipse/downloads/drops4/S-4.33RC2a-202409030240/ as API baseline and https://download.eclipse.org/eclipse/downloads/drops4/I20240904-0530/ as IDE.

SWT error is about changed execution environment.

I've rebuilt everything few times, no changes, errors do not disappear.
![image](https://github.com/user-attachments/assets/ebe52849-ed3d-44c7-8102-763973e8e40c)

Note: originally reported API errors for JDT were gone after switching default workspace JDK from 21 to 17 and back, only SWT error remains.

iloveeclipse:
Previous SWT related bug: https://github.com/eclipse-pde/eclipse.pde/issues/1283

merks:
I felt like this was a ground hog day issue that was fixed but now its back.  😱   That's what happens when there is no test. 😭

iloveeclipse:
Interestingly, I can't reproduce JDT errors after switching default workspace JDK to Java 17 version (I had 21 before and tried to change that because I've assued that would "mute" SWT errors) and back. But instead, JDT errors disappeared but SWT still has errors ???

iloveeclipse:
I see in debugger in `org.eclipse.pde.api.tools.internal.model.BundleComponent.init()`, that the binary fragment has empty array as `fdeclaredRequiredEE`, but workspace one gets `JavaSE-17` (either current EE of the workspace or the one from host). So we have a mismatch... And yes, that looks like regression from https://github.com/eclipse-pde/eclipse.pde/pull/1302.

merks:
FYI, I see both errors in my workspace:

![image](https://github.com/user-attachments/assets/a2f13e86-e58a-4416-9b18-8cc328b24500)

iloveeclipse:
Which JDK is set as default in the workspace? Could you try to change that?0

merks:
I have 17 as the default:

![image](https://github.com/user-attachments/assets/9a3294ea-f682-439a-95d6-ae9c076b06bd)

iloveeclipse:
> have 17 as the default:

Could you just try to change it to 21, restart & rebuild JDT? That helped me, and after that I was able to switch back and still no errors.

iloveeclipse:
[...]

PR #1387, opened 2024-09-04
Fix for API error on SWT on 4.34 master

iloveeclipse:
Don't try to add EE to bundles that don't have any requirements defined in manifest.

ManifestUtils.getRequiredExecutionEnvironments() currently returns "generic" bundle runtime requirements from OSGI for bundles that don't define anything, and it seem to be violation of the org.osgi.framework.wiring.BundleRevision.getDeclaredRequirements(String) contract that isn't supposed to return anything if nothing is specified in the manifest.

At least for API baseline tooling use of OSGI generic requirements not specified in the manifest do not make sense, as we compare what's written in the manifest and not what is derived from the platform or host.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1386


HannesWell:
Obsolete with https://github.com/eclipse-pde/eclipse.pde/pull/1395, thank you nevertheless!

PR #1388, opened 2024-09-04
Create test case for EE baseline warnings due to missing Manifest header

ptziegler:
This checks whether no error is produced when both the Require-Capability and the Bundle-RequiredExecutionEvironment headers are missing from the manifest.

See https://github.com/eclipse-pde/eclipse.pde/issues/1386


HannesWell:
Thank you for providing this test case!
I have incorporate your commit in https://github.com/eclipse-pde/eclipse.pde/pull/1388.
Therefore I'm closing this one.

PR #1390, opened 2024-09-05
IApiAnnotations.isExact for CompositeApiDescription.resolveAnnotations

merks:
- CompositeApiDescription.resolveAnnotations should be able to properly find the best match annotation using IApiAnnotations.isExact which is now also implemented by TypeAnnotations.isExact.

https://github.com/eclipse-pde/eclipse.pde/issues/1386

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.api.tools;singleton:=true
-Bundle-Version: 1.3.500.qualifier
+Bundle-Version: 1.3.600.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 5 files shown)


iloveeclipse:
I've just rebased on master to have the fix for API filter warning

merks:
I see.  The test were all green?

iloveeclipse:
> I see. The test were all green?

Yes, I just dislike merging PR's with partially red state.

iloveeclipse:
I've tried the patch and I can confirm it solves the jdt.core problem.

So if I understood it correct, the regression was in 40cf1f0bee0a50a9747436949b3a3be2905920c8 as the code checked for the annotation to be `instanceof ApiAnnotations` and of course `TypeAnnotations` weren't but should be (from very beginning probably).

iloveeclipse:
**Many** thanks Ed!

HannesWell:
Thank you Ed for this.

> So if I understood it correct, the regression was in [40cf1f0](https://github.com/eclipse-pde/eclipse.pde/commit/40cf1f0bee0a50a9747436949b3a3be2905920c8) as the code checked for the annotation to be `instanceof ApiAnnotations` and of course `TypeAnnotations` weren't but should be (from very beginning probably).

I wonder how this can be unnoticed for four years? :O

Issue #1392, opened 2024-09-05
org.eclipse.ua.tests.doc failure since I20240904-0530


iloveeclipse:
See

* https://download.eclipse.org/eclipse/downloads/drops4/I20240904-0530/testResults.php
* https://download.eclipse.org/eclipse/downloads/drops4/I20240905-0340/testresults/html/org.eclipse.ua.tests.doc_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

```
TocLinkChecker	testPdeUserGenerated()	Failure	Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/overview-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/api/tools/annotations/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/build/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/core/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/core/build/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/core/plugin/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/core/project/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/core/target/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/launching/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/ui/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/ui/launcher/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/ui/target/package-summary.html Invalid link in "/org.eclipse.pde.doc.user/topics_Reference.xml": reference/api/org/eclipse/pde/ui/templates/package-summary.html

[...]

PR #1393, opened 2024-09-05
fix recent javadoc "reference not found"

jukzi:
https://github.com/eclipse-pde/eclipse.pde/issues/1392


HannesWell (commented):
Can you please move the version bumps into a separate commit as it is our common practice with the usual message `Bump version(s) for 4.34 stream` so that version bumps are easier to discover in the history and are not reverted in case the actual change is reverted (unlikely in this case).

HannesWell:
I have included the commit of this PR into https://github.com/eclipse-pde/eclipse.pde/pull/1400 and applied the requested changes there.
Thanks for starting this initiative.

PR #1396, opened 2024-09-08
Return List<String> in IApiComponent.getExecutionEnvironments()

HannesWell:


jukzi:
this commit made all tycho builds fail: https://github.com/eclipse-tycho/tycho/issues/4221

HeikoKlare:
Also makes SWT tests fail, e.g., https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/10770510714/job/29864024982?pr=1456

laeubi:
I suggest not to use ibuilds for the API tools check:

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2327

HannesWell:
> I suggest not to use ibuilds for the API tools check:
> 
>     * [Don't use API-Tools from I-Builds for API checks in verification builds eclipse-platform/eclipse.platform.releng.aggregator#2327](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2327)

I agree.
To add some more context: The changed method is from a 'provisional' API of API-Tools that is still internal. But because there is no stable public API yet Tycho has to use that for the API-Check mojo.
This is per-se not a big issue besides that in case something is changed the Tycho mojo has to be adapted but since the same people are involved in both projects the knowledge is usually there.
The big problem is that the mojo is/was configured to use the latest API-tools jars from the I-builds. And that introduced a very tight dependency cycle without room to react to changes.
So I think the change linked above is the right solution to this problem.

PR #1191, opened 2024-03-09
Workaround for missing 4.31 SWT API descriptions

iloveeclipse:
Added SWT API description files from last good 4.31 build I20240207-0610 https://download.eclipse.org/eclipse/updates/4.31-I-builds/I20240207-0610/plugins/

The files will be loaded instead of "empty" SWT API description files shipped with 4.31 RC2 build (3.125.0.v20240227-1638).

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1187

PR #1397, opened 2024-09-08
Revert "Workaround for missing 4.31 SWT API descriptions"

HannesWell:
This reverts commit 2fd4268ab73151b8559fbd6b9c98dc97e014a5db and commit a6824f75205aca3e2fe3c802a406be4523c68830, introduced with https://github.com/eclipse-pde/eclipse.pde/pull/1191 and https://github.com/eclipse-pde/eclipse.pde/pull/1192

These workarounds for missing API description files are not necessary anymore because they are now correctly included again for some time.

@iloveeclipse or do you see a reason why this should be kept?


HannesWell:
@iloveeclipse any remarks from you?
If there is no objections I'll submit this latest tomorrow (Tuesday) evening.

iloveeclipse:
I had no time to review that in the IDE, and tomorrow I will be in meetings whole day, so I can't do a proper review.

Beside this, the workaround was for API checks in 4.33 IDE / 4.32 target only, which shouldn't be interested on master branch. Whoever develops on 4.32 maintenance branches wuld need to stay with 4.33 IDE.

HannesWell:
Ok, no need to rush from my side. Just want this to be forgotten.

> Whoever develops on 4.32 maintenance branches wuld need to stay with 4.33 IDE.

That's right, but is there a lot of maintenance work happening? Currently the `R4_32_maintenance` branch is still in its initial state after the release and AFAICT usually if at all only selected changes are backported. In that case I would say it's acceptable that the API tools are not available then.

iloveeclipse:
> is there a lot of maintenance work happening?

I don't know that and I personally don't care simply because a workaround is possible. I guess some bigger companies might have in house maintenance branches, but if they use API tooling I don't know. At least at Advantest we don't care about API checks on maintenance branch we have.

HannesWell:
> > is there a lot of maintenance work happening?
> 
> I don't know that and I personally don't care simply because a workaround is possible. I guess some bigger companies might have in house maintenance branches, but if they use API tooling I don't know. At least at Advantest we don't care about API checks on maintenance branch we have.

The only company I know of at all that cares about maintenance is IBM and AFAIK they play all maintenance work back to the official eclipse repos and only do back-ports of bug fixes or critical dependency updates.

So if I understand you correctly you are fine with this, aren't you?

iloveeclipse:
Yes

PR #1398, opened 2024-09-08
Fix more warnings in pde.ui and pde.ds.ui

HannesWell:
And apply minor clean-ups in ExtensionsSection.

The branch in `DSFileWizardPage.getProject(Object)` that checks if the passed object is a `ClassPathContainer` seems to be possible anymore.
In general creating OSGI Service Components through the UI is a quite uncommon nowdays since one can simply use the Decalarative service component annotations.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitTabGroup.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitTabGroup.java
@@ -48,9 +48,11 @@ public class JUnitTabGroup extends AbstractPDELaunchConfigurationTabGroup {
 		} catch (CoreException e) {
 			vmArgs = ""; //$NON-NLS-1$
 		}
-		vmArgs = AssertionVMArg.enableAssertInArgString(vmArgs);
-		if (vmArgs.length() > 0)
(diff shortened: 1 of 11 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitTabGroup.java
```diff
@@ -48,9 +48,11 @@ public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
 		} catch (CoreException e) {
 			vmArgs = ""; //$NON-NLS-1$
 		}
-		vmArgs = AssertionVMArg.enableAssertInArgString(vmArgs);
-		if (vmArgs.length() > 0)
-			configuration.setAttribute(IJavaLaunchConfigurationConstants.ATTR_VM_ARGUMENTS, vmArgs);
+		@SuppressWarnings("restriction")
```
HannesWell, line 51: This entire `src_samples` source folder somehow seems to be related to something like `sample.properties` files.
Does anybody know anything about that?

It also seems to be associated to a `org.eclipse.sdk.samples` provided via the site
http://www.eclipse.org/pde/samples/site.xml

But I have never heard of this and assume it's outdated if it is usable at all?

@merks, @iloveeclipse or @akurtakov do you know anything about that or can this entire source folder be removed?

PR #1394, opened 2024-09-05
fix some javadoc

jukzi:

PR #1400, opened 2024-09-08
Fail on javadoc errors and fix remaining javadoc errors

HannesWell:
Fix all javadoc errors except for the packages `org.eclipse.pde.internal.build.publisher` and `org.eclipse.pde.internal.core.ant` which contain only ANT task sources and therefore require special class-path configuration that I wasn't able to properly replicate for the javadoc build.

To avoid javadoc errors in the future also fail the build on javadoc errors.

Should fix https://github.com/eclipse-pde/eclipse.pde/issues/1392.

This also contains the commit/changes from the following PRs with the requested changes applied:
- https://github.com/eclipse-pde/eclipse.pde/pull/1394
- https://github.com/eclipse-pde/eclipse.pde/pull/1393

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/refactoring/FilterChange.java
+++ b/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/refactoring/FilterChange.java
@@ -96,26 +96,6 @@ public abstract class FilterChange extends Change {
 		return NLS.bind(RefactoringMessages.FilterChange_remove_used_filter, problem.getMessage());
 	}
 
-	/**
-	 * Returns the name to use for a {@link #RENAME} change operation
-	 *
-	 * @return the name for a {@link #RENAME} change
-	 */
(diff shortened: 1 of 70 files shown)


apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/refactoring/FilterChange.java
```diff
@@ -96,26 +96,6 @@ protected String getDeleteName() {
...
-	/**
-	 * Returns the name to use for a {@link #MOVE} change operation
-	 *
-	 * @return the name for a {@link #MOVE} change
-	 */
-	protected String getMoveName() {
-		IApiProblem problem = this.filter.getUnderlyingProblem();
-		return NLS.bind(RefactoringMessages.FilterChange_remove_used_filter, new Object[] { problem.getMessage() });
-	}
-
```
HannesWell, line 118: These methods return the very same result as `getDeleteName()` and are never called.

PR #1402, opened 2024-09-09
Adapt to changed output directory in maven-javadoc-plugin 3.10.0

HannesWell:
Probably as part of https://github.com/apache/maven-javadoc-plugin/pull/204, the default output directory of the `maven-javadoc-plugin` changed when updating the `maven-javadoc-plugin` from `3.8.0` to `3.10.0`. This adapts the doc-bundle build accordingly.

In a local test the missing `reference/api` folder was generated in the project as expected, but the built jar still didn't contain it. So maybe more adaption is necessary.

Hopefully fixes https://github.com/eclipse-pde/eclipse.pde/issues/1392


akurtakov:
Build artifacts look good.

laeubi:
Something similar might be required for other doc bundles as well!

HannesWell:
> Something similar might be required for other doc bundles as well!

I thought that as well, but at least the tests for the other doc bundles didn't report a problem. Not sure if this is false positive or if PDE is different.

PR #1408, opened 2024-09-16
Set up workflow for automated version increments in pull-requests 

HannesWell:
Set up the workflow for automated version increments in pull-requests, added via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352, for PDE.

Before this can fully work, the EF infra team has to add the required PAT of the PDE-bot to this repo as desribed in the linked PR.

PR #1413, opened 2024-09-24
[JUST TO TEST] Some first change in a release-cycle

HannesWell:
This just exists to test the workflow added via https://github.com/eclipse-pde/eclipse.pde/pull/1408.

THIS IS NOT INTENDED TO BE SUBMITTED!


HannesWell:
Looks like this works well!

@HeikoKlare if you are interested to test this, can you please create an own PR in PDE that contains the same two dummy changes from my first commit?
Then we can verify that this works for non-committers too.

HannesWell:
Test was successful, closing.

PR #1414, opened 2024-09-24
[JUST TO TEST] Some first change in a release-cycle by non-committer

HeikoKlare:
This just exists to test the workflow added via https://github.com/eclipse-pde/eclipse.pde/pull/1408 by some non-committer contributor.

Complements #1413

THIS IS NOT INTENDED TO BE SUBMITTED!


HannesWell:
Great. This works well too. 🎉 
Thanks for testing!

HannesWell:
Since this test was successful we can close this.
Thanks again Heiko!

PR #1421, opened 2024-09-28
Fix breakage caused by o.e.p.i.publishing.Utils changes

merks:
https://github.com/eclipse-equinox/p2/pull/550


merks:
@HannesWell @akurtakov @iloveeclipse 

Given the I-build is broken, this, or something like this, needs to be merged sooner rather than later.

iloveeclipse:
Should the bundle dependency be updated too?

merks:
I suppose that's a good idea too, yes.

merks:
@iloveeclipse 

I think I cannot change the bound because the I-build is broken so the new p2 thingy isn't available yet.  So I can either merge this change without verification or the bound needs to change later after a successful I-build. 

What shall it be?

akurtakov:
Merge and start new I-build. That way there is at least a chance to have it done at once.

merks:
Started https://ci.eclipse.org/releng/job/Builds/job/I-build-4.34/48/

HannesWell:
Thank you Ed and sorry for the trouble! One really has to be very careful with P2, even if its internal. :/

> I think I cannot change the bound because the I-build is broken so the new p2 thingy isn't available yet. So I can either merge this change without verification or the bound needs to change later after a successful I-build.

Even if the bound is satisfied the change can't succeed with a green verification, because without an I-build the changed method signatures are not available. So it's a chicken-egg problem not solvable with only green builds.
So I agree that doing it at once is the best: Close your eyes and go (to say it with a badly translated German quote).

Issue #1424, opened 2024-09-30
Compilation errors in `org.eclipse.tools.layout.spy` and `org.eclipse.ui.editors` bundles in the IDE


iloveeclipse:
Not sure whether this is related to the [split package issue](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3027) or not, but using I20240929-1800 I see compilation errors in `org.eclipse.tools.layout.spy` and `org.eclipse.ui.editors` bundles.

The reason is that the compiler can't resolve `org.eclipse.jdt.annotation` anymore.

The only way to "fix" the errors is to add `additional.bundles = org.eclipse.jdt.annotation` line to the `build.properties` for each not compilable bundle.

Without looking into PDE / JDT code I can't understand why that is needed, as both affected bundles require the bundle in question in manifests via `org.eclipse.jdt.annotation;resolution:=optional` and I have the annotation bundle in both the workspace / target platform.

For me it looks like some PDE hard-coded hack, however, I don't understand why did it work last week with build `I20240923-1800` but doesn't work this week with the new `I20240929-1800  SDK version.

I will provide patches for both bundle build files, but it would be better to understand why do we have such change first.

laeubi:
Don't use additional bundles for this, it is a completely different purpose see:
- https://github.com/eclipse-pde/eclipse.pde/issues/1407
- https://github.com/eclipse-pde/eclipse.pde/issues/1403
- https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Fguide%2Ftools%2Feditors%2Fmanifest_editor%2Feditor.htm

iloveeclipse:
> Don't use additional bundles for this

Which solution do you propose?

iloveeclipse:
I mean, which **currently available** solution exists to get an error free workspace?

laeubi:
> I mean, which **currently available** solution exists to get an error free workspace?

`jars.extra.classpath` is available "since ever" ...

iloveeclipse:
Could you be so kind and provide the patch please? If I understand the linked ticket, currently it requires bundle file name, which is not nice as it has to be updated on every change.

laeubi:
`jars.extra.classpath = platform:/plugin/org.eclipse.jdt.annotation` should work

laeubi:
[...]

PR #1425, opened 2024-09-30
Fix compilation error due unresolved org.eclipse.jdt.annotation

iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/issues/1424

### DIFF
--- a/ui/org.eclipse.tools.layout.spy/build.properties
+++ b/ui/org.eclipse.tools.layout.spy/build.properties
@@ -21,3 +21,4 @@ bin.includes = META-INF/,\
                about.html,\
                icons/
 src.includes = about.html
+jars.extra.classpath=platform:/plugin/org.eclipse.jdt.annotation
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.tools.layout.spy/build.properties
```diff
@@ -21,3 +21,4 @@ bin.includes = META-INF/,\
                about.html,\
                icons/
 src.includes = about.html
+additional.bundles = org.eclipse.jdt.annotation
```
laeubi, line 24: This is not correct to handle the situation, see 
- https://github.com/eclipse-pde/eclipse.pde/issues/1424#issuecomment-2382669427
use `jars.extra.classpath` instead!

laeubi (changes requested):

iloveeclipse:
Couldn't reproduce with I20240930-1800 anymore

PR #1428, opened 2024-10-03
[Build] Fix deployment on new Ubuntu images

HannesWell:


HannesWell:
Using the basic agent does not really help because we would have to copy built artifacts between agents to be able to publish them.
So we have to await until this issue is fixed by the EF-infra team: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5061

Issue #1423, opened 2024-09-28
I-build test failure since I20240924-1810: Unable to find plug-in: org.eclipse.equinox.launcher.cocoa.macosx_0.0.0


HannesWell:
Currently multiple I-build tests of PDE fail with a message like
https://download.eclipse.org/eclipse/downloads/drops4/I20240924-1810/testresults/html/org.eclipse.pde.build.tests_ep434I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

This is probably related to https://github.com/eclipse-equinox/equinox/pull/622.

```
The following error occurred while executing this line: /Users/genie.releng/workspace/AutomatedTests/ep434I-unit-macM1-java17/workarea/I20240927-0430/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.pde.build_3.12.600.v20240908-2035/scripts/productBuild/productBuild.xml:72: Unable to find plug-in: org.eclipse.equinox.launcher.cocoa.macosx_0.0.0. Please check the error log for more details.

/Users/genie.releng/workspace/AutomatedTests/ep434I-unit-macM1-java17/workarea/I20240927-0430/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.pde.build_3.12.600.v20240908-2035/scripts/productBuild/productBuild.xml:42: The following error occurred while executing this line:
/Users/genie.releng/workspace/AutomatedTests/ep434I-unit-macM1-java17/workarea/I20240927-0430/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.pde.build_3.12.600.v20240908-2035/scripts/productBuild/productBuild.xml:72: Unable to find plug-in: org.eclipse.equinox.launcher.cocoa.macosx_0.0.0. Please check the error log for more details.
at org.apache.tools.ant.ProjectHelper.addLocationToBuildException(ProjectHelper.java:582)
at org.apache.tools.ant.taskdefs.Ant.execute(Ant.java:440)
at org.apache.tools.ant.taskdefs.CallTarget.execute(CallTarget.java:106)
at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:299)
at jdk.internal.reflect.GeneratedMethodAccessor171.invoke(Unknown Source)
[...]

PR #1429, opened 2024-10-03
MacOS AArch64 launcher is its own fragment now.

akurtakov:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1423


HannesWell (approved):
Thank you for this!
This looks good and I also didn't found more references, so this should fix it.

I have also (again) searched past release repositories and at least down until [Eclipse 4.1](https://download.eclipse.org/eclipse/updates/4.1) the `org.eclipse.equinox.launcher.cocoa.macosx.x86_64` artifact exists. So this should not have a relevant impact if one works with older targets.

akurtakov:
New deprecations cause new warnings. Mergin despite them.

PR #1444, opened 2024-10-16
Remove no longer needed pde.docs releng scripts

akurtakov:
No longer needed since the move to maven.


akurtakov:
Failing windows test is not related to documentation build.

PR #1449, opened 2024-10-18
Various improvements of the TP-Editor's content-assist

HannesWell:
- Make RepositoryCache static instead of singleton
- Consider all repositories of location for proposals
- Completely parallelize P2 metadata fetching


ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
```diff
@@ -31,41 +42,73 @@
...
+		}).map(Map::values).flatMap(Collection::stream).flatMap(List::stream));
+	}
+
+	public static void prefetchP2MetadataOfRepository(String repository) {
+		fetchP2DataOfRepo(repository);
+	}
+
+	private static Future<Map<String, List<IVersionedId>>> fetchP2DataOfRepo(String repository) {
+		return CACHE.compute(repository, (r, f) -> {
+			if (f != null && f.isDone() && !f.isCompletedExceptionally() && !f.isCancelled()) {
```
laeubi, line 89: Job cancellation seem to to result in future cancellation ....

By the way this seems a good use-case for
- https://github.com/eclipse-platform/eclipse.platform/pull/327

ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
```diff
@@ -31,41 +42,73 @@
...
+				return f;
+			}
+			CompletableFuture<Map<String, List<IVersionedId>>> future = new CompletableFuture<>();
+			// Fetching P2 repository information is a costly operation
+			// time-wise. Thus it is done in a job.
+			Job job = Job.create(NLS.bind(Messages.UpdateJob_P2DataFetch, r), m -> {
+				Map<String, List<IVersionedId>> map = toSortedMap(P2Fetcher.fetchAvailableUnits(r));
```
laeubi, line 96: Maybe pass the monitor to P2Fetcher so it could give some progress there...

ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
```diff
@@ -31,41 +42,73 @@
...
+				future.complete(map);
+			});
+			job.setUser(true);
```
laeubi, line 99: I think Jobs should have a group here as there could be many of them.

ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
HannesWell: > Maybe pass the monitor to P2Fetcher so it could give some progress there...

Yes. I have noticed that as well and started to add it locally.

> 
> Beside that one should catch any runtime exceptions here and complete the future exceptionally, also P2 fetcher seem to catch and ignore some exceptions already so a future will never be finished with an error.

Yes the exceptional completion of the Future was indeed missing. Added it. But in this case it does not really make a difference if the job was canceled or another exception occurred. Therefore I just complete exceptionally in case of any exception, regardless of if it's a CancellationException.

`P2Fechter` catches and rethrows some exception, only the exception when creating the agent was swallowed, but lead to a subsequent NPE.
But I have now cleaned all that up to not catch any `CoreException` and just propagate all of them.

> 
> Also if error or cancelation occurs, it might be good to remove the future from the map to not retain it forever in the cache until it is queried next.

Don't think a failed future consumes much memory? And removing it could maybe lead to subtitle concurrency issues.
So I would just keep it simple and clean it up on the next content assist interaction.

ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
HannesWell: > By the way this seems a good use-case for
> 
>     * [Support executing tasks with the Job Framework using CompletableFuture eclipse-platform/eclipse.platform#327](https://github.com/eclipse-platform/eclipse.platform/pull/327)

That would help here, yes.
> 
> 
> Beside that the condition seems to be wrong here. This will result in a new future be created for each call if the previous one is not yet finished. Instead of having one big if, I would separate the cases:
> 
> ```
> if (f != null) {
>    if (!f.isDone()) {
>      //not finished yet...
>      return f;
>    }
>    if(!f.isCompletedExceptionally() && !f.isCancelled()) {
>      //finished without an error
>     return f;
>    }
> }
> ```

Indeed it was wrong, thanks for pointing that out.
But I'm not sure if necessary to break it up so much. I added a combined comment that should help if it's not clear.

ui/org.eclipse.pde.genericeditor.extension/src/org/eclipse/pde/internal/genericeditor/target/extension/model/RepositoryCache.java
HannesWell: I tried that but in the UI I couldn't see a difference and in code it's not needed, since the joint completion is handled via the Future. Therefore I cannot see a real advantage here.

What also should be considered that at maximum all declared repos of a location are queried in parallel.
So I assume the number of parallel jobs not to be too great.

Issue #1218, opened 2024-04-04
[scr 1.4] constructor for injection must be public


laeubi:
If a constructor is used for injection that is package protected Felix SCR complains:

> Constructor with 3 arguments not found. Component will fail.

example is 

```
MyComponent(BundleContext ctx, @Reference Service1 s1, @Reference Service2 s1) {

}
```

This can be seen [in the spec here](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-constructor.injection):

> If the constructor is not public, then the constructor must not be considered. 

PDE should emit an error in this case to the user.

Tilak-21:
I'm working on this.

HannesWell:
Great. You are one of the students from CodeDays, aren't you?

Much success on this task and don't hesitate to ask for help if you need any.

EverettHanke:
Hey there everyone. I am a code day student working with @Tilak-21 on this ticket. After some time searching we have found what we believe is the location to write our solution. AnnotationVisitor.java line 258. 

We were curious if anyone here might be able to confirm that this is the location we need to apply a solution to?

EverettHanke:
Reaching out here to see if we can clear up any misunderstandings about this ticket. I submitted what I believed to be a solution but it didn't fully match the parameters of what this ticket asks of. For me and the team I'm working with this ticket seems very vague and because of that it's been difficult to replicate exactly what we are trying to solve. 

If we can get some more specifics so we can clear up the miscommunication and have a clearer path to work with. Our CodeDay mentorship ended weeks ago but I still want to finish what my group started and see these issues through. CodeDay or no CodeDay. 

What I am misunderstanding can be found [here](https://github.com/eclipse-pde/eclipse.pde/pull/1480#issuecomment-2502972629) Any help is appreciated. 

I do want to involve myself further than just codeday and join the open source scene but I really need someone to take me under their wing so to speak.

HannesWell:
Sorry for the delayed reply but the time before Christmas was very hectic.
> For me and the team I'm working with this ticket seems very vague and because of that it's been difficult to replicate exactly what we are trying to solve.

Please see the example from my https://github.com/eclipse-pde/eclipse.pde/pull/1480#pullrequestreview-2452506187. With that it should be possible to resolve this issue.

> If you have a very simple OSGi DS component like the following
> 
> ```
> package pack;
> 
> import org.osgi.framework.BundleContext;
[...]

PR #1464, opened 2024-11-08
#1218 scr 1.4 constructor for injection

EverettHanke:
Annotating when having Injectable Constructor for ticket #1218


laeubi (changes requested):
This does not look right,

1. Error message is unclear
2. To solve the referenced ticket it has to check if constructor is public and then not give an error message.

EverettHanke:
Hi there @laeubi. I have a few questions regarding the changes needed to be made. Would you want me to give a problem report if the constructor is injectable but not public as in the constructor doesn't have the @activate annotation?

laeubi:
Yes the ticket describes the case that there is an injectable constructor (annotated with `@Activate`) but it is not public (what is mandatory).

PR #1465, opened 2024-11-09
Resolve #1308: Fix E4 editor boolean parameters

### LINKED ISSUE #1308: E4 editr use inconsitent style for boolean parameters
Currently the E4 Editor looks like this:

grafik

Also in contrast to other things the colon `:` is missing.

I think this should be unified to always use 

`Label: [ ]`

instead of 

`[ ] Label`

stewedbeef:


HannesWell:
Thank you for this PR aiming to resolve https://github.com/eclipse-pde/eclipse.pde/issues/1308.

Could you be so kind and provide an example to verify this change?

Furthermore please squash all your commits into one with a meaningful message that covers the change as whole and force push it to the branch of this PR.
If you want to record your team members in the commit you can add them as co-authors to the commit message using the format:
`Co-authored-by: name <name@example.com> `

stewedbeef:
Example as follows from the E4 spies window reachable by shift-alt-F9:
* Before: 
  pde_e4_fix_before
* After:
  pde_e4_fix_after

As for the squashing, does that include squashing the version bump made by the bot?

e4tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/internal/common/component/ControlFactory.java
```diff
@@ -745,12 +745,14 @@ public static <M> void createCheckBox(Composite parent, String label, IObservabl
 	public static <M> void createCheckBox(Composite parent, String label, String tooltip, IObservableValue<M> master,
 			EMFDataBindingContext context, IWidgetValueProperty<Button, Boolean> selectionProp,
 			IValueProperty<? super M, Boolean> modelProp) {
+		final Label l = new Label(parent, SWT.NONE);
```
laeubi, line 748:
```suggestion
		final Label label = new Label(parent, SWT.NONE);
```

stewedbeef: It is not possible to make the reviewed change without causing the code to have an error because naming the `Label` `label` would shadow the parameter `label` of type `String` passed into the function.

Since naming the variable of type `Label` `l` was also done by other people who wrote similar functions, for example in the functions `createXMIId` and `createTextField` in the same file `ControlFactory.java`, I was wondering if a change needs to be made.

HannesWell: That's right. This is an unfortunate situation and done like this in similar cases in this class.
So I think it's fine.

HannesWell (approved):
> Example as follows from the E4 spies window reachable by shift-alt-F9:
> 

Thanks for the example, that made it simpler to reproduce it.
The code looks good and the changes in the labels as well. Thanks for this diligent work!

> As for the squashing, does that include squashing the version bump made by the bot?

Yes. Please keep that commit separated.

In general this seems ready for submission, but unfortunately we are currently in the quite period before the December release and have to await the start of the next development cycle, which should start in one to two weeks, before this can be submitted.

In the meantime could you please enhance the commit message by appending a reference to the fixed issue, i.e. by appending to the message by the line:
`Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1308`

and additionally use your full name in the author and committer fields and for your co-authors. Ideally you also use real mails and not only github-noreply mails.
Then again please force push the updated commit to this PR's branch (and you fetch from this repository and rebase on top of the current master).

@laeubi are you fine with this as well?

stewedbeef:
Sorry for the delay. We rebased it and fixed the commit message.

PR #1466, opened 2024-11-09
Simplify event publication in TargetPlatformService

HannesWell:
Instead of implementing an asynchronous event dispatching mechanism in TargetPlatformService, simply use IEventBroker.post().


HannesWell:
Comparator differences are unrelated and happen on the master too.

PR #1461, opened 2024-11-07
Comparator errors in I20241106-1800

jarthana:
These are expected changes due to the compiler changes via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3254

PR #1467, opened 2024-11-10
Force qualifier update in org.eclipse.pde.api.tools

HannesWell:
to fix comparator errors/warnings in verification builds:
```
 [WARNING] MavenProject: org.eclipse.pde:org.eclipse.pde.api.tools:1.3.600-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.pde_master/apitools/org.eclipse.pde.api.tools/.polyglot.META-INF: baseline and build artifacts have same version but different contents
    no-classifier: different
       org/eclipse/pde/api/tools/internal/builder/IncrementalApiBuilder$ResourceDeltaVisitor.class: different
       org/eclipse/pde/api/tools/internal/util/Signatures.class: different
```

But since the I-build does not complain I fear it's a configuration difference and it will be a reoccurring issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,22 @@
     <tycho.scmUrl>scm:git:https://github.com/eclipse-pde/eclipse.pde.git</tycho.scmUrl>
     <failOnJavadocErrors>true</failOnJavadocErrors>
     <javadoc.excludePackageNames></javadoc.excludePackageNames>
+    <cbi-ecj-version>[3.40.0,4)</cbi-ecj-version>
   </properties>
+
+  <pluginRepositories>
+    <pluginRepository>
+      <id>eclipse-snapshots</id>
+      <url>https://repo.eclipse.org/content/repositories/eclipse-snapshots</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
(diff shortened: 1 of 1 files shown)


laeubi:
> But since the I-build does not complain I fear it's a configuration difference and it will be a reoccurring issue.

This is usually because one needs to deploy/use a more recent compiler see for example:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1960

HannesWell:
> > But since the I-build does not complain I fear it's a configuration difference and it will be a reoccurring issue.
> 
> This is usually because one needs to deploy/use a more recent compiler see for example:
> 
>     * [Class file diff in validation builds eclipse-platform/eclipse.platform.releng.aggregator#1960](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1960)

Thanks for that hint. That makes sense and fits to https://github.com/eclipse-pde/eclipse.pde/pull/1461.

I updated this PR to verify this in PDE, but if it's fixing it it should be done in the parent-pom as in the linked issue.

HannesWell:
This test was successful and the generalization is done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2538.

Issue #1407, opened 2024-09-14
Support `jars.extra.classpath` in the Manifest editory to avoid confusion with `additional.bundles`


laeubi:
Currently there is no UI support for `jars.extra.classpath` in the Manifest editor and du to its prominent placing people instead use often `additional.bundles` to add something that should be there for compile but not imported ([what serves a different purpose](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Fguide%2Ftools%2Feditors%2Fmanifest_editor%2Feditor.htm)) [leading to confusion](https://github.com/eclipse-pde/eclipse.pde/issues/1403).

Even though `jars.extra.classpath` theoretically supports different styles, we should only support the form where one can specify a single bundle `platform:/plugin/<Bundle-SymbolicName>` as this is also supported by Tycho and easy to adapt by other tools.

merks:
This is a good idea.

I'm not suggesting one needs to make things more complicated so this is just an FYI...

The scheme generally supports
```
platform:/plugins/<bsn>_<version>/
```
as parsed by `org.eclipse.core.internal.boot.PlatformURLConnection.parse(String)`.

laeubi:
> The scheme generally supports

Officially also relative path is supported as well as inner jars:

> extra classpaths used to perform automated build. Classpath can either be relative paths, or platform urls referring to plug-ins and fragments of your development environment (e.g. ../someplugin/xyz.jar, platform:/plugins/org.apache.ant/ant.jar). Platform urls are recommended over relative paths;

I just think that if one really requires a version or relative path one can do it manually, but e.g additional bundles do only support a BSN and people seem happy with that already (but using it for a different purpose).

merks:
I 100% agree to keep this simple for the common case!

Nicolasgarbarino:
I'm working on this.

HannesWell:
> I'm working on this.

Great. You are one of the students from CodeDays, aren't you?

Much success on this task and don't hesitate to ask for help if you need any.

Nicolasgarbarino:
Thank you, yes I am from CodeDay. I wanted to reach out for clarification. In build configuration I am able to add jars by using Extra ClassPath Entries and using the add Jars feature, is there anything specific you wanted on top of having this available in the build section?

HannesWell:
[...]

PR #1469, opened 2024-11-11
Updated BuildClasspathSection.java

anviik:
Rewrote the initialize() function in BuildClasspathSection.java to make the 'Extra Classpath Entries' expanded by default
Before:
buildclasspathbefore

After:
Screenshot 2024-11-10 at 1 21 32 PM


HannesWell (approved):
Thank you @anviik for this contribution to https://github.com/eclipse-pde/eclipse.pde/issues/1407.

This change looks good and works as desired.
I just enhanced the commit message a bit and added a reference to the issue. With that this can be submitted.

PR #1470, opened 2024-11-12
Automated management of dependencies

anviik:
Wasn't able to add a link to the editor object so just added a text description of the "Automated Management of Dependencies"

Before:
buildclasspathafter

After:
Screenshot 2024-11-12 at 1 06 52 AM


laeubi:
> added a text description of the "Automated Management of Dependencies"

I would better place it on the top description to make it more visible.

Issue #1306, opened 2024-06-18
Adding SWT + JFace when creating a plugin that "makes contribution to the UI"


laeubi:
Currently if one creates a new plugin project and chooses in the wizard 
![grafik](https://github.com/eclipse-pde/eclipse.pde/assets/1331477/4cb50753-f0a3-4f0a-9ea6-be6d60d5cb6e)

One of the first things I usually then do is adding swt + jface as required bundles, because otherwise it is hard to really "contribute to the UI".

As these are dependencies of any Eclipse IDE anyways and easy to remove in rare cases where it is not desired, I like to suggest we simply add them by default.

vogella:
Currently this flag is only used to select the templates. I would dislike having plug-in dependencies automatically added. Maybe you can better provide a JFace and SWT template in the template section if that flag is set?

laeubi:
> Currently this flag is only used to select the templates.

It marks this plugin as a singelton also

> I would dislike having plug-in dependencies automatically added. 

Why?

> Maybe you can better provide a JFace and SWT template in the template section if that flag is set?

It seems odd to require an additional step, beside that what should such "template" offer?

vogella:
> Why

Because I may want to make a menu contribution via model elements and in this case I do not need SWT and JFace.

laeubi:
> > Why
> 
> Because I may want to make a menu contribution via model elements and in this case I do not need SWT and JFace.

That's what I mean with "and easy to remove in rare cases where it is not desired", having only a menu contribution without defining any handler that do not reference anything (e.g. shells, dialogs, ...) is likely seldom. And even in that case there is a high chance one later adds new features to the plugin that still require SWT/JFace.

vogella:
Its is also easy to add dependencies, if you need it.

-1 from me for such adjustments.

laeubi:
> Its is also easy to add dependencies, if you need it.

No one questions it, but if I have two "easy" options and I need `Option 1` in 99% of the time and `Option 2` in 1% of cases it is clear what should be the obvious default... currently it is the reverse...

vogella:
You could add a template for your 99% use case.

vogella:
@laeubi if you want this, please ask a PDE project lead to resolve our conflict.

laeubi:
> You could add a template for your 99% use case.

Using a template is simply over engineered here and more complex, first it needs another step, then I need to find it (not to mention that I need to know about this).

[...]

PR #1472, opened 2024-11-12
Adding SWT + JFace when creating a plugin that "makes contribution to the UI" #1306 

cervantes-yadira:
According to the original issue, using JFace and SWT is almost necessary for plugins that contribute to the UI. Since these dependencies can be easily removed and are already dependencies of Eclipse, we decided to add them by default. 
  
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1306


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/plugin/NewProjectCreationOperation.java
```diff
@@ -554,6 +555,15 @@ private IPluginReference[] getDependencies() {
...
 		}
+		// Else if a plug-in project is being created...
+		else if (fData instanceof IPluginFieldData) {
+			// And the plug-in project makes contributions to the UI...
+			if (((IPluginFieldData) fData).isUIPlugin()) {
+				// Add SWT and JFace as dependencies
+				result.add(new PluginReference("org.eclipse.swt")); //$NON-NLS-1$
+				result.add(new PluginReference("org.eclipse.jface")); //$NON-NLS-1$
+			}
+		}
```
HannesWell, line 566: You can use Pattern matching for instanceof here and then combine the two if statements nicely into one.
Furthermore the comments can be reduced here. It's not necessary to write down every action in a corresponding comment. Comments should only be used to clarify the logic where it is not obvious from the code (ideally the code is readable and understandable in itself).

And last but not least the logic should be enhanced to to not add the dependencies if a user has selected a template.
If a template is selected that template already defines the necessary dependencies and nothing should be change here.
Just as described in https://github.com/eclipse-pde/eclipse.pde/issues/1306#issuecomment-2176489533.

```suggestion
		} else if (fContentWizard == null && fData instanceof IPluginFieldData pluginData && pluginData.isUIPlugin()) {
			// a plug-in project is being created and makes contributions to UI
			result.add(new PluginReference("org.eclipse.swt")); //$NON-NLS-1$
			result.add(new PluginReference("org.eclipse.jface")); //$NON-NLS-1$
		}
```

HannesWell (changes requested):
Hi @cervantes-yadira, thank you for this PR. It already looks quite good. Nice work.
I just have a comment in the code to further refine this.

If you apply the changes, please amend your commit locally and force push the amended commit to the branch of this PR so that we get a nice and clean history.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/plugin/NewProjectCreationOperation.java
cervantes-yadira: Thanks for the suggestions! I added the additional check for the use of templates and used pattern matching as you suggested. I have appended the commit and pushed it to this branch.

HannesWell (approved):
Thanks for the update, the code looks perfect now!

Unfortunately we are currently in the quite period before the December release and have to await the start of the next development cycle, which should start in one to two weeks, before this can be submitted.

In the meantime could you enhance the commit message by appending a reference to the fixed issue, i.e. by `appending` to the message by the line:
`Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1306`

and then again force push the updated commit to this PR's branch (and you fetch from this repository and rebase on top of the current master).

@laeubi, @vogella or @opcoach, since you discussed in the fixed issue, are you fine with this change as well?

vogella:
Change looks fine to me, thanks @cervantes-yadira 

I think this solution is fine, additional checkboxes in the UI might IMHO not be necessary.

HannesWell (approved):
Thanks for the update.

PR #1474, opened 2024-11-13
[scr 1.4] constructor for injection must be public #1218 Pull Request

EverettHanke:
Ticket #1218 solution. 

Gives a warning when a constructor is injectable but is not public/annotated with @activiate telling users to make constructor public.


HannesWell:
Hi @EverettHanke thanks for this update.

First please squash all your commits into one with a meaningful message that covers the change as whole and **force push it** to the branch of this PR. Please don't create a new PR, you can simply update its change by using git force-push!

Furthermore please revert all brace re-formatting.
And as far as I can tell the requested changes from https://github.com/eclipse-pde/eclipse.pde/pull/1464#pullrequestreview-2423149544 are not address, are they?

EverettHanke:
Hi there @HannesWell. I'm having an issue with the squash of the several commits. Also this commit should be addressing the requested changes asked on the #1464 changes requested. However if it appears that we didn't meet all the expectations for the ticket. Then I'd like to somehow get into contact with the community so I can discuss it and get a better understanding. Do you have a community discord by chance?

Issue #451, opened 2023-01-31
[target] Implicit Plugin Dependencies is broken


laeubi:
The target environment has a section called "Implicit Plugin Dependencies", where one can add a plugin, and then the target is marked as "dirty" but save has no effect.

![grafik](https://user-images.githubusercontent.com/1331477/215749308-4d239caa-4946-4b78-834d-990bc5d38327.png)

As this could become a very handy feature to include e.g. testing or annotation libs, it would be good to re-enable the feature, and possible have a way to select if this is a regular or test dependency.

laeubi:
If we fix this feature, I think it would be good to have the opportunity to select if a dependency should be added for test or for regular classpath. This would help in setting up certain test-frameworks much easier.

merks:
This seems like a cool and good feature.  

I can imagine though that many people use the running SDK (or even EPP package) as their target platform, because it works, they don't know any better, and the better/proper approach of defining a target platform requires quite a bit of knowledge and significant skills to set up.   I guess those people would have to add the right things manually for it to work if things only worked because of those (action at a distance) implicit dependencies.  

It would still seems nice if one could somehow declare the plug-ins needed at development-time, in the plug-in itself.  Because without that people will continue to use hacks, i.e., optional package imports or build.properties properties because they just work.  I also say this because Oomph's targlets looks at the plugins to figure out what they require to figure out what to put in the target platform, so having this information in the plug-ins would be beneficial.

Of course no one is obligated to implement such a nice feature...

laeubi:
> I can imagine though that many people use the running SDK (or even EPP package) as their target platform, because it works, they don't know any better, and the better/proper approach of defining a target platform requires quite a bit of knowledge and significant skills to set up. 

1. Create a new target
2. add a new location, choose "Installation"
3. use the drop-down to choose `${eclipse_home}`

You now have a target that is your eclipse installation but you can further customize that, so beside from people probably not know about that, it does not seem too hard.

> It would still seems nice if one could somehow declare the plug-ins needed at development-time, in the plug-in itself

I once suggested such a feature but it was rejected as "not interesting for PDE", nerveless the main disadvantage is that one has to configure it for each and every project, what is a bit cumbersome if one has not a rather static setup.

laeubi:
I have now added a fix here:
- https://github.com/eclipse-pde/eclipse.pde/pull/1475

[...]

PR #1475, opened 2024-11-13
Fix editing of Implicit Plugin-Dpendencies

laeubi:
Currently editing the Implicit Plugin-Dpendencies only seem to work randomly (e.g. if one change some other property along with the change).

This now do not refresh the full target (what seem to reset the underlying model and remove the dirty flag before changes are applied) but only updates the UI so finally these thing appear in the target file.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/451


laeubi:
Merging as the new issues are caused by:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3402

PR #1479, opened 2024-11-16
WIP: Fix NPE in PDEModelUtility.java (line 415) by adding null check (Fix #1404)

### LINKED ISSUE #1404: NPE in PDEModelUtility.java:415
```
eclipse.buildId=4.34.0.I20240908-1800
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.eclipse.ui
Error
Tue Sep 10 11:47:22 CEST 2024
Unhandled event loop exception

java.lang.NullPointerException: Cannot invoke "org.eclipse.pde.internal.ui.editor.context.InputContextManager.findContext(org.eclipse.core.resources.IResource)" because the return value of "org.eclipse.pde.internal.ui.editor.PDEFormEditor.getContextManager()" is null
	at org.eclipse.pde.internal.ui.util.PDEModelUtility.lambda$1(PDEModelUtility.java:415)
	at org.eclipse.ui.internal.PendingSyncExec.run(PendingSyncExec.java:68)
	at org.eclipse.ui.internal.UILockListener.doPendingWork(UILockListener.java:166)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4099)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3715)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:271)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:668)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:605)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
```

As far as i remember the last action i did before the error was logged: Close like 90 editors by "Close Others" from the editor tab context menu.

As far as i see in the code org.eclipse.pde.internal.ui.editor.PDEFormEditor.dispose() is the only thing that makes contextManager = null.

slvo-loading:
**Summary Of Change**
Added a null guard to editor.getContextManager() to prevent NullPointerException (NPE).

**Why this Change is Needed**
An NPE occurs when too many processes are running and the user uses the context menu to close all tabs. This is caused by a race condition where dispose() is called while modifyEditorModel() is still running, resulting in getContextManager() returning null.

**Related Issue**
Fix #1404 

**Testing**
Testing is in progress. We are working on a test where dispose() is called followed by modifyEditorModel() to verify that the null guard resolves the issue.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/util/PDEModelUtility.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/util/PDEModelUtility.java
@@ -54,6 +54,7 @@ import org.eclipse.pde.internal.ui.editor.build.BuildEditor;
 import org.eclipse.pde.internal.ui.editor.build.BuildInputContext;
 import org.eclipse.pde.internal.ui.editor.build.BuildSourcePage;
 import org.eclipse.pde.internal.ui.editor.context.InputContext;
+import org.eclipse.pde.internal.ui.editor.context.InputContextManager;
 import org.eclipse.pde.internal.ui.editor.plugin.ManifestEditor;
 import org.eclipse.pde.internal.ui.editor.schema.SchemaEditor;
 import org.eclipse.pde.internal.ui.editor.schema.SchemaInputContext;
@@ -412,9 +413,13 @@ public class PDEModelUtility {
 				for (IFile file : files) {
 					if (file == null)
 						continue;
-					InputContext con = editor.getContextManager().findContext(file);
-					if (con != null)
-						con.flushEditorInput();
+					InputContextManager manager = editor.getContextManager();
+					if (manager != null){
+						InputContext con = manager.findContext(file);
+						if (con != null) {
+							con.flushEditorInput();
+						}
+					}
 				}
 				if (mod.saveOpenEditor())
 					editor.doSave(monitor);


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/util/PDEModelUtility.java
```diff
@@ -412,9 +412,11 @@ private static void modifyEditorModel(final ModelModification mod, final PDEForm
...
 					if (file == null)
 						continue;
-					InputContext con = editor.getContextManager().findContext(file);
-					if (con != null)
-						con.flushEditorInput();
+					if (editor.getContextManager() != null){
+						InputContext con = editor.getContextManager().findContext(file);
+						if (con != null)
+							con.flushEditorInput();
+					}
```
HannesWell, line 419: The method `editor.getContextManager()` should only be called once and the null-check should be done on the saved result of that only call. Otherwise it might happen that the first call returns a non-null value while the second call returns null again:
```suggestion
					InputContextManager manager = editor.getContextManager();
					if (manager != null){
						InputContext con = manager.findContext(file);
						if (con != null) {
							con.flushEditorInput();
						}
					}
```

HannesWell (commented):
Hello @slvo-loading and thank you for your contribution.
First of all sorry for the late reply, I somehow lost track of this PR.

In general this change look reasonable, but one thing should be changed, see the remark below.
But it's a small change and I'm about to apply and push it directly to your branch to not further delay this change.

HannesWell (approved):

PR #1490, opened 2024-11-22
Ignore 'target' folders in e4tools

HannesWell:
Because in PDE we have packages that contain an element named 'target' we cannot generally ignore 'target'-folders and the patterns have to be more specific than usual.


HannesWell:
Submitting this to unblock https://github.com/eclipse-pde/eclipse.pde/pull/1489.
Also did a local test to be very sure this behaves as desired.

PR #1495, opened 2024-11-27
Updated bundles with comparator errors

trancexpress:
Changes to ecj are for handling of switch on String.

See: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595


trancexpress:
For some reason only `forceQualifierUpdate.txt` was changed by the `TouchBundles` script. I manually bumped `MANIFEST.MF`... So I'm not sure we need the version bump? I do see the last version bump was 2 days ago... So I think we do, to avoid the comparator errors?

akurtakov:
No version bump is the right thing to do as https://github.com/eclipse-pde/eclipse.pde/commit/86709bf05031ea83ee32fd103ac505e96ad19e4d already bumps the version. Even api tools complains about the version bump now - please remove it.

Issue #1426, opened 2024-10-02
Always add a version if exporting a package


laeubi:
Currently if one exports a package

1. Using a Quickfix
2. Using the Manifest editor

PDE adds the export without a version. Unversioned packages have the drawback that one can not reliable reference them with a proper import range (what recently was added to PDE: https://github.com/eclipse-pde/eclipse.pde/pull/1007 and enhanced with https://github.com/eclipse-pde/eclipse.pde/pull/1212).

I therefore would suggest that PDE exports packages by default with version `1.0.0` (what is also the default version for Bundles created by PDE). In the rare cases where the user actively choose to not using a version, it can then be edited manually.

merks:
Versioned packages are a good thing.  That being said, does PDE, or anything else, help maintain an up-to-date version as APIs in that package change?  At least for bundle versions, there are the API tools.   In the end, a versioned package is only as useful as the accuracy of the version...

laeubi:
Sadly API-Tools currently does not, but if we start more using packages (what already is done!) it might be good to enhance. Beside that Tycho offers [`tycho-baseline:verify`](https://tycho.eclipseprojects.io/doc/latest/tycho-baseline-plugin/verify-mojo.html) mojo that supports OSGi semantics for package (and bundle) versions, e.g. m2e is using it.

Beside that, no version is as bad as a default version that never increments, so it is not really making the situation worse to use one.

merks:
> Beside that, no version is as bad as a default version that never increments, so it is not really making the situation worse to use one.

One cannot specify a version constraint on a unversioned package can one?  It seems to me that an arbitrary version on a package that's not properly maintained will create a false illusion (for the consumer) that a unversioned package does not create.  So I hope not to see this show up in Platform projects (but I bet it will) unless something is done to properly maintain the versions.

None of this is arguing that we should not do this is PDE, it's a word of caution to those who version their packages that they are obligated to maintain those version properly.  Also, perhaps as those who encourage properly maintained version, we should consider how best to ensure that both are supported by our tools.

laeubi:
> One cannot specify a version constraint on a unversioned package can one? 

Correct.

> It seems to me that an arbitrary version on a package that's not properly maintained will create a false illusion

Without an version it can't be maintained at all :-)

> So I hope not to see this show up in Platform projects (but I bet it will) unless something is done to properly maintain the versions.

[...]

PR #1498, opened 2024-11-30
Always add a version if exporting a package #1426

EverettHanke:
Changed the default for get version to always return a version even if file is unversioned. Defaults at 1.0.0


merks:
This just seems completely wrong. This logic is supposed to make the import version correspond to ( a range around) what's actual exported, e.g., used here:

image

If the exported package has no version the import **must not** specify a version.

Issue #1500, opened 2024-12-03
selection of bnd template persists across different project types


nburnwal09:
Steps:

1. Create a plugin project
2. Select OSGI framework, then select equinox
3. Click next, then again next, observe there are a list of available plugin templates
4. Click back, then again back to 'Create a new plug-in project' page
5. Now select standard under OSGI framework and check the box for 'Generate OSGi metadata automatically' (bnd)
6. Click next, again next and observe the list of bundles this time.
7. Go back twice to  'Create a new plug-in project' and change the OSGi framework to Equinox
8. Click next, then again next, but observe this time, the plugin templates is still bnd's not the standard equinox's

Observe that even when the OSGi framework was changed from standard(with checked in 'Generate OSGi metadata automatically') to Equinox, the plugin template was still showing Bundles.

nburnwal09:
I am working on the solution

PR #1502, opened 2024-12-03
Updated selection when switched from bnd to standard equinox

nburnwal09:
This commit fixes the bnd in templates while switching the plugin project type from bnd(with checked 'Generate OSGi metadata automatically' ) to standard equinox 

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/1500

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/plugin/TemplateListSelectionPage.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/plugin/TemplateListSelectionPage.java
@@ -36,6 +36,7 @@ import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.jface.viewers.ViewerFilter;
 import org.eclipse.jface.wizard.IWizardNode;
 import org.eclipse.pde.bnd.ui.templating.RepoTemplateLabelProvider;
+import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.ui.IHelpContextIds;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
 import org.eclipse.pde.internal.ui.elements.ElementList;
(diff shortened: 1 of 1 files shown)


gireeshpunathil:
@laeubi @HannesWell - can you please have a look? thanks!

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/wizards/plugin/NewProjectCreationPage.java
```diff
@@ -162,6 +162,9 @@ protected void createFormatGroup(Composite container) {
 	private void updateAutomatic() {
 		boolean standardSelected = PDEUIMessages.NewProjectCreationPage_standard.equals(fOSGiCombo.getText());
 		useAutomaticMetadata.setVisible(standardSelected);
+		if (!standardSelected) {
+			useAutomaticMetadata.setSelection(false);
```
laeubi, line 166: I'm not sure it should be fixed in the UI but at the place where the value is evaluated.

nburnwal09: I will look into the possibility and get back.

laeubi: > As the problem occurs when moved from bnd to standard equinox, we need to make it False in the UI.

I don't agree here, the UI is just showing a state. If that state is not valid or irrelevant then it should be assumed to be `false` on this place (or not being evaluated at all).

nburnwal09: I have updated the fix accordingly. Kindly check

But I have a question here:
Does the checkbox 'This plug-in will make contributions to the UI' need to be enabled only for Eclipse projects? I notice that it is greyed out for the both OSGi framework -> standard Equinox and bnd(with checked or unchecked generate automatic metadata)

laeubi: > This plug-in will make contributions to the UI

Yes this is eclipse specific, in standard OSGi there is no mean of "UI", this is application specific. Still a "standard" plugin can of course contribute to the UI... it just controls what is offered to the user and some initial settings but nothing that is persisted.

laeubi (approved):
This looks fine to me!

Issue #1509, opened 2024-12-10
Browser View templates should use SWT.DEFAULT as browser style


wimjongman:
When using SWT.WEBKIT, SWT uses internet explorer as browser widget. As a result, the browser views don't work with Edge on windows.

Set the style to SWT.DEFAULT to force the latest embedded browser technology.

wimjongman:
The javascript game example template does not work when webkit selected with current i-builds (dec 10 2024).

It does work with SWT.DEFAULT:

![image](https://github.com/user-attachments/assets/bda2b6b7-4373-4929-a008-972f0971873f)

wimjongman:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1466

PR #1510, opened 2024-12-10
Browser templates view should not use SWT.WEBKIT as browser style #1509

wimjongman:
Switched to SWT.DEFAULT to activate default browser

### DIFF
--- a/ui/org.eclipse.pde.ui.templates/templates_3.1/browserView/java/$className$.java
+++ b/ui/org.eclipse.pde.ui.templates/templates_3.1/browserView/java/$className$.java
@@ -54,7 +54,7 @@ public class $className$ extends ViewPart implements ISelectionListener {
 
 	@Override
 	public void createPartControl(Composite parent) {
-		fBrowser = new Browser(parent, SWT.WEBKIT);
+		fBrowser = new Browser(parent, SWT.DEFAULT);
 		fBrowser.setText(getContent());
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.ui.templates/templates_3.1/browserView/java/$className$.java
```diff
@@ -54,7 +54,7 @@ public class $className$ extends ViewPart implements ISelectionListener {
 
 	@Override
 	public void createPartControl(Composite parent) {
-		fBrowser = new Browser(parent, SWT.WEBKIT);
```
wimjongman, line 57: @akurtakov should this style bit load internet explorer or does it need fixing in SWT?

merks: The default is being changed from IE to Edge during the current release cycle.

https://www.eclipse.org/lists/cross-project-issues-dev/msg19996.html

akurtakov: IMO the template should not choose any browser engine at all.

akurtakov: Yes, sorry for not giving clearer answer.

PR #1515, opened 2024-12-13
[Launching] Improve addition of default VM arguments for Eclipse apps

HannesWell:
Simplify the code and add arguments only if they are not even added with another value.


HannesWell:
Included in https://github.com/eclipse-pde/eclipse.pde/pull/1516.

PR #1516, opened 2024-12-13
[Launching] Don't add -Djava.security.manager=allow for Java-24 or later

HannesWell:
Since Java-24 the security-manager cannot be used anymore and launching a Java-24 VM fails to launch if the VM-argument
`-Djava.security.manager=allow` is specified.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2623

CC @merks

Depends on https://github.com/eclipse-pde/eclipse.pde/pull/1515.

### DIFF
--- a/features/org.eclipse.pde.unittest.junit-feature/feature.xml
+++ b/features/org.eclipse.pde.unittest.junit-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.unittest.junit"
       label="%featureName"
-      version="1.0.900.qualifier"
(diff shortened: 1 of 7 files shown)


HannesWell:
The complaint about new warnings is a JDT error:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3402

Everything else looks good, submitting.

HannesWell:
Actually I wanted to have the first commit submitted as part of https://github.com/eclipse-pde/eclipse.pde/pull/1515, but I confused the tabs... Anyways since this result is equal, that's not a problem.

PR #1517, opened 2024-12-14
Refresh all repositories when performing an update of IU container

laeubi:
Currently the IUBundleContainer#update only queries the current P2 metadata that might be cached. This can lead to situations where when a repository has "just changed" it does not see the latest metadata.

As we can assume that if the user mind to hit the update button it is meant to really get the latest, this now ask the manager to update the metadata (and possibly colocated artifacts) repository.

FYI @merks @HannesWell @phermsdorf (and @vogella as always being interested in target editor).

This currently handle the case where one press the `Update` button (to set all IUs to latest version) and something similar will be needed for `Refresh`, I'll provide a PR once this is finished to not cause any conflicts.

Just in case we still see issues, we need to change/fix P2, as PDE now explicitly express its intend to refresh the data I would expects the manager to take whatever actions are required.


merks (commented):
I think this logic assumes that metadata repositories and artifact repositories are always colocated.   While I can respect and appreciate that one is carefully trying to update/refresh only the relevant repositories I have a bad feeling that things will be missed.  E.g., when a composite is loaded, that can load a whole whack of children and any of those children my need to be refreshed, and so on recursively.  Does this logic really do that?  (I don't see that it does, but maybe I miss something.)

In any case, would not a flushCache be far simpler and accomplish the same thing (yes perhaps flushing more than necessary, but also not missing anything) without assuming the colocation of metadata and artifact?

laeubi:
> I think this logic assumes that metadata repositories and artifact repositories are always colocated. 

Target locations always assume that, I'm not aware one is able to specify different ones (and have never really seen this), also PDE assume this on multiple places.

>  Does this logic really do that?

The logic asks the repository manager to update all repositories (maybe referenced ones too) to refresh, if that "do not work" then this has to be considered at P2/Manger level and not at PDE/Client side.

> In any case, would not a flushCache be far simpler and accomplish the same thing (yes perhaps flushing more than necessary, but also not missing anything) without assuming the colocation of metadata and artifact?

`flushCache` is not an API (maybe for good reasons) and this will as you mentioned flush much more, e.g assume I update one target with one location it will result that immediately all other update sites need to be reloaded (what could be many more). Beside that a cache-flush do not mean it is really reloaded because refresh does much more including some events and enforce download the data from the server.

So either we should drop the caching at all or do it selectively, it does not make sense to cache something when on the other hand one then needs hacks to flush the cache unconditionally.

merks:
If a composites refresh recursively then that's a good thing!  I think non-collocated metadata and artifact data is rare enough we could just ignore that detail. So this approach looks reasonable.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/P2TargetUtils.java
```diff
@@ -984,44 +985,108 @@ private static <T> T getP2Service(Class<T> key, String absentErrorMessage) throw
...
+		IMetadataRepositoryManager metadataRepositoryManager = getRepoManager();
+		Collection<URI> existing = new LinkedHashSet<>(
+				Arrays.asList(metadataRepositoryManager.getKnownRepositories(IRepositoryManager.REPOSITORIES_ALL)));
 		if (repos.isEmpty()) {
-			repos = Arrays.asList(manager.getKnownRepositories(IRepositoryManager.REPOSITORIES_ALL));
+			repos = existing;
 		}
-		SubMonitor subMonitor = SubMonitor.convert(monitor, repos.size() * 2);
-
+		int work = repos.size() * (forceReload?4: 2);
```
merks, line 1027: Does the formatter not add white space around the `?`? 
```suggestion
		int work = repos.size() * (forceReload ?4  2);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/Messages.properties
```diff
@@ -42,6 +42,8 @@ LocalTargetHandle_2=Unable to generate memento for target platform
 LocalTargetHandle_3=Failed to delete target definition: {0}
 LocalTargetHandle_4=Error saving target definition {0}
 LocalTargetHandle_5=Error creating target file
+P2TargetUtils_cant_refresh_artifacts=Can't refresh artifact repository {0}
+P2TargetUtils_cant_refresh_metadata=Can't refresh metadata repository {0} reported metadata might be outdated
```
merks, line 46: It doesn't read like a grammatically complete sentence:
```suggestion
P2TargetUtils_cant_refresh_metadata=Can't refresh metadata repository {0} so the reported metadata might be outdated
```

laeubi:
If no one objects I want to merge this so we have enough time for finding any issues and base further changes on top of this.

merks (approved):
Go for it.

PR #1518, opened 2024-12-14
Change the name of the "version update" button in target editor

laeubi:
I see constant confusion about what the buttons in the target editor do and therefore like to suggest we use a different term, my proposal is to use "Upgrade" instead of "Update" (similar to `apt-get update` versus `apt-get (dist-)upgrade`).

Even though the file is (possibly) _updated_  this often seem to indicate to users that they can _update a stale state_ instead, while now we better express that one version is _upgraded_ to a newer one.

## Before
grafik

## After
grafik


HannesWell:
> I see constant confusion about what the buttons in the target editor do and therefore like to suggest we use a different term, my proposal is to use "Upgrade" instead of "Update" (similar to `apt-get update` versus `apt-get (dist-)upgrade`).
> 
> Even though the file is (possibly) _updated_ this often seem to indicate to users that they can _update a stale state_ instead, while now we better express that one version is _upgraded_ to a newer one.

While I think it won't solve all confusion, it could definitely help a bit. Especially the latter confusion should be solved. Ultimately I can just recommend and recommend to recommend to read the tool-tips.

PR #1519, opened 2024-12-14
Provide user feedback when updating versions

laeubi:
One current issue is that users are confused about the meaning of buttons, especially as pressing them gives no immediate feedback.

This now do the following:
- disable the update button while update is running (or user selects another location)
- after completion gives a message to the user what was the outcome of the most recent operation


vogella:
Would be nice to use the non blocking notification API here instead of a blocking MessageDialog.

vogella:
See https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/docs/JFaceSnippets.md#Snippet085---Notification-popup-with-user-interaction

PR #1525, opened 2024-12-17
Defer initialization of classpath to the background if called from UI

laeubi:
Currently there are some bad behaving UI components in the eclipse IDE that trigger resolving of the classpath containers in the UI, this leads to very bad startup performance and even deadlocks in startup.

This now detects the issue, logs a warning of the offending component and defer the initialization of classpath to a background job, this currently increase time from starting eclipse until UI is shown noticeable.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1481

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
@@ -13,19 +13,32 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.core;
 
+import java.lang.StackWalker.Option;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobChangeEvent;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.eclipse.jdt.core.ClasspathContainerInitializer;
 import org.eclipse.jdt.core.IClasspathContainer;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 
 public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
+
+	private static final AtomicBoolean WARNING_LOGGED = new AtomicBoolean();
+
+	private static final Map<IJavaProject, Job> JOB_MAP = new ConcurrentHashMap<>();
 
 	private static final Job initPDEJob = Job.create(PDECoreMessages.PluginModelManager_InitializingPluginModels,
 			monitor -> {
@@ -34,8 +47,52 @@ public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
 				}
 			});
 
+	private static StackWalker walker = StackWalker.getInstance(Option.RETAIN_CLASS_REFERENCE);
+
 	@Override
 	public void initialize(IPath containerPath, IJavaProject javaProject) throws CoreException {
+		if (isCalledWhileWorkbenchStartup()) {
+			// See https://github.com/eclipse-pde/eclipse.pde/issues/1481
+			if (WARNING_LOGGED.compareAndSet(false, true)) {
+				ILog.get().warn(
+						"RequiredPluginsInitializer called from within the Application startup thread this will badly impact your IDE performance!", //$NON-NLS-1$
+						new RuntimeException("Called from Application startup thread thread here")); //$NON-NLS-1$
+			}
+			JOB_MAP.compute(javaProject, (jp, oldjob) -> {
+				if (oldjob != null) {
+					oldjob.cancel();
+				}
+				Job job = Job.create(PDECoreMessages.PluginModelManager_InitializingPluginModels, m -> {
+					if (oldjob != null) {
+						try {
+							oldjob.join();
+						} catch (InterruptedException e) {
+						}
+					}
+					setClasspath(jp);
+				});
+				job.addJobChangeListener(new JobChangeAdapter() {
+					@Override
+					public void done(IJobChangeEvent event) {
+						JOB_MAP.remove(jp);
+					}
+				});
+				job.schedule();
+				return job;
+			});
+			return;
+		}
+		Job job = JOB_MAP.get(javaProject);
+		if (job != null) {
+			try {
+				job.join();
+			} catch (InterruptedException e) {
+			}
+		}
+		setClasspath(javaProject);
+	}
+
+	protected void setClasspath(IJavaProject javaProject) throws JavaModelException {
 		IProject project = javaProject.getProject();
 		// The first project to be built may initialize the PDE models, potentially long running, so allow cancellation
 		PluginModelManager manager = PDECore.getDefault().getModelManager();
@@ -83,6 +140,16 @@ public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
 	public void requestClasspathContainerUpdate(IPath containerPath, IJavaProject project, IClasspathContainer containerSuggestion) throws CoreException {
 		// The only supported update is to modify the source attachment
 		JavaCore.setClasspathContainer(containerPath, new IJavaProject[] {project}, new IClasspathContainer[] {containerSuggestion}, null);
+	}
+
+	private static boolean isCalledWhileWorkbenchStartup() {
+
+		return walker.walk(frames -> {
+			return frames.anyMatch(sf -> {
+				return "org.eclipse.e4.ui.internal.workbench.E4Workbench".equals(sf.getClassName()) //$NON-NLS-1$
+						&& "createAndRunUI".equals(sf.getMethodName()); //$NON-NLS-1$
+			});
+		});
 	}
 
 }


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
```diff
@@ -36,6 +48,48 @@ public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
...
+				});
+				job.schedule();
+				return job;
+			});
+			return;
+		}
+		Job job = JOB_MAP.get(javaProject);
+		if (job != null) {
+			try {
+				job.join();
```
HannesWell, line 85: Doesn't this have the effect that the method blocks again until the classpath-set returns?
How would this then improve the start-up performance? Or is it just done to resolve dead-locks?

HannesWell (commented):
Thank you Christoph for working on this issue, I'm very happy about that as well.
But the presented solution does seem to be very robust and I don't think pde.core should handle issues that arise from wrong usage in the UI. Therefore I would prefer a solution in the UI classes where this is caused. E.g. the editor should defer the initialization to a background thread. I would like to have this only as a last resort solution.

laeubi:
> But the presented solution does seem to be very robust and I don't think pde.core should handle issues that arise from wrong usage in the UI.

It is not nice but reflects reality in Eclipse IDE runtime environment and we can quite easy remove it once a better solution is there. Also there are sadly numerous places (and maybe some unknown places) where this happens and at least we get a warning in the log now when new things arise.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
laeubi: This codepath is only executed when it is called outside the UI (e.g. in a build thread) where it is fine to block, the goal is to join on a possible execution that previously was scheduled in an UI thread.

laeubi:
Interestingly the warning even triggers in the CI build already :-D

https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1525/2/maven-warnings/new/source.b23f7a3a-eedd-4a60-b012-8767a6761747/#6521

jukzi:
Thanks for working on the issue. But i think one should not make decisions based on thread name. Better solve it in the caller.

laeubi:
> Thanks for working on the issue. But i think one should not make decisions based on thread name. Better solve it in the caller.

Will you propose PR to do so? And as mentioned before the worst that can happen is it will work as before, so there is no real "danger" in this decision.

jukzi:
> Will you propose PR to do so for JDT?

no, but that is no reason to introduce dirty hacks.

laeubi:
> > Will you propose PR to do so for JDT?
> 
> no, but that is no reason to introduce dirty hacks.

This is not a dirty "hack" it is a workaround for the situation that is is very unlikely that changes will occur any time soon in an unrelated project (JDT) as it is "understaffed" (as every other project as well... So if we can improve the situation quite easy and make it better for all users of PDE there is no reason to wait for something that maybe never happens.

And even if it does, we can easily remove the code again without any problem.

vogella:
I agree that we should apply this PR. 

PDE should protect itself from bad callers which seems to be in this case JDT.

iloveeclipse:
> I agree that we should apply this PR.
> PDE should protect itself from bad callers which seems to be in this case JDT.

I'm not sure you understand possible implications of that PR. JDT will see classpath containers that are not fully initialized or use containers that can't be initialized at all. This may lead to unpredictable plugin compilation results.

Please check `org.eclipse.jdt.core.ClasspathContainerInitializer.initialize(IPath, IJavaProject)` javadoc and  `org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(IJavaProject, IPath)` code.

I agree with @jukzi this is a dirty hack and dangerous one too.

laeubi:
> I'm not sure you understand possible implications of that PR. JDT will see classpath containers that are not fully initialized or use containers that can't be initialized at all. This may lead to unpredictable plugin compilation results.

When the computation is done, we notify JDT that something has changed and then it will simply reevaluate. Also as **compilation** does not happen in the UI thread for this case nothing changes at all, this all calling from the UI is as far as I can see only to inspect the **classpath entries** and the init of the container is just a side-effect.

Also as you mention the javadoc it says:

> Binds a classpath container to a <code>IClasspathContainer</code> for a given project,
> **or silently fails if unable to do so**.

So it would be even valid to do nothing and simply return, I just wanted to be a little bit nice here ;-)

I would be happy to not require this so I suggest anyone who feels uncomfortable with this put a breakpoint in

`org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(IJavaProject, IPath)`

and fix all call sites that enter her in the main (== UI thread) when you start an eclipse, sadly those are very deep nested so its not trivial, and obviously those are not compile threads in any case.

laeubi:
I now have made some more tests:

1. Simply return when called from UI Thread --> The Plugin Container is not added (as expected) unless one changes something from the Manifest or something else so PDE set it again
2. Current solution I see the container is set due to the Job I started (like if someone changes the manifest or similar)
3. Throw an exception, then I see 
 ```
eclipse.buildId=unknown
java.version=21
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=de_DE
Framework arguments:  -product org.eclipse.sdk.ide

org.eclipse.jdt.core
Error
Fri Dec 27 14:15:20 CET 2024
Exception while initializing all containers

org.eclipse.core.runtime.CoreException: Called from Ui thread!
	at org.eclipse.pde.internal.core.RequiredPluginsInitializer.initialize(RequiredPluginsInitializer.java:81)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(JavaModelManager.java:3209)
	at org.eclipse.jdt.internal.core.JavaModelManager$11.run(JavaModelManager.java:3097)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2457)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2482)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeAllContainers(JavaModelManager.java:3155)
	at org.eclipse.jdt.internal.core.JavaModelManager.getClasspathContainer(JavaModelManager.java:2131)
	at org.eclipse.jdt.core.JavaCore.getClasspathContainer(JavaCore.java:3970)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3127)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3291)
	at org.eclipse.jdt.internal.core.JavaProject.getResolvedClasspath(JavaProject.java:2405)
	at org.eclipse.jdt.internal.core.JavaProject.buildStructure(JavaProject.java:488)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:246)
	at org.eclipse.jdt.internal.core.Openable.openAncestors(Openable.java:508)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:228)
	at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:569)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:292)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:278)
	at org.eclipse.jdt.internal.core.PackageFragmentRoot.getKind(PackageFragmentRoot.java:539)
	at org.eclipse.jdt.internal.core.PackageFragment.getKind(PackageFragment.java:367)
	at org.eclipse.jdt.internal.core.PackageFragment.validateExistence(PackageFragment.java:596)
	at org.eclipse.jdt.internal.core.Openable.exists(Openable.java:212)
	at org.eclipse.jdt.internal.core.PackageFragment.exists(PackageFragment.java:189)
	at org.eclipse.jdt.ui.StandardJavaElementContentProvider.exists(StandardJavaElementContentProvider.java:509)
	at org.eclipse.jdt.ui.StandardJavaElementContentProvider.getParent(StandardJavaElementContentProvider.java:246)
	at org.eclipse.jdt.internal.ui.navigator.JavaNavigatorContentProvider.getParent(JavaNavigatorContentProvider.java:118)
	at org.eclipse.ui.internal.navigator.extensions.SafeDelegateTreeContentProvider.getParent(SafeDelegateTreeContentProvider.java:105)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceContentProvider$4.run(NavigatorContentServiceContentProvider.java:646)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceContentProvider.findParents(NavigatorContentServiceContentProvider.java:623)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceContentProvider.findPaths(NavigatorContentServiceContentProvider.java:594)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceContentProvider.findPaths(NavigatorContentServiceContentProvider.java:603)
	at org.eclipse.ui.internal.navigator.NavigatorContentServiceContentProvider.getParents(NavigatorContentServiceContentProvider.java:330)
	at org.eclipse.jface.viewers.AbstractTreeViewer.getParentElement(AbstractTreeViewer.java:1793)
	at org.eclipse.jface.viewers.TreeViewer.getParentElement(TreeViewer.java:573)
	at org.eclipse.jface.viewers.AbstractTreeViewer.internalExpand(AbstractTreeViewer.java:1733)
	at org.eclipse.jface.viewers.AbstractTreeViewer.setSelectionToWidget(AbstractTreeViewer.java:2737)
	at org.eclipse.ui.navigator.CommonViewer.setSelectionToWidget(CommonViewer.java:439)
	at org.eclipse.jface.viewers.StructuredViewer.setSelectionToWidget(StructuredViewer.java:1706)
	at org.eclipse.jface.viewers.AbstractTreeViewer.setSelectionToWidget(AbstractTreeViewer.java:3253)
	at org.eclipse.jface.viewers.StructuredViewer.setSelection(StructuredViewer.java:1662)
	at org.eclipse.jface.viewers.TreeViewer.setSelection(TreeViewer.java:1091)
	at org.eclipse.ui.navigator.CommonViewer.setSelection(CommonViewer.java:369)
	at org.eclipse.ui.navigator.CommonNavigator.selectReveal(CommonNavigator.java:385)
	at org.eclipse.ui.internal.navigator.actions.LinkEditorAction$2$1.run(LinkEditorAction.java:104)
	at org.eclipse.ui.internal.navigator.actions.LinkEditorAction$2.runInUIThread(LinkEditorAction.java:93)
	at org.eclipse.ui.progress.UIJob.lambda$0(UIJob.java:148)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5042)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4522)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:663)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:570)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:178)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:271)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:668)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:605)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1454)
```

but the container is there because at UI startup the init of container is triggered automatically.

So an alternative could be to simply throw an exception...

**Review by HannesWell — CHANGES_REQUESTED** [review]
> I now have made some more tests:
> 
>     1. Simply return when called from UI Thread --> The Plugin Container is not added (as expected) unless one changes something from the Manifest or something else so PDE set it again
> 
>     2. Current solution I see the container is set due to the Job I started (like if someone changes the manifest or similar)
> 
>     3. Throw an exception, then I see

I still think this should be solved in JDT, i.e. the Java Editor should defer the initialization if called from the UI thread. Since it's a UI component it can also test if the main thread is really the UI thread. I'm really not inclined to workaround bad behavior of other UI parts in the core of PDE.
Concurrency in the intialization of JDT is already a problem and introducing more of it probably doesn't make it better. One also has to consider headless apps that just run in the main thread. They will also be affected of this change.

Therefore I want to ask the JDT committers again to help here as they know JDT better and propose a solution for JDT UI.

**Review by jukzi — DISMISSED** [review]
dirty hack

**Comment by laeubi** [comment]
> dirty hack

Please either abstain from reviews or leave actionable comments, this is not a hack but common practice to use reflection if direct calls are not possible or not wanted (due to coupling). [Dynamic Import Package](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.dynamicimportpackage) is also a standard OSGi technology for late binding of optional imports.

**Comment by jukzi** [comment]
> leave actionable comments,

the only legit action is to close this PR as there is no legit way to fix it in pde. any descision made here based on the caller or calling thread is just too bad.

**Review by vogella — COMMENTED** [review]
My review was requested. It is hard to decide what to do here for me. @laeubi maybe send an email to the JDT dev list so that the team can decide if they can fix it. If not, I would suggest to bring it to the PMC to discuss.

**Comment by laeubi** [comment]
@vogella I added you as reviewer as you are left a comment that one can interpret as a positive review but you haven't given explicit review here yet. So if you don't care (anymore) feel free to remove you from the list of reviewers again.

Beside that @HannesWell has already expressed positive feedback [here](https://github.com/eclipse-pde/eclipse.pde/pull/1525#pullrequestreview-2510430527) but (valid) concerns where expressed about how to check for an UI thread. This is now addressed with a check that is done hundred of times in our platform code (but because we do not want direct UI binding in an optional non intrusive way).

Still @HannesWell and me agree that fixing it "at the root" would be preferable of course, but now after 4 weeks I don't see much hope of any progress here and the current situation is simply unacceptable as it potentially blocking the IDE startup when one has package-explorer/java editor in an IDE (what we can assume as a very common use-case here).

And as PDE and JDT are distinct project there is nothing from any PMC perspective PDE can do to "force" movement in JDT, so having a "grace-period" of just another week seems fair to me instead of waiting an undefined amount of time for "something" to happen. **If this changes any time we can easily remove the workaround at any time**, because there is no API involved or any behavior change that would influence valid callers so there is really no need to worry even tough we all would wish a better alternative.

**Comment by vogella** [comment]
I'm still interested and I also still share the same opinion (PDE should protect itself from bad callers which seems to be in this case JDT.).

@HannesWell is PL in PDE so he can make the cut-off decision if two committers (you and @jukzi) disagree. If someone disagrees with this decision he can escalate this to the PMC.

File: ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java [hunk 2]
```diff
@@ -85,4 +140,15 @@ public void requestClasspathContainerUpdate(IPath containerPath, IJavaProject pr
 		JavaCore.setClasspathContainer(containerPath, new IJavaProject[] {project}, new IClasspathContainer[] {containerSuggestion}, null);
 	}
 
+	private static boolean isCalledFromUI() {
+		try {
+			Class<?> display = RequiredPluginsInitializer.class.getClassLoader()
+					.loadClass("org.eclipse.swt.widgets.Display"); //$NON-NLS-1$
```
iloveeclipse, line 146: If I see it right, this will trigger all static init code in the Display class to run if the display wasn't yet used (and also all the classes used by display static code), and that is potentially worse as the bug you are trying to workaround here.

Please don't do that.

Also please don't add (even if technically possible) dependency to SWT from PDE core. It is just not OK.

iloveeclipse (changes requested):

HannesWell:
> Still @HannesWell and me agree that fixing it "at the root" would be preferable of course, but now after 4 weeks I don't see much hope of any progress here and the current situation is simply unacceptable as it potentially blocking the IDE startup when one has package-explorer/java editor in an IDE (what we can assume as a very common use-case here).

Thank you for the update. I think the code is less 'dangerous' now, but yes I'm still strongly favor of fixing this at the root in JDT. I consider this as a really last resort solution. Async work can already be a problem now and introducing more (even just for special cases) can potentially cause more problems.

Therefore I can just repeat myself and ask the JDT committer that are active in this discussion to help here so that we don't have to decide if it's better to not fix the problem to to apply this change.
I can also try to have a look into this, but I'm actually busy with other things at the moment. So this might take longer than the desired week.

> I'm still interested and I also still share the same opinion (PDE should protect itself from bad callers which seems to be in this case JDT.).

I wouldn't sign this statement in general. While it's nice to protect callers from bad behavior it's practically not feasible because it would blow up a lot of code-parts if done everywhere. And what's right and what's wrong might also depend on the exact case (although I cannot tell if there is a reasonable use-case for calling the code discussed from the UI thread, but who knows...).
Eventually it's the responsibility of the caller to call the code in the 'correct' way. And in very general, calling this from the UI thread is not semantically wrong. It 'just' has the potential to give a bad user-experience due to a long blocked UI.

> @HannesWell is PL in PDE so he can make the cut-off decision if two committers (you and @jukzi) disagree. If someone disagrees with this decision he can escalate this to the PMC.

We can also have a vote on this among all PDE committers.

gireeshpunathil:
I support this PR. justification: in the absence of a better alternative, a working solution to the stated problem (deadlock) with no known side effects in the common code path is reasonable. people with concerns on (real / imaginary) side effects are best placed to refine the code base through future PRs.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
laeubi: I just replaced the check wit a stackwalker that detects if this is performed during the creation of the E4 Workbench (and therefore hinders the UI to show up).

laeubi:
I don't think a vote would be the right thing to do as this change does not affect PDE project as a whole. A simple PL decision by @HannesWell and @vik-chand would be enough.

Regarding the concern about "async" operation, a classpath container can change any time and it does for example if you edit the manifest, so this is really nothing to worry about and just usual usage.

vogella:
In case we are voting: +1 from me for this change (assuming JDT cannot / doesn't want to fix their code)

jukzi:
The whole eclipse plugin concept was designed such that an RCP app could be render a skeleton before any plugin code is executed. The plugin.xml was choosen to be a text file such that it was not needed to run any java code while already knowing which ui parts they would contribute. Somewhen that good initial idea was violated and the plugin code was executed before the workbench is shown. I do not know when or why it started but my best guess is that it is the e4.compatibility.CompatibilityEditor which got it wrong. So while it may be possible to be fixed in JDT it would be even better to solve it e4 for all UI plugins somewhere here:

```
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitEditor(org.eclipse.ui.texteditor.AbstractTextEditor).init(org.eclipse.ui.IEditorSite, org.eclipse.ui.IEditorInput) line: 3213	
	org.eclipse.ui.internal.EditorReference.initialize(org.eclipse.ui.IWorkbenchPart) line: 353	
	org.eclipse.ui.internal.e4.compatibility.CompatibilityEditor(org.eclipse.ui.internal.e4.compatibility.CompatibilityPart).create() line: 344	
```

vogella:
e4.compatibility.CompatibilityEditor is only a small wrapper to call the inherited methods in the 3x based editors.

jukzi:
e4 calls it a feature not a bug:
https://wiki.eclipse.org/Eclipse4/RCP/FAQ#Lazy_Instantiantion

vogella:
> e4 calls it a feature not a bug: https://wiki.eclipse.org/Eclipse4/RCP/FAQ#Lazy_Instantiantion

Sounds like you are mixing stuff here. The lazy instantiation feature applies to editors and views which are currently not visible / selected in a stack. So if you have 300 Java editors open in the editor area not all are intialized at startup but only the active one. This was also the case in the 3.x time and is one of the reason which the startup time of Eclipse is not linear dependent of the number of editors / views open.

iloveeclipse:
> agree that fixing it "at the root" would be preferable of course, but now after 4 weeks I don't see much hope of any progress here

Well, we had lot of vacations in this time. Now we see some signs of life in JDT:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3468#issuecomment-2574612149

So assuming the offending call would be removed 

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/17fc31849b5225c441e93a8cc7eec636cd05067f/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/PackageFragmentRoot.java#L616

we would not need this PR.

@laeubi : would you please push a PR to JDT core with this line removed, and assuming we would not have any JDT test fails, we can try to merge this for M1 and see which possible side effects in PDE/m2e/Xtext will arise.

laeubi:
Even without this call there is still one issue left in jdt ui:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1860

also platform has a pending issue:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2636

vogella:
> Even without this call there is still one issue left in jdt ui:

Would still makes sense to push the PR to JDT core I think. Are you planning to push it or shall someone else do it?  I'm surprised that @jarthana did not push a PR as he already had the change in his workspace.

iloveeclipse:
> Even without this call there is still one issue left in jdt ui:
> 
> * [[performance] JDT initialization blocks IDE startup eclipse-jdt/eclipse.jdt.ui#1860](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1860)
> 
> also platform has a pending issue:
> 
> * [[performance] CloseUnrelatedProjectsAction.buildConnectedComponents(IProject[]) blocks UI startup eclipse-platform/eclipse.platform.ui#2636](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2636)

So what? This shouldn't be an excuse to add workarounds in PDE? Let fix the problems where they are rooted, not add workarounds here and there at all possible JDT / Platform clients.

laeubi:
> Would still makes sense to push the PR to JDT core I think. Are you planning to push it or shall someone else do it? I'm surprised that @jarthana did not push a PR as he already had the change in his workspace.

but I'm not very confident I should push such changes forward, because beside from deleting the offending line I can't in any way answer question or decide what consequences this has. So actually a JDT committer would be much better person to "own" such change... anyways I have created a PR here:

- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3531

vogella:
> > So actually a JDT committer would be much better person to "own" such change... anyways I have created a PR here:

I also had the same impression and was surprised that neither @jarthana nor another JDT committer actively discussing this issue here created the PR

laeubi:
> So what?

You claimed 

> So assuming the offending call would be removed
> ...
> we would not need this PR.

what is simply wrong, because it will still be required unless someone

> fix the problems where they are rooted

because otherwise the problem will simply persist and this **has some serious consequences** as it possibly deadlocks the whole IDE startup (as complained by @jukzi).

Because of that I do not want to "wait" for too much things just to see we missed the M1 / whatever / ... and need to wait another release for that. If we are fast, the code can simply be removed without any effort, so we do not miss anything (beside not let people complain about  hanging Eclipse).

laeubi:
Just search for "eclipse freezes on startup splash screen" and you will see that this is far from being "not observable by many people"... I don't want to state that it is exactly *this* problem, but the people see _such problems_ and are annoyed, and the usual "workaround" is to kill the process and try again (and being annoyed)...

> if we can consistently work on improving things that cause the problem in the platform/JDT.

Well the first reaction on JDT was "can you fix it yourself" so I'm not very confident that it is "consistently work" on ... but the blame will be put on PDE.

jarthana:
> > Even without this call there is still one issue left in jdt ui:
> 
> Would still makes sense to push the PR to JDT core I think. Are you planning to push it or shall someone else do it? I'm surprised that @jarthana did not push a PR as he already had the change in his workspace.

Two reasons: And as you already know, I work in a different timezone. And then, I had clearly mentioned it was an experimental change and I had to run some tests. Just because it's a single line code, that doesn't mean we can just get rid of it without much thought.

jarthana:
> > > > doesn't mean we can just get rid of it without much thought.
> 
> Creating a PR does not prevent you from giving it thought nor from testing it further.

Lars, why don't you go over the history and come back? The change in question was already suggested in the first comment of the relevant issue. So, anyone, including you could have posted a PR. So, please stop wasting everyone's time by saying things that don't add any value.

laeubi:
I have now created
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1916

so we get one less call from the UI thread.

laeubi:
Closing in favor of

- https://github.com/eclipse-pde/eclipse.pde/pull/1552

PR #1528, opened 2024-12-18
Imported packages showing version of matching packages

### LINKED ISSUE #1501: Imported packages don't show version of matching plugins
If I open the properties of a package it is displayed like this:

grafik

on the other hand, properties of a plugin show this

grafik

it would be good if the same is shown for packages so one can select a package version

nburnwal09:
The commit fixes Properties dialog of Imported packages showing version of matching packages. Implemented similar to required plugin properties section
Fixes #1501

@laeubi 
One questions here:-
1. Under what condition can the list contains plug-in or packages with same name and multiple version in respective Required Plug-ins and Imported Packages section?

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyPropertiesDialog.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyPropertiesDialog.java
@@ -48,18 +48,20 @@ public class DependencyPropertiesDialog extends StatusDialog {
 	private String fPluginId;
 
 	public DependencyPropertiesDialog(boolean editable, IPluginImport plugin) {
-		this(editable, true, plugin.isReexported(), plugin.isOptional(), plugin.getVersion(), true, true, plugin.getId());
+		this(editable, true, plugin.isReexported(), plugin.isOptional(), plugin.getVersion(), true, true,
+				plugin.getId(), true);
 	}
 
 	public DependencyPropertiesDialog(boolean editable, ImportPackageObject object) {
-		this(editable, false, false, object.isOptional(), object.getVersion(), true, true, null);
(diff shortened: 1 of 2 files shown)


laeubi:
@nburnwal09 for example if you have two version of SLF4J API in your target platform you will get multiple providers for that bundle/package version.

nburnwal09:
> @nburnwal09 for example if you have two version of SLF4J API in your target platform you will get multiple providers for that bundle/package version.

I have verified this. It's working.
Could you please review it.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/DependencyPropertiesDialog.java
```diff
@@ -61,16 +62,17 @@ public DependencyPropertiesDialog(boolean editable, ExportPackageObject object)
 
 	public DependencyPropertiesDialog(boolean editable, boolean showReexport, boolean export, boolean optional, String version, boolean showOptional, boolean isImport, String pluginId) {
 		super(PDEPlugin.getActiveWorkbenchShell());
+		isPlugin = showReexport;
```
laeubi, line 65: If `isPlugin` equals `fShowReexport` it should not need a new field.

Beside this "fShowReexport" seem to indicate it is more about reexported packages?

nburnwal09: To find a unique value to identify for plug-in properties, I had assigned isPlugin with fShowReexport which is true only for plug-in case.
Now to avoid the confusion, I have passed True value in the 'this' call only in case of required plug-in initialisation and using is as isPlugin.

gireeshpunathil (approved):
I tested this, and works as desired, plus the changes look good to me!

PR #1529, opened 2024-12-19
Stop using deprecated URL constructors

akurtakov:
Changes tests only.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathresolver/ClasspathResolverTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/classpathresolver/ClasspathResolverTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2021 Sonatype, Inc. and others.
+ * Copyright (c) 2011, 2024 Sonatype, Inc. and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


merks:
@HannesWell 

I wonder is there some URIUtil somewhere that can be used safely with strings that are safe for new URL but not safe for new URI or URI.create?

akurtakov:
@merks Would you please share some links with details if you have handy? I'm surprised that the new recommended way is more fragile than the deprecated one.

merks:
I've seen folks getting bitten by this on several separate occasions.  Only the second one throws an exception.
```
		String string = "file:/Users/Joe Shmoe/";
		new URL(string);
		URI.create(string).toURL();
```

Perhaps/likely there are other characters, but the space is the most glaring and most likely to occur.

It's a bit of a disservice to deprecate something without a description of the gotchas involved with the proposed alternative. ☹️

merks:
FYI, this version of ecj was just build a short while ago:

https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/jdt/ecj/3.41.0-SNAPSHOT/

So while the build says it can't be found, you can see that it really does exist there...

HannesWell:
> I wonder is there some URIUtil somewhere that can be used safely with strings that are safe for new URL but not safe for new URI or URI.create?

I started an effort for a robust URL-> Path conversion in https://github.com/eclipse-equinox/equinox/pull/691, which probably could be extended to convert from String too. I just didn't have the time yet to complete it.

In this case I think this would better be handled by using the actual type of the value: Path; as proposed in:
- https://github.com/eclipse-pde/eclipse.pde/pull/1530

With the latter this should be obsolete.

akurtakov:
I'm closing this one as there seem to be multiple things that make this one just uninformed noise.

Issue #1523, opened 2024-12-16
Target Editor UI freeze / taking very long when refresh a target file


laeubi:
I added an example [test.zip](https://github.com/user-attachments/files/18150360/test.zip) file which is somehow similar to our internal target file, just for the sake of reproducing the described problem.

1. Start Eclipse
2. Open the Progress View
3. Open test.target with Target Editor
4. wait until all is resolved (takes for e about 2 minutes)
5. press Refresh Button (if you click somewhere you see the ui is frozen)
6. i see the above described dialog, which then disappears
7. the more you click, the more you see the dialog
8. now i have this "ui freeze state" for about 4:30 minutes (i guess the repositories are resolved in the background)
9. after that the target platform is resolved again (which is shown in the Progress View)
10. all in all the refresh takes about 5min vs 2min of the initial resoving

_Originally posted by @phermsdorf in https://github.com/eclipse-pde/eclipse.pde/discussions/1513#discussioncomment-11580987_

phermsdorf:
Great! Thanks a lot @laeubi .
I'll try it out next week with the nightly build.

phermsdorf:
I tried it with the last nightly build and now the refresh button works as expected and there is no ui freeze anymore.
Thanks a lot and merry x-mas!

PR #1531, opened 2024-12-20
Refresh targets in the background

laeubi:
A target refresh operation is currently run in the UI thread as it probably in the past where considered a fast operation. At least with P2 Updatesites this is no longer the case and we can'T really make any valid assumptions on how other implementation might perform.

This now schedule the work into a background job and also cleanup the code to reflect current code base. In addition to that, if a job is already running, the user is asked to cancel it before perform other actions.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1523


vogella:
During activator such a cancelation would also be nice, still possible to trigger activation multiple times white it is activating:

image

vogella:
Otherwise, this is great. Works fine for me.

vogella:
Would also be great to have the same for "Reload Target Platform" (similar to Activate).

laeubi:
@vogella please create a dedicated issue with reproducer, I'm sure there are lingering other things around but we can only hunt them down one-by-one :-)

laeubi:
Warnings are due to changing the Build to Java 21, I reset the quality gate now.

PR #1532, opened 2024-12-27
Fail if a workspace is disposed outside the manager

laeubi:
Currently one can obtain a workspace baseline and dispose it leading to hard to track errors.

This now fails if someone tries to dispose the baseline without using the manager.


laeubi:
One of the failing tests seem to be caused by this:
- https://github.com/eclipse-pde/eclipse.pde/pull/1534

PR #1534, opened 2024-12-27
Synchronize Target Node

laeubi:
The org.eclipse.pde.internal.genericeditor.target.extension.model.Node is accessed in a concurrent way but its internal implementation uses unsynchronized mutator methods.

This add synchronized to all methods and copy internal datastructures to make the class at least concurrent-safe and prevent unsyncronized data structures to escape the scope of a node.


laeubi:
Only network issues.

PR #1536, opened 2024-12-29
Move plugins depending on o.e.pde.ui.tests to Java 21

akurtakov:
The mentioned plugin already has BREE of Java-21 thus this fixes warnings in the workspace.


akurtakov:
Jenkins succeeds, GHA failure is apitools.

PR #1539, opened 2024-12-29
Use the project component api baseline to lookup the component

laeubi:
Currently the api-tools heavily rely on a common shared "workspace baseline" this baseline can be disposed and then changes at any time. While this can work in an IDE context, it makes it quite hard for other tools to use this standalone because there is no way to reliable use an alternative mean of "workspace baseline".

This now changes the ProjectApiDescription in a way where the baseline is passed in from the project component and it only fall back to the static singleton if the baseline is disposed.


laeubi:
One problem is `org.eclipse.pde.api.tools.internal.model.ProjectComponent.createLocalApiDescription()` want to cache API descriptions for a project.

PR #1540, opened 2024-12-31
Remove usage of the sequence number from target files

laeubi:
Currently there is an attribute `sequenceNumber` that should change 

> whenever something in the target that affects the set of features and bundles that would be resolved

but effectively this never worked very well and is nowhere used in such a way:

- A sequence number is not mandatory
- Target files are often edited by humans and it is easy to forget to increment those
- Updating it was never public API so only with some operations it was possible to implicitly change that
- it is only used to check if a P2 profile is up-to-date where the sequence number is a very bad indication as actually even things outside the target can influence it
- even if the sequence number does not matches, P2TargetUtils performed some other checks if maybe the profile has not changed, so P2TargetUtils is already aware of changes without it

**To summarize** if the sequence number did not changed the profile is considered clean (what might be wrong), if it changed we ignore it and check our self if it changed.

To prevent us from strange errors and uncertainties, this now completely removes the 'sequenceNumber' and corresponding code leaving the responsibility to detect something changes completely to the implementation of the target location.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/TargetDefinition.java
@@ -124,8 +124,6 @@ public class TargetDefinition implements ITargetDefinition {
 	private TargetFeature[] fFeatures;
 	private TargetBundle[] fOtherBundles;
 
-	private int fSequenceNumber = -1;
-
 	/**
 	 * Constructs a target definition based on the given handle.
 	 */
@@ -210,7 +208,6 @@ public class TargetDefinition implements ITargetDefinition {
 
 	@Override
 	public void setArch(String arch) {
-		incrementSequenceNumber();
 		fArch = arch;
 		if (fRoot != null && arch != null && !arch.isEmpty()) {
 			Element archNode = TargetDefinitionDocumentTools.getChildElement(fRoot,
@@ -224,7 +221,6 @@ public class TargetDefinition implements ITargetDefinition {
 
 	@Override
 	public void setNL(String nl) {
-		incrementSequenceNumber();
 		fNL = nl;
 		if (fRoot != null && nl != null && !nl.isEmpty()) {
 			Element nlNode = TargetDefinitionDocumentTools.getChildElement(fRoot,
@@ -250,7 +246,6 @@ public class TargetDefinition implements ITargetDefinition {
 
 	@Override
 	public void setOS(String os) {
-		incrementSequenceNumber();
 		fOS = os;
 		if (fRoot != null && os != null && !os.isEmpty()) {
 			Element nlNode = TargetDefinitionDocumentTools.getChildElement(fRoot,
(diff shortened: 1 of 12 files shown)


merks (approved):
We discussed my small concern that we might just change the checkProfile method to return false when the sequencer numbers are not equal.  But given the sequence number seems poorly managed and poorly behave since forever, I'm not sure keeping it has any real significant value.  I also checked that the Target DSL generator won't be broken by the changes, and that looks fine too.

laeubi:
This seem to affect a small area in API tools :-(

Therefore I decided to start first with a much smaller approach here:

- https://github.com/eclipse-pde/eclipse.pde/pull/1541

in the long run I still think we should get rid of the `sequenceNumber` altogether at it seems not working reliable.

HannesWell:
> in the long run I still think we should get rid of the `sequenceNumber` altogether at it seems not working reliable.

Yes, absolutely! For the reasons you listed in your initial message. In general I always found it odd that this was even 'necessary' and PDE should be smart enough to detect changes without such a counter that has to be persisted in the file.
Just in case you missed it: I recently extended the sync-check and everything relevant in IU-locations should now be checked:
https://github.com/eclipse-pde/eclipse.pde/pull/1438

I can also have a look at this in detail, probably at the end of this week, but I like the overall direction very much.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiModelFactory.java
```diff
@@ -351,9 +350,6 @@ public static IApiBaseline newApiBaselineFromTarget(String name, ITargetDefiniti
 	private static String generateTargetLocation(ITargetDefinition definition) {
 		StringBuilder sb = new StringBuilder(TARGET_PREFIX);
 		sb.append(IPath.SEPARATOR);
-		if (definition instanceof TargetDefinition target) {
-			sb.append(target.getSequenceNumber());
```
HannesWell, line 355: Could this be the cause of the new test failures? I.e. that it's somehow expected to get different identifier?

laeubi: Yes API tools require the sequence number for some "stale" check but it is only used in the UI (and not even reliable as the sequence number is never persisted...)

HannesWell: Understand. Couldn't then API-tools then reuse the `P2TargetUtils.checkProfile()`?

laeubi: The profile check is only one aspect and only valid for IU locations.

HannesWell: Yes. When looking at the removed calls of `incrementSequenceNumber()` I can only see the methods
- `setTargetLocations()`
- `setImplicitDependencies()`

that are currently not covered by leveraging `P2TargetUtils.checkProfile()`.

laeubi: I think for API tools it would be more useful to store a hash of the target content, but lets do it step-by step here...

laeubi:
With all the preliminary changes now we should be able to finally remove the sequence number all togehter!

PR #1541, opened 2024-12-31
Remove the check for sequence number from P2TargetUtils.checkProfile

laeubi:
Currently it is assumed that a profile is up-to-date if the sequence number does not change, but this has some risk to produce strange errors if for some reason there is no number or it is not changed. And even if the number changed, we still perform a check for actual changes.

Because of the constant problems in inconsistent behavior this now completely removes the check for the sequence number from the profile.


HannesWell:
In general this looks good and I appreciate the simplification.

But I wonder why this succeeds while https://github.com/eclipse-pde/eclipse.pde/pull/1540 fails a test. If possible I think it would better to make https://github.com/eclipse-pde/eclipse.pde/pull/1540 succeed as a whole.

laeubi:
As both of @HannesWell and @merks expressed consensus here already that it is the right direction:

- https://github.com/eclipse-pde/eclipse.pde/pull/1540#pullrequestreview-2526248106
- https://github.com/eclipse-pde/eclipse.pde/pull/1540#issuecomment-2566380272

and no test are broken I will merge this now so we can see if anything goes wrong with this smaller change.

PR #1542, opened 2024-12-31
Persist the sequence number if it changes to the document

laeubi:
Currently the targetdefinition is maintaining a sequence number that should roughly reflect the number of times a relevant property of the target file changes. Currently this number is never written out to the XML making it only effective at runtime but not e.g. when committed to a VCS if not maintained manually.

This now makes sure that the number is correctly persisted in the XML.


HannesWell:
Isn't this the opposite direction as started with your other PRs?
- https://github.com/eclipse-pde/eclipse.pde/pull/1540
- https://github.com/eclipse-pde/eclipse.pde/pull/1541

Therefore I'm not sure we should continue here.

laeubi:
Unless it is not removed, it should at least work... but it looks like making it work / persisted already breaks a lot of stuff, so it serves as another proof that maybe removing it completely is a better choice.

laeubi:
This one would be obsolete if we merge:
- https://github.com/eclipse-pde/eclipse.pde/pull/1544

Issue #406, opened 2022-11-15
Support deprecation of Plugins, Packages, Capabilities and Features


HannesWell:
For OSGi there is the proposal to introduce a `deprecated` directive to signal that a capability is deprecated: https://github.com/osgi/osgi/issues/509
This can be used to mark a Plugin/Bundle, an Exported-Package or a Generic-Capability as deprecated and to add a deprecation notice as the directives value.

PDE should support in the Manifest Editor, to mark a Plug-in (on the main-page) and each Exported-Package (in the Runtime tab as part of the Properties of a Package) deprecated, e.g. by providing a corresponding Checkbox and a then enabled text field to add a deprecation-notice.

Similar to when one references deprecated Java classes/methods/fields in code, deprecated Plugins/Packages should be presented with a specialized icon (striked through?) and, if possible/wanted, its name should be striked through in all Editors and Views.
Furthermore a warning should be issued if a deprecated Plug-in is referenced in a Product or Feature or required by another Plugin. The same applies if one Imports a depreacted package.

The same should be implemented for PDE Features, where a `deprecated` attribute could be added to the Feature element whose value is the deprecation notice. Analogously deprecated Features should be displayed striked through and referencing a deprecated Feature (in a Product or another Feature) should issue a warning.

References to deprecated Features and Plug-ins could be considered for Launch-Configs as well.

HannesWell:
The proposal has been incooperated into the spec in https://github.com/osgi/osgi/pull/513.

Furthermore we discussed in the last [OSGi spec-call](https://github.com/osgi/osgi/wiki/Minutes-for-OSGi-Specification-Project-calls#2022-12-07-call) how to handle if multiple providers of a package exists (e.g. in different version) but it is only deprecated in some.
Then consensus was to simply rely on what the resolver wires, since it tries to use the highest version numbers (if possible) and therefore reflects the latest state of the package (it may become deprecated, and un-deprecated in a later version). It also reflects what (likely) the real situation is in the eventually running application, given it is build with the same TP as in the workspace.

The same applies for bundles/Plug-ins itself.

gian-benito:
Hi I’m from CodeDay, I’ll be working on this.

PR #1543, opened 2024-12-31
Add support for the 'deprecated' directive on required bundles

laeubi:
OSGi recently has added support for a 'deprecated' directive that can be used by tools to inform users that the bundle supplier in one way or another has deprecated a capability and it will likely removed in the future.

This now adds support to PDE to show a deprecation warning if such a bundle is used in require bundles.

grafik


HannesWell:
That's a nice first step for https://github.com/eclipse-pde/eclipse.pde/issues/406.

PR #1544, opened 2024-12-31
Do not load sequence number from target files anymore

laeubi:
As the sequence number is not persisted into a target file by the target editor, there is actually no point in loading it as it only can lead to inconsistent state.


HannesWell (approved):
Looks good.

Can you tell if existing `sequenceNumber` attributes will now be removed (implicitly) on the next change? If not, it would be remove them automatically so that users don't wonder about them anymore.

laeubi:
They are not touched as before like any other custom attribute.

PR #1552, opened 2025-01-10
If JobManager is suspended delay init to a later time

laeubi:
Currently if the JobManager suspended all kind of unwanted effects might happen, especially the joining of a scheduled job will not work but PDE use that here to init the PluginModelManager.

The only feasible situation here is to schedule a new job (that will be executed once the manager is enabled again) and then perform the required action. this will the possibly trigger a rebuild if required or notify other listeners.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
@@ -23,6 +23,7 @@ import org.eclipse.jdt.core.ClasspathContainerInitializer;
 import org.eclipse.jdt.core.IClasspathContainer;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 
 public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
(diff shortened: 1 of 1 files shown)


merks:
Firstly, I ensure that I can reliably reproduce the original freeze problem like this currently:

image

Then I check out this PR and can see that we get to this in the debugger:

image

If I let it run at this point, the main thread is no longer frozen, the initialization continues, and eventually the job manager is resumed such that the scheduled job initializes the project.  No exceptions are logged.

---

I want to note that when we get to this point with the job manager suspended we are highly likely to end up using p2, i.e., p2's ECF framework, depending of course on the ITargetLocation implementation details, which are not known to the framework because the implementation is extensible. As such we are highly to hang waiting for jobs that cannot start because most implementations will want to do loading in parallel and are quite likely to use jobs for that purpose.

The fact that we probably should not be asked to initialize the classpath on the main thread during workbench startup is another problem, but that seems a can of worms.   If that problem is also fixed, then this change is irrelevant and harmless.  I believe @laeubi has already pointed out that the method says "... or silently fails if unable to do so" so one can definitely argue that we are unable to initialize and that deferring the action is a reasonable silent failure.  Simply opening the job manager's job flood gate during startup is also known to cause problems, so to me this looks like the lesser of many evils.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
```diff
@@ -36,6 +37,22 @@ public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
 
 	@Override
 	public void initialize(IPath containerPath, IJavaProject javaProject) throws CoreException {
+		if (Job.getJobManager().isSuspended()) {
+			// if the jobmanager is currently suspended we can't use the
+			// schedule/join pattern here, instead we must retry the requested
+			// action once jobs are enabled again, this will the possibly
+			// trigger a rebuild if required or notify other listeners.
+			Job job = Job.create(PDECoreMessages.PluginModelManager_InitializingPluginModels, m -> {
```
jukzi, line 45: I just tested this with my colleagues workspace and saw it creates 1 job per Project =>1 Worker Thread per Project. It works for me, but it looks dangerous many threads and each thread still (after https://github.com/eclipse-platform/eclipse.platform/pull/1720) burns significant time on synchronization. Could it be somehow refactored to not spam so many jobs, probably a single Job initializing all projects?
image

laeubi:
It create one job per invalid method call maybe but we can't do much about it (except get rid of the invalid calls). One can further complicate this by having a queue of tasks to process and a single job that is scheduled after adding something to the queue, similar to what the init manager job does.

PR #1553, opened 2025-01-10
DS processing crashes when Pattern matching is used in a switch statement

FlorianKroiss:
+ Use getJLSLatest instead of deprecated JLS4 constant to use latest AST Parser


laeubi (approved):
Looks sane to me there is no need to use "old" parsers here.

laeubi:
@FlorianKroiss thanks for this fix, you might want to add a testcase here so we do not run into such issue in the future.

PR #1554, opened 2025-01-10
Move bundles to Java 21 where needed

akurtakov:
Bundles with any Java 21 dependency in the stack are moved accordingly. Fixes warnings like:
```
Execution environment is lower than one of the plug-in's dependencies
(org.eclipse.jdt.core.manipulation) which has an execution environment
of JavaSE-21.	MANIFEST.MF	/org.eclipse.pde.api.tools.ui/META-INF	line
26	Plug-in Problem
```


laeubi:
@akurtakov moving just the BREE is kind of useless IMO and breaks the idea of @mickaelistria that one "simply" can use an older dependency. What directly leads to the conclusion that if we raise the BREE because of dependencies, we also should raise the required version of that dependency (as the older one now can't be used anyways).

merks:
I always thought the warning itself was kind of useless or at the very least misleading or misguided.  Most certainly I have been ignoring that warning in EMF for a very long time because definitely EMF can **and does** work with older versions of Platform dependencies, i.e., dependencies with a lower BREE than the BREE of the current version.  I think the only case where the warning makes good sense sense was for the dependency on Lucene 10.x where the older 9.x version of Lucene was excluded and Lucene 10.x definitely requires Java 21.  Of course when producing the warning, the framework cannot know the BREE of the actual/current lower bound's resolved bundle, only the BREE of the bundle to which which it currently resolves.  Anyway, blah, blah, blah, the warning is questionable in my opinion...

Moving the BREE as is done here is useful for getting rid of the warning (and not much else).  Not incrementing the lower bound on any dependency, unless I grossly misunderstand, does not break anything (any ideas).  The bundle can simply continue to use the older dependency but the older dependency will of course need to run on Java 21 long with the bundle whose BREE has been incremented.   Think of it another way.  If I increment a bundle to Java 21 to use some new must-have  Java feature and latter on one of its dependencies does the same thing, at that point we won't increment the dependency lower bound either.  So we need to increment lower dependency bounds if and only if we use new API from that bundle, or (new ruling) if and only if we are using that dependency to justify the need to increment the BREE.  That seems questionable at best, seems like a lot of work, and seems even more misguided that the warning.  But that's just my opinion...

laeubi:
> Moving the BREE as is done here is useful for getting rid of the warning (and not much else). Not incrementing the lower bound on any dependency, unless I grossly misunderstand, does not break anything (any ideas)

It depends how you define "break"... raising the BREE to the one of the dependency means that this bundle can never run with an older version of that dependency because this bundle now will always require Java 21 regardless of the dependency. So what is the point to retain "backward compatibility" if you never can run it with e.g. Java 17 anymore.

So if we already stub into the face of people requiring Java 17 support just for the sake of questionable `null` check, it makes sense to also being consistent and say we don't support the older dependencies anyways as it is _impossible to test_ (**even if we wanted to!**) that this bundle actually works with the older dependency with Java 17.

And given in the past we have seeing numerous ways this can break (code signing outdated, split packages, not incremented lower bounds), I would say we should not imply false expectations here.

And if we go even further to the OSGi specification, it recommends to use the lowest version the dependency was compiled against (for a good reason).

merks:
We have moved here beyond "breaking an idea" to the provability and testability of the validity of lower bounds only to recognize (yet again) how poorly we maintain lower bounds in general because we simply have no goods tools for that.    So we could indeed hit it with the giant hammer and simply set the lower bound to be the current resolved version because that's definitely not wrong, definitely what we used when compiling, and definitely what we tested against.  Probably almost all minor versions will increment with each release...

laeubi:
I just asked for if we

- increment the BREE **because a dependency** requires a higher one now
- we also raise the lower bound **of that dependency**

that is not a "big hammer" and does not require any special treatment and is the same as if we raise the requirement because we need a new API (one could see the java version as API too... ) **from a dependency**.

akurtakov:
For the sake of sharing same definitions in the project there is https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md#When-to-change-the-minor-segment . Aka BREE bump is seen the same way as API addition for the bundle. This doesn't mandate that the dependent requires the "new API"
The reason why "increment the BREE because a dependency requires a higher one now" happens is https://www.eclipse.org/lists/cross-project-issues-dev/msg17112.html. 
Also, setting min versions "correct" is more than welcome change from my side but requiring to happen together with the BREE bump is not "100% required" as the relation is not 1:1 and thus can be handled separately and when needed , much like handling deprecations from new dependency versions.
Last, what I think of as  "big hammer" is https://github.com/laeubi/p2/pull/80. If/when there is smth that detects and reports even a 1% chance of that the bundle with the bumped BREE can no longer use the old version of its dependencyI would be the one executing such changes whenever I see that warning much like I do here.

laeubi:
It is not require 100% but as said, saying we raise it **because** of the dependency for me implies that we from now on don't want to support any older versions anymore, because otherwise one would simply ignore the warning as it is not the only possible BREE of that dependency.

akurtakov:
It is raised in order to prevent a *known problem* (dependency indirectly requiring newer JVM classes). There is no *known problem* for using  the bundle with bumped BREE against the old version of the dependency - that's the report I look for.

PR #1558, opened 2025-01-13
Add support for checking package exports for changed API

laeubi:
Currently API tools only check for API changes on the bundle level, but even more important are checks on the exported packages. If the package version is not properly incremented this can lead to method not found or similar errors, also consumers of the package can not depend on the package version reliable to get new API.

This now adds some basic checks to check for API changes on the package level, compare it with the baseline and suggest a new version based on the resulting delta being a breaking or non breaking change.

This is a very basic first step. We need (see TODOs) and likely want to further improve this when we use it, for example currently all deltas are assumed to be important, we might discover places where it is actually not required to make a minor change but only a micro change (currently not used at all).


laeubi:
Only old issues (due to java 20) and mirror timeouts.

akurtakov:
> Only old issues (due to java 20) and mirror timeouts.

https://github.com/eclipse-pde/eclipse.pde/pull/1560 should make only issues due to the PR itself to fail it.

PR #1559, opened 2025-01-13
Suggest a micro change for added interface fields

laeubi:
If one adds a field to an interface this will not result in any problem for the consumer or the provider of the interface.

Because of this these changes can be made in a micro package version
increment.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/BaseApiAnalyzer.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/BaseApiAnalyzer.java
@@ -2307,14 +2307,41 @@ public class BaseApiAnalyzer implements IApiAnalyzer {
 				.collect(Collectors.toMap(ExportPackageDescription::getName, Function.identity(), (a, b) -> a));
 		// a mapping between a package name and a required change
 		Map<String, RequiredPackageVersionChange> requiredChanges = new HashMap<>();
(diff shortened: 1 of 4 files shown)


merks:
I just wanted to point out that this (will not result in a problem) isn't **strictly** true.   Here if we add this default method to Foo, bad things will definitely happen in Bar:

image

That being said, Java has been doing this regularly and I suspect the API tools checking has not adapted to the concept of default methods.  I wonder why a field in an interface was ever done this way in the first place?  I should also mention I did **not** read this in detail:

https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

laeubi:
The important part about package versions is binary compatibility, and java says [here](https://docs.oracle.com/javase/specs/jls/se9/html/jls-13.html):

> Adding a default method, or changing a method from abstract to default, **does not break compatibility with pre-existing binaries**, but may cause an IncompatibleClassChangeError if a pre-existing binary attempts to invoke the method. 

So even if it is binary compatible, I think its not sufficient for a micro change and therefore removed this now and only assume adding fields is safe.

laeubi:
Closing as

https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md#How-to-version-packages

says same rules for plugins apply to package versions as well

and 

https://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html

suggests micro is only for bug-fixes "that do not affect either consumers or providers of the API", while this is the case here, adding a field is merely a bugfix...

PR #1560, opened 2025-01-13
Make Jenkins complain on new warnings from the current PR

akurtakov:
That should prevent the need to manually reset the baseline when new warning is introduced by previous PR or even due to change outside of the project.


merks:
The dash failure seems odd given this version has been in Orbit for some years...

akurtakov:
Merging as the licensecheck is tool failure.

PR #1564, opened 2025-01-14
fix Unnecessary @SuppressWarnings("deprecation")

jukzi:


akurtakov:
Suppressing "removal" will get the nasty effect that no one will see any warning until it's actually removed and be "surpised" that build broke.
@HannesWell Are you fine with removing the usage of these deprecated for removal APIs?

jukzi:
> Suppressing "removal" will get the nasty effect that no one will see any warning until it's actually removed and be "surpised" that build broke.

You would only see such warning if you have pde in the workspace. In that case you will get a hard compilation error if you remove the implementation.

laeubi:
> > Suppressing "removal" will get the nasty effect that no one will see any warning until it's actually removed and be "surpised" that build broke.
> 
> You would only see such warning if you have pde in the workspace. In that case you will get a hard compilation error if you remove the implementation.

If it is PDE internally using deprecated PDE thing it should better be fixed than suppressed (or itself deprecated).

akurtakov:
IMO, usage of Platform.ARCH_X86 should be removed altogether now that Platform marked it for removal now rather than fixing yet another I-build failure in 2 years.

laeubi:
A good alternative is to simply inline the constant to these places.

HannesWell:
> IMO, usage of Platform.ARCH_X86 should be removed altogether now that Platform marked it for removal now rather than fixing yet another I-build failure in 2 years.

The affected part of the code is for the Product Editor and it's ability to add arch-specific arguments. If we remove these code paths one cannot target older versions of the Eclipse-runtime with a recent development Eclipse.

> A good alternative is to simply inline the constant to these places.

I would also be in favor of this approach as it's simple doesn't add much overhead and extends the compatibility of PDE with older Eclipse releases to target.

jukzi:
> > A good alternative is to simply inline the constant to these places.
> 
> I would also be in favor of this approach as it's simple doesn't add much overhead and extends the compatibility of PDE with older Eclipse releases to target.

inlined ARCH_X86

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/product/PropertiesSection.java
```diff
@@ -130,8 +131,8 @@ private class PropertyDialog extends StatusDialog {
 
 		private final String[] COMBO_OSLABELS = new String[] { PDEUIMessages.PropertiesSection_All, Platform.OS_LINUX,
 				Platform.OS_MACOSX, Platform.OS_WIN32 };
-		@SuppressWarnings("deprecation")
-		private final String[] COMBO_ARCHLABELS = new String[] { PDEUIMessages.PropertiesSection_All, Platform.ARCH_X86,
+		private final String[] COMBO_ARCHLABELS = new String[] { PDEUIMessages.PropertiesSection_All,
+				IArgumentsInfo.ARCH_X86,
```
laeubi, line 135: What makes me wonder if we should add the other supported archs now here? so what about arm, longaarch and alike?

laeubi (approved):
Looks fine to me

PR #1565, opened 2025-01-15
Fix launch -configuration param to always be proper URI 

akurtakov:
Ensures that the parameter can always we be converted back to URI.
It's been already done for -dev param value.
Fixed PluginBasedLaunchTest to no longer try to "workaround"
-configuration values that can't be converted to URI.


akurtakov:
This is continuation of https://github.com/eclipse-pde/eclipse.pde/pull/541 to ensure that the arguments are actually proper URIs always rather than putting a workaround in the test.

PR #1574, opened 2025-01-18
Remove dependency to icu4j for html exports

laeubi:
The only reason for icu4j currently is to determine if a RTL language is used as the default locale on html exports.

This replaces the fat (12MB!) dependency by a simpler variant that seems suitable for the sake of generating html reports.

The list of codes is taken from https://raw.githubusercontent.com/shadiabuhilal/rtl-detect/2eed8a33276461a24e7033d1d3a115ee64aee3f5/lib/rtl-detect.js and adapted to java.


laeubi:
@waynebeaton I'm not really sure about the "IP" here, is a list of strings already something we need a vetting from the IP team? The source where I found the list mentions New BSD License  Theoretical I could extract the code into an on class and license it under that as well... I'm just not sure if this is a bit exaggerated here...

laeubi:
It would maybe better to move the code to https://github.com/eclipse-equinox/equinox/tree/master/bundles/org.eclipse.equinox.bidi so it can be reused.

waynebeaton:
Sorry for the delay.

My recommendation is that you indicate to the source of the content in a comment. Further investigation isn't a good use of anybody's time.

HannesWell:
This is great! Thanks.
IIRC there are one or two more usages of ICU4J in the Eclipse SDK. Do you plan to get rid of them too?

PR #1585, opened 2025-01-28
Add High Resolution Icons for Wizard Banners

BeckerWdf:
These Icons already existed in the images repo. So let's also use them in PDE


vogella:
Thanks @BeckerWdf

Issue #820, opened 2023-10-26
ClasspathUpdaterTest randomly fails


laeubi:
See https://github.com/eclipse-pde/eclipse.pde/runs/18054276230

```
Update Classpath Job failed: Status ERROR: org.eclipse.pde.ui code=0 Updating failed. See log for details. Java Model Exception: Error in Java Model (code 1006): library.jar [in classpathupdater] is not on its project's build path
java.lang.AssertionError: Update Classpath Job failed: Status ERROR: org.eclipse.pde.ui code=0 Updating failed. See log for details. Java Model Exception: Error in Java Model (code 1006): library.jar [in classpathupdater] is not on its project's build path
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.runUpdateClasspathJob(ClasspathUpdaterTest.java:185)
	at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.test_CreateFromScratch(ClasspathUpdaterTest.java:119)
	at org.eclipse.pde.ui.tests.classpathupdater.ClasspathUpdaterTest.test_CreateFromScatch_WithTestPluginName(ClasspathUpdaterTest.java:141)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
[...]

PR #1601, opened 2025-02-05
fix ClasspathComputer: ignore if oldEntry not found  #820

jukzi:
fixes random failing ClasspathUpdaterTest

When IJavaProject.setRawClasspath() is explicitly called with ClasspathEntries distinct from PDE (like in the Test), and then later PDEs UpdateClasspathJob is executed it can't compute the "oldEntry" and fails with "... is not on its projects build path". The Message is correct - it was manually removed from the projects build path. That state should just be ignored so that PDE continues to set the IClasspathEntries computed by PDE.

https://github.com/eclipse-pde/eclipse.pde/issues/820

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
@@ -255,7 +255,7 @@ public class ClasspathComputer {
 	}
 
 	private static void addLibraryEntry(IPluginLibrary library, Map<String, IPath> sourceLibraryMap,
-			ClasspathConfiguration context) throws JavaModelException {
+			ClasspathConfiguration context) {
 		String name = ClasspathUtilCore.expandLibraryName(library.getName());
 		IResource jarFile = context.javaProject.getProject().findMember(name);
 		if (jarFile == null) {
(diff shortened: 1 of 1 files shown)


jukzi:
how about that? @haubi @laeubi @iloveeclipse @HannesWell ?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -267,17 +268,23 @@ private static void addLibraryEntry(IPluginLibrary library, Map<String, IPath> s
...
+				// - is not defined: the default could be available now, or
+				// - is overridden with a different value.
+				if ((sourceAttachment == null && oldEntry.getSourceAttachmentPath() != null)
+						|| (sourceAttachment != null && sourceAttachment.equals(oldEntry.getSourceAttachmentPath()))) {
+					context.reloaded.add(oldEntry);
+					return;
+				}
+				isExported = oldEntry.isExported();
+			} catch (JavaModelException e) {
+				if (e.getStatus().getCode() != IJavaModelStatusConstants.ELEMENT_NOT_ON_CLASSPATH) {
```
iloveeclipse, line 284: So we see cases where classpath entry is not on the classpath.`IJavaModelStatusConstants.ELEMENT_NOT_ON_CLASSPATH` because someone modified classpath during this method execution?

What about to throw dedicated exception so that the caller `addSourceAndLibraries() `could re-iterate over the classpath again?

iloveeclipse: It sounds like we need a real fix here for `PerProjectInfo.rootPathToRawEntries`?
Ideally you could update PR description what happened and how the change is supposed to fix that. Right now I'm clueless.

haubi: Requiring UpdateClasspathJob to deal with the old state to potentially change _while_ the job is running somehow fails to make good feelings.

jukzi: I updated the PR description https://github.com/eclipse-pde/eclipse.pde/pull/1601#issue-2832463808
If that is understandable i would take it as comfit message.

haubi: Things seem to become more clear now - at least this question pops up now:
Would it make sense, or be possible at all, for this test to _notify PDE_ about this "manual" removal of classpath entries before launching UpdateClasspathJob, to avoid the inconsistent knowledge of "old classpath"?

laeubi: I think we can simply ignore any exceptions here, if it does not work PDE can not do much here and likely something will happen short after that refresh the classpath and run a new `UpdateClasspath` job. If not the worst that can happen is that we can't update the source location.

jukzi: ok, done

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -267,17 +267,23 @@ private static void addLibraryEntry(IPluginLibrary library, Map<String, IPath> s
...
-			IClasspathEntry oldEntry = root.getRawClasspathEntry();
-			// If we have the same binary root but new or different source, we
-			// should recreate the entry. That is when the source attachment:
-			// - is not defined: the default could be available now, or
-			// - is overridden with a different value.
-			if ((sourceAttachment == null && oldEntry.getSourceAttachmentPath() != null)
-					|| (sourceAttachment != null && sourceAttachment.equals(oldEntry.getSourceAttachmentPath()))) {
-				context.reloaded.add(oldEntry);
-				return;
+			try {
```
laeubi, line 270: should we probabbly even move the try outside the if check `getKind` is documented to also throw `JavaModelException` 

> if this element does not exist or if an
> exception occurs while accessing its corresponding resource

jukzi: i am all in for that

laeubi (approved):
This looks like something we can do to be more defensive in error handling.

Issue #1608, opened 2025-02-11
Pressing Browse... in Extensions tab when identifier is of type Nature causes NPE and no dialog to open


jonahgraham:
Pressing "Browse" in the plugin.xml extensions tab when the type is of nature causes an NPE. [example of definition](https://github.com/eclipse-platform/eclipse.platform/blob/72a51171cc347868f3ed6b610c804e970532025c/resources/bundles/org.eclipse.core.resources/schema/natures.exsd#L194) [example of where in a plugin.xml](https://github.com/eclipse-pde/eclipse.pde/blob/ab28042c7d6616e422203cf4d731fea9a923b062/apitools/org.eclipse.pde.api.tools/plugin.xml#L132-L133)

![Image](https://github.com/user-attachments/assets/4157673f-22f1-4940-a352-c3a478335819)

<details><summary>full stack trace</summary>

```
!ENTRY org.eclipse.ui 4 0 2025-02-11 14:40:08.358
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IConfigurationElement.getAttribute(String)" because "value" is null
	at org.eclipse.pde.internal.ui.editor.plugin.rows.IdAttributeRow$IdAttributeLabelProvider.getText(IdAttributeRow.java:56)
	at org.eclipse.ui.dialogs.FilteredList.setElements(FilteredList.java:262)
	at org.eclipse.ui.dialogs.AbstractElementListSelectionDialog.setListElements(AbstractElementListSelectionDialog.java:191)
	at org.eclipse.ui.dialogs.ElementListSelectionDialog.createDialogArea(ElementListSelectionDialog.java:63)
	at org.eclipse.jface.dialogs.Dialog.createContents(Dialog.java:766)
	at org.eclipse.jface.window.Window.create(Window.java:431)
	at org.eclipse.jface.dialogs.Dialog.create(Dialog.java:1092)
	at org.eclipse.ui.dialogs.SelectionStatusDialog.create(SelectionStatusDialog.java:155)
	at org.eclipse.ui.dialogs.AbstractElementListSelectionDialog.access$superCreate(AbstractElementListSelectionDialog.java:462)
	at org.eclipse.ui.dialogs.AbstractElementListSelectionDialog.lambda$1(AbstractElementListSelectionDialog.java:469)
[...]

PR #1609, opened 2025-02-11
Resolve NPE when Ids don't have associated configuration elements

jonahgraham:
This fixes a regression introduced in 34b8e38ef25242a4992f61c8676569a8dcffc7f7 When that commit was done some instanceof checks were removed. The result of removing the instanceof checks had the side effect of removing the implicit null check that instanceof essentially does.

This change restores the null check for when IConfigurationElement is null.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1608


HannesWell:
Thank you Jonah for this fix.

Issue #1614, opened 2025-02-13
Suppress resolution if 'Configure problem severity' is the only available options


nburnwal09:
Currently, the context menu option "Quick fix" is visible for problems where no fix is available, which then leads to a pop-up menu saying No fixes available. 
There are many such issues for this condition. Attaching a sample here for the issue - a space added after the header name before the ':'.

<img width="935" alt="Image" src="https://github.com/user-attachments/assets/6cc81ca4-16fa-42b8-adad-dca7ed6600dc" />

Steps to reproduce:
1. Introduce a minor issue in the code, such as adding an extra space after the header name before the ':'.
2. Right-click on the problem in the Problems view.
3. Select the "Quick fix" context menu option.
4. Observe the pop-up menu indicating 'No fixes available.'

This behaviour misleads developers into thinking that a quick fix is available for the problem. After clicking the 'Quick fix' context menu option, it states 'No fixes available,' which can waste developers' time and cause frustration.

To avoid this misleading, one solution could be to disable the context menu option 'Quick fix' where no fixes are available like this:

![Image](https://github.com/user-attachments/assets/f30300ef-68b3-4607-a6bb-6d49ccb0efdb)

laeubi:
The QuickFix menu is provided by platform, not by PDE. Beside that the main problem is that computing Quickfixes can be an expensive task so disable the menu item might not be possible without performance issues when opening the menu.

nburnwal09:
Upon right click on the problem to get the context menu pop-up, the code passes through this function((before the context menu pop-up is showed))

https://github.com/eclipse-pde/eclipse.pde/blob/975f72b145b7effd0d9df342b3ad2bcb1a1cacb6/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ResolutionGenerator.java#L159-L165

via function hasResolution

https://github.com/eclipse-pde/eclipse.pde/blob/975f72b145b7effd0d9df342b3ad2bcb1a1cacb6/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/ResolutionGenerator.java#L257-L259

And in cases where it reaches to Line 164, where marker id is getting set with PDEMarkerFactory.NO_RESOLUTION, the `Quick fix` menu option gets disabled today as well. We just would have to work around similar way for other no quick fix available problems as well.

laeubi:
[...]

PR #1615, opened 2025-02-13
Updating marker to disable Quick Fix option if 'Configure problem severity' is the only available options

nburnwal09:
This commit fixes the Quick fix option when 'Configure problem severity' is the only available options for the selected problem in the problems tab.

It in turns greys out the Quick fix option when there is no fix attached to the respective problem. This will into misleads developers into thinking that a quick fix is available for the problem when there is no available fix
Previously, it was showing a pop-up saying 'No quick fixes available.'
Attaching the screenshot of the fix
image

Fixes: #1614


nburnwal09:
Yes, that part will work fine.

nburnwal09:
@laeubi I have verified the "working of configure the severity of the problem." 
Can others review this PR too.

nburnwal09:
@laeubi 
Could you these PR's when you have some time.

gireeshpunathil (approved):

akurtakov:
@gireeshpunathil Why do you merge things for RC2 without required approvals? @HannesWell I leave it to you as PL to decide what to do here.

gireeshpunathil:
oops.. missed the fact on the rc2. should this be reverted?

Issue #1604, opened 2025-02-09
Include required plugins automatically seem not consider already added capabilties


laeubi:
If I use the "Include required plugins automatically" option I noticed that it does seem to add more than required in some situations.

For example assume a product hat has `slf4j.api` and two slf4j provider implementations are in the target then PDE adds another slf4j provider even if one was already chosen in the product, I noticed that when working on

- https://github.com/eclipse-equinox/equinox/pull/740

with the product here:

```
<?xml version="1.0" encoding="UTF-8"?>
<?pde version="3.5"?>

<product uid="testslf4j" version="1.0.0.qualifier" type="bundles" includeLaunchers="true" autoIncludeRequirements="false">

   <configIni use="default">
   </configIni>

   <launcherArgs>
      <programArgs>-console
      </programArgs>
      <vmArgs>-Declipse.ignoreApp=true
-Dosgi.noShutdown=true
      </vmArgs>
      <vmArgsMac>-XstartOnFirstThread -Dorg.eclipse.swt.internal.carbon.smallFonts
      </vmArgsMac>
   </launcherArgs>

   <launcher>
      <win useIco="false">
         <bmp/>
      </win>
   </launcher>

   <vm>
   </vm>

   <plugins>
      <plugin id="my.logging.bundle"/>
      <plugin id="org.apache.aries.spifly.dynamic.bundle"/>
      <plugin id="org.apache.felix.gogo.command"/>
      <plugin id="org.apache.felix.gogo.runtime"/>
      <plugin id="org.apache.felix.gogo.shell"/>
      <plugin id="org.apache.felix.scr"/>
      <plugin id="org.eclipse.equinox.common"/>
      <plugin id="org.eclipse.equinox.console"/>
      <plugin id="org.eclipse.osgi"/>
      <plugin id="org.objectweb.asm"/>
      <plugin id="org.objectweb.asm.commons"/>
      <plugin id="org.objectweb.asm.tree"/>
      <plugin id="org.objectweb.asm.tree.analysis"/>
      <plugin id="org.objectweb.asm.util"/>
      <plugin id="org.osgi.service.component"/>
      <plugin id="org.osgi.util.function"/>
      <plugin id="org.osgi.util.promise"/>
      <plugin id="slf4j.api"/>
   </plugins>

   <configurations>
      <plugin id="org.apache.aries.spifly.dynamic.bundle" autoStart="true" startLevel="1" />
[...]

PR #1618, opened 2025-02-15
Don't include more bundles if the requirement is already fulfilled

laeubi:
Currently PDE includes everything wired in the target platform even though a requirement might already be fulfilled and only has single cardinality.

This now tracks all capabilities provided and check if a single cardinality is already fulfilled by some of the bundles chosen.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1604

@HannesWell something for RC1?

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -20,10 +20,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 
@@ -38,12 +36,10 @@ import org.eclipse.pde.core.target.NameVersionDescriptor;
 import org.osgi.framework.Constants;
 import org.osgi.framework.Version;
 import org.osgi.framework.namespace.HostNamespace;
-import org.osgi.framework.wiring.BundleCapability;
 import org.osgi.framework.wiring.BundleRequirement;
 import org.osgi.framework.wiring.BundleRevision;
 import org.osgi.framework.wiring.BundleWire;
 import org.osgi.framework.wiring.BundleWiring;
-import org.osgi.resource.Capability;
 import org.osgi.resource.Namespace;
 import org.osgi.resource.Resource;
 
@@ -176,13 +172,15 @@ public class DependencyManager {
 			throw new AssertionError("Cannot combine INCLUDE_ALL_FRAGMENTS and INCLUDE_NON_TEST_FRAGMENTS"); //$NON-NLS-1$
 		}
 
+
+
 		Set<BundleDescription> closure = new HashSet<>(bundles.size() * 4 / 3 + 1);
 		Queue<BundleDescription> pending = new ArrayDeque<>(bundles.size());
-		Map<String, List<BundleCapability>> provided = new HashMap<>();
+		Capabilities capabilities = new Capabilities();
 
 		// initialize with given bundles
 		for (BundleDescription bundle : bundles) {
-			addNewRequiredBundle(bundle, closure, pending, provided);
+			addNewRequiredBundle(bundle, closure, pending, capabilities);
 		}
 		// perform exhaustive iterative bfs for required wires
 		while (!pending.isEmpty()) {
@@ -197,7 +195,7 @@ public class DependencyManager {
 				// A fragment's host is already required by a wire
 				for (BundleDescription fragment : bundle.getFragments()) {
 					if (includeAllFragments || !isTestWorkspaceProject(fragment)) {
-						addNewRequiredBundle(fragment, closure, pending, provided);
+						addNewRequiredBundle(fragment, closure, pending, capabilities);
 					}
 				}
 			}
@@ -218,7 +216,7 @@ public class DependencyManager {
 			List<BundleWire> requiredWires = wiring.getRequiredWires(null);
 			for (BundleWire wire : requiredWires) {
 				BundleRequirement requirement = wire.getRequirement();
-				if (isSingle(requirement) && isAlreadyProvided(requirement, provided)) {
+				if (isSingle(requirement) && capabilities.findCapabilities(requirement).size() > 0) {
 					continue;
 				}
 				BundleRevision declaringBundle = requirement.getRevision();
@@ -231,7 +229,7 @@ public class DependencyManager {
 				// Use revision of required capability to support the case if
 				// fragments contribute new packages to their host's API.
 				if (provider instanceof BundleDescription requiredBundle && (includeOptional || !isOptional(requirement))) {
-					addNewRequiredBundle(requiredBundle, closure, pending, provided);
+					addNewRequiredBundle(requiredBundle, closure, pending, capabilities);
 				}
 			}
 		}
@@ -243,29 +241,12 @@ public class DependencyManager {
 				.getOrDefault(Namespace.REQUIREMENT_CARDINALITY_DIRECTIVE, Namespace.CARDINALITY_SINGLE));
 	}
 
-	protected static boolean isAlreadyProvided(BundleRequirement requirement,
-			Map<String, List<BundleCapability>> provided) {
-		List<BundleCapability> list = provided.get(requirement.getNamespace());
-		if (list != null && !list.isEmpty()) {
-			for (BundleCapability bundleCapability : list) {
-				if (requirement.matches(bundleCapability)) {
-					return true;
-				}
-			}
-		}
-		return false;
-	}
-
 	private static void addNewRequiredBundle(BundleDescription bundle, Set<BundleDescription> requiredBundles,
-			Queue<BundleDescription> pending, Map<String, List<BundleCapability>> provided) {
-		if (bundle != null && bundle.isResolved() && !bundle.isRemovalPending() && requiredBundles.add(bundle)) {
+			Queue<BundleDescription> pending, Capabilities capabilities) {
+		if (bundle != null && bundle.isResolved() && !bundle.isRemovalPending() && requiredBundles.add(bundle)
+				&& !"org.eclipse.osgi".equals(bundle.getSymbolicName())) { //$NON-NLS-1$
 			pending.add(bundle);
-			List<Capability> capabilities = bundle.getCapabilities(null);
-			for (Capability capability : capabilities) {
-				if (capability instanceof BundleCapability bc) {
-					provided.computeIfAbsent(capability.getNamespace(), nil -> new ArrayList<>()).add(bc);
-				}
-			}
+			capabilities.addCapabilities(bundle);
 		}
 	}
 
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -20,8 +20,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
```diff
@@ -207,18 +216,42 @@ public static Set<BundleDescription> findRequirementsClosure(Collection<BundleDe
...
+
+	protected static boolean isAlreadyProvided(BundleRequirement requirement,
+			Map<String, List<BundleCapability>> provided) {
+		List<BundleCapability> list = provided.get(requirement.getNamespace());
+		if (list != null && !list.isEmpty()) {
+			for (BundleCapability bundleCapability : list) {
+				if (requirement.matches(bundleCapability)) {
+					return true;
+				}
+			}
```
HannesWell, line 240: This will increase the runtime characteristics of the collection of dependencies by one or orders as for each single requirement the entire namespace is searched again. And namespaces like `osgi.wiring.package` or `osgi.wiring.bundle` the namespace is huge, basically all packages or bundles available reside their and it would be checked with a linear search for each single-requirement.

To mitigate that this check should either exclude certain known large namespaces like `package` and `bundle` that we are usually not interested anyways (ok, well excluding duplicates could also be interesting for bundles/packages).
Or by introducing a second look-up level that includes the bundle/package name, but I think that's not trivial just on general requirements/capabilities as one has to decompose the filters.

HannesWell (changes requested):
The problem that a product only contains a sub-set of all bundles in the TP and that therefore the wiring can be different than in the TP is something I thought about a while ago.
While filtering out already provided capabilities as you suggested should have the desired effect (since it's performing a breath-first-search through the complete wiring to compute the closure), I'm not sure if there are cases where just skipping capabilities and thus bundles could break something (plus we have the potential performance problem described below).

When I thought about this general problem, the solution alternative solution I found was to create a new state/wiring resolution for the launch where the bundles explicitly listed in the product (directly or through feature containments) are the 'roots' and are therefore preferred by the resolver. But I havn't checked if it's possible to define such 'roots' or maybe adding them first is already sufficient?

> @HannesWell something for RC1?

Better not, as this area has been shown to contain surprises in the past that one doesn't really expect.

laeubi:
>  I'm not sure if there are cases where just skipping capabilities and thus bundles could break something

I can't think about it right now, but the users always has the chance to influence "the roots" (but can't exclude things what leads to the problem described here).

Basically this is how Tycho/P2 works when building a product, first include the roots and then everything what is required from there (just that it uses deep first if I rember correctly).

HannesWell:
> Basically this is how Tycho/P2 works when building a product, first include the roots and then everything what is required from there (just that it uses deep first if I rember correctly).

I think that's what we should do in PDE as well to get a clean new state for a launch.
Unfortunately I didn't have the time to look into this before my vacation, but I'll do it when I'm back. Sorry for the delay.

merks:
I made changes to this version to build an index using information extracted via the APIs:

[DependencyManager.java.txt](https://github.com/user-attachments/files/19687139/DependencyManager.java.txt)

The performance is then `elapsed=236` and `elapsed=100` for the second time.

HannesWell:
Thanks for having a look Ed.

I also had a look at this, respectively my suggestion to create a new state were the explicit content is preferred, and noticed that the launch validation already creates a new state.So I think that wouldn't take too long, at least I never noticed that. Furthermore I noticed that the `DependencyManager` is called more often than it seems necessary when launching a product.
I'll try to work out a specific proposal as soon as possible, because I think that would be the cleaner solution. But if that doesn't work out eventually, I think this is the way we have to go.

merks:
@HannesWell 

M2 is next week so I'm concerned that if this doesn't move forward that we will miss this release cycle due to riskiness.  Whatever we change, we should ensure it's tested out in the wild sooner rather than later.  

I don't really have a clue of what you think would be a better alternative implementation so I cannot help with that.

HannesWell:
Thanks for the heads-up.

> I don't really have a clue of what you think would be a better alternative implementation so I cannot help with that.

I have just create a PR with my alternative solution: https://github.com/eclipse-pde/eclipse.pde/pull/1725
Unfortunately it does not yet work as expected.

laeubi:
I'm now reusing a class from Equinox for the index. I'll prepare a PR with the adjustments as well.

Given that using the resolver will not likely be faster and is currently having issues, I would propose we merge this one for M2 and maybe improve / enhance / replace it with something else later on.

merks:
Testing this with my EMF workspace I have this one problem:

image

Which leads to real problems in the log:
```
!ENTRY org.eclipse.pde.api.tools.ui 4 0 2025-05-03 07:29:28.485
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.pde.api.tools.ui [315]
  Unresolved requirement: Require-Bundle: org.eclipse.pde.ui; bundle-version="[3.14.200,4.0.0)"
    -> Bundle-SymbolicName: org.eclipse.pde.ui; bundle-version="3.16.100.v20250320-0551"; singleton:="true"
       org.eclipse.pde.ui [336]
         Unresolved requirement: Import-Package: org.bndtools.templating; version="[2.0.0,3.0.0)"
           -> Export-Package: org.bndtools.templating; bundle-symbolic-name="org.bndtools.templating"; bundle-version="7.1.0.202411251545"; version="2.0.0"; uses:="aQute.service.reporter,org.eclipse.core.runtime,org.osgi.framework,org.osgi.service.metatype,org.osgi.util.promise"
              org.bndtools.templating [73]
                Unresolved requirement: Import-Package: com.google.appengine.api; resolution:="optional"
                Unresolved requirement: Import-Package: com.google.apphosting.api; resolution:="optional"
                Unresolved requirement: Import-Package: javax.annotation
         Unresolved requirement: Require-Bundle: org.eclipse.pde.bnd.ui; bundle-version="1.0.0"
           -> Bundle-SymbolicName: org.eclipse.pde.bnd.ui; bundle-version="1.2.100.v20250322-0546"; singleton:="true"
              org.eclipse.pde.bnd.ui [316]
                Unresolved requirement: Import-Package: org.bndtools.templating; version="[2.0.0,3.0.0)"
                  -> Export-Package: org.bndtools.templating; bundle-symbolic-name="org.bndtools.templating"; bundle-version="7.1.0.202411251545"; version="2.0.0"; uses:="aQute.service.reporter,org.eclipse.core.runtime,org.osgi.framework,org.osgi.service.metatype,org.osgi.util.promise"

	at org.eclipse.osgi.container.Module.start(Module.java:495)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:2111)
	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:146)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:2102)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:2042)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:2004)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1916)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:341)
```
This problem does not happen with the *.txt version I attached above:

image

So something is not quite greedy enough...

merks:
If I read this correct, it appears that the requirement is ignored because something believes that it's coming from the system bundle:

image

image

merks:
The system bundle exports the union of all EEs:

image

So once the system bundle is added to the capabilities, then many javax things that were removed from the JDK to jakarta will become problematic.

Avoiding to add the system bundle to the capabilities avoids this problem.

```
	private static void addNewRequiredBundle(BundleDescription bundle, Set<BundleDescription> requiredBundles,
			Queue<BundleDescription> pending, Capabilities capabilities) {
		if (bundle != null && bundle.isResolved() && !bundle.isRemovalPending() && requiredBundles.add(bundle)
				&& !"org.eclipse.osgi".equals(bundle.getSymbolicName())) { //$NON-NLS-1$
			pending.add(bundle);
			capabilities.addCapabilities(bundle);
		}
	}
```

Yes, that might lead to bundles being included that are not strictly needed because the JDK provides the required package.  If we properly populated the state with the actual EE, we'd not need this, but that's a much bigger design change than simply being a bit more pessimistic here.

HannesWell:
> Given that using the resolver will not likely be faster and is currently having issues, I would propose we merge this one for M2 and maybe improve / enhance / replace it with something else later on.

What issues do you mean https://github.com/eclipse-pde/eclipse.pde/pull/1725 has? You have mentioned some points that are not ideal, but I think that has to be accepted and in my testing it worked as desired. But it would be nice if @merks would confirm that as well.

The fundamental problems I see with this approach still exists:
1. it affects all callers of `DependencyManager.findRequirementsClosure()`. So even if using the resolve is not faster, it only affects the launching not all other callers (but from my experiences with the validation and TP resolution it is usually fast).
2. Just ignoring providers could lead to resolution errors at runtime due to class-space violations (as mentioned in https://github.com/eclipse-pde/eclipse.pde/pull/1725#issuecomment-2848538474)
3. The solution in https://github.com/eclipse-pde/eclipse.pde/pull/1725 is also less complex.

merks:
I tested #1725 and it works well for my EMF scenario...

laeubi:
> What issues do you mean https://github.com/eclipse-pde/eclipse.pde/pull/1725 has?

You previously wrote 

> I have just create a PR with my alternative solution: https://github.com/eclipse-pde/eclipse.pde/pull/1725
> **Unfortunately it does not yet work as expected**.

so I assumed it is/was not ready yet.

> it affects all callers of DependencyManager.findRequirementsClosure(). So even if using the resolve is not faster, it only affects the launching not all other callers (but from my experiences with the validation and TP resolution it is usually fast).

All callers of `findRequirementsClosure` currently getting wrong results, so I'm not sure its a bad thing to fix it.

> The solution in https://github.com/eclipse-pde/eclipse.pde/pull/1725 is also less complex.

If we ignore the copied code from Equinox, it is 30 simple lines versus 70 lines of highly complex lamda and streaming so I doubt 'less complex' is really true here anyways my main goal is that we get some solution to the problem. It was initially meant to be merged/fixed as soon as master opens, now we have already missed M1 and maybe even will miss M2 as well.

merks:
Let's not miss M2!

HannesWell:
> You previously wrote
> 
> > I have just create a PR with my alternative solution: #1725
> > **Unfortunately it does not yet work as expected**.
> 
> so I assumed it is/was not ready yet.

That was referring to the very first state of that PR two weeks ago and was fixed soon afterwards.

> > it affects all callers of DependencyManager.findRequirementsClosure(). So even if using the resolve is not faster, it only affects the launching not all other callers (but from my experiences with the validation and TP resolution it is usually fast).
> 
> All callers of `findRequirementsClosure` currently getting wrong results, so I'm not sure its a bad thing to fix it.

That statement is not true. They get the correct result according to their current wiring. But as the wiring state is not necessarily unambiguous different solutions are possible. In the general case of the TP-state there is simply not a preference for a specific solution(yet, theoretically we could apply a similar schema when creating the TP-state based on their definition). But in case of a launch the already included bundles are (kind of) preferred, as other providers are (maybe) not necessary and could potentially lead to problems. 

> my main goal is that we get some solution to the problem. It was initially meant to be merged/fixed as soon as master opens, now we have already missed M1 and maybe even will miss M2 as well.

From my POV https://github.com/eclipse-pde/eclipse.pde/pull/1725#issuecomment-2848815745 is ready :)

laeubi:
Then lets close this and merge the other PR asap!

Issue #877, opened 2023-11-03
First class OSGi Testing Support


laeubi:
[OSGi Testing Support](https://github.com/osgi/osgi-test#osgi-testing-support) is a great library for testing OSGi applications, PDE should add first-class support as it does for OSGi Annotations already.

gireeshpunathil:
@laeubi - could you explain what the support entails? like a break-down of activities. I can then attempt expand those as check lists to track coverage etc. that way new and enthusiastic contributors can pick those up and add contributions.

PR #1619, opened 2025-02-16
Add OSGi test classpath support

laeubi:
Similar to what we have for OSGi annotations, PDE should have OSGi Testing Support as it is a great library for testing OSGi applications. The most hindering thing in this regard is that it is rater complex to setup until one can make the first steps.

This now adds a new classpath contributor that detects if a PDE project is already using JUNIT classpath container and then adds OSGi test dependencies automatically as test dependencies if they are part of the target platform or alternatively from the running platform.

See https://github.com/eclipse-pde/eclipse.pde/issues/877

### DIFF
--- a/features/org.eclipse.pde-feature/feature.xml
+++ b/features/org.eclipse.pde-feature/feature.xml
@@ -28,6 +28,13 @@
       <import plugin="org.bndtools.templates.template"/>
       <import plugin="biz.aQute.repository"/>
       <import plugin="bndtools.jareditor"/>
+      <import plugin="org.osgi.test.common"/>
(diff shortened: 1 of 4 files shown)


features/org.eclipse.pde-feature/feature.xml
```diff
@@ -28,6 +28,13 @@
       <import plugin="org.bndtools.templates.template"/>
       <import plugin="biz.aQute.repository"/>
       <import plugin="bndtools.jareditor"/>
+      <import plugin="org.osgi.test.common"/>
```
laeubi, line 31: Should probably become a dedicated feature for easier inclusion in other target platforms and to have an opt-in.

Issue #1621, opened 2025-02-18
Tracing settings disappear after changing launch config


iloveeclipse:
4.35 M3

1) Create Eclipse application launch config, enable tracing and check at least one bundle to enable trace flags.
2) "Apply" and close launch configuration dialog. Launch config file will contain something like
```
<stringAttribute key="checked" value="com.company.bundle1,com.company.bundle2"/>
```
3) Open launch configuration dialog, on the first page change something (like clear workspace data) and "Apply".
4) Run application or inspect saved launch file => **tracing information is removed**.
The launch config will contain
```
<stringAttribute key="checked" value="[NONE]"/>
```

Basically the problem is that every time launch configuration that contains tracing options will be saved **without** opening "Tracing" tab, all tracing data will be lost and tracing will be disabled.

iloveeclipse:
@fedejeanne : probably related to https://github.com/eclipse-pde/eclipse.pde/issues/1250 ?

I remember seeing this problem before, and I assumed it was fixed, but now while switching from 4.30 to 4.35 I see this problem again.

iloveeclipse:
Originally reported https://github.com/eclipse-platform/eclipse.platform/issues/859
Fixed via https://github.com/eclipse-pde/eclipse.pde/pull/925
Some changes reverted in https://github.com/eclipse-pde/eclipse.pde/pull/946
More changes in https://github.com/eclipse-pde/eclipse.pde/pull/1233
Another regression reported https://github.com/eclipse-pde/eclipse.pde/issues/1250
Reverted in https://github.com/eclipse-pde/eclipse.pde/pull/1251

Not clear what caused current regression.

HannesWell:
> Not clear what caused current regression.

I can also reproduce this with I20241120-1800, i.e. the 4.34 release so actually at least not a recent regression.

HannesWell:
The root for the problem you have described is that for the `TracingTab.initializeFrom()` is empty because it can be a long-running operation that blocks potentially blocks the UI for quite some time:
https://github.com/eclipse-pde/eclipse.pde/blob/515f99602838c4f7a7bf848a4f6a719044a7d212/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/TracingTab.java#L82-L85

So it's indeed a regression from https://github.com/eclipse-pde/eclipse.pde/pull/946, but actually not a recent one.
[...]

PR #1252, opened 2024-04-24
Skip the method PluginsTab::performApply until after activation

fedejeanne:
# What it does

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1250 (a regression introduced by #1233) by skipping the whole method `PluginsTab::performApply` before the _Plugins_ tab was activated. The reason for this is that this method changes some values in the `configuration` it receives as a parameter and it does it too early. If the method is called **after** having completely activated the tab then the configuration already has the proper values and they will **not** be overwritten.

# Bug description
When a _Launch configuration_ that has a tab called _Plugins_ (_e.g._ an _Eclipse Application_  or a _JUnit Plug-In Test_) is opened and the _Plugins_ tab is selected, the default values are wrong (see screenshots in https://github.com/eclipse-pde/eclipse.pde/issues/1250#issuecomment-2074775823)

## How to reproduce
* Launch a _JUnit Plugin Test_ with _Right click > Run As... > JUnit Plugin Test_
* Open the _Run configurations_ dialog and select that run configuration that was just created
* Switch to the _Plugins_ tab --> You'll see that the usual default values for **Default Start Level** (`4`) and **Default Auto-Start** (`false`) will be replaced by `1` and `true`, like this:

325209441-c43e1345-675d-40aa-9aff-4c628c2e78b2

## After applying this PR
* The default values should be `4` and `false` again

325209240-92438081-ae90-4ac5-88f3-706c626344ef

DISCLAIMER: the screenshots are from this https://github.com/eclipse-pde/eclipse.pde/issues/1250#issuecomment-2074775823

PR #1622, opened 2025-02-18
Don't apply empty settings from not initialized TracingBlock

iloveeclipse:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1621

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/TracingBlock.java
@@ -122,6 +122,7 @@ public class TracingBlock {
 	private final Map<IPluginModelBase, TracingPropertySource> fPropertySources = new HashMap<>();
 	private FormToolkit fToolkit;
 	private ScrolledPageBook fPageBook;
+	private boolean activated;
 
 	/**
 	 * The last selected item in the list is stored in the dialog settings.
@@ -351,6 +352,7 @@ public class TracingBlock {
 					pluginSelected(null, false);
 				}
 			}
+			activated = true;
 		} catch (CoreException e) {
 			PDEPlugin.logException(e);
 		}
@@ -361,6 +363,11 @@ public class TracingBlock {
 	}
 
 	public void performApply(ILaunchConfigurationWorkingCopy config) {
+		if (!activated) {
+			// nothing to apply as we weren't initialized before,
+			// all the data is empty anyway
+			return;
+		}
 		boolean tracingEnabled = fTracingCheck.getSelection();
 		config.setAttribute(IPDELauncherConstants.TRACING, tracingEnabled);
 		if (tracingEnabled) {


iloveeclipse:
@HeikoKlare, @fedejeanne, @HannesWell : could you please check this PR and #1621?
For me it seem to fix the scenario I described in #1621.
However after so many fixes / reverts in this area the fix here seem to be too simple, so there must be a catch?
Do I miss something?

iloveeclipse:
PS: my goal is to have **some** fix in 4.35 RC1 for #1621, whether it is this PR or revert of something again, or something completely different.

HannesWell:
After a very quick look this change seems sound.
But I'll have a detailed review later this evening and will submit it, if it's fine and save.

HannesWell (approved):
After an in detail review this still looks save and the right thing to do.

fedejeanne:
Hi, sorry for the late reply, I was out-of-office yesterday.

@iloveeclipse thank you for the fix. I actually run into a similar issue with the `PluginsTab` while working on #1252 and introduced a similar solution. The only difference I see with this PR is that you may initialize the tracing block with the same input when you switch back and forth between the tracing tab and other tabs. I avoided this to improve performance in https://github.com/eclipse-pde/eclipse.pde/pull/1252/files#diff-508c84ed339b4ddeadd5db68b30fc9a66cfd8db2c9ebd080678483782bbfaab7

image

Maybe that's also worth doing in your case?

HannesWell:
> The only difference I see with this PR is that you may initialize the tracing block with the same input when you switch back and forth between the tracing tab and other tabs.

Can you explain this in more detail? The initialization of each tab from an existing launch-config should only happen once and at least the change in this PR doesn't change it. I assume you meant this as a general remark or do you think this is not save or wrong?



> I honestly have no time digging into performance optimization in this particular area, as I never observed any performance problems here nor anyone reported that for our product. If you know what to do, please provide a PR.

The main pain-point here is that initializing the 'registry' of available tracing options scans all bundles in the TP and depending on it's size this can take some time. But it's only done once per Eclipse execution (which is actually wrong, because changing the TP may change the list of tracable bundles and each bundles options).
Maybe we should only postpone reading the registry and not the UI initialization (e.g. by using some kind of proxies) to avoid this problem in a more elegant way with less state. But all of this is something for https://github.com/eclipse-pde/eclipse.pde/pull/1252 I think. Sorry Federico I know you are awaiting a review and I'll try my best to do it soon.

fedejeanne:
> Can you explain this in more detail? The initialization of each tab from an existing launch-config should only happen once and at least the change in this PR doesn't change it.

That's correct and this PR doesn't change that so **it's safe** 👍 . The detailed explanation is: 
- The _initialization_ (call to `org.eclipse.debug.ui.ILaunchConfigurationTab.initializeFrom(ILaunchConfiguration)`) happens only **once** and it happens **right at the beginning**, when one selects the run configuration
- The _activation_ (call to `org.eclipse.debug.ui.ILaunchConfigurationTab.activated(ILaunchConfigurationWorkingCopy)`) can happen **any number of times** (0+) and it happens **every time the tab is selected**

> I assume you meant this as a general remark or do you think this is not save or wrong?

It was just a general remark about performance. This PR is safe and I don't see any need to add any changes to it. If performance every becomes a problem, we may add the early return (_Zukunftsmusik_).

> The main pain-point here is that initializing the 'registry' of available tracing options scans all bundles in the TP and depending on it's size this can take some time. But it's only done once per Eclipse execution (which is actually wrong, because changing the TP may change the list of tracable bundles and each bundles options).
> Maybe we should only postpone reading the registry and not the UI initialization (e.g. by using some kind of proxies) to avoid this problem in a more elegant way with less state. But all of this is something for #1252 I think. Sorry Federico I know you are awaiting a review and I'll try my best to do it soon.

I'd say we can leave the detailed discussion for #1252 :-)

PR #1624, opened 2025-02-20
Add support for using a hosted eclipse platform as target

laeubi:
Currently if one uses the "Running Platform" in an Eclipse launched from within an SDK it can happen that a project from the host workspace is chosen as a dependency. In such case there are some issues that only the root folder is added and then compilation fails.

This now enhances the PDEClasspathContainer to detect the situation and adding the output folders instead of the root.


HannesWell:
> Currently if one uses the "Running Platform" in an Eclipse launched from within an SDK it can happen that a project from the host workspace is chosen as a dependency. In such case there are some issues that only the root folder is added and then compilation fails.

Isn't this already handled by the target-platform weaving done by PDE based on the dev-classpath of Equinox?

laeubi:
> Isn't this already handled by the target-platform weaving done by PDE based on the dev-classpath of Equinox?

If then it seem to not work in all situations.

laeubi:
Closing as I haven't seen recent problems here

PR #1625, opened 2025-02-21
Quick fix added for invalid characters in MANIFEST header

nburnwal09:
This is to provide quick fix for invalid header name in MANIFEST.MF
It will provide fix by removing all the invalid characters (characters other than alpha numeric characters, '-' and '_')
for eg. in the below image, added a space after 'Require-Bundle' header
the added quick fix label is 'Remove invalid characters from header'. Once user clicks on 'Finish', it updates the header

image

Attaching another example
image

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2017, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.jface.text.BadLocationException;
@@ -59,6 +60,15 @@ public class JarManifestErrorReporter extends ErrorReporter {
 			}
 		}
 		return null;
+	}
+
+	private boolean isInvalidHeader(String line) {
+		int colonInd = line.indexOf(':');
+		if (colonInd > 0) {
+			String headerPart = line.substring(0, colonInd);
+			return headerPart.matches(".*[^a-zA-Z0-9-_].*"); //$NON-NLS-1$
+		}
+		return false;
 	}
 
 	protected int getPackageLine(IHeader header, ManifestElement element) {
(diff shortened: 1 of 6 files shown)


iloveeclipse:
> invalid header name

You probably mean **attribute**? See official wording at java.util.jar.Manifest.
If so, please change the wording overall - e.g. "Remove invalid characters from attribute name" etc.

nburnwal09:
As far as I have understood, he attribute would mean both, the key: value pair. But I have provided fix for removing invalid characters only from key(Header part).
So, have labeled it as 'Remove invalid characters from header'
Kindly confirm on this @laeubi

laeubi:
[The Jar Specification](https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Name-Value_pairs_and_Sections) explicitly allows both values:

> information contained within the manifest file and signature files is represented as so-called "`name: value`" pairs inspired by the RFC822 standard.  **We also call these pairs headers or attributes.**
> Groups of name-value pairs are known as a "section"

So I think both can be used but "Header" is much more commonly understood.

nburnwal09:
@HannesWell 
Can you please review this PR.

HannesWell:
> Can you please review this PR.

Yes, I'll try all your open PRs and answer your questions in the next days. Sorry for the delay, it has been a busy time.

nburnwal09:
@HannesWell 
Could you please check and review the PR's.

HannesWell:
> Could you please check and review the PR's.

Sorry, same as before. I try my best to review it as soon as possible.

merks:
We can't incorporate merge commits.  Best to force push a rebased commit.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/JarManifestErrorReporter.java
```diff
@@ -61,6 +62,15 @@ private String getHeaderName(String line) {
 		return null;
 	}
 
+	private boolean isInvalidHeader(String line) {
+		int colonInd = line.indexOf(':');
+		if (colonInd > 0) {
+			String headerPart = line.substring(0, colonInd);
+			return headerPart != null && headerPart.matches(".*[^a-zA-Z0-9-_].*"); //$NON-NLS-1$
```
gireeshpunathil, line 69: String.substring doesn't return null - either returns a valid string or throws OOB exception. given you are checking the bounds already, I think the null check is redundant?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/UpdateCorrectHeaderName.java
```diff
@@ -0,0 +1,49 @@
...
+			int colonInd = getLine.indexOf(':');
+
+			if (colonInd > 0) {
+				String userHeader = getLine.substring(0, colonInd);
+				String correctHeader = removeInvalidChars(userHeader);
+				doc.replace(offset, colonInd, correctHeader);
+			}
+
+		} catch (Exception e) {
+
```
gireeshpunathil, line 44: can we log the issue if there is an exception? similar to what you did in the other PR?

Issue #1630, opened 2025-02-27
Unable to launch Eclipse launch config if name contains spaces


iloveeclipse:
Regression in 4.35 via https://github.com/eclipse-pde/eclipse.pde/pull/1565.

1) Create new launch configuration to start Eclipse (default name `New_configuration`)
2) Try to start Eclipse - works.
3) Duplicate launch config to get a name `New_configuration (1)` (note the space)
4) Try to Start Eclipse - fails with error 
```
An error has occurred.  See the log file
/data/runtime-Java_21/.metadata/.plugins/org.eclipse.pde.core/New_configuration%20(1)/1740652776602.log
```

Log file contains this:
```
ESSION Thu Feb 27 11:39:36 CET 2025 ------------------------------------------
!ENTRY org.eclipse.equinox.launcher 4 0 2025-02-27 11:39:36.619
!MESSAGE Exception launching the Eclipse Platform:
!STACK
java.io.FileNotFoundException: Could not find framework under /data/git/equinox/bundles/org.eclipse.equinox.launcher/bin/plugins
        at org.eclipse.equinox.launcher.Main.getBootPath(Main.java:990)
        at org.eclipse.equinox.launcher.Main.basicRun(Main.java:585)
        at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
        at org.eclipse.equinox.launcher.Main.main(Main.java:1454)
```

Reverting e62be0b66a92e07709cea8a13d77ccfd6b99bc47 fixes the launch.

iloveeclipse:
Note, the encoding of the `-dev`  attribute was broken before and not fixed by reverting https://github.com/eclipse-pde/eclipse.pde/commit/e62be0b66a92e07709cea8a13d77ccfd6b99bc47. I'm looking at commit that broke this.

akurtakov:
I'm looking into it. Reverting a series of commits is not what I aim at .

iloveeclipse:
> I'm looking into it. Reverting a series of commits is not what I aim at .

It is one commit + one change from other commit. Should be safe.

akurtakov:
@iloveeclipse I am unable to reproduce the problem on Linux . Do you see it on windows?

merks:
FYI, I see a problem here where the launcher simply doesn't decode a properly encoded URI:

![Image](https://github.com/user-attachments/assets/eec2b9cf-f343-433e-a214-ec081dec6a3e)

[...]

PR #1631, opened 2025-02-27
Issue 1630 fix

iloveeclipse:
Partial revert of 04785c95dbef752959495c19958c8db664f78e21
Revert commit e62be0b66a92e07709cea8a13d77ccfd6b99bc47 because it
caused regression

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1630


iloveeclipse:
Commit 64ac7b0 causes problems which I can't understand right now.

Anyway, https://github.com/eclipse-equinox/equinox/pull/850 seem to be a better candidate as it is smaller and seem to be exactly the missing piece needed for original changes in  e62be0b and 04785c9.

Issue #1632, opened 2025-03-03
Ambiguous button names in dependency properties dialog


ptziegler:
Whenever I need to update the version range of a plugin/import, I find myself having to do the same change twice, because I accidentally clicked on "Cancel", rather than "Close".

![Image](https://github.com/user-attachments/assets/f6157058-4922-4543-bc3c-ac6fa0faecbd)

I find it very confusing to use words with a very similar meaning right next to each other and I think it makes sense to rename "Close" to e,g, "Apply".

ptziegler:
This should be where the name is defined:

https://github.com/eclipse-pde/eclipse.pde/blob/515f99602838c4f7a7bf848a4f6a719044a7d212/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties#L1456-L1458

vogella:
+1 for this proposal. PR welcome

PR #1634, opened 2025-03-03
Rename "Close" to "Apply" in DependencyPropertiesDialog

ptziegler:
This changes the name of the OK button to "Close", to avoid any confusion with the similarly named "Cancel" button.

Closes https://github.com/eclipse-pde/eclipse.pde/issues/1632

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2024 IBM Corporation and others.
+ * Copyright (c) 2014, 2025 IBM Corporation and others.
  *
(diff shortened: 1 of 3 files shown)


ptziegler:
I think the buttons are sorted lexicographically, so the OK button is now the first button. This might mess with muscle memory, but I'm not sure if it's worth changing, especially because you have the same in e.g. the property pages.

image

merks:
The button order is different (opposite) on MacOs than on Windows/Linux.  Buttons are not sorted...

https://stackoverflow.com/questions/6667271/standard-placement-order-for-common-buttons

ptziegler:
> Did you test / develop on different platforms? As @merks said, button are OS specific

The order is definitely OS specific. On Windows, the order is "Apply" "Cancel" but on Linux it's "Cancel" "Apply". Weird... 🤷

image

vogella:
Test / build failure seems unrelated.

Thanks @ptziegler

akurtakov:
Please don't merge things until we open the stream. I don't want such merges to break build due to missing version bumps or etc.

PR #1639, opened 2025-03-05
Quick fix to provide an option to update user directive to mandatory singleton true directive(if not present)

nburnwal09:
This is to provide quick fix:
In case, user has not entered any singleton directive to true, this change will provide an option to update the same to singleton:=true
this will provide the update option only when there is one set of directive. For multiple sets, we already have an option to add 'singleton' directive to true.

for example: 
before the fix:
image

after applying the **Update** fix:
image


nburnwal09:
Have one question here:

Have added the if check in UpdateSingletonToSymbolicName.java(the new class) similar to this in AddSingletonToSymbolicName.java file

https://github.com/eclipse-pde/eclipse.pde/blob/f941e23e3707b3f237d275ce34eb2d6cb3589ea9/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/AddSingletonToSymbolicName.java#L55-L58

I guess this is for maintaining the backward compatibility.
Can someone clarify on this?

nburnwal09:
@HannesWell 
Can you review this PR when you some time

gireeshpunathil (approved):

PR #1706, opened 2025-03-24
Jenkinsfile: Use Tycho 4.0.13 way to display api tools warnings

sratz:
Also make sure the issues are published even in case of a failed build, as otherwise they are not visualized.


HannesWell (approved):
Looks good. Thank you.

PR #1707, opened 2025-03-26
Remove preference listener if manager is shut down or project is deleted

ptziegler:
The PreferenceChangeListener that is added when a project is created/opened or when the plugin is started needs to be removed when the project is deleted, in order to make it eligible for garbage collection.


ptziegler:
The memory leak that is caused by this as seen in https://github.com/eclipse-tycho/tycho/pull/4906

image

ptziegler:
I'm fairly certain the new Maven warnings are unrelated...

Issue #1708, opened 2025-03-28
Plug-in dependencies for runtime workspaces is broken in Eclipse (Modeling) 2025-03 R on Windows


maxkratz:
Hi,
I found a bug when using plug-in dependencies in runtime workspaces with Eclipse (Modeling) 2025-03 R on Windows. I opened an [issue](https://github.com/eclipse-emf/org.eclipse.emf/issues/65) at the EMF repository but I [was told](https://github.com/eclipse-emf/org.eclipse.emf/issues/65#issuecomment-2760831255) this could be a PDE problem.

Copy paste from the other issue:

> Bug description:
> 
> Normally, you can create a plug-in project in your development workspace and let it export some Java packages. The import and export of such projects are managed via the MANIFEST.MF file(s). On your runtime workspace, you can have other projects that have the projects of the development workspace as a dependency defined in their MANIFEST.MF. Eclipse should normally resolve the dependencies automatically and compile the project(s) within the runtime workspace correctly.
> 
> However, since updating to Eclipse Modeling 2025-03 R this seems to be broken on Windows-based systems only.
> 
> I documented it with a description and an MWE consisting of two Eclipse projects in this repository: https://github.com/maxkratz/eclipse-2025-03-runtime-bug-mwe

**If this is not the right place to raise this issue, please let me know.** I would happily get a pointer to the correct repository/mailing list, etc.
I appreciate any help you can provide.

merks:
Is it possible that your launch configuration has a space in the name?  I see that the existing logic is not properly decoding the %20 when converting the URL to a path pass to the File constructor:

![Image](https://github.com/user-attachments/assets/afbac1e6-100a-4d1e-8e2f-14f408e8da16)

I know there have been changes in other places to create well-formed URLs.  A URL tolerates a space, a URI does not and throws an exception....

maxkratz:
> Is it possible that your launch configuration has a space in the name?

Yes, my home folder on my Windows systems contains a space: `C:\Users\Maximilian Kratz\...`. I will try to reproduce the issue with workspaces that do not have a space in their paths.

maxkratz:
> > Is it possible that your launch configuration has a space in the name?
> 
> Yes, my home folder on my Windows systems contains a space: `C:\Users\Maximilian Kratz\...`. I will try to reproduce the issue with workspaces that do not have a space in their paths.

[...]

PR #1709, opened 2025-03-28
Fix TargetWeaver.getDevProperties() to decode fgDevPropertiesURL

merks:
https://github.com/eclipse-pde/eclipse.pde/issues/1708


HannesWell:
Thanks for this fix Ed!
Once available, we can probably use https://github.com/eclipse-equinox/equinox/pull/691 as a unified solution.

PR #1712, opened 2025-04-01
Remove listeners in MinimalState if bundle is stopped

ptziegler:
This change adapts the listeners that have been added with 24e9bb548b0de8c8ea79ee70d28096e28202da31 so that they are removed again, once the bundle is shut down and thus makes the classloader eligible for garbage collection.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2024 IBM Corporation and others.
+ * Copyright (c) 2005, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -37,6 +37,7 @@ import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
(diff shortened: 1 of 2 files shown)


ptziegler:
One of the other memory leaks in PDE I've noticed:
image

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -268,39 +301,23 @@ protected boolean initializePlatformProperties() {
...
-			public void defaultVMInstallChanged(IVMInstall previous, IVMInstall current) {
-				triggerSystemPackagesReload();
-			}
-		});
+	private static IEclipsePreferences getJdtLaunchingPreferences() {
+		// Listen to changes in the available VMInstalls and
+		// ExecutionEnvironment defaults
+		@SuppressWarnings("restriction")
+		String nodeQualifier = org.eclipse.jdt.internal.launching.LaunchingPlugin.ID_PLUGIN;
+		return InstanceScope.INSTANCE.getNode(nodeQualifier);
```
HannesWell, line 320: Copying the comment doesn't seem to be necessary and if we have this part extracted into a method, it can become a one-liner and the warning-suppression can be moved to the method.
```suggestion
	@SuppressWarnings("restriction")
	private static IEclipsePreferences getJdtLaunchingPreferences() {
		return InstanceScope.INSTANCE.getNode(org.eclipse.jdt.internal.launching.LaunchingPlugin.ID_PLUGIN);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -268,39 +301,23 @@ protected boolean initializePlatformProperties() {
...
+		IEclipsePreferences launchingNode = getJdtLaunchingPreferences();
+		launchingNode.addPreferenceChangeListener(PREF_CHANGE_LISTENER);
+		JavaRuntime.addVMInstallChangedListener(VM_CHANGED_LISTENER);
+	}
 
-			@Override
-			public void vmAdded(IVMInstall vm) {
-				triggerSystemPackagesReload();
-			}
+	public static void shutdown() {
```
HannesWell, line 309: No need to have this public.
```suggestion
	static void shutdown() {
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -82,6 +83,38 @@ public class MinimalState {
...
+		public void vmAdded(IVMInstall vm) {
+			triggerSystemPackagesReload();
+		}
+
+		@Override
+		public void defaultVMInstallChanged(IVMInstall previous, IVMInstall current) {
+			triggerSystemPackagesReload();
+		}
+	};
+
```
HannesWell, line 117: Could you please move this static final fields to the very top of this class to follow the usual convention?

HannesWell (commented):
Thank you for the enhancement. Technically it looks good, I just have a few code-style nit-picks below.

HannesWell (approved):
I have now applied the requested changes to this PR.

Thank you for this contribution!

ptziegler:
Whoops. Apologies, I didn't see you comment. Thanks for applying the the changes and merging everything.

iloveeclipse:
This caused regression, please check https://github.com/eclipse-pde/eclipse.pde/issues/1714

Issue #1714, opened 2025-04-07
BundleException on shutdown


iloveeclipse:
Using https://download.eclipse.org/eclipse/downloads/drops4/I20250406-1800/.
Stated SDK, installed few bundles, restarted.
After restart, I see error log from previous session contains this error:
```
eclipse.buildId=4.36.0.I20250406-1800
java.version=21.0.6-7
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -os linux -ws gtk -arch x86_64

org.eclipse.pde.core
Error
Mon Apr 07 07:10:34 CEST 2025
FrameworkEvent ERROR

org.osgi.framework.BundleException: Exception in org.eclipse.pde.internal.core.PDECore.stop() of bundle org.eclipse.pde.core.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.stop(BundleContextImpl.java:916)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.stopWorker0(EquinoxBundle.java:1080)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.stopWorker(EquinoxBundle.java:399)
	at org.eclipse.osgi.container.Module.doStop(Module.java:699)
	at org.eclipse.osgi.container.Module.stop(Module.java:557)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.decStartLevel(ModuleContainer.java:2154)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:2023)
	at org.eclipse.osgi.container.SystemModule.stopWorker(SystemModule.java:281)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule.stopWorker(EquinoxBundle.java:225)
	at org.eclipse.osgi.container.Module.doStop(Module.java:699)
	at org.eclipse.osgi.container.Module.stop(Module.java:557)
	at org.eclipse.osgi.container.SystemModule.stop(SystemModule.java:212)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule$1.run(EquinoxBundle.java:244)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.ExceptionInInitializerError
[...]

PR #1716, opened 2025-04-07
Inline call to BundleHelper.getPlatformAdmin().getFactory()

ptziegler:
As a result of 14f66ae9e57d5f7a03973a41f01ef77a6f77c474, it might happen
that the MinimalState is shut down, even though it was never started. In
such a case, the class is implicitly loaded while stopping the plug-in.

When loading this class, the "stateObjectFactory" is initialized using
the PlatformAdmin provided by the PDE Build bundle. If this bundle is
stopped before the PDE Core bundle, an exception is thrown because the
PlatformAdmin is no longer available.

The BundleHelper is already cached by the PDE Build plug-in and can
therefore be called directly, making the "stateObjectFactory" field
obsolete.

Closes https://github.com/eclipse-pde/eclipse.pde/issues/1714

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
@@ -49,7 +49,6 @@ import org.eclipse.jdt.launching.environments.IExecutionEnvironment;
 import org.eclipse.osgi.service.resolver.BundleDescription;
 import org.eclipse.osgi.service.resolver.State;
 import org.eclipse.osgi.service.resolver.StateDelta;
-import org.eclipse.osgi.service.resolver.StateObjectFactory;
 import org.eclipse.osgi.util.ManifestElement;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
@@ -111,16 +110,10 @@ public class MinimalState {
 
 	private boolean fNoProfile;
 
-	protected static StateObjectFactory stateObjectFactory;
-
 	protected String fSystemBundle = IPDEBuildConstants.BUNDLE_OSGI;
 
-	static {
-		stateObjectFactory = BundleHelper.getPlatformAdmin().getFactory();
-	}
-
 	protected MinimalState(MinimalState state) {
-		this.fState = stateObjectFactory.createState(state.fState);
+		this.fState = BundleHelper.getPlatformAdmin().getFactory().createState(state.fState);
 		this.fState.setPlatformProperties(state.fState.getPlatformProperties());
 		this.fState.setResolver(BundleHelper.getPlatformAdmin().createResolver());
 		this.fId = state.fId;
@@ -213,8 +206,8 @@ public class MinimalState {
 		try {
 			// OSGi requires a dictionary over any map
 			Dictionary<String, String> dictionaryManifest = FrameworkUtil.asDictionary(manifest);
-			BundleDescription descriptor = stateObjectFactory.createBundleDescription(fState, dictionaryManifest,
-					bundleLocation.getAbsolutePath(), bundleId == -1 ? getNextId() : bundleId);
+			BundleDescription descriptor = BundleHelper.getPlatformAdmin().getFactory().createBundleDescription(fState,
+					dictionaryManifest, bundleLocation.getAbsolutePath(), bundleId == -1 ? getNextId() : bundleId);
 			// new bundle
 			if (bundleId == -1 || !fState.updateBundle(descriptor)) {
 				fState.addBundle(descriptor);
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -111,16 +115,10 @@ public void defaultVMInstallChanged(IVMInstall previous, IVMInstall current) {
...
-
 	protected String fSystemBundle = IPDEBuildConstants.BUNDLE_OSGI;
 
-	static {
-		stateObjectFactory = BundleHelper.getPlatformAdmin().getFactory();
-	}
-
 	protected MinimalState(MinimalState state) {
-		this.fState = stateObjectFactory.createState(state.fState);
+		this.fState = SingletonHolder.STATE_OBJECT_FACTORY.createState(state.fState);
```
laeubi, line 121:
```suggestion
		this.fState = BundleHelper.getPlatformAdmin().getFactory().createState(state.fState);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -213,8 +211,8 @@ public BundleDescription addBundle(Map<String, String> manifest, File bundleLoca
 		try {
 			// OSGi requires a dictionary over any map
 			Dictionary<String, String> dictionaryManifest = FrameworkUtil.asDictionary(manifest);
-			BundleDescription descriptor = stateObjectFactory.createBundleDescription(fState, dictionaryManifest,
-					bundleLocation.getAbsolutePath(), bundleId == -1 ? getNextId() : bundleId);
+			BundleDescription descriptor = SingletonHolder.STATE_OBJECT_FACTORY.createBundleDescription(fState,
```
laeubi, line 214:
```suggestion
			BundleDescription descriptor = BundleHelper.getPlatformAdmin().getFactory().createBundleDescription(fState,
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEState.java
```diff
@@ -77,7 +77,7 @@ public PDEState(URI[] target, boolean addResolver, boolean removeDuplicates, IPr
 	}
 
 	private void createNewTargetState(boolean resolve, URI[] uris, IProgressMonitor monitor) {
-		fState = stateObjectFactory.createState(resolve);
+		fState = SingletonHolder.STATE_OBJECT_FACTORY.createState(resolve);
```
laeubi, line 80:
```suggestion
		fState = BundleHelper.getPlatformAdmin().getFactory().createState(resolve);
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
```diff
@@ -66,6 +66,10 @@
 
 public class MinimalState {
 
+	protected static class SingletonHolder {
+		protected static final StateObjectFactory STATE_OBJECT_FACTORY = BundleHelper.getPlatformAdmin().getFactory();
+	}
```
laeubi, line 71:
```suggestion
	}
```

laeubi (commented):
Given that `BundleHelper.getPlatformAdmin().getFactory()` is already cached and we only have three call sites I wonder if it would not be better to replace calls directly with

`BundleHelper.getPlatformAdmin().getFactory()`

as in the suggestions.

ptziegler:
> Given that `BundleHelper.getPlatformAdmin().getFactory()` is already cached and we only have three call sites I wonder if it would not be better to replace calls directly with
> 
> `BundleHelper.getPlatformAdmin().getFactory()`
> 
> as in the suggestions.

That makes sense. The singleton in the BundleHelper is implicitly synchronized by the `Bundle.start()` and `Bundle.stop()` methods and removing excessive indirections is always a good thing.

I've applied your changes and updated the commit title/description.

laeubi (approved):
Looks good to me, `factory` can become a (protected) instance field then we even have only one reference at all (but not a static one), but even with this for I don't think we will see any issues.
Only "addBundle" is something that might be called more often here.

ptziegler:
The four new compiler warnings are about the parameters of the [RequiredPackageVersionRange](https://github.com/eclipse-pde/eclipse.pde/blob/master/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/builder/RequiredPackageVersionChange.java) record, which is unrelated to this PR from my perspective also wrong.

laeubi:
@srikanth-sankaran can you probably have a look? Looks like all constructor parameter of a record are not marked as "unused".

iloveeclipse:
> That seems to be already fixed via [eclipse-jdt/eclipse.jdt.core#3905](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3905). So I think it's save to merge the PR?

If there are no other pending actions, yes.

HannesWell (approved):
Looks good. Thanks for the fix and I'm glad that the review lead to an even simpler solution. :)

PR #1718, opened 2025-04-08
Fix Inefficient regular expression

laeubi:
Some regular expressions take a long time to match certain input strings to the point where the time it takes to match a string of length n is proportional to nk or even 2n. Such regular expressions can negatively affect performance, or even allow a malicious user to perform a Denial of Service ("DoS") attack by crafting an expensive input string for the regular expression to match.

This replaces the problematic regular expression by a little parser that checks for the same results as the regular expression.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDETextHelper.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDETextHelper.java
@@ -15,8 +15,6 @@ package org.eclipse.pde.internal.core.util;
 
 import java.util.Map;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public class PDETextHelper {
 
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDETextHelper.java
```diff
@@ -283,13 +281,49 @@ private static boolean isValidTagAttributeList(String text) {
...
+			return false;
+		}
+		int length = key.length();
+		for (int i = 0; i < length; i++) {
+			char c = key.charAt(i);
+			if (!isValidAttributeKeyChar(c)) {
+				return false;
+			}
+		}
+		return true;
```
HannesWell, line 317: Would a stream-pipeline fit here?
```suggestion
		return key.chars().allMatch(PDETextHelper::isValidAttributeKeyChar);
```
With the suggestions below one could maybe even inline the isValidAttributeKeyChar method?

HannesWell (commented):
This change seems reasonable to me. Especially as the Pattern is currently very difficult to understand, it's probably easier to read the sequential code (therefore I haven't compared the logi ...).

I just have two nitpicks below.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDETextHelper.java
```diff
@@ -283,13 +281,49 @@ private static boolean isValidTagAttributeList(String text) {
...
+	}
 
-		//			              space  attributeName      space = space  "attributeValue" space /
-		String patternString = "^([\\s]+[A-Za-z0-9_:\\-\\.]+[\\s]?=[\\s]?\".+?\")*[\\s]*[/]?$"; //$NON-NLS-1$
-		Pattern pattern = Pattern.compile(patternString);
-		Matcher matcher = pattern.matcher(text);
-		// Determine whether the given attribute list matches the pattern
-		return matcher.find();
+	private static boolean isValidAttributeKeyChar(char c) {
+		return (c >= 'A' && c <= 'Z') || //
```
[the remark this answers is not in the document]
laeubi: I tried to replicate what the RegEx is doing here, Character.isUpperCase/isLowerCase/ and alike might include more characters as before and I haven really checked why we test this particular things.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PDETextHelper.java
HannesWell: What about this part?

PR #1721, opened 2025-04-15
Allow  Eclipse 4 contributions to select svg files via the find dialog

vogella:
With the svg support in SWT we should also allow the selection of svg files for Eclipse 4 application model elements.

Fixes #https://github.com/eclipse-platform/eclipse.platform.ui/issues/2911


merks:
Yes it’s really great. Plus @laeubi’s cleanups and tp updater.  These guys make things that streamline our work. 🥇

HannesWell:
> @HannesWell I know its been a while since you created it, but your version update tool is still amazing

Thank you and thank you for this contribution.

vogella:
> Yes it’s really great. Plus @laeubi’s cleanups and tp updater. These guys make things that streamline our work. 🥇

Thanks for pointing this out @merks . I agree that @laeubi clean-up automation is also amazing, especially since @jjohnstn continues to improve and fix the JDT clean-ups. Together with the Github build action validation the current automation of the work is outstanding.

PR #1723, opened 2025-04-17
Use svg files in the Eclipse RCP template for e4

vogella:
Adds Require-Capability: eclipse.swt;filter:="(image.format=svg)" to the MANIFEST and replaces png with svg files.


vogella:
Require-Capability: eclipse.swt;filter:="(image.format=svg)" prevents the plug-in from starting correctly, if I remove it its starts and svg files are shown correctly.

I used the approach supported by @laeubi from https://github.com/eclipse-platform/eclipse.platform.ui/pull/2925/files#diff-2e3ec9ee6782044d4a3f43dcaff53f6d43fd66f7a91a804ba02246ca9be74b4dR26

Anyone wrong with the MANIFEST.MF you can spot? @Michael5601

````
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: F1
Bundle-SymbolicName: f1;singleton:=true
Require-Bundle: org.eclipse.swt,
 org.eclipse.e4.ui.model.workbench,
 org.eclipse.jface,
 org.eclipse.e4.ui.services,
 org.eclipse.e4.ui.workbench,
 org.eclipse.e4.core.di,
 org.eclipse.e4.ui.di,
 org.eclipse.e4.core.contexts,
 org.eclipse.core.runtime
Bundle-RequiredExecutionEnvironment: JavaSE-21
Automatic-Module-Name: f1
Import-Package: jakarta.annotation;version="[2.1.0,3.0.0)",
 jakarta.inject;version="[2.0.0,3.0.0)"
Require-Capability: eclipse.swt;filter:="(image.format=svg)"
````

Michael5601:
> 
> Anyone wrong with the MANIFEST.MF you can spot? @Michael5601

Should be correct as you just need to add the single `Requre-Capability` line. @HannesWell also had this problem and assumed it was a PDE bug. We still need to investigate this as far as I know.

vogella:
As for now I will take out Require-Capability: eclipse.swt;filter:="(image.format=svg)" until it works it is no use to add this to the template.

vogella:
https://github.com/eclipse-pde/eclipse.pde/issues/1727

HannesWell:
> #1727

@vogella that issue is resolved now. Can we proceed here, respectively can you update the code to add the discussed header?

vogella:
@HannesWell using this new template I still get:

image

vogella:
Strange. I only see this error message is a runtime Eclipse,  not in my latest I-Build. I merge this and test it tomorrow in the next I-build.

HannesWell:
> Strange. I only see this error message is a runtime Eclipse, not in my latest I-Build. I merge this and test it tomorrow in the next I-build.

Could it be that the TP in your Dev-Eclipse doesn't contain the latest I-build respectively was on an old state of PDE?

However looking at what I have described in https://github.com/eclipse-pde/eclipse.pde/issues/1727#issuecomment-2825159901 and your screenshot from here, we are actually talking about different errors.
But I have never seen your problem, if the `swt.svg` fragment was in the TP.

vogella:
> > Strange. I only see this error message is a runtime Eclipse, not in my latest I-Build. I merge this and test it tomorrow in the next I-build.
> 
> Could it be that the TP in your Dev-Eclipse doesn't contain the latest I-build respectively was on an old state of PDE?

I had no explicit target platform defined, so I assumed that it would pick up the new runtime plug-ins but maybe it was using an older state. I don't know how a runtime Eclipse sources its target environment if it is using the IDE as target.

> However looking at what I have described in [#1727 (comment)](https://github.com/eclipse-pde/eclipse.pde/issues/1727#issuecomment-2825159901) and your screenshot from here, we are actually talking about different errors. But I have never seen your problem, if the `swt.svg` fragment was in the TP.

Lets see in tomorrows I-Build.

vogella:
Works fine in latest I-Build (Using template to create RCP app with content -> Start via product -> No error dialog)

vogella:
Thanks @HannesWell

PR #1724, opened 2025-04-21
Reduce re-computation of plugins included in Eclipse/OSGi launches

HannesWell:
This ensures consistency and reduces calls of `DependencyManager.findRequirementsClosure()`.

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
@@ -78,6 +78,7 @@ public abstract class AbstractPDELaunchConfiguration extends LaunchConfiguration
 
 	protected File fConfigDir = null;
 	String launchMode;
+	private Set<IPluginModelBase> launchedPlugins;
 
 	/**
 	 * This field will control the addition of argument --add-modules=ALL-SYSTEM in the VM arguments
@@ -101,8 +102,9 @@ public abstract class AbstractPDELaunchConfiguration extends LaunchConfiguration
 		String commandLine = ""; //$NON-NLS-1$
 		launch.setAttribute(PDE_LAUNCH_SHOW_COMMAND, "true"); //$NON-NLS-1$
 		fConfigDir = null;
-		SubMonitor subMonitor = SubMonitor.convert(monitor, 4);
+		SubMonitor subMonitor = SubMonitor.convert(monitor, 5);
 		try {
+			launchedPlugins = computeLaunchedPlugins(configuration, subMonitor.split(1));
 			preLaunchCheck(configuration, launch, subMonitor.split(2));
 		} catch (CoreException e) {
 			if (e.getStatus().getSeverity() == IStatus.CANCEL) {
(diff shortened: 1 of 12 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/JREBlock.java
```diff
@@ -227,7 +231,11 @@ private void initializeJRESection(ILaunchConfiguration config) throws CoreExcept
 			}
 			// Try to get a default EE based on the selected plug-ins in the config
 			if (eeId == null) {
-				eeId = VMHelper.getDefaultEEName(config);
+
+				// TODO: determine this right from the config?!
```
HannesWell, line 235: This has to be fixed before this can be submitted.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
```diff
@@ -166,7 +169,11 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
 		} else {
 			subMonitor.setCanceled(true);
 		}
+	}
 
+	Set<IPluginModelBase> computeLaunchedPlugins(ILaunchConfiguration configuration, IProgressMonitor monitor) throws CoreException {
```
HannesWell, line 174: I wonder if this should become protected so that sub-classes may override this as well, if they inherit this class directly and don't inherent `EclipseApplicationLaunchConfiguration` or `EquinoxLaunchConfiguration`. But I don't know existing extenders and can therefore not say if that's needed.

ui/org.eclipse.pde.launching/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.launching;singleton:=true
-Bundle-Version: 3.13.400.qualifier
+Bundle-Version: 3.14.0.qualifier
```
HannesWell, line 5: The API-tools currently say this would be necessary, but in the current state I don't see a reason, because only package-private methods are added.
But let's see if we these methods become `protected` eventually.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/JREBlock.java
HannesWell: Done.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java
HannesWell: For now I think we should keep this package-private. If there is a demand for sub-classes to override it, we can still make it protected, but I expect it not to be necessary.

ui/org.eclipse.pde.launching/META-INF/MANIFEST.MF
HannesWell: As far as I can tell, API-tools considers the removal of the overridden `EquinoxLaunchConfiguration.preLaunchCheck()` method falsely as an addition. I have now suppressed this error.

merks:
This definitely does not help solve the problem I have with EMF's workspace.

PR #1728, opened 2025-04-23
Touch bundles affected by compiler changes

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2995


iloveeclipse:
Build was fine, but we had ~ 40 test failures related to baseline mismatch. I assume they are unrelated.

PR #1729, opened 2025-04-23
Add SVGs to org.eclipse.pde.api.tools.ui

Michael5601:
This PR adds SVGs for all icons in the bundle `org.eclipse.pde.api.tools.ui` except for the following as it does not exist as SVG yet:

~~obj16/eclipse16.svg~~

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools.ui/META-INF/MANIFEST.MF
@@ -37,3 +37,4 @@ Export-Package: org.eclipse.pde.api.tools.ui.internal;x-friends:="org.eclipse.pd
  org.eclipse.pde.api.tools.ui.internal.views;x-internal:=true,
  org.eclipse.pde.api.tools.ui.internal.wizards;x-friends:="org.eclipse.pde.api.tools.tests"
 Automatic-Module-Name: org.eclipse.pde.api.tools.ui
+Require-Capability: eclipse.swt;filter:="(image.format=svg)"
--- a/apitools/org.eclipse.pde.api.tools.ui/icons/full/elcl16/compare_apis.svg
+++ b/apitools/org.eclipse.pde.api.tools.ui/icons/full/elcl16/compare_apis.svg
@@ -1,0 +1,764 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   height="16"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   sodipodi:docname="compare_apis.svg">
+  <defs
+     id="defs4">
+    <linearGradient
+       inkscape:collect="always"
+       id="linearGradient4817">
+      <stop
+         style="stop-color:#7695b8;stop-opacity:1"
+         offset="0"
+         id="stop4819" />
+      <stop
+         style="stop-color:#8daac8;stop-opacity:1"
+         offset="1"
+         id="stop4821" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient4805">
+      <stop
+         style="stop-color:#4b82b6;stop-opacity:1"
+         offset="0"
+         id="stop4807" />
+      <stop
+         id="stop4815"
+         offset="0.79358917"
+         style="stop-color:#9ab7d5;stop-opacity:1" />
+      <stop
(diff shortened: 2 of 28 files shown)


HeikoKlare (approved):
The change looks sound.

Regarding missing:
> obj16/eclipse16.svg

I wonder whether we really don't have the Eclipse logo as SVG to place here?

HannesWell (commented):
> > obj16/eclipse16.svg
> 
> I wonder whether we really don't have the Eclipse logo as SVG to place here?

I also assume that we already have such an icon, even if not explicitly defined in the images repo. Maybe it can be found out from which SVG the current PNGs are generated when looking at their git-history?
@Michael5601 I would prefer to add that SVG with this change.
Besides that this change looks good.

merks:
I’m not sure from where I got this from

https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/eclipse.svg

somewhere in the images repo.

Michael5601:
I intentionally marked icons like eclipse16/32/48... as missing as there is no SVG with exactly this name. I proceeded like this for every icon that does not have an SVG with the same name even though there could be duplicates with a different name so we can discuss this topic.

For the eclipse icons in particular I am aware of one special SVG file (`eclipse-svg/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/icons/Eclipse.icns.svg`) which is 1024x1024px in size that looks like the needed icons. The same folder in the `eclipse-png` package contains the same icon in many different sizes as PNG. We would need to check if we can just reduce the SVG in size and use it.

@merks I couldn't find the SVG that you sent in the images repo. The SVG you sent can of course also be used for the eclipse16/32/48 icons but it would also need to be adjusted in size.

HeikoKlare:
> I intentionally marked icons like eclipse16/32/48... as missing as there is no SVG with exactly this name.

Might be that the Eclipse icons is actually the only one that really used in a "shared" way, i.e., placed as SVG at some central place and then used ot generated PNGs at different sizes out of it.
In addition, most of the eclipse16/32/48 PNGs are probably fine as they not used as icons within the application but, if I am not mistaken, also embedded into the application to be used by the OS as program icon and the like. That's why those can probably not be replaced easily. Those will usually be placed in the root folder of the application rather than the `icons` folder.

Michael5601:
I now added the missing SVG `obj16/eclipse16.svg` by adjusting the SVG @merks sent.

HannesWell:
I think what has been said is right. When search the `eclipse-platform` organization for eclipse16 (https://github.com/search?q=org%3Aeclipse-platform%20eclipse16&type=code) I found. The two branding plugins that provide it:
- [org.eclipse.sdk](https://github.com/eclipse-platform/eclipse.platform/tree/master/platform/org.eclipse.sdk)
- [org.eclipse.platform](https://github.com/eclipse-platform/eclipse.platform/tree/master/platform/org.eclipse.platform)

But for both I didn't find an SVG source, but I think that's what you said before.
Btw. I noticed that in [`org.eclipse.ant.ui`'s `icons/full/obj16` folder](https://github.com/eclipse-platform/eclipse.platform/tree/1ff6f0130de3a7289459eb1538b82c4a0637466b/ant/org.eclipse.ant.ui/icons/full/obj16), there is one `eclipse16.png`, one `eclipse_obj.svg` and one `eclipse_obj@2x.png`. This group doesn't fit in the usual pattern and the latter two Eclipse icons looks much older than the former. @Michael5601 could you please have a look at that and check what's correct respectively used?

HannesWell (approved):
This change looks good. Thank you.

Michael5601:
> I noticed that in [`org.eclipse.ant.ui`'s `icons/full/obj16` folder](https://github.com/eclipse-platform/eclipse.platform/tree/1ff6f0130de3a7289459eb1538b82c4a0637466b/ant/org.eclipse.ant.ui/icons/full/obj16), there is one `eclipse16.png`, one `eclipse_obj.svg` and one `eclipse_obj@2x.png`. This group doesn't fit in the usual pattern and the latter two Eclipse icons looks much older than the former. @Michael5601 could you please have a look at that and check what's correct respectively used?

I don't know if the `eclipse16.png` and `eclipse_obj.png` are connected. I viewed them as different icons as both of them have duplicates that are used in other bundles. I think both should have a 100% and a 200% size, which is not the case. In the end also both SVGs should be in the folder but I could only add the SVG for `eclipse_obj.svg` as the other one was missing (documented in this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1811)).

HannesWell:
> I don't know if the `eclipse16.png` and `eclipse_obj.png` are connected. I viewed them as different icons as both of them have duplicates that are used in other bundles. I think both should have a 100% and a 200% size, which is not the case.

I think we have to check the exact usage and see what that reveals. If you don't have time, I can try to do it in the next days.

Michael5601:
> > I don't know if the `eclipse16.png` and `eclipse_obj.png` are connected. I viewed them as different icons as both of them have duplicates that are used in other bundles. I think both should have a 100% and a 200% size, which is not the case.
> 
> I think we have to check the exact usage and see what that reveals. If you don't have time, I can try to do it in the next days.

I can do it on monday if this has time to wait. :) If it doesn't you can do it if you want.

Michael5601:
>
> I think we have to check the exact usage and see what that reveals. If you don't have time, I can try to do it in the next days.

I think you are right that the `eclipse_obj.png` is the old variant. I can only find it used in the fieldassist menu.
image

Also I found the SVG in the bundle `org.eclipse.ant.ui` in the folder `icons/full/obj16/` but it is not used anywhere. 

The icon `eclipse16.png` is used in various places like the help menu:
image

or in the top right corner of the window:
image

The icons are placed in 
org.eclipse.ant.ui/obj16/
org.eclipse.pde.api.tools.ui/obj16/

Also there is a `eclipse.png` in `org.eclipse.pde.ui/etool16/eclipse.svg` and in `org.eclipse.pde.ui/obj16/eclipse.svg` that looks like the `eclipse16.png`.

Issue #1736, opened 2025-04-23
ClassCastException in "Import Package" dialog


ptziegler:
One edge-case for #1501 was missed.

![Image](https://github.com/user-attachments/assets/f26b095a-c807-4a63-aeb3-80c501bc0e45)

To reproduce: Double-click on one of the available versions in the "Import Package" dialog.

```
java.lang.ClassCastException: class org.eclipse.pde.internal.core.text.bundle.PackageObject cannot be cast to class org.eclipse.pde.core.plugin.IPluginModelBase (org.eclipse.pde.internal.core.text.bundle.PackageObject and org.eclipse.pde.core.plugin.IPluginModelBase are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @94029be)
	at org.eclipse.pde.internal.ui.parts.PluginVersionPart$PluginVersionTablePart.handleDoubleClick(PluginVersionPart.java:73)
	at org.eclipse.pde.internal.ui.parts.TablePart.lambda$1(TablePart.java:43)
	at org.eclipse.jface.viewers.StructuredViewer$1.run(StructuredViewer.java:781)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.viewers.StructuredViewer.fireDoubleClick(StructuredViewer.java:778)
	at org.eclipse.jface.viewers.StructuredViewer.handleDoubleSelect(StructuredViewer.java:1094)
	at org.eclipse.jface.viewers.ColumnViewer.handleDoubleSelect(ColumnViewer.java:1033)
	at org.eclipse.jface.viewers.StructuredViewer$4.widgetDefaultSelected(StructuredViewer.java:1207)
	at org.eclipse.jface.util.OpenStrategy.fireDefaultSelectionEvent(OpenStrategy.java:271)
	at org.eclipse.jface.util.OpenStrategy$1.handleEvent(OpenStrategy.java:328)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5864)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5074)
[...]

PR #1738, opened 2025-04-23
Add plugin/package distinction to double-click listener

ptziegler:
Closes https://github.com/eclipse-pde/eclipse.pde/issues/1736


HannesWell (approved):
Nice catch. Thank you for this fix, it looks good and works well.

HannesWell:
Ignoring linux-build and version check failures failures as they are probably related to the ongoing infra-structure outage.

PR #1742, opened 2025-04-28
Replace pre-generated disabled icon with on-the-fly generated ones

HeikoKlare:
SWT's algorithm to calculate disabled icons on the fly has been enhanced to produce results that are equal to the existing, pre-generated disabled icons being used. This makes the usage of pre-generated disabled icons obsolete. This is particularly important for the introduction of SVG icons, as the quality of on-the-fly generated disabled icons based on the original SVGs is better for arbitrary zooms than the scaled up or scaled downs versions of the PNG-based disabled icons.

This change removes all explicit specifications of disabled icons unless it's a usage of one of Platform's shared images.


HannesWell (approved):
Looks good.
Thank you for this clean-up.

PR #1746, opened 2025-05-03
Migrate swt spies to pde

laeubi:
See
- https://github.com/eclipse-platform/eclipse.platform.swt/discussions/2068#discussioncomment-12979813


HannesWell (approved):
Looks good, thank you. I just updated the commit message of your last commits to include the affected project's name.

It seems like we can just submit this using `Rebase and merge` without changing any repo-settings (and ignoring the merge commit check). Or will this destroy the merge commit?

@merks I also tried to have the new project included in the Oomph setup, but somehow it does not work with the current setup and I also failed to force it's inclusion somehow. Could you help here?

As a follow-up step we could also consider to move all the spies into a dedicated sub-folder.

HannesWell:
> It seems like we can just submit this using `Rebase and merge` without changing any repo-settings (and ignoring the merge commit check). Or will this destroy the merge commit?

Ok, I just tested this in my fork and using `Rebase and merge` will rebase the entire merged branch on top of master, so this is not what we want!

I think we can temporarily enable Merge submission in the self-service at:
https://github.com/eclipse-pde/.eclipsefdn/blob/f9ed9c8b036c72aec56bfb82e5a275d09aec2188/otterdog/eclipse-pde.jsonnet#L35-L38

I can look into this tomorrow morning or whenever you say this is ready.

laeubi:
> I can look into this tomorrow morning or whenever you say this is ready.

Yes please merge this as soon as possible, there is one new compiler warning but I thing any further cleanup should be performed as separate steps as migrating git histories is already clumsy enough and I don't want to extend this more than required.

merks:
Yes, I think you need to use otterdog to temporarily allow the merge commit. 

I'm not sure the setup problem but  if you are moving a project and that project from the other repo is already in the workspace, of course there cannot be two projects with the same name in the workspace.  If there are two projects with the same name in the git repositories, of course only one can and will be imported.   So likely any problems you see are related to this.

HannesWell:
> > I can look into this tomorrow morning or whenever you say this is ready.
> 
> Yes please merge this as soon as possible, there is one new compiler warning but I thing any further cleanup should be performed as separate steps as migrating git histories is already clumsy enough and I don't want to extend this more than required.

That can be done subsequently. I just fixed one mail and created the mentioned Otterdoc change (it currently takes long to build):
- https://github.com/eclipse-pde/.eclipsefdn/pull/5

PR #1747, opened 2025-05-05
Fix the about.html for org.eclipse.pde.doc.user to be for EPL-2.0

merks:


merks:
Quality gate failures are unrelated to this change.

PR #1748, opened 2025-05-05
OSGI Repositories view: lower required version ranges for bndtools

chrisrueger:
- to be compatible with bndtools which uses Eclipse 2022-09


laeubi:
Currently Eclipse Infra is down, I'll respin the build once it is back to operational.

laeubi (approved):
New compiler issues are unrelated to the PR

PR #1759, opened 2025-05-06
improve readme

chrisrueger:


ui/org.eclipse.pde.bnd.ui/README.MD
```diff
@@ -26,7 +26,7 @@ transform an (Eclipse) `IProject` into a (bndlib) `Project` (from were the Works
 ```
 @Component
 @AdapterTypes(adaptableClass = IProject.class, adapterNames = Project.class)
-public class BndPluginAdapter implements IAdapterFactory {
+public class PdeBndAdapter implements IAdapterFactory {
```
laeubi, line 29: As this is an example for such an adapter, should be really mention PDE here?

```suggestion
public class MyBndProjectAdapter implements IAdapterFactory {
```

ui/org.eclipse.pde.bnd.ui/README.MD
```diff
@@ -46,6 +46,8 @@ public class BndPluginAdapter implements IAdapterFactory {
 }
 ```
 
+Also see `BndProjectManager.java~.
```
laeubi, line 49: Can you tell what exactly this reference is meant for? It should also probably be a link to the source in the repo.

ui/org.eclipse.pde.bnd.ui/README.MD
chrisrueger: Ah ok, I thought it was supposed to be the current PdeBndAdapter.java but just with an initial name. I was searching for the class in the readme (BndPluginAdapter) and did not find something. So I wanted to use a name from something which really exists, so I can have a look

ui/org.eclipse.pde.bnd.ui/README.MD
chrisrueger: As mentioned earlier. I as a noob wished the readme points me to existing files which are relevant. 
Sure I can link. But maybe let's talk about about your desired readme style.

PR #1760, opened 2025-05-06
BndWorkspace improvements

chrisrueger:
WIP
- find correct workspace for cnf

The current change has the following effect:
If you click in a project called "cnf" (which is kind of a marker project name for a bnd workspace) then the correct Bnd Workspace is returned. 
In the RepositoriesView this has the effect that the Bnd Workspace Repository and all other bnd repos are also shown if you click on "cnf" project. 
image

Before this change clicking on a "cnf" project did not show any of the Bnd Workspace repositories (just the PDE Target Repo).


laeubi (approved):
Looks reasonable. Do you think it needs further enhancements? If not just adjust the Commit message and set this ready for review.

chrisrueger:
> Looks reasonable. Do you think it needs further enhancements? If not just adjust the Commit message and set this ready for review.

I think this is fine for now. Probably not the last change, but ok for now.

PR #1761, opened 2025-05-06
Support svg-images in more Editors

HannesWell:
- Support SVG icons as custom Extension-Element icons
- Support SVG images as About-dialog image of products
- Support SVG images as Windows image of products

and use the SVG-format in extension-point schema examples.


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/product/AboutSection.java
```diff
@@ -147,6 +147,7 @@ private void handleBrowse() {
 		FileExtensionsFilter filter = new FileExtensionsFilter();
 		filter.addFileExtension("gif"); //$NON-NLS-1$
 		filter.addFileExtension("png"); //$NON-NLS-1$
+		filter.addFileExtension("svg"); //$NON-NLS-1$
```
HannesWell, line 150: I'm not sure if SVG can be generally allowed for `images that are associated with the application window`? This might be renderer natively by the OS.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExtensionsSection.java
```diff
@@ -151,7 +152,7 @@ public class ExtensionsSection extends TreeSection implements IPropertyChangeLis
 			"commandId", "property", "activityId", //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 			"attribute", "value"}; //$NON-NLS-1$ //$NON-NLS-2$
 
-	private static final String[] VALID_IMAGE_TYPES = {"png", "bmp", "ico", "gif", "jpg", "tiff"}; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$ //$NON-NLS-6$
+	private static final Set<String> VALID_IMAGE_TYPES = Set.of("svg", "png", "bmp", "ico", "gif", "jpg", "tiff"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$ //$NON-NLS-6$ //$NON-NLS-7$
```
HannesWell, line 155: Actually every image format that SWT supports could be listed here.
Maybe we should consider to introduce some kind of `ImageFileDialog` as extension of `FileDialog` that allows to load and save images of the formats that SWT supports and returns the path or even the `Image/ImageData` directly as convenience.

laeubi: AWT allows to enumerate supported images (with Java SPI + `ImageReader` interface) in what case e.g. JFace could offer such smarter FileChooser Dialogs...

Issue #1766, opened 2025-05-07
PDE launch configurations miss icons


iloveeclipse:
Most likely regression from https://github.com/eclipse-pde/eclipse.pde/pull/1734

All Eclipse PDE launch configurations have "red" icons as a sign of missing right one.
New workspace:

![Image](https://github.com/user-attachments/assets/7c7d2838-c1be-4889-862d-01a9ae214926)

Existing workspace:

![Image](https://github.com/user-attachments/assets/786bdeef-3a6c-4878-ba5b-ebb2ef0b8c00)

Also "Run/Debug" drop-down is affected:

![Image](https://github.com/user-attachments/assets/9e35b2e0-1716-424c-a9de-8bb656cb441f)

merks:
This is in the log
```
!MESSAGE The image could not be loaded: org.eclipse.jface.resource.DeferredImageDescriptor@30c085f3
!STACK 0
org.eclipse.jface.resource.DeviceResourceException: Unable to create resource org.eclipse.jface.resource.DeferredImageDescriptor@30c085f3
	at org.eclipse.jface.resource.ImageDescriptor.createResource(ImageDescriptor.java:273)
	at org.eclipse.jface.resource.DeviceResourceManager.allocate(DeviceResourceManager.java:56)
	at org.eclipse.jface.resource.AbstractResourceManager.create(AbstractResourceManager.java:92)
	at org.eclipse.jface.resource.LazyResourceManager.create(LazyResourceManager.java:88)
	at org.eclipse.jface.resource.LocalResourceManager.allocate(LocalResourceManager.java:71)
	at org.eclipse.jface.resource.AbstractResourceManager.create(AbstractResourceManager.java:92)
	at org.eclipse.jface.resource.ResourceManager.createImageWithDefault(ResourceManager.java:192)
	at org.eclipse.jface.action.ActionContributionItem.updateImages(ActionContributionItem.java:1047)
	at org.eclipse.jface.action.ActionContributionItem.updateMenuItem(ActionContributionItem.java:918)
	at org.eclipse.jface.action.ActionContributionItem.update(ActionContributionItem.java:743)
	at org.eclipse.jface.action.ActionContributionItem.fill(ActionContributionItem.java:293)
	at org.eclipse.debug.ui.actions.AbstractLaunchHistoryAction.addToMenu(AbstractLaunchHistoryAction.java:160)
[...]

PR #1767, opened 2025-05-07
Add missing eclipse.svg

merks:
- This was copied here https://github.com/eclipse-pde/eclipse.pde/blob/master/apitools/org.eclipse.pde.api.tools.ui/icons/full/obj16/eclipse16.svg
- The SVG is also missing here https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images/eclipse-svg/org.eclipse.pde.ui/icons/etool16

https://github.com/eclipse-pde/eclipse.pde/issues/1766

### DIFF
--- a/ui/org.eclipse.pde.ui/icons/etool16/eclipse.svg
+++ b/ui/org.eclipse.pde.ui/icons/etool16/eclipse.svg
@@ -1,0 +1,230 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="16"
+   height="16"
+   viewBox="0 0 1.3617021 1.3617021"
+   version="1.1"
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
The change to add a proper SVG at this place is correct.
The SVG itself seems to contain a different "state" of the Eclipse icon than the original PNG one. In this screenshot, you see the old on the left and the new on the right:
image

I am not sure which of them is "correct"/up-to-date, but except for that the new one looks good to me and is definitely better than not having an icon at all :-)

HannesWell:
Thank you Ed for this fix. Highly appreciated.

> I am not sure which of them is "correct"/up-to-date, but except for that the new one looks good to me and is definitely better than not having an icon at all :-)

Absolutely.

> Linux seems to have stalled. I believe nothing tests this change anyway.

I noticed that as well in the recent past. No clue why.

PR #1769, opened 2025-05-07
Fix missing about.html and includes missing from build.properties

merks:


merks:
Test failures unrelated.

PR #1772, opened 2025-05-09
Don't generate a source bundle with only an about.html

merks:


merks:
The source bundle contains only an about.html so effectively empty/useless.

image

merks:
Actually I'm not really sure this is the right way. 

Other places seem to set the property to not generate, but even that seems insufficient and so there also appears to be a need to configure the pom.xml to exclude it:

image

I.e., it appears to be necessary to add `org.eclipse.core.filesystem.linux.aarch64` to the pom configuration.

merks:
This is related:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3038

For the record, these are all bundles that I noticed have source bundle that don't actually contain any useful sources:
```
pde/org.eclipse.pde/3.13.3100/org.eclipse.pde-3.13.3100.jar
platform/org.eclipse.core.filesystem.linux.aarch64/1.4.200/org.eclipse.core.filesystem.linux.aarch64-1.4.200.jar
platform/org.eclipse.equinox.console.jaas.fragment/1.2.0/org.eclipse.equinox.console.jaas.fragment-1.2.0.jar
platform/org.eclipse.osgi.util/3.7.300/org.eclipse.osgi.util-3.7.300.jar
platform/org.eclipse.swt.tools.base/3.107.700/org.eclipse.swt.tools.base-3.107.700.jar
platform/org.eclipse.ui.themes/1.2.2700/org.eclipse.ui.themes-1.2.2700.jar
```

mickaelistria:
Could Tycho be made smarter and not generate source bundles if it identifies no .class file was generated?

laeubi:
> Could Tycho be made smarter and not generate source bundles if it identifies no .class file was generated?

A bundles does not need to contain classes to have sources, e.g. html, images, ... so its a bit a matter of definition and in the past some checks in platform even complained when there is no source bundle at all.

I think instead of trying to be smart, it would be easier to add a `skip` option to https://tycho.eclipseprojects.io/doc/latest/tycho-source-plugin/plugin-source-mojo.html that the can be set in the build.properties with pomless options.

The second part is about source *features* what will exclude it from the feature, but as we use `includeAllSources` since a while it is mostly irrelevant for inclusion of such things.

laeubi:
If i where to solve it I would go to [OsgiSourceMojo](https://github.com/eclipse-tycho/tycho/blob/main/tycho-source-plugin/src/main/java/org/eclipse/tycho/source/OsgiSourceMojo.java) and add a new `@Parameter` with name `skip` and property `tycho-plugin-source.skip`, then one can add `pom.model.property.tycho-plugin-source.skip=true` to any `build.properties` file and the generation would be skipped.

laeubi:
@stephan-herrmann thanks for the hints, I finally sorted it out, it was the other way round than I initially thought.

Because the about to compiled names *are set* the compiler ignores to lookup these types from the compiled binary locations! So what was needed instead is to limit the names to the types currently compiled for the current target level!

Now your provided testcase passes without a problem!

HannesWell:
> If i where to solve it I would go to [OsgiSourceMojo](https://github.com/eclipse-tycho/tycho/blob/main/tycho-source-plugin/src/main/java/org/eclipse/tycho/source/OsgiSourceMojo.java) and add a new `@Parameter` with name `skip` and property `tycho-plugin-source.skip`

Created a corresponding PR at
- https://github.com/eclipse-tycho/tycho/pull/4976

HannesWell:
> > If i where to solve it I would go to [OsgiSourceMojo](https://github.com/eclipse-tycho/tycho/blob/main/tycho-source-plugin/src/main/java/org/eclipse/tycho/source/OsgiSourceMojo.java) and add a new `@Parameter` with name `skip` and property `tycho-plugin-source.skip`
> 
> Created a corresponding PR at
> 
>     * [Add flag to skip execution of tycho-source-plugin eclipse-tycho/tycho#4976](https://github.com/eclipse-tycho/tycho/pull/4976)

@merks the required change in Tycho is now available, also as backport to Tycho 4.0.13-SNAPSHOT.

merks:
The build seems to hang with this failure which looks unrelated to this change.
```
 Results:
Failures: 
  UpdateUnitVersionsCommandTests.testVersionSort:55->confirmVersionUpdates:93 ID: org.eclipse.fake.3 has the incorrect version. updatedText=<target>
<locations>
<location>
<unit id="org.eclipse.fake.1" version="0.0.0"/>
<unit id="org.eclipse.fake.2" version="0.0.0"/>
<unit id="org.eclipse.fake.3" version="0.0.0"/>
<unit id="org.eclipse.fake.4" version="0.0.0"/>
<unit id="org.eclipse.fake.5" version="0.0.0"/>
<unit id="org.eclipse.fake.6" version="0.0.0"/>
<unit id="org.eclipse.fake.7" version="0.0.0"/>
<repository location="bundleentry://358.fwk1077351082/testing-files/testing-sites/MultipleUnitsConfirmSorting/"/>
</location>
</locations>
</target> expected:<[1.1.2]> but was:<[0.0.0]>
Tests run: 29, Failures: 1, Errors: 0, Skipped: 0
```

@HannesWell  @laeubi 

Is this the change you are expecting?

I see this in the log
```
07:16:03  [INFO] --- tycho-source:4.0.13-SNAPSHOT:plugin-source (plugin-source) @ eclipse.pde ---
07:16:03  [INFO] 
07:16:03  [INFO] --- javadoc:3.11.2:jar (attach-javadocs) @ eclipse.pde ---
07:16:03  [INFO] Not executing Javadoc as the project is not a Java classpath-capable package
07:16:03  [INFO] 
```
Does that tell us anything useful?

I think not because in https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1772/lastSuccessfulBuild/artifact/repository/target/repository/plugins/ I see this

image

What might I be doing incorrectly?

ui/org.eclipse.pde/build.properties
```diff
@@ -24,3 +24,4 @@ bin.includes = plugin.properties,\
                eclipse32.png,\
                images/
 src.includes = about.html
+pom.model.property.tycho.plugin.source.skip = true
```
laeubi, line 27:
```suggestion
pom.model.property.tycho.source.skip = true
```

merks:
It's still there:  😱 

image

merks:
Still there:

image

But now I realize this thing has a pom and I bet the property only works on a pomless bundle and here I need to configure this in the pom.xml.  Or?  How best to do that?

HannesWell:
> Yes if there is a pom (didn't realized that) you need to configure it in the pom in the build section.

Wouldn't it be sufficient to define it as property in the properties section? I.e.
```
  <properties>
    <tycho.source.skip>true</tycho.source.skip>
  </properties>
```

merks:
It's really gone this time:

image

merks:
To make sure this doesn't cause downstream problems and correctly produces the expect result in the final p2 repository I kicked off this I-build:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.36/60/

Then I can apply this solution to the other bundle mentioned in https://github.com/eclipse-pde/eclipse.pde/pull/1772#issuecomment-2866481730

Issue #1561, opened 2025-01-14
Provide some buttons to perform usual version increments


laeubi:
If one edits a package version it looks like this:

![grafik](https://github.com/user-attachments/assets/39028b79-f3b3-4c84-bf59-3b71112bee69)

below the version there is some free space in the dialog, it should contain three links to change the version:

1. New Major Version (changing it to `<major+1>.0.0`) [in this example `3.0.0`]
2. New Minor Version (changing it to `<major>.<minor+1>.0` [in this example `2.13.0`]
3. New Micro Version (changing it to `<major>.<minor>.<micro+1>` [in this example `2.13.1`]

so one do not needs edit the text field manually.

anviik:
@jackson-drew and I be working on it.

braedonbillingsley:
My team and I are working on this.

anviik:
Hi @braedonbillingsley , @jackson-drew and I are still working on this issue however, since it hasn't been done yet we would love to possibly collaborate. Feel free to reach out to me at my email: anvikudaraya417@gmail.com

PR #1773, opened 2025-05-09
Created buttons for PluginVersionPart to increment the exported version

Shyesta:
Original Issue: [#1561 ](https://github.com/eclipse-pde/eclipse.pde/issues/1561)

My team and I created three additional buttons that will allow for incrementing of the major, minor, and micro versions of an exported package by 1. We did this by:
- Defining Strings for each of the text properties attached to these buttons in [PDEUIMessages.java](https://github.com/Shyesta/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java)
- Attaching values to the defined Strings at [pderesources.properties](https://github.com/Shyesta/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties)
- Adding buttons to [PluginVersionPart.java](https://github.com/Shyesta/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java) that have all relevant text and layout properties.

These changes can be viewed in Eclipse by building the project with all dependencies, creating a plugin project with a plugin template, entering the Runtime section, and then viewing the properties of a plugin.

One thing my team and I noticed was that the versions of these exported packages are not validated, meaning we could enter a clearly incorrect version number like 10000.0.0 and it would still apply. This validation could be a potential enhancement made in the future.

Here is a preview of the changes made: 

image

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java
@@ -288,6 +288,38 @@ public class PluginVersionPart {
 		fMinVersionText = new Text(parent, SWT.SINGLE | SWT.BORDER);
 		fMinVersionText.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
 		fMinVersionText.setEnabled(editable);
+
+		if (!fRangeAllowed) {
+			Label incrLabel = new Label(parent, SWT.NONE);
+			incrLabel.setText(PDEUIMessages.PluginVersionPart_incrementTitle);
+			incrLabel.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
+
+			Composite btnBar = new Composite(parent, SWT.NONE);
+			btnBar.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+
+			GridLayout btnBarLayout = new GridLayout(3, true);
+			btnBarLayout.marginWidth = 0;
+			btnBar.setLayout(btnBarLayout);
+
+			Button majorBtn = new Button(btnBar, SWT.PUSH);
+			majorBtn.setText(PDEUIMessages.PluginVersionPart_bumpMajor);
+			majorBtn.setEnabled(editable);
+			majorBtn.addListener(SWT.Selection, e -> bumpVersion(Type.MAJOR));
+			majorBtn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+
+			Button minorBtn = new Button(btnBar, SWT.PUSH);
+			minorBtn.setText(PDEUIMessages.PluginVersionPart_bumpMinor);
+			minorBtn.setEnabled(editable);
+			minorBtn.addListener(SWT.Selection, e -> bumpVersion(Type.MINOR));
+			minorBtn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+
+			Button microBtn = new Button(btnBar, SWT.PUSH);
+			microBtn.setText(PDEUIMessages.PluginVersionPart_bumpMicro);
+			microBtn.setEnabled(editable);
+			microBtn.addListener(SWT.Selection, e -> bumpVersion(Type.MICRO));
+			microBtn.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+		}
+
 	}
 
 	public void preloadFields() {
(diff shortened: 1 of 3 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java
```diff
@@ -454,4 +486,43 @@ public void addListeners(ModifyListener minListener, ModifyListener maxListener)
...
+                    mic = 0;
+                    break;
+                case MINOR:
+                    min++;
+                    mic = 0;
+                    break;
+                case MICRO:
+                    mic++;
+                    break;
+            }
```
HannesWell, line 512: Instead defining a new Enum-Type and adding a lot of logic in this method, I think it would be better to pass a `UnaryOperator<Version> versionModifier` instead of `type` and just use that to compute the new version after the current version String was parsed into a `Version` object. That new Version object could then be serialized back into the text field as it's already done now.
With that new approach the caller would pass a lambda that return a corresponding new version object from the current version that was passed to it.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java
```diff
@@ -454,4 +486,43 @@ public void addListeners(ModifyListener minListener, ModifyListener maxListener)
...
+			String originalText = getMinVersion();
+			String qualifier = ""; //$NON-NLS-1$
+			int qualifierIndex = originalText.indexOf(v.toString()) + v.toString().length();
+			if (qualifierIndex < originalText.length()) {
+				qualifier = originalText.substring(qualifierIndex);
+			}
+
+			target = new Version(maj, min, mic).toString() + qualifier;
+        } catch (IllegalArgumentException e) {
+            target = "1.0.0"; //$NON-NLS-1$
```
HannesWell, line 523: Instead of handling exception, maybe the input should be validated in advance and for example the buttons should be disabled when the field is empty/no version is specified yet?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/parts/PluginVersionPart.java
```diff
@@ -288,6 +288,38 @@ private void createSingleField(Composite parent, boolean createGroup, boolean ed
...
+			btnBar.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+
+			GridLayout btnBarLayout = new GridLayout(3, true);
+			btnBarLayout.marginWidth = 0;
+			btnBar.setLayout(btnBarLayout);
+
+			Button majorBtn = new Button(btnBar, SWT.PUSH);
+			majorBtn.setText(PDEUIMessages.PluginVersionPart_bumpMajor);
+			majorBtn.setEnabled(editable);
+			majorBtn.addListener(SWT.Selection, e -> bumpVersion(Type.MAJOR));
```
HannesWell, line 307: Please use a typed-listener here, i.e. call `addSelectionListener()` instead and create a `SelectionListener` using `SelectionListener.widgetSelectedAdapter()`.
Together with the suggestion below to pass a lambda, this code would be
```suggestion
			majorBtn.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> bumpVersion(v -> {
				return new Version(v.getMajor() + 1, 0, 0, v.getQualifier());
			})));
```
The other button definitions would looks correspondingly.
I'm not sure if it's necessary to handle the qualifier here. In general versions of exported-packages don't have a qualifier anyway. But it could also be handled uniformly in the `bumpVersion()` (if really necessary).

HannesWell (changes requested):
Thank you for this contribution and sorry for the delayed reply, the result looks good already but the implementation should be enhanced in a few aspects, please see my comments below.

Also please only use tabs for indentation to match the existing style in the changed class. 

Furthermore please squash all your commits locally (including the new changes), with a message covering the change as a whole, and force push the new commit to the branch of this PR (don't create a new branch). Eventually this change will be submitted as one commit anyway.

Shyesta:
Hi @HannesWell sorry for taking so long to implement the changes you requested.

I believe I have fixed everything you requested, including:
- Adhering to tabs instead of spaces for line starts
- Disabling the buttons when there is not a valid version input
- Using UnaryOperator instead of an enum Type and SelectionListeners for some of the logic
- Removing the exception check and instead validating the input differently

I did run into issues with squashing the commits and am unsure if that was done properly as you requested. Let me know if there are any issues there that I can resolve, or if anything else needs to be adjusted. Thank you!

HannesWell (approved):
Thank you for the update I did another review of this change and applied some more required changes:
- fix a bug in the (optional and usually absent) qualifier handling
- rearrange method boundaries to reduce code duplication.

> I did run into issues with squashing the commits and am unsure if that was done properly as you requested. Let me know if there are any issues there that I can resolve, or if anything else needs to be adjusted. Thank you!

I have squashed the changes for you locally and force pushed the result together with the other adjustments I mentioned to this PRs branch.
If you want to do it yourself in the future, you can do it on the CLI following for example this [guide](https://www.baeldung.com/ops/git-squash-commits) or by using the EGit History view via the context menu:
grafik


Thank you for this contribution. This is ready for submission now.

PR #1776, opened 2025-05-13
Whitespace clean-up in pde ds

vogella:
Using the JDT clean-up for unnecessary whitespace


HannesWell:
Shouldn't we rely on the clean-up job to apply this, which is about to be set up in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3039 ?

vogella:
> Shouldn't we rely on the clean-up job to apply this, which is about to be set up in [eclipse-platform/eclipse.platform.releng.aggregator#3039](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3039) ?

Please merge it, if you want to enable this flow. I only look from time to time at Eclipse and don't want to create a bunch of PR for others to look at. My plan is to reduce the number of automatically created PR by doing some pre-work before enabling it, but if someone else enables the workflow that is even better.

PR #1788, opened 2025-05-19
Add Missing SVGs to PDE Bundles

Michael5601:
This PR adds missing SVGs to `org.eclipse.pde` bundles.

Please note that this PR does not include all missing SVGs documented in this [issue](https://github.com/eclipse-platform/eclipse.platform.images/issues/129) but contributes to providing all PDE Icons as SVGs. All Icons added in this PR are either copied from another location where a duplicate with another name already existed or created by myself. 

Example content in `.exsd` files is also changed to .svg even though this has no direct impact on any icons.

One reference of the icon `ant.png` in the bundle  `org.eclipse.ant.ui` is changed to SVG as this bundle is already updated for using SVGs.
_____________________________________________________
Appart from the added SVG in this PR the bundles `org.eclipse.pde.spy.preferences`, `org.eclipse.e4.tools.emf.ui` and `org.eclipse.pde.bnd.ui` contain many GIFs and PNGs for which no SVGs exist yet. 

The bundle `org.eclipse.pde.doc.user` contains many "icons" that are screenshots of editors. I don't think SVGs are needed here.


HannesWell (approved):
Thank you for this contribution, this looks good.

While testing this I noticed (again) that the 'toolbar' icons in the E4 Spies window are still blurry, but I believe that's an E4 issue respectively that it scales up already rasterized images and does not scale properly them if possible:
grafik

HeikoKlare:
> While testing this I noticed (again) that the 'toolbar' icons in the E4 Spies window are still blurry, but I believe that's an E4 issue respectively that it scales up already rasterized images and does not scale properly them if possible:

Yes, there seems to be some strange scaling being involved. At least it's not related to SVGs. I replaced the OSGi SVG with the PNG again and it still looks that weird:
image

HeikoKlare:
I took a further look but I have to admit that I have no clue.

My findings so far:
- On a 100% monitor, those icons are not rendered at 100% zoom (16x16) but at some larger scale (something like 20x20), which is probably the reason for them being blurry
- I am aware of two ways how images can be scaled that badly:
  1. `ImageData#scaledTo()`: this method is never called when running an application and opening the spies window
  2. `GC#drawImage()` with anti-aliasing and different source/destination width and height: this is also never called when running an application and opening the spies windows for any of the affected images
- No handles at different zooms than 100 are retrieved for the according images. This would require `Image#getImageMetadata()` to be called on those images, but a breakpoint only triggered for zoom=100
- The images are properly assigned to the `HandledContributionItem`s making up the toolbar entries. They use the proper SVG (or, if changed, PNG) URL, create a proper image descriptor and properly create and set and image at 100%

I guess this must be related to how those items are rendered, but I did not find yet where something special is made for them. But the unusal size of the rendered images looks suspicious.

PR #1797, opened 2025-06-04
Add version bumps for 4.37

HannesWell:
This just adds a dummy change to provoke the automated version-bumps.
Once they are applied I plan to update the PR and remove the dummy commit so that only the version bump commit remains.


HannesWell:
> #1798 could serve same purpose, right?

Yes, both works.

Issue #1722, opened 2025-04-17
Support svg in the Plug-in Image Browser


vogella:
The Plug-in Image Browser is a very nice tool to review the existing icons or Eclipse plug-ins. It would be great if it could also be used to show and download the new SVG icons.

To open the view use CTRL+3 -> Plug-in Image Browser 

![Image](https://github.com/user-attachments/assets/5459d384-f3f3-4f17-842d-9726e9548d85)

vogella:
cc @HeikoKlare  @Michael5601

HeikoKlare:
Thank you for the pointer, @vogella! I didn't think about the image browser yet, but definitely makes sense to extend that to also show SVGs.

Supporting SVGs in generals seems to be easy, as there is an explicit allowlist for considered extensions:
https://github.com/eclipse-pde/eclipse.pde/blob/4f898d144a1e67254d2c36fdbb12fccbb3775af5/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java#L58

When doing so, I would expect the SVGs to show up next to the existing PNGs/Gifs. There are several filter configurations in place to, in particular the "Icon" filter only showing one version of each icon (no disabled ones, no 2x versions etc.), but such a filter probably matches both PNGs and SVGs in the original size. In particular, the 2x-filtering is done by matching the 16x16 size, which will also hold for SVGs. So extending the filtering is probably the bit more tricky part:
https://github.com/eclipse-pde/eclipse.pde/blob/4f898d144a1e67254d2c36fdbb12fccbb3775af5/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java#L129-L146

From a user's perspective, when choosing the "Icons" filter, I would expect SVGs to be shown be precedence:
- if there is an SVG and one or multiple PNGs, take the SVG
- if there is no PNG for an image, take that one as done right now

Filtering currently seems to be limited to a single image, but given that a PNG is only to be shown when there is no matching SVG, this will probably need to be extended.

PR #1809, opened 2025-06-09
Add SVG support to ImageBrowser

Michael5601:
This PR enables the display of SVGs in the `ImageBrowser`. 

When both a PNG and an SVG version of an icon are present, only the SVG will be shown.
To allow this behavior a new filter `PNGReplacementFilter` is introduced which decides depending on the icons in the cache (future icons) and existing filters if a PNG should be added or suppressed. If an SVG with the same path exists in the cache and if it passes all active filters the PNG will be discarded.

Using the existing text pattern filter to only view specific icons (e.g. "*.png" for only PNGs) continues to work as expected with this implementation showing only PNGs for all available icons. 

For this change to work a new public method `public getElementsCache()` is added to `AbstractRepository`. This should be no problem as the package is marked as internal. 

This change follows the recommeded behaviour for the `ImageBrowser` regarding SVGs by @HeikoKlare in this [comment](https://github.com/eclipse-pde/eclipse.pde/issues/1722#issuecomment-2830868565).

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1722.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.commands.NotEnabledException;
@@ -91,6 +92,7 @@ public class ImageBrowserView extends ViewPart implements IImageTarget {
 
 	private static final String COMMAND_SAVE_TO_WORKSPACE = "org.eclipse.pde.ui.imagebrowser.saveToWorkspace"; //$NON-NLS-1$
 	protected final static String VIEW_ID = "org.eclipse.pde.ui.ImageBrowserView"; //$NON-NLS-1$
+	private static final Pattern FILE_EXTENSION = Pattern.compile("\\.[^.]+$"); //$NON-NLS-1$
 
 	private final UpdateUI mUIJob = new UpdateUI();
 
@@ -118,6 +120,7 @@ public class ImageBrowserView extends ViewPart implements IImageTarget {
 
 	private Text txtFilter;
 	private IFilter textPatternFilter;
+	private IFilter svgFilter = null;
 	private PageNavigationControl pageNavigationControl;
 
 	private ImageElement imageElement;
@@ -144,6 +147,20 @@ public class ImageBrowserView extends ViewPart implements IImageTarget {
 		mFilters.add(enabledIcons);
 		textPatternFilter = new StringFilter("*"); //$NON-NLS-1$
 		mFilters.add(textPatternFilter);
+		svgFilter = element -> {
+			String fileName = element.getPath();
+			if (!fileName.endsWith(".svg")) { //$NON-NLS-1$
+				String replacementSVGName = FILE_EXTENSION.matcher(fileName).replaceAll(".svg"); //$NON-NLS-1$
+				ImageElement replacementSVG = repository.getImageElement(replacementSVGName);
+				if (replacementSVG != null) {
+					return new ArrayList<>(mFilters).stream()
+							.anyMatch(filter -> filter != svgFilter && !filter.accept(replacementSVG));
+				}
+				return true;
+			}
+			return true;
+		};
+		mFilters.add(svgFilter);
 	}
 
 	@Override
@@ -191,10 +208,14 @@ public class ImageBrowserView extends ViewPart implements IImageTarget {
 			mFilters.clear();
 			mFilters.add(textPatternFilter);
 			Combo source = (Combo) e.getSource();
-			switch (source.getSelectionIndex()) {
+			int selectionIndex = source.getSelectionIndex();
+			switch (selectionIndex) {
 				case 0 -> mFilters.add(enabledIcons);
 				case 1 -> mFilters.add(disabledIcons);
 				case 2 -> mFilters.add(wizard);
+			}
+			if (selectionIndex != source.getItemCount() - 1) {
+				mFilters.add(svgFilter);
 			}
 			page = 0; // reset to 1st page
 			scanImages();
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
@@ -22,10 +22,10 @@ import java.io.InputStream;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.SequencedMap;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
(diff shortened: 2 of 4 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
```diff
@@ -48,14 +50,15 @@ public abstract class AbstractRepository extends Job {
 	protected List<ImageElement> mElementsCache = new LinkedList<>();
 
 	private final IImageTarget mTarget;
+	protected Map<String, ImageElement> allImageElements = new HashMap<>();
```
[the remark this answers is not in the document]
Michael5601: great idea. This really improves the code, thanks :).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
```diff
@@ -190,11 +205,15 @@ public void createPartControl(final Composite parent) {
 		typeCombo.addSelectionListener(widgetSelectedAdapter(e -> {
 			mFilters.clear();
 			mFilters.add(textPatternFilter);
-			Combo source = (Combo) e.getSource();
-			switch (source.getSelectionIndex()) {
-				case 0 -> mFilters.add(enabledIcons);
-				case 1 -> mFilters.add(disabledIcons);
-				case 2 -> mFilters.add(wizard);
+			List<String> options = List.of("Enabled", "Disabled", "Wizard", "AllIcons"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
```
HannesWell, line 208: I think the indirection through this list doesn't make the code easier to maintain, as we now have to keep that list in sync with the list of labels defined for the combo above.

I agree it's not ideal to just rely on hard-coded indices, but I think currently we have not really completely clean and simple solution, because the labels are not constants (for the compiler).
Therefore I would leave this part as it is.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
```diff
@@ -45,7 +45,7 @@
 
 public abstract class AbstractRepository extends Job {
 
-	protected List<ImageElement> mElementsCache = new LinkedList<>();
+	private Map<String, ImageElement> mElementsCache = new LinkedHashMap<>();
```
HannesWell, line 48:
```suggestion
	private SequencedMap<String, ImageElement> mElementsCache = new LinkedHashMap<>();
```
Then casting the field to `LinkedHashMap` below should also not be necessary.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
```diff
@@ -408,6 +420,38 @@ public synchronized void notifyImage(final ImageElement element) {
...
+			String correspondingSVGName = element.getPath().replaceAll("\\.[^.]+$", ".svg"); //$NON-NLS-1$ //$NON-NLS-2$
+			ImageElement svgElement = repository.getImageElement(correspondingSVGName);
+			if (svgElement != null) {
+				return checkSVGRejectedByFilters(svgElement);
+			}
+			return true;
+		}
+
+		private boolean checkSVGRejectedByFilters(ImageElement cachedSvg) {
+			return new ArrayList<>(filters).stream()
```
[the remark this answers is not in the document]
HannesWell: Acknowledged, thanks for the explanation.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
```diff
@@ -408,6 +420,38 @@ public synchronized void notifyImage(final ImageElement element) {
 		imageIndex++;
 	}
 
+	private class SvgFilter implements IFilter {
```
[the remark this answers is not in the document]
HannesWell: > Is there a better way? Or should we avoid using lambdas?

Indeed, I'm also not aware of a better way. But then I suggest to keep the lambda and also keep the field non-final.
Or do you prefer the separate class?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
```diff
@@ -142,7 +149,9 @@ protected void searchJarFile(final File jarFile, final IProgressMonitor monitor)
 			while ((entries.hasMoreElements()) && (!monitor.isCanceled())) {
 				ZipEntry entry = entries.nextElement();
 				if (isImageName(entry.getName())) {
-					addImageElement(new ImageElement(() -> createImageData(jarFile, entry), jarFile.getName(), entry.getName()));
+					ImageElement imageElement = new ImageElement(() -> createImageData(jarFile, entry),
+							jarFile.getName(), entry.getName());
+					mElementsCache.put(imageElement.getPath(), imageElement);
```
HannesWell, line 154: Can't this code be left unchanged?
```suggestion
					addImageElement(new ImageElement(() -> createImageData(jarFile, entry), jarFile.getName(), entry.getName()));
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/repositories/AbstractRepository.java
```diff
@@ -177,7 +186,10 @@ protected void searchDirectory(File directory, final IProgressMonitor monitor) {
 
 						} else {
 							if (isImage(resource)) {
-								addImageElement(new ImageElement(() -> createImageData(resource), pluginName, resource.getAbsolutePath().substring(directoryPathLength)));
+								ImageElement imageElement = new ImageElement(() -> createImageData(resource),
+										pluginName, resource.getAbsolutePath().substring(directoryPathLength));
+								mElementsCache.put(imageElement.getPath(), imageElement);
+								addImageElement(imageElement);
```
HannesWell, line 192: Can't this code be left unchanged?
```suggestion
								addImageElement(new ImageElement(() -> createImageData(resource), pluginName, resource.getAbsolutePath().substring(directoryPathLength)));
```

HannesWell (commented):
Thanks for the update.
I have a few more remarks, please see them below.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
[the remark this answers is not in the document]
Michael5601: As the lambda is not huge I think it is still relatively easy to comprehend so we can avoid introducing the inner class SVGFilter.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/imagebrowser/ImageBrowserView.java
Michael5601: I agree. I changed the switch to hard indices and changed the condition for adding the SVGFilter to the following (only add SVGs if the selectionIndex is the last possible index):

if (selectionIndex != source.getItemCount() - 1) {
				mFilters.add(svgFilter);
			}

HannesWell (approved):
Thank you for the update and in general for this nice contribuiton.

The change looks fine now, I just pushed an update to extract the `FILE_EXTENSION` Pattern to a static final constant in order to avoid it's recompilation on each call.
This should make the filter a bit quicker (as this is potentially a hot-loop I think we should consider such simple optimizations).

HeikoKlare:
Thank you two for creating and processing this extensions! I have just re-tested it in latest I-Build and it works very well.

PR #1810, opened 2025-06-11
Remove references to org.eclipse.platform.source.feature.group

merks:


iloveeclipse:
@merks : was that the fix for this:

image

https://download.eclipse.org/eclipse/downloads/drops4/I20250610-1800/testResults.php

or should I open a ticket that pde.build.tests are not executed?

PR #1826, opened 2025-06-19
[target] Remove Content and Environment Section from the Definition page

laeubi:
Currently the Target-Editor contains two sections "Content" and "Environment" that roughly explain what these things can be used for if one clicks on the corresponding tabs. In practice, these information is barley usable for todays work an users uncertain would better consult the Eclipse Help or read a tutorial on PDE. More worse, these clutter the UI and steal valuable space not available for the actual content of the location leading to scrollbars appearing.

This now removes the sections entirely ro reclaim the space for something the user is actually interested in.

## Before
grafik

## After
grafik


akurtakov:
Really good call! More content is better than such labels.

PR #1835, opened 2025-06-23
Print framework state on failing bundles

laeubi:
If plugin test do not work as expected the user is currently quite blind. At best there is an exception in the log about a failing bundle, at worst there is only a message "No tests found" and then nothing is executed at all.

This now first checks if there are any unresolved plugins and print the current bundles and their state if anything is failing, and also makes the additional manifest to use non optional imports as otherwise things are possibly missing.


trancexpress:
@laeubi can we add a system property (or any type of switch) to disable the outputs here (default would be showing the outputs)?

We have a hack to unresolve Copilot bundles via `ResolverHookFactory` (for reasons) and the output here prints ~1000 lines that developers have to scroll through - without much we can do about it for now.

Issue #1838, opened 2025-06-27
TargetDefinitionResolutionTests.testSourceBundleRecognition fails since I20250627-0350


iloveeclipse:
https://download.eclipse.org/eclipse/downloads/drops4/I20250627-0350/testresults/html/org.eclipse.pde.ui.tests_ep437I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
Wrong severity expected:<0> but was:<4>

java.lang.AssertionError: Wrong severity expected:<0> but was:<4>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotEquals(Assert.java:835)
at org.junit.Assert.assertEquals(Assert.java:647)
at org.eclipse.pde.ui.tests.target.TargetDefinitionResolutionTests.testSourceBundleRecognition(TargetDefinitionResolutionTests.java:451)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

iloveeclipse:
> [@laeubi](https://github.com/laeubi) : related to [1ba2338](https://github.com/eclipse-pde/eclipse.pde/commit/1ba2338220919a4479fb314ba1501ff814cb9bf6) ?

Nope.
The test tries resolve `"org.eclipse.equinox.p2.core.feature"` with source features and that of course fails if the sources are removed via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3160

@akurtakov : could you please check?

PR #1839, opened 2025-06-27
Fix TargetDefinitionResolutionTests

akurtakov:
Test is supposed to verify that the corresponding source bundles are there and this should be independent of whether it's coming from source feature or not.
Test has been extended to actually check that there are source bundles for every bundle in the feature as it was only checking that there are target bundles for all source bundles but if a source bundle was missing that could have easily be missed.
Side effect assertion about non-source bundle in source feature is being dropped as source features are on their way out.
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1838 .


merks:
Yes I recall this test being a pending problem.

PR #1844, opened 2025-06-29
Avoid deprecated `org.eclipse.jdt.ui.actions.ShowInPackageViewAction`

ruspl-afed:


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/ShowInPackageExplorer.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.ui.JavaUI;
+import org.eclipse.jface.viewers.StructuredSelection;
+import org.eclipse.pde.internal.ui.PDEPlugin;
+import org.eclipse.ui.IViewPart;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.part.IShowInTarget;
+import org.eclipse.ui.part.ShowInContext;
+
+public final class ShowInPackageExplorer implements Runnable {
```
HannesWell, line 27: Wouldn't it be simpler to just have a static method `showInPackageExplorer(IPackageFragment fragment)`?
Then the callers would not first have to map an optional to a new instance of this class and run it if present, instead it could just be `...ifPresent(ShowInPackageExplorer::showInPackageExplorer)`.

And encapsulating this one action in a whole class is IMHO a bit heavy-weight. I think the static method would then be sufficiently small to add it e.g. to `EditorUtilities`, where many similar methods already reside.

HannesWell (commented):
Thanks for helping to get rid of deprecated code.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/ShowInPackageExplorer.java
laeubi: I fail to see what benefits it gives us to implement runnable here? Also the class does not seem to encapsulate any data except it being a data-carrier for the `IPackageFragment`. So using a static method in a place that already exits and could be reused by other place to not look "bad" to me, even though I think there are no strict rules we are forced to follow.

In general of course try to follow the existing style of a project (regardless of its considered bad or not) could have some like this results usually in smaller patch sets that are easier to review and the project has not to learn new styles.

So I think that can be sen more as a guidance or maybe a personal preference than some hard rules we like to enforce.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
```diff
@@ -357,16 +354,8 @@ private IPackageFragment getPackageFragment(ISelection sel) {
...
-		if (frag != null) {
-			try {
-				IViewPart part = PDEPlugin.getActivePage().showView(JavaUI.ID_PACKAGES);
-				ShowInPackageViewAction action = new ShowInPackageViewAction(part.getSite());
-				action.run(frag);
-			} catch (PartInitException e) {
-			}
-		}
+	private void handleGoToPackage(IStructuredSelection selection) {
+		Optional.ofNullable(getPackageFragment(selection)).map(ShowInPackageExplorer::new).ifPresent(r -> r.run());
```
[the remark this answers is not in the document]
ruspl-afed: > I would have expected it that getPackageFragment(selection); already returns an optional here.

It was so in initial version, but then the fight with `null` spreads more and more and I reverted it to avoid loosely related changes.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/ShowInPackageExplorer.java
ruspl-afed: Strange GitHub infra issue
image

ruspl-afed: did another force push to my fork and now it is updated

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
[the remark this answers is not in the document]
ruspl-afed: @laeubi I suggest to revwork nullable `IPackageFragment getPackageFragment` to `Optional<IPackageFragment> extractPackageFragment` with another PR

HannesWell (commented):
Thanks for the updates, the code looks good to me now.

I just think we should not add generic comments to the contributors list.
But besides that I'm fine with this change.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/EditorUtilities.java
```diff
@@ -10,6 +10,7 @@
  *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
+ *     see git history
```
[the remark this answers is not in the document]
ruspl-afed: ok, this was another point to clarify, thank you for your feedback @HannesWell 
this line was reverted in all files

PR #1845, opened 2025-06-30
Add getAdditionalEntries to IClasspathContributor

steffenschmitt1:
Currently, it is only possible to add classpath entries at the beginning of the calculation process. However, in order to properly take AccessRule.IgnoreIfBetter into account from all other entries, it is necessary to add entries at the end of the calculation.

This PR introduces a new method that allows classpath contributors to append additional entries at the end of the calculation process, enabling more accurate handling of access rules.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -52,6 +52,7 @@ import org.eclipse.osgi.service.resolver.HostSpecification;
 import org.eclipse.osgi.service.resolver.ImportPackageSpecification;
 import org.eclipse.osgi.service.resolver.StateHelper;
 import org.eclipse.pde.core.IClasspathContributor;
+import org.eclipse.pde.core.IClasspathContributor2;
 import org.eclipse.pde.core.build.IBuild;
 import org.eclipse.pde.core.build.IBuildEntry;
 import org.eclipse.pde.core.build.IBuildModel;
@@ -259,6 +260,11 @@ class RequiredPluginsClasspathContainer {
 
 			addJunit5RuntimeDependencies(added, entries);
 			addImplicitDependencies(desc, added, entries);
+
+			// Add any additional library entries contributed via classpath
+			// contributor
+			entries.addAll(getClasspathContributors().filter(IClasspathContributor2.class::isInstance)
+					.map(IClasspathContributor2.class::cast).flatMap(cc -> cc.getAdditionalEntries(desc)).toList());
 
 		return entries;
 	}
--- a/org.eclipse.pde.doc.user/forceQualifierUpdate.txt
+++ b/org.eclipse.pde.doc.user/forceQualifierUpdate.txt
@@ -1,1 +1,2 @@
 Added new method to IBuildEntry
+https://github.com/eclipse-pde/eclipse.pde/issues/1846
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IClasspathContributor2.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IClasspathContributor2.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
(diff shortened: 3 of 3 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -276,6 +276,11 @@ private List<IClasspathEntry> computePluginEntriesByModel() {
 
 			addJunit5RuntimeDependencies(added, entries);
 
+			// Add any additional library entries contributed via classpath
+			// contributor
+			getClasspathContributors().map(cc -> cc.getAdditionalEntries(desc)).flatMap(Collection::stream)
+					.forEach(entries::add);
+
```
merks, line 283: I’m looking on my phone so maybe a dumb question. Can addAll be used directly?

steffenschmitt1: I would use it to address an issue in Eclipse that's unrelated to the PDE. The problem is that JrtPackageFragmentRoots are associated with the first project in which they are opened. As a result, if the dependencies of a JAR file are not on the classpath of that associated project, features like the type hierarchy and conditional breakpoints don't work correctly.

Therefore, I would add them to the classpath using a classpath contributor, marking them as non-accessible. By using AccessRule.IGNORE_IF_BETTER, these entries are only considered if the dependency does not already exist—ensuring they're only added as a fallback at the end of the classpath.

Solving the underlying problem would require a major refactor, which is unlikely to happen.

merks: > Like this?
> 
> ```
> entries.addAll(getClasspathContributors().map(cc -> cc.getAdditionalEntries(desc))
> 					.flatMap(Collection::stream).toList());
> ```

Sorry to comment on my phone.  I just thought that map and forEach with addAll would be fewer calls than flattening the stream into individual elements.

steffenschmitt1: Yes, of course, you're right.
I was too influenced by the original version.
`getClasspathContributors().forEach(cc -> entries.addAll(cc.getAdditionalEntries(desc)));`
i will change it

laeubi:
This needs to be suppressed here sadly.

ui/org.eclipse.pde.core/src/org/eclipse/pde/core/IClasspathContributor.java
```diff
@@ -68,4 +68,24 @@ public interface IClasspathContributor {
...
+	 *
+	 * @param project
+	 *            the bundle descriptor for the plug-in project having its
+	 *            classpath computed
+	 * @return additional classpath entries to add to the project at the end of
+	 *         the classpath calculation, possibly empty, must not be
+	 *         <code>null</code>
+	 * @since 3.21
+	 */
+	default List<IClasspathEntry> getAdditionalEntries(BundleDescription project) {
```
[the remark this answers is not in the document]
steffenschmitt1: I'm not sure if much is gained by converting this into a stream. The call changes from
`getClasspathContributors().forEach(cc -> entries.addAll(cc.getAdditionalEntries(desc)));`
to
`entries.addAll(getClasspathContributors().flatMap(cc -> cc.getAdditionalEntries(desc)).toList());`
Or is there a better way to use it as Stream?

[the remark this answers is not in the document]
steffenschmitt1: How would the suppression work?
I don’t really have a problem with adding it as an extension interface, but I think using default methods in interfaces is probably the more elegant way to handle extensions.
That said, if changing it to an extension interface is necessary for it to be accepted, I can certainly adjust it accordingly.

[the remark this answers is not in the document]
laeubi: The stream is for the implementer not for the caller...

Streams has many benefits as return types instead of plain collections:

1. Can be collected into any type of collection required by the caller
2. Can be filtered and are lazy
3. No need to care if collection can be modified or not

The IDE should allow you suppress the error with a quickfix.

laeubi:
> @laeubi do you know anything about that build error?

The forcequalifierupdate file must be enhanced when new API is added.

laeubi (approved):
It looks fine for me, anyone else has concerns or should we merge it?

laeubi:
@steffenschmitt1 sadly in the meanwhile there is a merge conflict, please rebase on master and resolve the conflict then from my side everything seems fine.

laeubi:
I have documented the testfailure here:
- https://github.com/eclipse-pde/eclipse.pde/issues/1903

laeubi:
@steffenschmitt1 thanks for the enhancement and congratulation for your first PDE contribution :1st_place_medal:

PR #1849, opened 2025-07-06
Support special IDREF validation via ExtensionsErrorReporter.isPresent

merks:
Support special IDREF validation via ExtensionsErrorReporter.isPresent

- The quicklinks.exsd defines two attributes that are specified as IDREFs to command IDs but they actually permit a command spec and even wildcards.
- This leads to warnings when those aspects are used and those warnings are impossible for the user to fix.
- If we disable the basedOn type for those, the extension attributes no longer support Browse... to find a command, which makes authoring very inconvenient.
- Ideally PDE is made aware of these special attribute so support specialized validation.

In EPP there are a bunch of quicklinks definitions, e.g., 

https://github.com/eclipse-packaging/packages/blob/d0d154a80c075e857dcf531396546686ec844dba/packages/org.eclipse.epp.package.committers/plugin.xml#L390-L439

So we end up with lots of warnings that we can't eliminate:

image

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionsErrorReporter.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionsErrorReporter.java
@@ -20,6 +20,8 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.StringTokenizer;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IMarker;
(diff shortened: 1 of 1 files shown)


merks:
For 

https://github.com/eclipse-packaging/packages/blob/d0d154a80c075e857dcf531396546686ec844dba/packages/org.eclipse.epp.package.committers/plugin.xml#L390-L439

without this fix we have these warnings:

image

With the fix we have only these warnings:

image

These are legitimate because those are not in the target platform in the self-hosted debug launch.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/ExtensionsErrorReporter.java
```diff
@@ -657,15 +659,54 @@ private void validateIdentifierAttribute(Element element, Attr attr, ISchemaAttr
...
+			int index = value.indexOf("("); //$NON-NLS-1$
+			if (index != -1) {
+				baseValue = value.substring(0, index);
+			}
+			// If a simple wildcard is permitted and is present...
+			if (allowsWildcard && baseValue.contains("*")) { //$NON-NLS-1$
+				try {
+					// Convert it to a pattern and match against all keys.
+					Pattern pattern = Pattern.compile(baseValue.replace(".", "\\.").replace("*", ".*")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+					return attributes.keySet().stream().anyMatch(key -> pattern.matcher(key).matches());
```
HannesWell, line 699: This could be slightly simpler:
```suggestion
					return attributes.keySet().stream().anyMatch(pattern.asMatchPredicate());
```

HannesWell (commented):
From what you have described, it sounds like there is no good solution for the problem, besides reworking the specification and validation of extension-points in general.
Of course the latter would be better then to have special handling of certain extension-point values in PDE, but I would be ok with this.

OTOH, a relatively simple generic solution could be introducing the possibility to declare a custom validator (for attributes?) when defining a new extension-point? Then platform.ui could inject the custom validation for the affected points.

PR #1852, opened 2025-07-07
Allow configure initial values for new JUnit launches with preferences

laeubi:
Currently PDE creates a new JUnit launch with quite broad default settings that often lead to large launches and users has to adjust them after the first run to their own needs.

This now adds new preference options to control these values chosen for new launch configurations and defaults to the smallest possible configuration that is only the test-plugin and include all requirements automatically.

grafik


iloveeclipse:
I believe it would make sense to add this to the extra "Launching" sub-page, similar what is done in JDT.
Current page is already overloaded.

laeubi:
Do mean there should be a new Launching Subpage for PDE? Yes the current layout is not optimal but as the other options are also configured here it seems it would require a complete overhaul here the to be consistent.

iloveeclipse:
> but as the other options are also configured here 

With "other options" you mean "Workspace Location" block above? Can also be moved to a subpage.

> it seems it would require a complete overhaul here the to be consistent.

Not sure what complete overhaul you mean? Just moving some preferences on a dedicated page shouldn't require a complete overhaul?

PR #1853, opened 2025-07-07
Include fragments when computing dependencies for Extensible-API bundles

laeubi:
Currently there is an option to either include all fragments, or only non-test fragments. But there is some kind of special bundles (like SWT) that declare a custom 'Eclipse-ExtensibleAPI' header, these are handled special in the way that PDE attaches their fragments to the classpath of projects requiring these. This behavior is currently not reflected when computing required dependencies leading to ClassNotFoundException when launching them.

This now adds another check, that if such special bundle is found its fragments are considered when it is not a test bundle project.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -33,6 +33,7 @@ import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.plugin.PluginRegistry;
 import org.eclipse.pde.core.target.ITargetPlatformService;
 import org.eclipse.pde.core.target.NameVersionDescriptor;
+import org.osgi.framework.Bundle;
 import org.osgi.framework.Constants;
 import org.osgi.framework.Version;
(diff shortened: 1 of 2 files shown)


HannesWell:
The change looks reasonable, especially to solve the SWT fragment problem.

I found the following uses of the `Eclipse-ExtensibleAPI` header within the Eclipse TLPs:
- https://github.com/eclipse-equinox/equinox/blob/08c5cd3ca093d6f8236621b1f4d7e388c45dfcf4/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF#L113
- https://github.com/eclipse-equinox/equinox/blob/08c5cd3ca093d6f8236621b1f4d7e388c45dfcf4/bundles/org.eclipse.equinox.registry/META-INF/MANIFEST.MF#L32
- https://github.com/eclipse-jdt/eclipse.jdt.core/blob/44b8793d6ce993a99b4eadd1ba08a48b0a95a7b2/org.eclipse.jdt.core/META-INF/MANIFEST.MF#L51
- https://github.com/eclipse-platform/eclipse.platform/blob/77879f393b9cb79882346cf4380a27cb08539748/runtime/bundles/org.eclipse.e4.core.services/META-INF/MANIFEST.MF#L47

I hope none of these drags in undesired fragments.

PR #1876, opened 2025-07-16
Problem during restore of ClassFileEditor

mehmet-karaman:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2347

- Changed the RequiredPluginsClasspathContainer interface from IClasspathContainer to IAsyncClasspathContainer

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -42,8 +42,8 @@ import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExtensionRegistry;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.jdt.core.IAsyncClasspathContainer;
 import org.eclipse.jdt.core.IClasspathAttribute;
-import org.eclipse.jdt.core.IClasspathContainer;
 import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.osgi.service.resolver.BaseDescription;
@@ -71,7 +71,8 @@ import aQute.bnd.build.Project;
 import aQute.bnd.build.Workspace;
 import aQute.bnd.osgi.Constants;
 
-public class RequiredPluginsClasspathContainer extends PDEClasspathContainer implements IClasspathContainer {
+public class RequiredPluginsClasspathContainer extends PDEClasspathContainer
+		implements IAsyncClasspathContainer {
 
 	@SuppressWarnings("nls")
 	private static final Set<String> JUNIT5_RUNTIME_PLUGINS = Set.of("org.junit", //
@@ -747,5 +748,15 @@ public class RequiredPluginsClasspathContainer extends PDEClasspathContainer imp
 		} finally {
 			addImportedPackages = false;
 		}
+	}
+
+	@Override
+	public boolean isInitialized() {
+		return fEntries != null;
+	}
+
+	@Override
+	public void initialize() {
+		getClasspathEntries();
 	}
 }


laeubi:
@mehmet-karaman I really doubt this is the "right" thing to came up with something new here, can you please explain why we need a new interface here?

iloveeclipse:
> @mehmet-karaman I really doubt this is the "right" thing to came up with something new here, can you please explain why we need a new interface here?

Because somebody merged https://github.com/eclipse-pde/eclipse.pde/pull/1552...

But seriously: the problem is that JDT UI after https://github.com/eclipse-pde/eclipse.pde/pull/1552 needs to know whether classpath containers, that were always fully initialized by JDT core before, now sometimes don't do anything on init and so JDT core provides basically "empty" classpaths for PDE based projects.

laeubi (changes requested):
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4208#pullrequestreview-3025653132

laeubi:
> now sometimes don't do anything on init and so JDT core provides basically "empty" classpaths for PDE based projects.

It only does when callers ask for classpath while the Jobframework is offline what can lead to infinite blocking of the UI, but sure its easier to "blame" someone else PR instead of fixing this really bad behavior.

Beside that, a classpath can change **anytime** but some code seem not to account for that fact, what immediately would fix the problem as well because then you would only see an empty editor for the time until the classpath is actually updated.

laeubi:
> Which was like this since ~20 years and no one complained...

Yes and 20 years before people where using 32 bit CPUs and 56k Modems and have their target platforms local on the hard disk... Times are changing and now we have Clouds and Servers that not always respond in time and Firewalls that block request.

And that no one complained is simply wrong see here:
- https://github.com/eclipse-pde/eclipse.pde/issues/1481

also people are getting more used to kill their Eclipse instead to bother reporting a problem that gets closed because its not easy to reproduce.

> The fix is right here, and I would expect the fix would be provided by the same person who introduced regression.

I can tell that PDE will not accept that "fix" as it introduces the same bad behavior as before just at another call-path.

I have documented the problem here:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1860
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3468
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2636

so if someone is eager to work on that please do so, unless this is fixed it is unlikely the situation will improve. Just because it is wrong for 20 years (what might or might not be the case as no one has checked since JDT is calling that code from UI during startup) does not mean it is acceptable to be wrong for the next 20 years.

iloveeclipse:
> I can tell that PDE will not accept

**You** are not PDE.
See https://projects.eclipse.org/projects/eclipse.pde/who

merks:
This exchange is going in a poor direction.  Please keep in mind that all of us want eclipse to be excellent and each of us invests a lot of time toward that goal. Opinions will sometimes differ on the best course of action.  We can all agree there is some very unfortunate timing behavior on startup.  It would seem best to address this without going back to doing long running network/disk io on the UI thread on startup.  I suppose one might argue that is just my opinion which indeed it is.

iloveeclipse:
> Beside that, the problem it is claiming to solve can be reproduced without PDE involved:
> 
> * [Problem during restore of ClassFileEditor #2347 eclipse-jdt/eclipse.jdt.core#4208 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4208#issuecomment-3079319265)

This is a different problem : https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4208#issuecomment-3079406033

If you are confused by "async", let's name it "lazy". That would surely better match PDE container behavior.

laeubi:
> If you are confused by "async", let's name it "lazy". That would surely better match PDE container behavior.

The classpath container is neither Lazy nor Async, the whole build process, the whole PDE tooling can work with it without a problem. Also this "new" API does nothing else than calling the same API that is already avable, so if you think it is the right fix, then simply call `getClasspathEntries()` when you currently call `initialize()` and you are done without any change in API or inside PDE or any other Classpathproviders out there.

While I strongly believe that this will open the door for many concurrency issues and only shows that code is called before things are properly initialized at some other place, it seems the "right" thing to do here instead of putting the burden for such hacks to other projects.

iloveeclipse:
> The classpath container is neither _lazy_ nor _async_ and the whole build process and the whole PDE tooling can work with it without a problem

Well, except that now we have regression https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245. 

This regression is caused by the [hack in PDE](https://github.com/eclipse-pde/eclipse.pde/pull/1552), and now we providing a way to fix the regression via API, making the hack not a hack anymore.

iloveeclipse:
> I’m really not sure how to judge this change without an overview of where and how this is used. How does this resolve problems? At this point it doesn’t even compile. I suppose one question is what happens in the caller when isInitialized returns false? Perhaps we can reserve judgment until a complete picture has been sketched…

All valid points. I apologize the PR's were created without proper explanation.

This PR is one from three PR's needed to fix regression introduced by https://github.com/eclipse-pde/eclipse.pde/pull/1552, see
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245#issuecomment-3079178540

@mehmet-karaman will provide an overview over the problem we fix : steps to reproduce, reason why https://github.com/eclipse-pde/eclipse.pde/pull/1552 caused that and how the proposed patches are supposed to fix this particular regression.

mehmet-karaman:
Hello, 

The problem can easily reproduced with the following steps:
- Create a plugin project
- Add any plugin as a plugin dependency to the MANIFEST.MF (for example: org.eclipse.jface)
- Navigate to the newly created Project in the Package Explorer 
- Expand "Plug-in Dependencies"
- Find the newly introduced depencies entry and open any class file in the ClassFileEditor (it just must be a class inside the Plug-in Dependencies container)
- Restart Eclipse

After the restart you will see that the ClassFileEditor coulnd't be restored as before the changes in PDE.

My investigations on that side was that the PDE classpath container wasn't resolved before the editor was restored. From jdt.ui or jdt.core side I don't have access to PDE so I don't know whether the PDE classpath container is initialized or not. So I had to introduce a new classpath container interface which is able to tell me if its initialized or not. @laeubi I guess this is the answer for your question why there is a new Interface introduced.

On JDT side i have implemented a delayed IEditorInput which delays any UI action until the PDE classpath container is initialized. After the PDE classpath container is initialized I retrieve the original IEditorPart and execute the delayed UI executions. This solves the problem. 

@iloveeclipse sorry that you got confronted with these critical comments. For all others your questions and reviews all welcome. Feel free to ask further questions I am willing to answer them. Thank you in advance.

Best regards
Mehmet

laeubi:
See comment here:

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2348#pullrequestreview-3028084194

and supporting PR:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2349

nothing of that requires new API... if the classfile is not on the classfile it does not means the container is "lazy" it does means the initialization is in progress (in a background thread). In the worst case this simply fetches all content twice.

laeubi:
I have now submitted a fix for the ClassFileEditor editor here:

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2351

as it does not require any changes to API or PDE at all I would assume this PR obsolete now, and therefore closing it. If later something more is needed we can reopen and/or create a new one.

PR #1882, opened 2025-07-19
Add a final check for references of unused packages

laeubi:
Currently there is one category of usages that are now no longer detected since we use bndlib to find referenced packages that is if something is only needed at compiletime because the compiler otherwise inline the value.

This now restores the previous java search but only for these packages that are selected to be removed. This is faster as we do not need to query all packages and discovers cases where an import is only needed for a compile time constant.


HannesWell:
> This now restores the previous java search but only for these packages that are selected to be removed. This is faster as we do not need to query all packages and discovers cases where an import is only needed for a compile time constant.

Also when looking at https://github.com/eclipse-pde/eclipse.pde/pull/1862, I was wondering if search provided by JDT should be enhanced somehow in order to be faster in general?

laeubi:
One problem is that currently we also gather all searches where we would be happy with one match... in any case performance is not that important than correctness, so if it work reliable I'm all happy with that at a first step. Also it might be good to cache the searchscope and such kind of optimization.

merks:
Yes, great work!

PR #1887, opened 2025-07-23
Update the project with an empty classpath container on initialize

laeubi:
Currently there are some cases where an open JavaEditor is not updated, even though the build of the project is fine with that. Strange enough one can make it "work" by simply make some actions slower(!) e.g. setting breakpoints somewhere in the code that initialize classpath containers. The reason is that if initialize happens "too fast" it leads to missing classpath change notification because it looks like the saved to disk state is up-to-date with the new state and no event happens.

This now set an empty intermediate container state to the JDT model so that any code arising at this point in time will see something has changed (from n > empty), and if then at any later time an update is performed from empty > n.

**Special thanks and kudos** to @merks that debugged the case and lead me on the right track by finding out that PDE _is to fast_ where one would assume its a problem to be _to slow_ when something is not updated.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
@@ -27,6 +27,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.ClasspathContainerInitializer;
 import org.eclipse.jdt.core.IClasspathContainer;
+import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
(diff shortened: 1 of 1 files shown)


iloveeclipse:
> missing classpath change notification because it looks like the saved to disk state is up-to-date with the new state and no event happens

If that is the case (classpath saved is same to classpath set) there shouldn't be any notification, because otherwise we would have unwanted rebuilds. Please don’t merge the patch without extensive testing. For me it seems to be yet another unwanted side effect of https://github.com/eclipse-pde/eclipse.pde/pull/1552.

laeubi:
> If that is the case (classpath saved is same to classpath set) there shouldn't be any notification, because otherwise we would have unwanted rebuilds. 

I plan to further improve this to let PDE restore classpath state from saved state (like m2e does) but first priority is to get _correct_ state. I can't tell if its a bug or a feature, but obviously some code in JDT is not recovering correct state and than falls apart. So if you can suggest a patch in JDT that makes it unnecessary that would of course be good, but in the meanwhile we need to make progress and as the actual causing issues in JDT seem not be addressed its only thing PDE can do at the moment.

iloveeclipse:
> but first priority is to get _correct_ state

Where it is incorrect? Probably this is missing in the bug description? I'm not aware about opened bugs related ro wrong PDE project state.

What I mean is: have few plugin projects in the workspace. Start Eclipse, do a clean build, shutdown. On next restart no build should be triggered (autobuild would start on startup but the delta should be empty so they will not recompile). To verify that, either start with tracing at compiler (so one can see no class files were written), or check timestamps or simply use bigger projects and check build time.

If sou see now rebuilds with this change, this would be a blocker.

laeubi:
> Where it is incorrect? Probably this is missing in the bug description?

It seems JDT requires for proper function that we set an empty classpath as otherwise it reports NO_DELTA, what does not trigger any events if PDE is too fast later on. So unless there is no "Empty Delta" event there is no chance for components to react to classpath changes (and obviously it makes a difference here).

So if we can fix any of those things it would not need any change in PDE of course.

laeubi:
> Proper function of what? 

e.g. Eclipse Java Editor is showing error markers even though there are no errors at all if it is opened and you restart the IDE and the InitClasspathJob takes longer than PDE requires to update the classpath.

merks:
@iloveeclipse 

I have another very simple change pending that also eliminates the transient error annotations in the editor.  But that's in this same file on top of this change...

laeubi:
Thanks @merks again for verification, lets merge and do other improvements on top, as said I plan to further go ahead to cache the last computed state and return it here so we can check for updates in the background entirely.
This then will likely improve initial startup times as currently much time can be spend to resolve the target platform and compute the container just then to see nothing has changed.

PR #1888, opened 2025-07-23
Improve DeferredClasspathContainerInitializerJob.belongsTo

merks:
- Add ClasspathContainerInitializer.class to the test for families in belongsTo to allow a client to wait for classpath container initialization without knowing or caring about PDE specifically.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsInitializer.java
@@ -104,7 +104,7 @@ public class RequiredPluginsInitializer extends ClasspathContainerInitializer {
 
 		@Override
 		public boolean belongsTo(Object family) {
-			return family == PluginModelManager.class;
+			return family == PluginModelManager.class || family == ClasspathContainerInitializer.class;
 		}
 	}
 


iloveeclipse:
> Both those scenarios are ones where the main thread would have blocked which is definitely worse than the reconcile thread blocking.

Sure. 

> I think it's important to keep in mind that there will be a wait if and only if PDE is doing deferred initialization. 

Sure.

That's what we see now with startup if a Java editor is opened, and I believe everyone has some editors opened.
So we are talking about use case of workspaces with mixed projects and slow PDE init.

What I mean is: now we would deferring reconciler for **every** Java file opened, independently whether it is related to PDE or not. Ideally, I would expect it should only affect Java editors for PDE projects, the rest shouldn't be affected, but the envisioned patch would do that.

Why I'm concerned: originally proposed patch from Mehmet would not have this problem...

merks:
> What I mean is: now we would deferring reconciler for every Java file opened.

Yes, if an editor on a PDE project is opened **and is visible** then all editors will be block.  Of course this is the scenario where the entire UI thread was blocked on startup for the entire period of time.  It was that way for decades, and then very few people seemed to care about such a serious problem. So you'll probably understand why I'm confused by corner cases of mixed PDE and non-PDE scenarios that were completely a disaster in the past.   I.e., why is it a concern now but never was a concern before.  Anyway, I feel like I'm falling into the trap of dredging up history, and I'm tired of seeing that because it doesn't move forward.

In any case, this whole discussion is about scenarios is concerning what JDT can do, should do, or might do.  It's not about about this specific PDE PR which simply provides the ability to wait if desired; it could use a constant, I don't really care. Also, if you prefer to actually solve the problem in some other way in JDT, that's of course your choice as JDT committers to make and I'll not get int he way.

What I would like is is to put this is place, even if it's useless.  And I would like to create PR for JDT, which does solve the problem and which you can choose to ignore in favor of something better.  Unfortunately the time I have available only allows these options for me.    I've become completely confused by the competing different solutions, and class editor versus java editor, with or without PDE, all of which are rejected for various reason along with way too much emotional/personal baggage for my taste.   Way too much.

So, we either move forward with something really simple that does work, or folks can invest their own time on other solutions as they see fit.

iloveeclipse:
Ed, this is all clear, no doubts. I just wanted to make sure the problem is discussed and understood, and maybe something could be improved for that specific case.

iloveeclipse (approved):

PR #1898, opened 2025-07-25
Save and restore the current plugin classpath container state from disk

laeubi:
Computing the bundle plugin classpath can be an expensive operation for different reasons. In many cases these computation even yields the same results as before especially on workspace restart without target changes.

This now stores the last successful computed state per project and read it back on initialization of the classpath container, scheduling a check for actual changes in the background. Only if this check yields a different result or the update is requested as part of a change to the manifest or target state for example the current values are updated.

This yields faster startup times without the risk of blocking (other than file I/O) until PDE can deliver the classpath to consumers delaying the heavy work to a later time.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
@@ -13,6 +13,13 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.core;
 
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -23,6 +30,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.regex.Pattern;
@@ -32,6 +40,8 @@ import java.util.stream.Stream;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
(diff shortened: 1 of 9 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathComputer.java
```diff
@@ -598,10 +618,92 @@ protected IStatus run(IProgressMonitor monitor) {
...
+			if (project == null) {
+				return;
+			}
+			workQueue.add(new UpdateRequest(project, classpathContainer));
 			schedule();
 		}
 
 	}
 
+	private static boolean isUptoDate(IProject project, IClasspathEntry[] currentEntries,
```
merks, line 637: Just a nit, but upto is not a word so this should be isUpToDate.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PdeClasspathContainerSaveHelper.java
```diff
@@ -0,0 +1,218 @@
...
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Sonatype, Inc. - initial API and implementation
+ *******************************************************************************/
+
```
merks, line 13: Was this copied from somewhere?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PdeClasspathContainerSaveHelper.java
```diff
@@ -0,0 +1,218 @@
...
+import org.eclipse.jdt.core.IClasspathContainer;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.JavaCore;
+
+/**
+ * BuildPath save helper
+ *
+ * Adapted from org.eclipse.m2e.jdt.internal.MavenClasspathContainerSaveHelper
+ */
+public class PdeClasspathContainerSaveHelper {
```
merks, line 37: Does it need a PDE prefix?  If so, it should be PDE not Pde.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/PdeClasspathContainerState.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.Serializable;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.jdt.core.IClasspathContainer;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.pde.internal.core.PDECore;
+import org.eclipse.pde.internal.core.PDECoreMessages;
+
+class PdeClasspathContainerState implements IClasspathContainer, Serializable {
```
merks, line 24: PDE not Pde please.

merks (changes requested):
Some minor cosmetic changes.

It appears PDE uses tracing for other things:

image

merks (approved):

laeubi:
As it might has an impact due to different timings now I started an I-Build to see if anything goes wrong in the large scale:
- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/81/

PR #1904, opened 2025-07-28
Write trace messages to the trace log for ClasspathComputer

laeubi:
Currently PDE uses System.out to print trace messages, that has the problem that id the user has configured a trace file it does not show up there.

This now acquire a DebugTrace instance that can be used and use that in the ClasspathComputer.


laeubi:
Triggered Ibuild to get proper tracing messages now:
- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/85/

Issue #1850, opened 2025-07-06
Add a "Show Launch Bundles" button


laeubi:
Currently one can enable the option to include all bundles in a launch that are transitively required what is really helpful, but now there is the problem that if I want to see what bundles will be launched I can not easily do that.

It would be good to have an option similar to "show commandline" that shows all bundles that will be launched (at best in a tree like in maven where one can see what pull ins what other things).

nburnwal09:
I will pick this one.

HannesWell:
And just in case you are looking for a method to get the bundles included in a launch, you should have a look at:
https://github.com/eclipse-pde/eclipse.pde/blob/260640d9827f44452bd0e03d1baea7efe3ab191b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java#L110

laeubi:
> As Christoph said a filterable simple list is more than sufficient. But if there are some long-hanging fruits that can be reused, it's even better.

To clarify:

1. A plain list is better than nothing, it should just highlight what is _selected_ and what is added _additional_.
2. A tree that shows bundles from their roots (= selected in the launch) to what they pull in is better (but maybe improved later)
3. Showing packages and locations like in the provided screenshot is not required

The purpose is to help users with two things:

1. Understand what is included without launching the prodcut / test / ...
2. (optional) what includes a possible problematic dependency  can be removed from the launch and its impact.

nburnwal09:
> As Christoph said a filterable simple list is more than sufficient. But if there are some long-hanging fruits that can be reused, it's even better. What would be nice to have is if one selects some (or all bundles in the launch) that can can copy the names (plus version or the path), so that one can for example compare different launches in a simple diff-editor. But this is just an idea and mandatory.

Essentially this means, I need to implement the copy handler and for the selected plug-ins I would need to get the name, id and location of the plug-in onto the clipboard. Did I catch it correctly?
Can I do this in a follow up PR?

PR #1922, opened 2025-08-07
Added a button to show launch bundles

nburnwal09:
Added a button to show launch bundles: #1850 
It has filter option and shows number of bundles selected(attached screenshot)
image

image

image

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -2475,6 +2475,8 @@ public class PDEUIMessages extends NLS {
 
 	public static String AbstractPluginBlock_counter;
 
+	public static String PluginsTabToolBar_show_launch_bundles;
+
 	public static String AbstractRepository_ErrorLoadingImageFromJar;
 
 	public static String AbstractRepository_ScanForUI;
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java
@@ -42,6 +42,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.util.Util;
 import org.eclipse.jface.viewers.CheckStateChangedEvent;
@@ -67,6 +68,7 @@ import org.eclipse.pde.internal.ui.PDEPlugin;
 import org.eclipse.pde.internal.ui.PDEPluginImages;
 import org.eclipse.pde.internal.ui.PDEUIMessages;
 import org.eclipse.pde.internal.ui.SWTFactory;
+import org.eclipse.pde.internal.ui.dialogs.PluginSelectionDialog;
 import org.eclipse.pde.internal.ui.elements.NamedElement;
 import org.eclipse.pde.internal.ui.shared.CachedCheckboxTreeViewer;
 import org.eclipse.pde.internal.ui.shared.FilteredCheckboxTree;
@@ -142,6 +144,7 @@ public abstract class AbstractPluginBlock {
 	private Label fCounter;
 
 	private Button fValidateButton;
+	private Button fShowPlugin;
 
 	private HashMap<Object, String> levelColumnCache = new HashMap<>();
 	private HashMap<Object, String> autoColumnCache = new HashMap<>();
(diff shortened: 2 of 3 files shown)


gireeshpunathil (approved):

nburnwal09:
>     For #1850 I initially thought about a separate tab or something like that.

thanks @HannesWell . I have these questions:

1. What would be an added benefit of having a separate tab as opposed to the current checkbox? is it to accommodate more scenarios than just the bundles view, or to improve the user experience, or something else?

In case we add a new separate tab:

2. are you suggesting to have a single view (in the new tab) for both features and plugins? and the tab embedding the tree view / list view in itself, as opposed to a popup dialog? and in the tree / list view, selection of one ore more features in the left shows the associated plugins? (with a default of all features and all plugins in the launch configuration displayed)
3. Can you please clarify on what you mean by re using the selection tree in the *centre*? if we want a feature-to-plugin mapping to be displayed, shouldn't the features be in the left, as opposed to be in the center?
4. What would happen about the unbounded plug-ins which doesn't correspond to any features? should we list them as is
5. `>     Maybe we could instead also re-use the selection tree in the center and just have a check-box on the right to list all included (features and) bundles. Of course selecting them should then not be possible.` - can you pls clarify why selecting them should not be possible?

nburnwal09:
@HannesWell 

> 1. The button should also be available for Eclipse-apps based on features (at least personally I'm hardly using bundle-based apps). This can be done easily by repeating the same code of the `AbstractPluginBlock` in the `FeatureBlock`, around here:

I added the same for Eclipse-apps based on features. Also, with current changes, if we select one/more features and click on 'show launch bundles', it is showing the list of plug-ins(direct and transitive dependent plug-ins both) respective to the features selected.

> 2. While using the `PluginSelectionDialog` is certainly a quick solution, I'm not sure if it's really suitable. At least the title of the dialog should be changed to something more meaningful in this use-case.

Yeah I missed that. I have changed the title to 'Launch Plug-ins list'

>    What I'm currently missing is the ability to copy the list of plugins (ideally just the selected sub-set) and being able to past it e.g. into a text- or diff-editor. This way I often try to figure what has changed in the full application. But that's not working at the moment.

Yeah I have that in my mind(based on our conversation in https://github.com/eclipse-pde/eclipse.pde/issues/1850#issuecomment-3139231513). I will be taking the copy/paste option in another PR.

laeubi:
> Ideally features would then also be listed (in addition to bundles), but I assume the currently used dialog is not capable of that.

Features are exploded and do not remain in the final runtime, so I would say this is not really useful here. One might want to show what a feature pulls in but then we need a tree to represent the data and it would possibly complicate things a lot.

So I would vote for starting simple and improve on that iteratively.

nburnwal09:
@laeubi 
> Features are exploded and do not remain in the final runtime, so I would say this is not really useful here. One might want to show what a feature pulls in but then we need a tree to represent the data and it would possibly complicate things a lot.

With my changes, we will still be able to see what a feature pulls in(the direct and indirect dependent plug-in). 

> I added the same for Eclipse-apps based on features. Also, with current changes, if we select one/more features and click on 'show launch bundles', it is showing the list of plug-ins(direct and transitive dependent plug-ins both) respective to the features selected.

laeubi:
While this can be interesting I don't think the selection should matter and would expect to always see the final result of what will happen when I click `Run`

nburnwal09:
Hi @HannesWell 
Kindly have a look at this PR.

HannesWell:
> > ```
> > For #1850 I initially thought about a separate tab or something like that.
> > ```
>     1. What would be an added benefit of having a separate tab as opposed to the current checkbox? is it to accommodate more scenarios than just the bundles view, or to improve the user experience, or something else?

Good question. I cannot answer this clearly, but I would say that it would allow to support more scenarios or 'tools' (later).
This could also be a general UI/UX questions: Why are things located in tabs and not just a wizard or preferences like page that opens when you click a button. I would say it 'feels' better from what I'm used to, but I'd need to ask an UX expert for a more clear and scientific answer.

But to come back to the current proposal and as my counter-suggestion from https://github.com/eclipse-pde/eclipse.pde/pull/1922#pullrequestreview-3113329612, is indeed much more complex, what about re-using the approach used for the `Show
Command Line` button?
This button just uses a not editable, but selectable text block/field to show the resolved command line:


If we use that to show the list of resolved bundles in CSV style (if we want to have more then the name) and also add a search field to for the text, then we would have, what would be provided with your proposal plus the ability to copy (parts) of the list.
Plus you could easily add start-level infos or the file-system path to the jar using a CSV style.

To feed the content of that text field you could actually use what's generated into the `<configuration-path>/org.eclipse.equinox.simpleconfigurator/bundles.info` nowadays (if the app contains P2). Otherwise you could use the value of the `osgi.bundles` field in the `<configuration-path>/config.ini`. Or you just use the current approach and add the suggested additional information from it (this would avoid the need to check two possible files, with different format.

Overall I think it should still be not too complicated to achieve and should provide more information than today.

What do you think?

nburnwal09:
> If you can make it fully copy-able where one gets the bundle-symbolic name and the version (and maybe even the file-system path) speparated by space or semicolon or similar copied into the clipboard on copy, a table like UI would definitively be nicer. But as it would probably be more complex I would also be fine with a copyable, read-only text field, maybe just as a first step. As mentioned in [#1922 (comment)](https://github.com/eclipse-pde/eclipse.pde/pull/1922#issuecomment-3330896188), I personally use the simple-configurator file for the designated purpose of the new button and it gives me everything I need. Although I agree that it feels quite low-level.

@HannesWell 
Yeah, but if I use simple-configurator - bundles.info, I would need to filter all the selected options from the plugins list. 

Earlier you had mentioned [this](https://github.com/eclipse-pde/eclipse.pde/issues/1850#issuecomment-3128844375)  to fetch all the details which gives the list of selected plug-ins. 
I will check if I can get the bundles details along with start level and file-system path to the jar from here. If not, I would filter and get it from bundles.info

nburnwal09:
@HannesWell 
Please have a look at this commit. I have made necessary changes which allows selecting and copying the bundle values.
This is the view now:
image

nburnwal09:
Hi @HannesWell 
Could you please look into the changes whenever you get sometime?

gireeshpunathil:
this is ready to land, but would benefit from a final review by @HannesWell . @HannesWell - pls have a look?

HannesWell:
I just looked into this and in general it looks good. Thanks for your work on this.

But I have multiple minor points that should be addressed.
- The pop-up listing all bundles should be wider or at least should not break lines and instead show a horizontal scroll-bar
- In the headline of the pop-up, i.e. above the text field, there should be an extra line explaining the format of the text below.
- For `OSGi Framwork` launch configs the formatting seems off and the code in `createControl()` should be stream-lined, but that's not related to this PR and I'll provide a dedicated one this evening.

I'll provide more details later today as I have to leave now.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/ShowBundlesDialog.java
```diff
@@ -0,0 +1,103 @@
+package org.eclipse.pde.internal.ui.launcher;
```
HannesWell: This class is missing the copy-right header.

HannesWell (approved):
Thank you @nburnwal09 for the update.

I made a few more adjustments, mostly minor and the most relevant one was to move the explanation for the new dialog into a separate label above the list of bundles.
With that this is ready for submission for my side.

PR #1925, opened 2025-08-08
Add all required bundles to the classpath

datho7561:
When building the classpath, don't skip over bundles that aren't re-exported; when using `javac`, `javac` requires that the classes referenced in `.class` files are present on the classpath.

Fixes eclipse-jdtls/eclipse-jdt-core-incubator#421

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -404,9 +404,7 @@ class RequiredPluginsClasspathContainer {
 
 		BundleSpecification[] required = desc.getRequiredBundles();
 		for (BundleSpecification element : required) {
-			if (element.isExported()) {
-				addDependency((BundleDescription) element.getSupplier(), added, map, entries, useInclusion);
-			}
+			addDependency((BundleDescription) element.getSupplier(), added, map, entries, useInclusion);
 		}
 
 		if (addImportedPackages) {


laeubi (changes requested):
While I understand the general demand behind your change this actually will open access to classes that are never imported, so we can not do it (and that's likely why the tests fail).

What would be needed if we want a complete transitive classpath is the following:

1. We need to keep track of all Bundle description we add during the classpath resolution
2. now we need to iterate over all of the items and collecting their full set of transitives while ignoring what already was added before
3. then add all additional items with a restriction rule that forbids actual access to everything
4. This should also be covered by additional test and include case where require bundle is used, but also import package

**This leads us to a much more serve problem with using `javac` and PDE:**

`javac` does not understand the concept of access restriction, so using `javac` + PDE will lead to the situation that you can access any type even if not imported what will **guarantee to fail at runtime** and we can not give any hints to the user that something is wrong! So unless one uses the automatic manifest generation (that is not touched by this change at all) it needs the user is finding possible issues and fixes manually what is not something sensible at all!

So we either need to make `javac` supporting some kind of access restrictions (e.g. through a compiler plugin?) or add support in PDE at another level e.g. in the form of a [CompilationParticipant](https://help.eclipse.org/latest/topic/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/core/compiler/CompilationParticipant.html).

 Because of this I would recommend the following:

1. try to evaluate if we can have some kind of feature detection if `ecj` or `javac` is used as a compiler backend, then we can enable/disable some things already now so you can make progress
2. investigate if/how we can possibly extends `javac` either by a compiler specific extension mechanism, or even better if the `javac` integration can support [CompilationParticipant](https://help.eclipse.org/latest/topic/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/core/compiler/CompilationParticipant.html) because this is used in many other places, so we would need that rather sooner than later
3. creates some test-cases that show a failure with the current approach.

I know that's probably much to ask compared to your at first sight easy finding to fix circumvent the issue you see, but I think it is mandatory due to the mentioned restriction of `javac` if we want to add real OSGi support in the IDE.

I will in the meantime evaluate how we can best add the validation through a `CompilationParticipant` as this might be a good thing to have in both worlds as well given that the JDT restrictions support currently [do not match well](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293) so we maybe need to came up with something on our own anyways. I'll be only partly available the next two weeks but try to look into this if possible and catch up with the topic afterwards, so please be patient with me not reacting immediately sometimes.

datho7561:
Okay, that makes sense. This is the wrong solution. I verified that after this change, I can use classes from bundles that aren't reexported without any compiler warnings. 

I'll take a look into how ECJ handles access restrictions.

We've been able to add an additional compilation step when using `javac` that reimplements `ecj` s unused variable/unused import problem markers (since `javac` doesn't provide those). Maybe we can do something similar for the access restrictions.

As for the differences in the classpath needed for `ecj` and `javac`, I'll need to think about it a bit. Would it okay to contribute an implementation that collects the classpath as required by `javac` that's not used by default, then add a flag to enable this implementation (via a system property)?

mickaelistria:
I am wondering whether this limitation of javac could/should be reported to javac.
@datho7561 can you share a very minimal example of a project that would build with ECJ command-line and would fail with Javac? If we agree that it would make sense for Javac to support it, we could report it and hopefully even improve javac.

mickaelistria:
Thanks for sharing this test.

>  an optimization of type lookups.

These days, memory and CPU optimizations are very welcome in Javac. A lazier loading classes/types would probably be perceived as a strong improvement by javac maintainers.
So that's worth trying IMO.

PR #1932, opened 2025-08-17
TestPDETemplates.createProjectWithTemplate use only supported EE

merks:
- The tests fail when running on Java 25 because that's not currently a supported execution environment.

### DIFF
--- a/ui/org.eclipse.pde.ui.templates.tests/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.ui.templates.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for PDE templates
 Bundle-SymbolicName: org.eclipse.pde.ui.templates.tests
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
 Bundle-Vendor: Eclipse.org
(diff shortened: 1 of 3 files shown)


merks:
@akurtakov 

I'm not sure this is the best way to ensure that the tests are configured only with a supported execution environment.  The unit tests pass if they avoid configuring JavaSE-25 when running on Java 25 but rather configure it to use Java 24 in that case.

HannesWell (approved):
> I'm not sure this is the best way to ensure that the tests are configured only with a supported execution environment. The unit tests pass if they avoid configuring JavaSE-25 when running on Java 25 but rather configure it to use Java 24 in that case.

I don't have a better idea than this or just accepting the tests to fail on not yet supported target versions.

akurtakov:
This brings the failure to https://download.eclipse.org/eclipse/downloads/drops4/I20250820-1800/testresults/html/org.eclipse.pde.ui.templates.tests_ep437I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html :
```
Marker [on: /TestPDETemplates_org.eclipse.pde.ui.pluginContent.choice, id: 16, type: org.eclipse.jdt.launching.jreContainerMarker, severity: WARNING(1), attributes: [location: Build path, message: Build path specifies execution environment JavaSE-24. There are no JREs installed in the workspace that are strictly compatible with this environment.], created: 8/20/25, 11:40 PM] expected:<0> but was:<1> 
```

Issue #1929, opened 2025-08-12
Launching: 'Include required Plug-ins automatically while launching' should include fragments like when pressing the 'Select Required' button


sratz:
In the launch config dialog,

<img width="779" height="460" alt="Image" src="https://github.com/user-attachments/assets/b948768e-0bd6-472a-af8e-831a9ca1c70b" />

the set of bundles included when _[x] Include required Plug-ins automatically while launching_ is checked does not include fragments:

https://github.com/eclipse-pde/eclipse.pde/blob/66a95fd33ca04efe7f0772c312b7177ca7e25d80/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java#L560-L562

But when pressing the _Select Required_ button, fragments are included:

https://github.com/eclipse-pde/eclipse.pde/blob/66a95fd33ca04efe7f0772c312b7177ca7e25d80/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/launcher/AbstractPluginBlock.java#L875-L877

This is confusing.

And especially, since the recent changes in #1852, that made the default _'The plugin of the Test'_, this has become fragile, because now needed fragments might be missing and the test may not launch at all.

My proposal would be to make _[x] Include required Plug-ins automatically while launching_ behave like the _Select Required_ button, and include fragments.

sratz:
@laeubi @HannesWell what do you think? I can provide a PR if you agree.

HannesWell:
> My proposal would be to make [x] Include required Plug-ins automatically while launching behave like the Select Required button, and include fragments.

The current behavior is intentional and I strongly discourage from changing it. Fragments can have new dependencies and would then pull in new bundles. And there are cases where the set of new bundles is very large.
The strategy for fragments (just like for optional dependencies) is that one has to include them explicitly in a launch if they should be in the launch. This has been discussed at length when this was introduced via [Bug 544838](https://bugs.eclipse.org/bugs/show_bug.cgi?id=544838) and related issues (I don't find the point about fragments anymore, maybe this was discussed in Gerrit and is now lost).

> But when pressing the Select Required button, fragments are included

That's just for historic reason as it was like that since ever. If you think the difference is a problem, I'm open to change that as well.

[...]

PR #1939, opened 2025-08-20
Always consider fragments that specify a platform filter

laeubi:
A fragment can declare a Eclipse-PlatformFilter to be only installed on a given platform. If such fragment is currently wired to a host bundle, it is quite likely that fragment is needed but the host usually do not declare an Eclipse-ExtensibleAPI so we currently miss to add such fragment making the native code loading fails.

This now adds some code to store the platform filter and if one is given the fragment is always included.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/1929


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDEAuxiliaryState.java
```diff
@@ -97,6 +97,7 @@ static class PluginInfo {
 		String localization;
 		String bundleSourceEntry;
 		boolean exportsExternalAnnotations;
+		public String platformFilter;
```
merks, line 100: I wonder why this is public but the others aren't?

laeubi: That's what happening when using quick-fix to create the field and devloper not checking the results :-)

I just changed this, this might even can become a record these days, but don't want to change to much at once to keep the change count low.

HannesWell:
> For me this is not really any heuristic, it is actually the reverse of Extensible-Api header, I never seen a case where one defines a Eclipse-PlatformFilter but don't want the fragment included, it is more the other way round that we take many efforts (e.g. additional `p2.inf`) to make sure the fragment is not badly missed.

All of this is in my opinion just a workaround because it's currently not possible to declare a requirement from a host to one of it's fragments and building the same in one reactor with Maven/Tycho.

Please see my answer in https://github.com/eclipse-pde/eclipse.pde/issues/1929#issuecomment-3212484648 

> > Furthermore this now affects all calls to the computation method that don't specify any options.
> 
> I now introduced a new option that is only activated in the `BundleLauncherHelper` so other callers should not be affected (even though I could not find it would harm anything on their call sites).

As written in the issue I'll probably reluctantly approve this, if this is only enabled for shrieked `JUnit Plugin Test` runtimes. But for general product's that's a no-go for me, as explicit means to include required fragments already exists e.g. for Products.

laeubi:
> All of this is in my opinion just a workaround because it's currently not possible to declare a requirement from a host to one of it's fragments and building the same in one reactor with Maven/Tycho.

Actually native fragments (and Eclipse-PlatformFilter) are the "workaround", OSGi has standard means to declare and select appropriate native code by platform, anyways its an Eclipse Extension and PDE can/should support it.

"required fragments" is really something completely alien from OSGi point-of-view ... if it is required then it should be part of the bundle, if it is truly optional then it can be a fragment, but both does not really make sense.

> As written in the issue I'll probably reluctantly approve this, if this is only enabled for shrieked JUnit Plugin Test runtimes. But for general app/product's that's a no-go for me, for the mentioned reasons and as explicit means to include required fragments already exists e.g. for Products.

The whole purpose (and power) of "include required automatically" is that it works without specify it explicitly. SO of course one wants this for products and plain OSGi launches as well!

I really don't understand the concerns mentioned in the ticket they are talking about fragments that pull in more dependencies, about test fragments and so one none of this applies to the proposed solution here:

1. Native fragments only carry native code for the whole purpose of not shipping a full set of natives to get smaller bundles and have well defined meta-data carried with
2. They never have any additional dependencies
3. We perform additional work to make sure they are always installed
4. I have never seen a native test-fragment

PR #1941, opened 2025-08-21
Possible misspelling of function name for changesForModelModication()

sderazo:
The change from changesForModelModication() to changesForModelModification() (originally declared in the PDEModelUtility.java file) is being made to improve the coding and debugging experience for developers. One of the parameters that this function takes in is final ModelModification modification, which makes me think that Modication should be Modification instead. This would help save time with debugging since when I was working on Issue 1813, I saw that there was an error message with the function and I could not figure out what was wrong until I saw that I misspelled the function by one letter.


gireeshpunathil (approved):

laeubi (approved):
Only internal so change looks fine

PR #1943, opened 2025-08-22
Fix split package between swt.spies and swt.tools

akurtakov:
Move Sleak class to o.e.swt.internal.tools.views. It's more correct (or at least less wrong as per
https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Naming_Conventions.md ).
Removes the nasty problem of split packages signed with different certificates.


merks (approved):
+1

This change is related to 

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3190

because it address this problem in a permanent by avoiding the split package altother.

image

akurtakov:
There is PMC member +1 thus merging in order to get the bundle signed proper for RC2.

PR #1944, opened 2025-08-24
Add htmlvalidator GHA

akurtakov:


akurtakov:
Like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3238 , it will be kept to always run till docs are cleaned up and filter will be added later.

laeubi (approved):
Looks fine to me

.github/workflows/htmlvalidator.yml
```diff
@@ -0,0 +1,17 @@
...
+    paths:
+      - '**.htm*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5 # Requried will all actions
+    - name: Checks HTML5
+      uses: Cyb3r-Jak3/html5validator-action@v8.0.0
```
HannesWell, line 14: To improve security it would probably be good to pin all actions to the specific release commit, similar to how it's done in the releng.aggregator workflows.

PR #1947, opened 2025-08-27
Fix faulty cheat sheet editor ID constant

HeikoKlare:
The constant does not match the actual editor ID (misses a dot).

### DIFF
--- a/ua/org.eclipse.pde.ua.ui/META-INF/MANIFEST.MF
+++ b/ua/org.eclipse.pde.ua.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.ua.ui;singleton:=true
-Bundle-Version: 1.4.100.qualifier
+Bundle-Version: 1.4.200.qualifier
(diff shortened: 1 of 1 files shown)


HeikoKlare:
The only official usage is the editor's ID getter:
https://github.com/eclipse-pde/eclipse.pde/blob/379558b67b266e7251752826951de1cd7c93db8a/ua/org.eclipse.pde.ua.ui/src/org/eclipse/pde/internal/ua/ui/editor/cheatsheet/simple/SimpleCSEditor.java#L44-L47
which in turn only seems to be used for storing dialog settings:
https://github.com/eclipse-pde/eclipse.pde/blob/379558b67b266e7251752826951de1cd7c93db8a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/PDEFormEditor.java#L521-L539

Since there are no other (official) consumers, I guess it did not produce any problems so far and would also not produce issues in the future but just be a bit inconsistent.

I only came across it because we actually use such editors but had no explicit reference/dependency to the bundle providing them (as we just replicate the editor ID) and I wanted to reference the (internal) constant instead to have an explicit dependency to the bundle, such that it gets properly auto-included. But that failed as the ID did not match actual editor ID. But since such references to internals are reasonably discouraged anyway, it's not an argument for actually doing this change.

PR #1951, opened 2025-09-01
[RelEng] Add individual release preparation script for PDE

HannesWell:
It will be called by the pipeline to prepare a new development cycle and automates the enforced qualifier update and clean-up of the 'forceQualifierUpdate.txt' files of the PDE doc bundle.

This will avoid doing it manually as done for the current cycle in
- https://github.com/eclipse-pde/eclipse.pde/pull/1948#issuecomment-3238182237

### DIFF
--- a/prepareNextDevCycle.sh
+++ b/prepareNextDevCycle.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -xe
+
+#*******************************************************************************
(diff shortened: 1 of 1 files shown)


HannesWell:
Testing was successful but the verification build doesn't test this, thus submitting this without await it.

Issue #1927, opened 2025-08-10
Missing SVG icons


HeikoKlare:
The following is a (not necessarily complete) list of icons in PDE that miss an SVG version. It is extracted from a list in the `eclipse.platform.images` repository that was created by @Michael5601 when adding SVG support to SWT and integrating the SVG files into the actual bundles:
- https://github.com/eclipse-platform/eclipse.platform.images/issues/129

Since the work of Michael is done and since that repository is about to be archived and the issue with the list will be closed, I repost it here. If it's not of interest for you, feel free to just close it.

**org.eclipse.pde.spy.preferences**:

- [ ] almost all icons from this bundle. SVGs with the same name are available but they look different.

**org.eclipse.pde.bnd.ui**:

- [ ] arrow_down.svg
- [ ] arrow_refresh.svg
- [ ] connect.svg
- [ ] connections.svg
- [ ] disconnect.svg
- [ ] file.svg
- [ ] link.svg
- [ ] loading_16x16.svg
- [ ] locked_state.svg
- [ ] repository_xml.svg
- [ ] repoindex.svg
- [ ] repository_maven.svg
- [ ] repository_pom.svg
- [ ] service.svg
- [ ] star_small.svg
- [ ] template_empty.svg
- [ ] tick.svg
- [ ] whiteboard.svg

**org.eclipse.pde.doc.user**:

- [ ] Many "icons" (PNG) that show screenshots of editors. I don't think we need SVGs for these. 

**org.eclipse.e4.tools.emf.ui**:

- [ ] Many GIFs and some PNGs that mostly do not exist as SVGs.

**org.eclipse.e4.tools**:

- [ ] Some of these GIFs and PNGs do not exist and need to be created. For the icons `newclass_wiz`, `class_obj` duplicates with identical name exist in other bundles.
- [ ] For the icon `application_view_tile` a duplicate with the name `e4.svg` was added by me in the bundle `org.eclipse.pde.ui.templates`.

**org.eclipse.e4.tools.emf.editor3x**:

- [ ] The icon `application_view_tile` does not exist as SVG. A duplicate with the name `e4.svg` was added by me in the bundle `org.eclipse.pde.ui.templates`.

PR #1952, opened 2025-09-03
Add SVG to org.eclipse.pde.spy.preferences

BeckerWdf:
Use SVGs instead of PNGs as platform now supports this.

All SVG already existing and were simply copied over.

For the preference_spy.png an SVG still needs to be created. This will be done in a separate commit.

Contributes to: https://github.com/eclipse-pde/eclipse.pde/issues/1927

### DIFF
--- a/features/org.eclipse.pde.spies-feature/feature.xml
+++ b/features/org.eclipse.pde.spies-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.spies"
       label="%featureName"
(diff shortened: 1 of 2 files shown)


BeckerWdf:
The expand / collape all, delete / delete all where greyed out in the current state (PNGs). This looked like these actions are disabled. See:
Screenshot 2025-09-03 at 08 22 49

I think they should be colorful like the other versions of these in platform. So I changed this.
See: 
image
image

@HannesWell: Can you pls. review and and manually test?

PR #1955, opened 2025-09-03
Replace PNGs With SVGs that already existed.

BeckerWdf:
This re-uses existing SVGs for the PNGs used here.


akurtakov:
Please use ISharedImages.IMG_ELCL_COLLAPSEALL and etc. that would make the change automatic in the future and requiring less copied files.

Issue #1815, opened 2025-06-14
Add an option to automatically add missing versions on exported packages


laeubi:
If packages are exported without a version this can be problematic, it would be good to have an option in the "Organize Manifest" Wizard to add missing package version on exports, so it can be applied in batch to many projects and in combination with the new [tycho-cleancode:manifest](https://tycho.eclipseprojects.io/doc/main/tycho-cleancode-plugin/manifest-mojo.html) mojo.

A user likely wants to have the following options:

1. Use a fixed version (defaults to 1.0.0)
2. Using the current bundle version

Vonna9:
Hello @laeubi, 

My teammate @hilla-ryy and I are still working on the issue #1815. We ran into a few issues importing a forked version of the  repository into Eclipse IDE for committers and have provided some of the errors we encountered in the Q&A Discussion: https://github.com/eclipse-pde/eclipse.pde/discussions/1878. We would be happy to open a Pull Request soon to share our current state of work. So far, we have been reviewing the codebase and creating a plan for how we are going to add the missing package versions to exports.

To make sure we are on the right track, we would like to verify with you a plan we created based on our understanding of the codebase. We plan on: 

1. Adding a checkbox button to the **Organize Manifests Wizard Page.java** file that says "add missing package versions on exports" and two check box options to either “use a fixed version (defaults to 1.0.0)” or “using the current bundle version”. 
2. Editing the **Organize_Manifests.html** file to specify the options we added to the Organize Manifests Wizard Page.
3. Editing code in the **Organize Manifest Settings.java** file to add our new “add missing version packages to exports” as an available setting. 
4. Editing code in **Organize Manifests Processor.java file** to add an if statement case to check if the file being exported is missing a package version. We can use a Boolean value in case there are no missing package versions. 
5. Editing code in the **Organize Manifest.java** file to write a function that adds the missing package version for exports based on whether the user checked that they wanted a fixed version or a current bundle version. In the code we can reference from **ExportPackageObject.java** and use the versionAttribute or a custom object to assign the packages with the correct versions. Otherwise we assign a default version. 


[...]

PR #1961, opened 2025-09-07
Issue #1815 -  Added options that add missing versions to export packages

Vonna9:
The following pull request draft addresses [issue #1815](https://github.com/eclipse-pde/eclipse.pde/issues/1815) to add a fixed or bundled version to missing export packages. The following code changes reflect the solution my teammate @hilla-ryy and I have worked on so far. 

Changes:
- We have added buttons and text to allow the feature to be displayed in the Organize Manifests Wizard. 
- [Work in Progress] We have added a function that adds missing versions to export packages in the Organize Manifests.java file to sort through each export package, detect if the version number is missing and add the correct versions bundled or fixed to the export package.

 Organize Manifests Wizard Preview

Obstacles: One of the obstacles my teammate @hilla-ryy and I have come across was adding the fixed and bundled versions to each package. After reviewing the codebase, we found within the Export Package Header.java file there is not a specific method for getting either a bundled or fixed version to add to the export package other than "getVersionAttribute". And in the Export Package Object.java file it looks like a bundled or fixed version is added when the export package object is created which means a new version can only be added by recreating the object. Is there a way we can add the correct version without recreating the export package object? For instance, we are working on adding the correct version directly into the manifest file of the export package.


Vonna9:
@HannesWell Good afternoon, I am following up to check the status of the pull request. My team and I have finished our solution. Is there anything that should be added to the following code? Thank you. Have a wonderful day.

PR #1962, opened 2025-09-09
Quick fix to add available matching version for Imported packages

### LINKED ISSUE #1623: Create a quickfix to aadd missing version information
If one imports a bundle without version PDE shows a warning in the manifest:

Image

but the only Quickfix offered is to configure the problem severity. Instead it should offer to add a version (range) to the import.

nburnwal09:
This is to provide a quick fix for adding the available matching version for Imported packages in MANIFEST.MF warning.
Similar to the fix: #1623 and as suggested in [here](https://github.com/eclipse-pde/eclipse.pde/pull/1791#issuecomment-3181261395)

Attaching screenshot for the quick fix:

image

After applying:

image

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/VersionUtil.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/VersionUtil.java
@@ -15,6 +15,7 @@
 package org.eclipse.pde.internal.core.util;
 
 import java.util.Comparator;
+import java.util.Optional;
 
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
@@ -118,6 +119,16 @@ public class VersionUtil {
 			return pvi.getMajor() + "." + pvi.getMinor() + "." + pvi.getMicro(); //$NON-NLS-1$//$NON-NLS-2$
 		}
 		return version;
+	}
+
+	public static Optional<VersionRange> createConsumerRequirementRange(Version version) {
+		if (version != null && !Version.emptyVersion.equals(version)) {
+			return Optional.ofNullable(new VersionRange(VersionRange.LEFT_CLOSED, //
+					new Version(version.getMajor(), version.getMinor(), 0), //
+					new Version(version.getMajor() + 1, 0, 0), //
+					VersionRange.RIGHT_OPEN));
+		}
+		return Optional.empty();
 	}
 
 }
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
@@ -1,0 +1,105 @@
+/*******************************************************************************
+ *  Copyright (c) 2025, 2025 IBM Corporation and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.pde.internal.ui.correction;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import org.eclipse.core.resources.IMarker;
+import org.eclipse.osgi.service.resolver.BundleDescription;
+import org.eclipse.osgi.service.resolver.ExportPackageDescription;
+import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.core.plugin.PluginRegistry;
+import org.eclipse.pde.internal.core.text.bundle.Bundle;
+import org.eclipse.pde.internal.core.text.bundle.BundleModel;
+import org.eclipse.pde.internal.core.text.bundle.ImportPackageHeader;
+import org.eclipse.pde.internal.core.text.bundle.ImportPackageObject;
+import org.eclipse.pde.internal.core.util.VersionUtil;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+import org.osgi.framework.Constants;
+import org.osgi.framework.Version;
+import org.osgi.framework.VersionRange;
+
+/**
+ * Resolution to add available matching version for Imported package in MANIFEST
+ */
+public class VersionMatchImportPackageResolution extends AbstractManifestMarkerResolution {
+
+	public VersionMatchImportPackageResolution(int type, IMarker marker) {
+		super(type, marker);
+	}
+
+	public Version getVersion(Object inputElement) {
+		IPluginModelBase[] models = PluginRegistry.getActiveModels();
+		Version highest = null;
+		for (IPluginModelBase pluginModel : models) {
+			BundleDescription desc = pluginModel.getBundleDescription();
+
+			String id = desc == null ? null : desc.getSymbolicName();
+			if (id == null) {
+				continue;
+			}
+			ExportPackageDescription[] exported = desc.getExportPackages();
+			for (ExportPackageDescription exportedPackage : exported) {
+				String name = exportedPackage.getName();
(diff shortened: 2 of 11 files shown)


nburnwal09:
@HannesWell 
Can you please review this PR?
Also, to provide the version range for Imported packages and Required bundles, I am working on a separate PR.

gireeshpunathil (approved):

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/VersionUtil.java
```diff
@@ -120,4 +120,14 @@ public static String computeInitialPluginVersion(String version) {
...
 
+	public static String computeInitialRequirementVersionRange(String version) {
+		if (version != null && VersionUtil.validateVersion(version).isOK()) {
+			Version pvi = Version.parseVersion(version);
+			return new VersionRange(
+					"[" + pvi.getMajor() + "." + pvi.getMinor() + "," + (pvi.getMajor() + 1) + ")") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$//$NON-NLS-4$
+							.toString();
+		}
+		return ""; //$NON-NLS-1$
+	}
```
HannesWell: Recently I added a method that does this to `ManifestUtils`:
https://github.com/eclipse-pde/eclipse.pde/blob/6a28c28a41ebf2634f1ef960283cfb0014491d76/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/util/ManifestUtils.java#L402-L410

Admittedly this method would probably better be moved to this class and probably be used instead of the existing method. That would hopefully add proper version ranges in more places.
I can try to have a look at unifying these as soon as possible. Or if you want to have a look at it, it would be more than welcome.

HannesWell (commented):
Just one point from me at the moment, I'll try to do a full review this evening.

nburnwal09:
@HannesWell 
Could you please check this whenever you get sometime?

gireeshpunathil:
this is ready to land, but would like another quick look from you, @HannesWell !

HannesWell:
> this is ready to land, but would like another quick look from you, @HannesWell !

Yes I will do it soon.
But please note that we are in RC1 week now and in the freeze period. So even when this is ready and reviewed we probably have to wait until the next dev-cycle opens in a bit more than one and a half week.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
```diff
@@ -0,0 +1,98 @@
...
+				continue;
+			}
+			ExportPackageDescription[] exported = desc.getExportPackages();
+			for (ExportPackageDescription exportedPackage : exported) {
+				String name = exportedPackage.getName();
+				if (("java".equals(name) || name.startsWith("java."))) { //$NON-NLS-1$ //$NON-NLS-2$
+					// $NON-NLS-2$
+					continue;
+				}
+				if (name.equalsIgnoreCase(inputElement.toString())) {
```
noopur2507: Package names in OSGi are case-sensitive, so this can lead to incorrect matches.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
```diff
@@ -0,0 +1,98 @@
...
+					return exportedPackage.getVersion();
```
noopur2507: This returns the first matching exported package. This may produce incorrect results when multiple bundles export the same package. IMO, the fix should select the highest available version instead.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Resolution to add available matching version for Imported package in MANIFEST
+ */
+public class VersionMatchImportPackageResolution extends AbstractManifestMarkerResolution {
+
+	public VersionMatchImportPackageResolution(int type, IMarker marker) {
+		super(type, marker);
+	}
+
+	public Version getVersion(Object inputElement) {
```
noopur2507, line 44: Can this parameter be of type String instead of Object?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	@Override
+	protected void createChange(BundleModel model) {
+		String bundleId = Objects.requireNonNull(marker.getAttribute("bundleId", (String) null)); //$NON-NLS-1$
+		Bundle bundle = (Bundle) model.getBundle();
+		ImportPackageHeader header = (ImportPackageHeader) bundle.getManifestHeader(Constants.IMPORT_PACKAGE);
+		if (header != null) {
+			for (ImportPackageObject importPackage : header.getPackages()) {
+				if (bundleId.equals(importPackage.getName())) {
```
noopur2507: "bundleId" is misleading here as the value represents a package name. Please update it across all usages of "bundleId" in the PR.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
```diff
@@ -0,0 +1,98 @@
...
+					Version version = getVersion(bundleId);
+					if (version != null) {
```
noopur2507: Currently, if no matching exported package is found i.e. if version is null, the quick fix will silently fail. So, the quick fix will appear but will have no effect. Better to avoid offering it if no version is available.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -1303,7 +1303,8 @@ UpdateClasspathResolution_label=Update the classpath and compliance settings
 UpdateExecutionEnvironment_label=Update the execution environment based on JRE on the classpath
 UpdateClasspathJob_task = Update classpaths...
 UpdateClasspathJob_title = Updating Plug-in Classpaths
-AddMatchingVersion_RequireBundle = Require latest available version
+AddMatchingVersion = Require latest available version
+AddMatchingVersionDescription = Add a version range aligned with the latest compatible available version
```
noopur2507: Please review both strings to reflect the intended change for the latest version / version range as these are now used by both Require-Bundle and Import-Package quick fixes.

noopur2507 (changes requested):
Thanks, Neha. The overall approach looks good and aligns well with the existing Require-Bundle quick fix. I’ve added a few comments to improve the correctness.

Also, please add tests if applicable for PDE quick fixes.

@HannesWell / @gireeshpunathil – please take a look and suggest if anything else should be adjusted.

Once the above comments are addressed, this should be in good shape for merging.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/VersionMatchImportPackageResolution.java
nburnwal09: versionRange is calculated based on version object(in line 79). If we change the type to string, we will have to change the type again there.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
nburnwal09: @noopur2507 
I have updated the msg. Attached the screenshot in the PR description

noopur2507:
@nburnwal09, I still see old changes in the PR. Can you please update it?

noopur2507 (approved):
Thanks, Neha.

nburnwal09:
> @nburnwal09, I still see old changes in the PR. Can you please update it?

Yeah sorry about that. That pushed got rejected in my local and I didn't notice. Have updated just now

PR #1969, opened 2025-09-17
Fix/improve details of org.eclipse.pde.osgitest.dependencies.feature

merks:


laeubi (approved):
Thanks for fixing, build  failure seems unrelated.

PR #1972, opened 2025-09-29
Don't pass null values to Signature.createArraySignature(String, int)

iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4448


laeubi (approved):
In general good, one might still ask what happens when null is returned but an array type is encountered.

laeubi:
> As said on the ticket, hard to understand how the problem manifests in higher levels if there is no reproducer.

Yes that's why I wonder if it is actually a reason or just a symptom of something else being wrong, but as you said it should not harm.

PR #1974, opened 2025-09-29
Code Cleanup o.e.e4.tools, o.e.e4.tools.emf.editor3x and o.e.e4.tools packages

nburnwal09:
This PR adds code cleanup in org.eclipse.e4.tools,
org.eclipse.e4.tools.emf.editor3x and org.eclipse.e4.tools packages

### DIFF
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/UndoAction.java
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/UndoAction.java
@@ -75,19 +75,22 @@ public class UndoAction extends Action {
 	 */
 	public static String getCommandLabel(Command cmd) {
 
-		if (cmd instanceof SetCommand) {
-			SetCommand sc = (SetCommand) cmd;
+		switch (cmd) {
+		case SetCommand sc -> {
 			return sc.getLabel() + " " + sc.getFeature().getName() + " on " + sc.getOwner().eClass().getName(); //$NON-NLS-1$ //$NON-NLS-2$
-		} else if (cmd instanceof AddCommand) {
-			AddCommand ac = (AddCommand) cmd;
+		}
+		case AddCommand ac -> {
 			return ac.getLabel() + " " + getFirstClassName(ac.getCollection()); //$NON-NLS-1$
-		} else if (cmd instanceof DeleteCommand) {
-			DeleteCommand dc = (DeleteCommand) cmd;
+		}
+		case DeleteCommand dc -> {
 			Collection<?> deleted = dc.getCollection();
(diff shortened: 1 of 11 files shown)


e4tools/bundles/org.eclipse.e4.tools.persistence/src/org/eclipse/e4/tools/internal/persistence/impl/PersistencePackage.java
```diff
@@ -92,8 +92,8 @@ public static IPersistencePackage init() {
 
 		// Obtain or create and register package
 		Object registeredPersistencePackage = EPackage.Registry.INSTANCE.get(eNS_URI);
-		PersistencePackage thePersistencePackage = registeredPersistencePackage instanceof PersistencePackage
-				? (PersistencePackage) registeredPersistencePackage
+		PersistencePackage thePersistencePackage = registeredPersistencePackage instanceof PersistencePackage p
+				? p
```
merks, line 96: I don’t think it’s a good idea to modify generated code without verifying that regenerating will not undo the changes.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/UndoAction.java
```diff
@@ -75,20 +75,23 @@ private void update() {
...
-			DeleteCommand dc = (DeleteCommand) cmd;
+		}
+		case DeleteCommand dc -> {
 			Collection<?> deleted = dc.getCollection();
 			if (deleted.size() == 1) {
 				return dc.getLabel() + " " + getFirstClassName(deleted); //$NON-NLS-1$
 			}
 			return dc.getLabel() + " " + dc.getCollection().size() + " Objects"; //$NON-NLS-1$ //$NON-NLS-2$
 		}
+		case null, default -> {
```
merks, line 92: I'm not sure I understand why null and default are specified, Isn't default sufficient.  Is something generating these changes?

e4tools/bundles/org.eclipse.e4.tools.persistence/src/org/eclipse/e4/tools/internal/persistence/impl/PersistencePackage.java
nburnwal09: @merks 
I was not aware of the @generated codebase. But I have reverted this and taken care that no code clean-up changes exist for generated codebase.

merks:
Actually, I think those other models were already 'cleaned up' and in fact regenerating doesn't work until I change compliance to > 16 and then lots of files are different.

image

We are busy creating work and problems for ourselves (with little substantial benefit).  😢 

I expect better for this case is to change the template in EMF to use the pattern instanceof depending on the genmodel compliance setting...

PR #1977, opened 2025-10-02
Add implicit requirements in TestSuiteHelper.addAllRequired

merks:
- The algorithm only considers bundle requirements so the new Require-Cpability in org.eclipse.core.contenttype misses dependencies causing resolution errors that fail some tests.

https://github.com/eclipse-platform/eclipse.platform/pull/2162


merks:
I see only the one test failure now so this PR does fix 3 of the test failures:

image

PR #2026, opened 2025-10-17
Support SVG icons in extension point editor

Dani-Hub:
Add "svg" to list of supported file extensions for images


Dani-Hub:
I'm not so familar with squashing, therefore I decided to close this puull request in favour for a new one (https://github.com/eclipse-pde/eclipse.pde/pull/2030)

HannesWell:
> I'm not so familiar with squashing, therefore I decided to close this pull request in favour for a new one (#2030)

Please try to avoid it for future contributions as this causes extra noise and makes it harder to follow discussions if they are distributed over multiple PRs.

You can squash existing PRs with EGit for example in the History view of Eclipse using the context menu. Just select the consecutive commits you want to squash, right-click on them and select via `Modify` -> `Squash`:



Alternatively/additionally you can also _Amend_ an existing commit in EGit's `Git staging` view:


And to force push you can for example select the corresponding check-box in the EGit push dialog  (the details depend a bit on your exact workflow):


On the CLI you can of course push with `--force/-f`.

Don't hesitate to ask if you have more questions :)

Dani-Hub:
Thanks very much about the explanation about squashing! Btw.: I actually did amend all my original commits of my first pull request, I'm surprised that it appeared as individual commits to you. Anyway, the problem is now solved, thanks again for your support!

PR #2029, opened 2025-10-18
Pr 2013

merks:
Fix problems in JUnitLaunchRequirements

- The JUNIT5_VERSIONS VersionRange should be [1, 6) because [1, 5) excludes Junit 5.
- The case for TestKindRegistry.JUNIT5_TEST_KIND_ID should add JUNIT5_JDT_RUNTIME_PLUGIN not JUNIT4_JDT_RUNTIME_PLUGIN.
- The search range `new VersionRange(VersionRange.LEFT_OPEN, version, version, VersionRange.RIGHT_OPEN))` is always empty so should use `newVersionRange(VersionRange.LEFT_CLOSED, version, version, VersionRange.RIGHT_CLOSED))`.


HannesWell:
This is now in-cooperated into https://github.com/eclipse-pde/eclipse.pde/pull/2013.
Thank you for your help!

PR #2035, opened 2025-10-21
Pr 2013

merks:


merks:
The problem is that the method org.eclipse.pde.internal.core.BundleValidationOperation.run(IProgressMonitor) create a new state, and new bundle descriptions are created for that state that reuse the bundleId of the orginal bundle which do in fact clash in general:

With this hack the tests pass:
```
	public void run(IProgressMonitor monitor) throws CoreException {
		if (FACTORY == null) {
			FACTORY = BundleHelper.getPlatformAdmin().getFactory();
		}
		SubMonitor subMonitor = SubMonitor.convert(monitor, fModels.size() + 1);
		fState = FACTORY.createState(true);
		int id = 0;

		for (IPluginModelBase fModel : fModels) {
			BundleDescription bundle = fModel.getBundleDescription();
			if (bundle != null) {
				BundleDescription bundleDescription = FACTORY.createBundleDescription(bundle);
				try {
					Method declaredMethod = bundleDescription.getClass().getDeclaredMethod("setBundleId", long.class); //$NON-NLS-1$
					declaredMethod.setAccessible(true);
					declaredMethod.invoke(bundleDescription, id++);
				} catch (NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException
						| InvocationTargetException e) {
					e.printStackTrace();
				}
				fState.addBundle(bundleDescription);
			}
			subMonitor.split(1);
		}
		fState.setPlatformProperties(fProperties);
		fState.resolve(false);
		subMonitor.split(1);
	}
```

PR #2036, opened 2025-10-22
Add support to search bundles in the host by version range

laeubi:
Currently PDE#findPluginInHost assumes that a bundle is unique by its id what is not true in all cases.

This now do the following:

- index the plugins by id but keep them in a list to support multiple
- if more than one version is found return the highest
- allow to further restrict by version range

This is in preparation of the new JUnit 6 support where the host will contain two bundles with same name but different version, but this can happen today as well e.g. different guava and whatever.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
@@ -15,9 +15,12 @@ package org.eclipse.pde.internal.core;
 
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


HannesWell:
> This is in preparation of the new JUnit 6 support where the host will contain two bundles with same name but different version, but this can happen today as well e.g. different guava and whatever.

I have to admit that I fail to see what the advantage is of effectivly extracting this change from https://github.com/eclipse-pde/eclipse.pde/pull/2013, which is hopefully ready very soon?
Technicaly this is valid, but as this change is also included in https://github.com/eclipse-pde/eclipse.pde/pull/2013, but in a slightly different variant, we should not further complicate that PR and not have this change in a separate PR.

PR #2037, opened 2025-10-22
Only select compatible providers during collection of JUnit requirements

laeubi:
Currently PDE searches JUnit requirements primary by its id and adds them unconditionally. This does not work well when there are multiple options and versions.

This now changes the selection process in the following way:

1) Select all required junit runtimes
2) Compute their dependencies and add them if required 
3) Among all possible options select those engines and launchers that actually match a requirement in this set
4) if none of the matching are already selected, use the highest matching version

FYI @trancexpress

### DIFF
--- a/features/org.eclipse.pde.unittest.junit-feature/feature.xml
+++ b/features/org.eclipse.pde.unittest.junit-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.unittest.junit"
       label="%featureName"
-      version="1.0.1100.qualifier"
+      version="1.0.1200.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/ui/org.eclipse.pde.unittest.junit/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.unittest.junit/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.pde.unittest.junit
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.unittest.junit;singleton:=true
-Bundle-Version: 1.2.100.qualifier
+Bundle-Version: 1.2.200.qualifier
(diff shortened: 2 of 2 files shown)


ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
```diff
@@ -831,6 +837,117 @@ public static Map<IPluginModelBase, AdditionalPluginData> getAdditionalPlugins(I
...
+			// compatible ones...
+			addIfAbsent(id, collectedModels, startLevelMap).or(() -> collectedModels.getOrDefault(id, List.of()).stream().filter(m -> m.getBundleDescription().isResolved()).findFirst()).ifPresent(collected::add);
+		}
+		// Now collect the dependencies of the collected...
+		Set<BundleDescription> closure = DependencyManager.findRequirementsClosure(collected.stream().map(IPluginModelBase::getBundleDescription).toList());
+		for (BundleDescription description : closure) {
+			collected.add(addIfAbsent(description, collectedModels, startLevelMap));
+		}
+		// Now add any indirectly required items
+		if (collectedModels.containsKey("junit-platform-runner")) { //$NON-NLS-1$
```
HannesWell: This entire block is actually not necessary as the dependencies added below are already part of the transitive hull of the o.e.jdt.junitX.runtime bundles. Therefore it's remove din #2013.

HannesWell (commented):
For this I also have to admit that I fail to the exact difference to https://github.com/eclipse-pde/eclipse.pde/pull/2013. At the same time this changes many things differently and will conflict with #2013.
And as the latter is hopfully is ready soon, my suggestion is to not delay https://github.com/eclipse-pde/eclipse.pde/pull/2013 further and keep the attribution with the author of it. If this contains changes that are eventually missing from the other PR, they still can be applied afterwards in this or a new PR.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
laeubi: Is this really always the case? What bundles require this already? Of course if we can omit this case that's even better.

PR #2039, opened 2025-10-22
Make sure new states assign unique ID to bundles

laeubi:
Currently it can happen that a bundle has an ID clash and then it is missing from a resolve operation afterwards.

This now uses the new API to copy a bundledescription with a specific id to circumvent the problem.

Requires the I-Build to compile:

- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.38/81/

FYI @trancexpress 

This should solve the validation dialog issues we have seen.

For selection the proper provider / junit bundles there is this PR (that was closed by the version bump!?!):

- https://github.com/eclipse-pde/eclipse.pde/pull/2037


HannesWell:
I have just pushed my proposal for these small adjustments to speed-up the resolution.
From my side this is ready for submission.

laeubi:
The Github Actions succeed and I would trust them enough to merge but wanted to know what others think?

laeubi:
I restarted the build (again) this morning, the problem is more that Jenkins config was changed to use shallow copy what did not work well and confuses the timestamp plugin.

PR #2053, opened 2025-10-24
Restrict junit-jupiter-api version to [5,6) when adding JUnit 5 bundle requirements

trancexpress:


HannesWell (commented):
Thanks for this enhancement. 

In general this looks good, still I have one comment blow.
Could you also please squash all the refactorings in one commit (and not multiple).

HannesWell:
I had another in-depth look at this and the whole import error resolution facility and instead of adding version ranges to the added JUnit-5 requirements (and implementing this for JUnit-6, instead I wonder if we should just remove the special handling for the JUnit-5 bundles
It's currently implemented in
https://github.com/eclipse-pde/eclipse.pde/blob/98f11cb229a6aa8f7013dfd865c8de5b4f94d548/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/UnresolvedImportFixProcessor.java#L100-L103
and while it works the code seems to have wrong assumption as it checks for bundle names (at least that's what it looks to me).
But the names that are actually checked are names of types that could be used as solution for a compilation error, like:
```
org.junit.Test
org.junit.jupiter.api.Test
```
So checking for `org.junit.jupiter` or `org.junit` should be sufficient.
And looking at which bundles are added at the same time, the checked names are actually not correct:
https://github.com/eclipse-pde/eclipse.pde/blob/98f11cb229a6aa8f7013dfd865c8de5b4f94d548/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java#L238-L240

Why is `org.junit` added, when only the `org.junit.jupiter.api.Test` can be resolved with the current code. And if it would resolve a `org.junit.platform`, requiring these bundles is actually not sufficient.
With the current state of affairs (and required capabilities of `junit-jupiter-api`), I think it would be sufficient to just propose the package import.

The solutions that are currently proposed are:


With all said in my mind proposal is:

1. Drop special JUnit-5 (and a future JUnit-6) proposal
2. Move proposal of package-imports to the top if the exported package has a version that's not empty or 0.0.0
This would advertise the recommendation to use import-package over require bundle
3. Propose multiple versions if more than one is available (maybe with a limit of the two or three latest).
4. And don't propose junit-3 in this case as `Test` is a class, not an annotation in junit3.

This would make the code simpler and would improve the situation in general.
@trancexpress, @merks or @laeubi what do you think?

HannesWell:
> So if there's no objection, I'll create PRs in the next days to implement my suggestions (at least one to three, No 4 might be more complicated).

The PR is now
- https://github.com/eclipse-pde/eclipse.pde/pull/2104

Issue #2045, opened 2025-10-23
Show dialog if JUnit 5 plug-in test launch detects unrestricted junit-jupiter-api in plug-in dependencies


trancexpress:
See: https://github.com/eclipse-pde/eclipse.pde/pull/2013#issuecomment-3422452639

In essence, we have a problem when JUnit 5 and JUnit 6 are in the target platform (which will soon be the case for the Eclipse SDK, see: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2535).

Given a `MANIFEST.MF` like this:

```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: TestPlugin
Bundle-SymbolicName: TestPlugin
Bundle-Version: 1.0.0.qualifier
Require-Bundle: org.junit,
 junit-jupiter-api
Automatic-Module-Name: TestPlugin
Bundle-RequiredExecutionEnvironment: JavaSE-21
```

Running a JUnit 5 plug-in test launch configuration will result in an odd error in the launched JVM:

```
Caused by: java.util.ServiceConfigurationError: org.junit.platform.engine.TestEngine: org.junit.jupiter.engine.JupiterTestEngine not a subtype
```

This is due to adding JUnit 5 to the launch (we fix this in https://github.com/eclipse-pde/eclipse.pde/pull/2013), but the launch already having JUnit 6 due to the unrestricted `junit-jupiter-api` in the `MANIFEST.MF`.

We opted for aborting the launch with a specific error message, which indicates what the problem is and how to fix it.

See also the long discussion here for more context: https://github.com/eclipse-pde/eclipse.pde/pull/2013#discussion_r2442422171

The problem is avoided by restricting the `junit-jupiter-api` version to `[5,6)` (thanks to https://github.com/eclipse-pde/eclipse.pde/pull/2013), which the new error should indicate.

laeubi:
> Furthermore we should error if there are two (different major) versions of junit-jupiter bundles present in the runtime (maybe checking for the jupiter-engine is sufficient). That will probably always cause problems at runtime.

Basically having Junit 5+6 in the launch config and even in the tests should not be a problem, I thought that was the goal of the recent changes in PDE? If not wee need to further adapt this! The problem is often more to have a mixture of JUnit 5 versions because then you get resolution errors but thats the case for other libraries as well and we can't fix the world.

[...]

PR #2055, opened 2025-10-24
Check for mixed JUnit 5 and JUnit 6 launch type and required bundles

trancexpress:
This changes ensures an error dialog is shown,
if a user runs a JUnit 5 launch with required JUnit 6 bundles.

Fixes: #2045

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.launching;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.jdt.internal.junit.launcher.TestKindRegistry;
+import org.eclipse.osgi.service.resolver.BundleDescription;
+import org.eclipse.osgi.util.NLS;
+import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.internal.launching.launcher.LaunchValidationOperation;
+import org.osgi.framework.Version;
+
+public class JUnitLaunchValidationOperation extends LaunchValidationOperation {
+
+	private static final Set<String> JUNIT_PLATFORM_ENGINE_BUNLDES = Set.of(new String[] { //
+			"junit-platform-engine", //$NON-NLS-1$
+			"org.junit.platform.engine", //$NON-NLS-1$
+	});
+
+	private final Map<Object, Object[]> fErrors = new HashMap<>(2);
+
+	public JUnitLaunchValidationOperation(ILaunchConfiguration configuration, Set<IPluginModelBase> models) {
+		super(configuration, models, null);
+	}
+
+	@Override
+	public void run(IProgressMonitor monitor) throws CoreException {
+		try {
+			checkJunitVersion(fLaunchConfiguration, fModels);
+		} catch (CoreException e) {
+			PDELaunchingPlugin.log(e);
+		}
+	}
+
+	@SuppressWarnings("restriction")
+	private void checkJunitVersion(ILaunchConfiguration configuration, Set<IPluginModelBase> models) throws CoreException {
+		org.eclipse.jdt.internal.junit.launcher.ITestKind testKind = org.eclipse.jdt.internal.junit.launcher.JUnitLaunchConfigurationConstants.getTestRunnerKind(configuration);
+		if (testKind.isNull()) {
+			return;
+		}
+		Set<Version> junitPlatformBundlesVersions = junitPlatformBundleVersions(models);
+		String testKindId = testKind.getId();
+		switch (testKindId) {
+			case TestKindRegistry.JUNIT3_TEST_KIND_ID, TestKindRegistry.JUNIT4_TEST_KIND_ID -> {
+			} // nothing to check
+			case TestKindRegistry.JUNIT5_TEST_KIND_ID -> {
+				// JUnit 5 platform bundles have version range [1.0,2.0)
+				junitPlatformBundlesVersions.stream().map(Version::getMajor).filter(i -> i.intValue() != 1).findFirst().ifPresent(otherVersion -> {
+					String message = NLS.bind(PDEMessages.JUnitLaunchConfiguration_error_JUnitLaunchAndRuntimeMissmatch, 5, otherVersion);
+					addError(message);
+				});
+			}
+			default -> throw new CoreException(Status.error("Unsupported test kind: " + testKindId)); //$NON-NLS-1$
+		}
+	}
+
+	private void addError(String message) {
+		fErrors.put(message.replaceAll("\\R", " "), null); //$NON-NLS-1$//$NON-NLS-2$
+	}
+
+	@Override
+	public boolean hasErrors() {
+		return !fErrors.isEmpty();
+	}
+
+	@Override
+	public Map<Object, Object[]> getInput() {
+		Map<Object, Object[]> map = new LinkedHashMap<>();
+		map.putAll(fErrors);
+		return map;
+	}
+
+	private static Set<Version> junitPlatformBundleVersions(Set<IPluginModelBase> models) {
+		return models.stream().map(IPluginModelBase::getBundleDescription) //
+				.filter(d -> JUNIT_PLATFORM_ENGINE_BUNLDES.contains(d.getSymbolicName())) //
+				.map(BundleDescription::getVersion).collect(Collectors.toSet());
+	}
+}
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java
@@ -71,6 +71,7 @@ import org.eclipse.pde.internal.core.util.CoreUtility;
 import org.eclipse.pde.internal.core.util.VersionUtil;
 import org.eclipse.pde.internal.launching.IPDEConstants;
 import org.eclipse.pde.internal.launching.JUnitLaunchRequirements;
+import org.eclipse.pde.internal.launching.JUnitLaunchValidationOperation;
 import org.eclipse.pde.internal.launching.PDELaunchingPlugin;
 import org.eclipse.pde.internal.launching.PDEMessages;
 import org.eclipse.pde.internal.launching.launcher.BundleLauncherHelper;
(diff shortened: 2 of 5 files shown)


laeubi (commented):
This does not feels like the right place and produces and ugly exception in the log.

We should check this maybe in the launch delegate or in the Run Config dialog and show a proper Messagedialog.... 

Also why we need replicate the bundle names here and can't reuse them from JDT?

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
```diff
@@ -0,0 +1,112 @@
...
+import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.internal.launching.launcher.LaunchValidationOperation;
+
+public class JUnitLaunchValidationOperation extends LaunchValidationOperation {
+
+	private static final Set<String> JUNIT_JUPITER_BUNLDES = new HashSet<>(Arrays.asList(new String[] {"junit-jupiter-api", //$NON-NLS-1$
+			"org.junit.jupiter.api", //$NON-NLS-1$
+			"junit-jupiter-engine", //$NON-NLS-1$
+			"org.junit.jupiter.engine", //$NON-NLS-1$
+	}));
```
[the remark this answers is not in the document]
HannesWell: Furthermore I think it's sufficient to just check for the `junit-platform-engine` bundle (respectivly it's old Orbit name), as two version of that are seem to be the root problem. As desribed in https://github.com/eclipse-pde/eclipse.pde/pull/2013#discussion_r2442793735 (and other places), two versions of `org.junit.platform.engine.TestEngine` will make the discovery of test-engine implementations at runtime fail.
Of course different major versions of `junit-jupiter-engine` will imply different versions of the former bundles due to their version ranges. But also in general other implementations (I'm not aware of them) should suffer the same problem, from my understanding of it.
However haing to check less names also simplifies this a bit.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	private void checkJunit5Version(Set<IPluginModelBase> models) throws CoreException {
+		List<String> junit6Bundles = junitBundles(models, 6);
+		if (!junit6Bundles.isEmpty()) {
+			String bundleIds = String.join(System.lineSeparator(), junit6Bundles);
+			String message = NLS.bind(PDEMessages.JUnitLaunchConfiguration_error_JUnit5LaunchJUnit6Runtime, bundleIds);
+			IStatus status = Status.error(message);
+			IStatusHandler statusHandler = DebugPlugin.getDefault().getStatusHandler(status);
```
HannesWell: Can you explain to me that the `statusHandler` is doing exactly when present?

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
```diff
@@ -0,0 +1,112 @@
...
+	public boolean hasErrors() {
+		return super.hasErrors() || fErrors.size() >= 1;
+	}
+
+	@Override
+	public Map<Object, Object[]> getInput() {
+		Map<Object, Object[]> map = super.getInput();
+		map.putAll(fErrors);
+		return map;
+	}
```
HannesWell: I'd move such unintersting management methods to the bottom of this class.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
```diff
@@ -0,0 +1,112 @@
...
+		validateJunitBundles();
+	}
+
+	private void validateJunitBundles() {
+		try {
+			checkJunitVersion(fLaunchConfiguration, fModels);
+		} catch (CoreException e) {
+			PDELaunchingPlugin.log(e);
+		}
+	}
```
HannesWell: For me, this separate method isn't necessary and I would have inlined it.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
```diff
@@ -0,0 +1,112 @@
...
+	private static final Object[] EMPTY = new Object[0];
+	private final Map<Object, Object[]> fErrors = new HashMap<>(2);
+
+	public JUnitLaunchValidationOperation(ILaunchConfiguration configuration, Set<IPluginModelBase> models) {
+		super(configuration, models, null);
+	}
+
+	@Override
+	public void run(IProgressMonitor monitor) throws CoreException {
+		super.run(monitor);
```
HannesWell: Is the validation performed in the super-class actually necessary for this?
It's at least called later in the launch procress again and we should avoid running the same validation twice for performance reasons.

HannesWell (changes requested):
Thanks for this one.
I think this goes in the right direction, still I have some comments below (of which some are just cosmetic).

Depending on the kind of validation error, i.e. if it's just a simple missmatch between junit bundles in the runtime and the test-kind, it would be very convenient for users to have a third button to immediatly correct the launch configuration to use the expected JUnit kind.
But this might not trivial with the current code structure and I see in https://github.com/eclipse-pde/eclipse.pde/pull/2055#issuecomment-3444618559 that there is at least already a link to edit the configuration.

laeubi:
> Any specific place you have in mind? org.eclipse.jdt.internal.junit.buildpath.BuildPathSupport.JUnitPluginDescription doesn't seem to expose bundleId. It could though, that could be useful. Should I set a PR for JDT UI with a getter for the field?

I would probably just extract the IDs itself into a String constant, but having a getter seems aslo suitable. For JUnit 6 we do not need to care about old orbit bundle names

>

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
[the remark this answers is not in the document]
trancexpress: I prefer defining arrays, since then adding/removing lines at the end doesn't result in a diff in two lines (due to the comma on the previous line).

For `junit-platform-engine` and `junit-jupiter-api`, does it hurt also checking for `junit-jupiter-api`? `junit-jupiter-api` is highly likely required in `MANIFEST.MF`, as it defines the `@Test` annotation.

What is made more simple by checking for fewer bundles?

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
trancexpress: AFAIK its responsible for showing a dialog on errors. E.g. tests can define their own status handler, that logs instead of showing a dialog (to avoid e.g. tests waiting forever on dialogs, or having to write test code that closes dialogs).

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java
```diff
@@ -518,6 +519,9 @@ protected void preLaunchCheck(ILaunchConfiguration configuration, ILaunch launch
 		fAllBundles = fModels.keySet().stream().collect(Collectors.groupingBy(m -> m.getPluginBase().getId(), LinkedHashMap::new, Collectors.toCollection(ArrayList::new)));
 		launchMode = launch.getLaunchMode();
 
+		JUnitLaunchValidationOperation junitValidation = new JUnitLaunchValidationOperation(configuration, fModels.keySet());
+		LaunchPluginValidator.runValidationOperation(junitValidation, monitor);
```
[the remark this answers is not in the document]
trancexpress: If we check after `addRequiredRuntimePlugins()`, how will we know if the user has mismatched JUnit Jupiter requirements in the `MANIFEST.MF` requirements?

We wouldn't know if we added bundles due to the launch type, or if the user actually has weird requirements (e.g. `junit-platform-engine;bundle-version="[5,6)"` *and* `junit-platform-engine;bundle-version="[6,7)"`, for whatever reason).

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
HannesWell: When I tried this in a secondoary Eclipse, but it always returned null for me.
Can you make it return something else? If the handler is always null, I suggest to just remove this code-path.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/JUnitLaunchValidationOperation.java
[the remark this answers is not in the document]
HannesWell: > `junit-jupiter-engine` is pulled by `junit-jupiter-api` apparently.

Yes. But as far as I understood the metadata, the api's requirement should also be satisfied by any engine bundle of the same major version.
So for example junit-api 5.10 plus 5.12 should both be satisfied by a jupiter-engine 5.12.
But that maybe also just a theoretical problem.

With the changes in the logic made so far, I still think we can and should reduce the IDs in the set to the proposed one, i.e. `junit-platform-engine` plus `org.junit.platform.engine` .

HannesWell (approved):
Thanks for the update @trancexpress.
I have already applied the few adjustments I wanted to remark.
First of all I changed the bundle that's checked to check to `junit-platform-engine` as suggested in https://github.com/eclipse-pde/eclipse.pde/pull/2055#discussion_r2471353601 and before as this is the really problematic bundle.

Besides that I made a few style adjustments and collected all engine bundle versions in a set so that we can safely reuse it, if we extend the check in the future (currently it's not necessary but future changed might forget to change this if necessary).

Last but not least I tried to further fine tune the error message. I agree with you that the sentences shouldn't be too long so I made the first sentence short and tried to make it on point. At the moment it doesn't seem to be ideal IMHO, that the last sentence is broken that often.
Maybe @merks do you want to check the new message as a native speaker?

But the bigger problem is that, on Windows the message is still cut-off after the first line break. I did some debugging and also tried to replace `\n` with `\r\n`, but nothing helped and I got the impression that this is an SWT issue...
Maybe we should change the facility to use something more simple, e.g. based on an ordinary error-dialog (with the link to edit the launch config)?

However with the adjusted first link I think it would be ok do this in a follow-up to get this off the table.

trancexpress:
Looks like we have a hang (maybe the same as before):

```
 [INFO] --- tycho-surefire:5.0.1-SNAPSHOT:test (default-test) @ org.eclipse.pde.junit.runtime.tests ---
[INFO] Selected test framework junit (5.0.0) with provider junit5vintage [5.12.0,6.0.0)
[INFO] Executing test runtime with timeout (seconds): 0, logs, if any, will be placed at: /home/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.junit.runtime.tests/target/work/data/.metadata/.log
[INFO] Command line: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.9-10/x64/bin/java -Dosgi.noShutdown=false -Dosgi.os=linux -Dosgi.ws=gtk -Dosgi.arch=x86_64 --add-modules=ALL-SYSTEM -Dosgi.clean=true -ea -jar /home/runner/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.7.100.v20251014-1222/org.eclipse.equinox.launcher-1.7.100.v20251014-1222.jar -data /home/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.junit.runtime.tests/target/work/data -install /home/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.junit.runtime.tests/target/work -configuration /home/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.junit.runtime.tests/target/work/configuration -application org.eclipse.tycho.surefire.osgibooter.uitest -testproperties /home/runner/work/eclipse.pde/eclipse.pde/ui/org.eclipse.pde.junit.runtime.tests/target/surefire.properties
WARNING: Using incubator modules: jdk.incubator.vector
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.

(java:5238): dbind-WARNING **: 23:34:33.036: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files
WARNING: Annotation classes from the 'javax.inject' or 'javax.annotation' package found.
It is recommended to migrate to the corresponding replacements in the jakarta namespace.
The Eclipse E4 Platform will remove support for those javax-annotations in a future release.
To suppress this warning, set the VM property: -Declipse.e4.inject.javax.warning=false
To disable processing of 'javax' annotations entirely, set the VM property: -Declipse.e4.inject.javax.disabled=true

Running org.eclipse.pde.junit.runtime.tests.JUnitRuntimeTests
Running JUnitRuntimeTests org.eclipse.pde.junit.runtime.tests.JUnitExecutionTest
Error: The operation was canceled.
```

trancexpress:
> But the bigger problem is that, on Windows the message is still cut-off after the first line break. I did some debugging and also tried to replace `\n` with `\r\n`, but nothing helped and I got the impression that this is an SWT issue... Maybe we should change the facility to use something more simple, e.g. based on an ordinary error-dialog (with the link to edit the launch config)?
> 
> However with the adjusted first link I think it would be ok do this in a follow-up to get this off the table.

How does the dialog look like on Windows?

Here is how it looks like on Linux:

Screenshot_20251031_103717

HannesWell:
Test failures are unrelated. Submitting.

Thanks @trancexpress  for this enhancement!

PR #2059, opened 2025-10-25
AbstractBundleContainer.java - Caching & Performance

vogella:
Optimizations:
  - Line 72: Renamed poorly-named static field hash to fVMArgsCache with proper documentation - much clearer intent
  - Line 229-233: Replaced inefficient loop through hash entries (for (Entry<...> entry : hash.entrySet())) with direct HashMap get() - O(n) → O(1) lookup
  - Line 268, 271: Updated all references to use the renamed fVMArgsCache

  Impact:
  - Major performance improvement: Changed from O(n) iteration through all cache entries to O(1) HashMap lookup
  - Better code clarity with descriptive naming
  - No API changes - purely internal optimization

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/AbstractBundleContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/target/AbstractBundleContainer.java
@@ -17,7 +17,6 @@ package org.eclipse.pde.internal.core.target;
 import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
-import java.util.Map.Entry;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -66,7 +65,7 @@ public abstract class AbstractBundleContainer extends PlatformObject implements 
 	 */
 	private String[] fVMArgs;
 
-	static private HashMap<AbstractBundleContainer, String[]> hash = new HashMap<>();
+	private static final HashMap<AbstractBundleContainer, String[]> fVMArgsCache = new HashMap<>();
 
 	/**
 	 * Resolves any string substitution variables in the given text returning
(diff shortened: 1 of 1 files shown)


laeubi (dismissed):
@vogella I don't think that such "modernization" currently bring us forward, if you absolute want this can you place make sure that:

1. We can not cover it by cleanup action, if we can, enable them and let it happen automatically
2. Batch other changes in one PR after cleanups where run.

vogella:
> @vogella I don't think that such "modernization" currently bring us forward, if you absolute want this can you place make sure that:
> 
> 1. We can not cover it by cleanup action, if we can, enable them and let it happen automatically
> 2. Batch other changes in one PR after cleanups where run.

I'm currently not looking into clean-up actions and have no plans to do so. Thanks for the suggestions but AFAIK according to the Eclipse policy you are not entitled to tell other committers how to work.

laeubi:
> I'm currently not looking into clean-up actions and have no plans to do so. Thanks for the suggestions but AFAIK according to the Eclipse policy you are not entitled to tell other committers how to work.

You can work as you like, but as soon as you open a PR you need to accept to get it reviewed and probably there are some discussion, suggestions and best practices.

For a while we have enabled cleanups to get a consistent codebase and to improve JDT leanups and you even agreed and asked it to rool out to platform, so it feel a bit odd now to complain.

So if cleaning up the codebase is important to you I can only ask to follow what is established as it works on a larger scale and we are already familiar with that and it makes sure the same style is applied also in the future.

HannesWell (commented):
This performance optimazation seems useful.

But for my taste this adds too many comments and I also think the commit message isn't suitable.
Instead of listing the changes line-by-line it should give an high level overview over the change and/or explain the rational behind it.
The headline could also be more expressive, like
`Speed-up vm-arguments lookup in AbstractBundleContainer` or similar.

Due to this style I assume this change is AI generated, isn't it?
I know they are good at generating text, but I think we should still only add what's necessary and avoid what's not. Even comments have to be maintained.

PR #2070, opened 2025-10-28
Limit Test Engines by using only compatible ones

laeubi:
Currently the RemotePluginTestRunner uses Class.forName to check for compatible engines but with JUnit 5/6 this now depends on the order of the (flat) classpath.

This now changes the code to use a predicate and handle the case better by loading the class trough the provider or limiting the version when we know to run on JUnit 5.

For some reason I was not yet able to really test this change in a launched-from-eclipse because it seem to not hit the breakpoint, but this is similar to fixes on other places now.

FYI @iloveeclipse 

See https://github.com/eclipse-platform/eclipse.platform/issues/2213#issuecomment-3454413667

### DIFF
--- a/ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.junit.runtime; singleton:=true
-Bundle-Version: 3.8.200.qualifier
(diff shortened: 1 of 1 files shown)


trancexpress:
Will we need a follow-up for JUnit 6 support? Seeing the hard-coded `< 6`.

iloveeclipse:
There is new compiler warning shown: 
https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-2070/6/eclipse/new/source.d87280e1-2b5c-4441-9d83-334d1ac9973e/#189
will be fixed via https://github.com/eclipse-pde/eclipse.pde/pull/2072

laeubi:
This is more generic:
- https://github.com/eclipse-pde/eclipse.pde/pull/2071

PR #2089, opened 2025-11-02
Improve display of multi-line single String launch validation errors

HannesWell:
mainly on Windows.

Follow-up on
- https://github.com/eclipse-pde/eclipse.pde/pull/2055

On Windows 11 this now looks like

Before it was

Alternatively we could set the Text field enabled=false instead of editable=false/READ_ONLY:

or could just use a plain label (`Label message = new Label(container, SWT.BORDER | SWT.WRAP);`, with modified text to be longer):

@trancexpress can you compare it for Linux and tell what works best?


HannesWell:
Just pushed another refinement. On Win11 it now looks like:


Except for the cursor in the error message, this looks good to me. at least for my case.

HannesWell:
> There is also a horizontal scrollbar when the message is just one long line, so IMO it looks fine.

Thanks for checking.
Lets give others a day to comment and if there are no objections I'll submit it.

Issue #1881, opened 2025-07-19
[manifest-editor] Add a button to remove unused dependencies to the toolbar


laeubi:
Currently removing unused dependencies is quite hidden in the "Dependencies" Tab under "Dependency Analysis" among other things.

I think it deserves to be placed in the toolbar so it is easier to find and use.

saumyapandey1998:
Hi @laeubi ! I would love to work on this issue 
Can you please assign it to me?

laeubi:
@saumyapandey1998 please make sure you have a working IDE setup see https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment and make you familiar with the codebase, then just create a PR with some initial proposal for this.

HannesWell:
> Currently removing unused dependencies is quite hidden in the "Dependencies" Tab under "Dependency Analysis" among other things.

Multiple of the things in the `Dependency Analysis` section are useful but adding all of them to the toolbar could make it very croweded.
So maybe we should just expand the section by default to make the tools more visible (and seeing them in the Dependencies section makes IMO sense) or we could also add them at the overview page?

sderazo:
I am from CodeDay and I am working on this.

sderazo:
> At least opening the dependency hierarchy is often useful:

I found the correct location to place the `Remove Unused Dependencies` button ( in progress PR: #1975 ) and I added a button for `Show the plug-in dependency hierarchy` as well. Both currently have sample data and are not functional yet.

<img width="232" height="92" alt="Image" src="https://github.com/user-attachments/assets/a51e5494-cad3-47f8-8333-1d50a811ff57" />

<img width="179" height="92" alt="Image" src="https://github.com/user-attachments/assets/80d8a962-3d77-48fb-bfeb-a5d386a24d1b" />

(text shows when cursor hovers over button)

**Original approach** (PR: #1975 ):
For adding the initial button in the main toolbar, I followed the following [article](https://www.vogella.com/tutorials/EclipseCommands/article.html) from vogella. The change of the button from the main toolbar to the toolbar of the editor (correct location) should be straightforward with a change of the `locationURI`. 

**Challenge I encountered**:
[...]

PR #2091, opened 2025-11-03
Expand Dependency Analysis section of the Manifest Editor by default

HannesWell:
Part of https://github.com/eclipse-pde/eclipse.pde/issues/1881


laeubi:
The Form Framework is really quite wasting space for little benefit. So if we really want to make this more accessible we better move it somewhere else and remove the section all together.

laeubi:
Just mentioning it, you can have text in toolbars as well.

Apart form that, no one has the need to have text / a big expandable section for all the other functions in Eclipse we have at the toolbars all over the place.

PR #2094, opened 2025-11-04
Fix ImageBrowserView combos popup transparency

akurtakov:
On Fedora 43 system SWT.COLOR_LIST_BACKGROUND comes with transparency set and this misrenders combos popup as the color is inherited by children. Not setting any color at all is the simplest and most naturally fitting choice to fix the problem.


akurtakov:
Before:
Screenshot From 2025-11-04 18-50-09

After:
image

akurtakov:
Merge despite Mac tests not starting due to brownout of Mac 13.

Issue #2082, opened 2025-10-31
Host bundles (even org.eclipse.osgi) leak into test launch config


akurtakov:
Steps to reproduce:

1. In a simple plugin project put the following java file
```java
package test;

import static org.junit.jupiter.api.Assertions.*;

class Test {

	@org.junit.jupiter.api.Test
	void test() {
		fail("Not yet implemented");
	}

}
```

2. Set the following as target platform
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="Eclipse" sequenceNumber="528">
<locations>
<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
	<unit id="org.eclipse.platform.feature.group"/>
	<unit id="org.eclipse.jdt.junit5.runtime"/>
	<repository location="https://download.eclipse.org/releases/2025-09/"/>
</location>
</locations>
</target>
```

3. Launch the test via Run As-> JUnit plugin test

4. You will be presented with 

<img width="1506" height="1102" alt="Image" src="https://github.com/user-attachments/assets/b0b279bf-3637-4572-969d-35941d56a9fd" />

5. If you go and validate in the launch configuration there are no problems 

<img width="1540" height="1951" alt="Image" src="https://github.com/user-attachments/assets/14af3ff9-5a79-4834-9faa-6745df65d8d8" /> but launching again gives the problem again

akurtakov:
If one puts `org.eclipse.pde.feature.group` in the target platform the problem is gone.

akurtakov:
So the bug is workarounded  by reverting one condition see https://github.com/akurtakov/eclipse.pde/commit/3cce11d137b793a057224f3572f881404438c755 .

laeubi:
I think this is a recurrence of 

- https://github.com/eclipse-pde/eclipse.pde/pull/1618

(regardless of weather we should include the system bundle at all), I also noticed such problem in another case where multiple singletons where selected.

So it seems the resolve operation performed during launch does not resolve singleton bundles correctly or at laest we need to handle it in some way.

In the meantime I would add a simple check to not expose the equinox system bundle from the running platform, it seems not useful as we can assume that a framework is that basic that it has to be provided by whatever target platform we use to be useful at all.

akurtakov:
With #2083 I end up with 

[...]

PR #2097, opened 2025-11-05
Filter bundles with preference for the target state

laeubi:
Currently it can happen that one bundle (in this particular case org.eclipse.pde.junit.runtime) is not part of the target platform and therefore selected from the running host. This then pull in even more from the host through the DependencyManager collection leading to duplicates all over the place.

As long as target and host are equal enough it does not cause any problem as id+version are the same and so are canceled out but if not this leads to multiple version unnecessary included and this is especially a problem for singletons.

This now:

- collects all items as before
- put everything into a new state and performs a resolve operation
- during the resolve make sure we filter out providers whenever there is an alternative from the target platform.

Contributes to https://github.com/eclipse-pde/eclipse.pde/issues/2082

~There is just one issue I'm currently investigate that the determination of the EE currently is faulty if one creates a new runconfiguration from scratch. If one edits the run config one more time (e.g. open it and then run) it works because then the right EE is set and found.~ Fixed that... the general problem remains and is reported [here](https://github.com/eclipse-pde/eclipse.pde/issues/2098).

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECore.java
@@ -47,7 +47,6 @@ import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.project.IBundleProjectService;
 import org.eclipse.pde.core.target.ITargetDefinition;
 import org.eclipse.pde.core.target.ITargetPlatformService;
-import org.eclipse.pde.internal.build.IPDEBuildConstants;
 import org.eclipse.pde.internal.core.bnd.BndResourceChangeListener;
 import org.eclipse.pde.internal.core.bnd.BndWorkspaceServiceFactory;
 import org.eclipse.pde.internal.core.builders.FeatureRebuilder;
@@ -245,11 +244,6 @@ public class PDECore extends Plugin implements DebugOptionsListener {
 	 *         {@link #VERSION} comparator.
 	 */
 	public Stream<IPluginModelBase> findPluginsInHost(String id) {
-		if (IPDEBuildConstants.BUNDLE_OSGI.equals(id)) {
-			// Do not expose the OSGi framework from the host see
-			// https://github.com/eclipse-pde/eclipse.pde/issues/2082
-			return Stream.empty();
-		}
 		Map<String, List<IPluginModelBase>> hostPlugins = getHostPlugins();
 		if (hostPlugins == null) {
 			return null;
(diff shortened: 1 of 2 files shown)


laeubi:
I don'T see the problems with your provided example so "something" seems to be broken in your config, I would recommend to completely delete the runconfig and create one fresh, make sure to only select the testbundle itself

laeubi:
> The good news is that it let's the test run with a whole of log in the beginning about resolving state.

It might be that just the validation dialog is broken but with the [example here](https://github.com/eclipse-pde/eclipse.pde/issues/2082#issuecomment-3479972239) ([test.me.zip](https://github.com/user-attachments/files/23300971/test.me.zip))  and latest SDK I get no error, so I would need another example to see the problem you reported.

HannesWell:
Since we actually try to reduce and eventually get rid of the ancient `org.eclipse.osgi.service.resolver` API, I wonder if this couldn't be solved simpler. For example by computing the dependency closure immediately only of the junit-requirements (be they from the host or the target) and then by checking if a _compatible_ version (i.e. with the same major segment) is already in the launch or at least in the target (if the bundles are from the host).
This way we should get the same effect but without using a significant part of the `org.eclipse.osgi.service.resolver` API.

Due to the additional `engine` requirement of the `jupiter-api` bundle, which is satisfied by the `jupiter-engine` bundle, at least the junt-bundles should always be in the TP as otherwise they would be missing for compilation.
So the bundles that may be looked up in the host are the jdt/pde.junit.runtime bundles and their non-junit requirements.

In general I would have appreciated, if more time would have been given to other reviewers to bring up for example this suggestion before this is submitted.

laeubi:
> Since we actually try to reduce and eventually get rid of the ancient org.eclipse.osgi.service.resolver API, I wonder if this couldn't be solved simpler. 

I tried to use the new Resolver API but it does not worked out well so I fear even though the "old" is discouraged the new one is not really an alternative as it lacks some features we use in PDE.

> For example by computing the dependency closure immediately only of the junit-requirements (be they from the host or the target) and then by checking if a compatible version (i.e. with the same major segment) is already in the launch or at least in the target (if the bundles are from the host).

This code actually do it, but without assuming what could possibly be "compatible", also we can't check the launch because it does effectively contains *nothing* at that point (except the test bundle itself)

> In general I would have appreciated, if more time would have been given to other reviewers to bring up for example this suggestion before this is submitted.

We can always change it and I have already considers some alternatives, the original change was weeks in the pipeline and seems still flawed as reported by @akurtakov so we need to at least make some progress as we are approaching M3 in a few weeks.

If one would aim for something "simpler" (while I think the problem in general is not simple to solve) we would need to modify the DependencyManager considerably what I like to avoid in this late stage and in the past was a guarantee for a PR to be stalled because of concerns because it is used elsewhere.

So current solution is:

1. fast (we only have a very small subset of bundles that we know should resolve)
2. correct (we use OSGi itself so no need to depend on names of bundles or assume things, e.g. especially Junit has some incompatible things in their change of package imports that are not trivial to get right manually)
3. does not require any assumptions about the nature of dependencies, if you debug it you will see that it will pull in a whole lot of duplicates here (including the OSGi framework itself) so I really don't think we can (or should) fix it manually in a "simple" way.

Issue #2099, opened 2025-11-05
Failing GHA builds due to JavaSE-1.8 BREE


akurtakov:
Bundle org.eclipse.pde.junit.runtime has BREE 1.8 (https://github.com/eclipse-pde/eclipse.pde/blob/668886ef6a47715c37daea7a7b6c1fa48b80fd4e/ui/org.eclipse.pde.junit.runtime/META-INF/MANIFEST.MF#L12 ) but this is no longer resolvable due to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/d619752a04691158da90966cd2b9dc9dc7ded8a8 which is pushed for fixing MacOS 13 runners being removed from GHAs https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3458 .
@HannesWell (as a PL) and whoever cares what should we do  - move bundles with BREE 1.8 to Java 11(or newer) or stop running builds on MacOS ? We can't have both IMO.

laeubi:
> @HannesWell (as a PL) and whoever cares what should we do - move bundles with BREE 1.8 to Java 11(or newer) or stop running builds on MacOS ? We can't have both IMO.

I will fix the build issue, but neither of booth is needed...

PR #2100, opened 2025-11-05
Set tycho.useJDK=SYSTEM globally for verification builds

laeubi:
We already use that for the API tools and as java 1.8 is vanish from github we want to disable it globally, it will still be checked in the ibuilds anyways.

Fix https://github.com/eclipse-pde/eclipse.pde/issues/2099

### DIFF
--- a/ui/org.eclipse.pde.junit.runtime/build.properties
+++ b/ui/org.eclipse.pde.junit.runtime/build.properties
@@ -19,3 +19,4 @@ bin.includes = .,\
                META-INF/
 src.includes = about.html
 output.. = bin/
+pom.model.property.tycho.useJDK=SYSTEM


laeubi:
@HannesWell adjusted to only enable for `org.eclipse.pde.junit.runtime` like we did for API tools already...

laeubi:
> Doesn't this also influence the jre used to excute tests?

Surefire has [its own option](https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/test-mojo.html#useJDK) but I really doubt that this is usable here as almost all test will pull in something that requires Java 21 so effectively it will never be the case to run a test with an "older" jdk anyways apart form some very special cases.

> And even if the release option works well most of the time, there can be differences between the JDK versions especially if the gap is larger.

I use the release option all the time and never seen an issue and it seems quite reliable. I also don't see how different JDK version should behave differently when it comes to compilation apart from that I expects never ones to have fewer bugs. As we are not using the compiler form the JDK anyways I even doubt it makes a practical difference anyways from what JDK we read the signature files.

> If we are absolutely sure that there are zero differences under all circumstances now and in the future, then we should better disabled the feature in the parent pom

What is absolutely sure? The question is more how long we can (or want) easily grab all the older JVMs, this restriction currently makes it quite hard to build the code and then the "solution" (like initially proposed here) is often to bum the java version so I don't really see what ever this option buys us.

> or even in Tycho.

Sure I would happily do this. Tycho should just be "smart" and use whatever JDK is there and best matches (and already do it for the system packages) in the toolchains. But this option is to ENFORCE the policy and since Platform is using it many project assume its "good practice" and also do it even if they are not sure what it does or if they actually need it ;-)

PR #2092, opened 2025-11-03
Remove error dialog for JUnit 5 and 6 conflicts

trancexpress:
Fixes: #2045

PR #2101, opened 2025-11-05
Validate that a JUnit 5 launch has no dependencies to JUnit 6 bundles

trancexpress:
This pull request introduces support for custom OSGi `ResolverHook`s in the PDE launch and validation workflow, specifically enabling exclusion of JUnit 6 bundles when running JUnit 5 tests. The changes add a mechanism to inject a `ResolverHook` into the bundle validation process and implement a hook that filters out JUnit 6 bundles, improving compatibility and reliability of JUnit 5 plug-in test launches.

### DIFF
--- a/features/org.eclipse.pde.unittest.junit-feature/feature.xml
+++ b/features/org.eclipse.pde.unittest.junit-feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.pde.unittest.junit"
       label="%featureName"
-      version="1.0.1200.qualifier"
+      version="1.0.1300.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/ui/org.eclipse.pde.launching/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.launching/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.launching;singleton:=true
-Bundle-Version: 3.13.600.qualifier
+Bundle-Version: 3.13.700.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-Vendor: %provider-name
 Require-Bundle: org.eclipse.jdt.junit.core;bundle-version="[3.6.0,4.0.0)",
--- a/ui/org.eclipse.pde.unittest.junit/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.unittest.junit/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.pde.unittest.junit
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.unittest.junit;singleton:=true
-Bundle-Version: 1.2.200.qualifier
+Bundle-Version: 1.2.300.qualifier
 Bundle-Activator: org.eclipse.pde.unittest.junit.JUnitPluginTestPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName


trancexpress:
@laeubi do you mean something like this? Coming from your comment here: https://github.com/eclipse-pde/eclipse.pde/pull/2092#issuecomment-3489590138

Or we report errors in some way instead of removing the candidate?

We do enter the new code with this in the `MANIFEST.MF`:

```
 org.junit,
 junit-jupiter-api
```

But removing the JUnit 6 bundle doesn't seem to fix the launch. There is no error dialog but still no tests are executed.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitEclipsePluginValidationOperation.java
```diff
@@ -0,0 +1,80 @@
...
+					String error = requirementName + " depends on " + name + "_" + version; //$NON-NLS-1$ //$NON-NLS-2$
+					fJunitErrors.put(error, EMPTY);
+				}
+			}
+		}
+	}
+
+	@Override
+	public boolean hasErrors() {
+		return super.hasErrors() || fJunitErrors.size() >= 1;
```
laeubi:
```suggestion
		return super.hasErrors() || fJunitErrors.size() > 0;
```

or 

```suggestion
		return super.hasErrors() || !fJunitErrors.isEmpty();
```

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitEclipsePluginValidationOperation.java
```diff
@@ -0,0 +1,80 @@
...
+	public void filterMatches(BundleRequirement requirement, Collection<BundleCapability> candidates) {
+		String requirementName = requirement.getRevision().getSymbolicName();
+		if (!requirementName.startsWith("junit")) { //$NON-NLS-1$
+			Iterator<BundleCapability> iterator = candidates.iterator();
+			while (iterator.hasNext()) {
+				BundleCapability candidate = iterator.next();
+				BundleRevision candidateRevision = candidate.getRevision();
+				String name = candidateRevision.getSymbolicName();
+				Version version = candidateRevision.getVersion();
+				if (junitVersion == 5 && version.getMajor() == 6 && (name.startsWith("junit"))) { //$NON-NLS-1$
```
laeubi:
```suggestion
				if (version.getMajor() > junitversion && (name.startsWith("junit"))) { //$NON-NLS-1$
```

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitEclipsePluginValidationOperation.java
```diff
@@ -0,0 +1,80 @@
...
+					// TODO: externalize
+					String error = requirementName + " depends on " + name + "_" + version; //$NON-NLS-1$ //$NON-NLS-2$
+					fJunitErrors.put(error, EMPTY);
```
laeubi: Can we not better put in the requirement resource here and as a value the error message?


```suggestion
					fJunitErrors.put(requirement.getRevision(), "Depends on a different JUnit Version than the launch");
```

trancexpress: That doesn't seem to include the error message:

Screenshot_20251106_140259

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java
```diff
@@ -609,8 +616,41 @@ protected void validateProjectDependencies(ILaunchConfiguration configuration, I
...
+
+		@Override
+		public void filterMatches(BundleRequirement requirement, Collection<BundleCapability> candidates) {
+			Iterator<BundleCapability> iterator = candidates.iterator();
+			while (iterator.hasNext()) {
+				BundleCapability candidate = iterator.next();
+				String name = candidate.getResource().getSymbolicName();
+				Version version = candidate.getResource().getVersion();
+				if (version.getMajor() == 6 && (name.startsWith("junit") || name.startsWith("org.junit"))) { //$NON-NLS-1$ //$NON-NLS-2$
+					iterator.remove();
```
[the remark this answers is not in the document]
trancexpress: I've opted for extracting the code in a static inner class, it doesn't make things much more complicated.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitEclipsePluginValidationOperation.java
```diff
@@ -0,0 +1,93 @@
...
+			if (!"optional".equals(directives.get("resolution"))) { //$NON-NLS-1$ //$NON-NLS-2$
+				String requirementName = requirement.getRevision().getSymbolicName();
+				if (!requirementName.startsWith("junit")) { //$NON-NLS-1$
+					Iterator<BundleCapability> iterator = candidates.iterator();
+					while (iterator.hasNext()) {
+						BundleCapability candidate = iterator.next();
+						BundleRevision candidateRevision = candidate.getRevision();
+						String name = candidateRevision.getSymbolicName();
+						Version version = candidateRevision.getVersion();
+						if (version.getMajor() > junitVersion && name.startsWith("junit")) { //$NON-NLS-1$
```
trancexpress: Technically combinations of 5 and 6 are a problem, which is not covered well by `version.getMajor() > junitVersion`. But we can change this once JUnit 6 support is there.

laeubi: > Technically combinations of 5 and 6 are a problem, which is not covered well by version.getMajor() > junitVersion. But we can change this once JUnit 6 support is there.

Let us first cover one thing at a time, the actual technical background does not matter so mutch but we should let the user knwo it uses a different version than the one is selected to launch with. SO at best we would use version ranges for Junit 5 + 6 and ... while 5 is a bit messy because it has different ones with 1.x and 5.x

ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java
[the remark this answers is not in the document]
trancexpress: Looks like static inner class won't work, since we need to retrieve the bundle, which we do with `getState()`.

laeubi:
The text might be a bit improved but that's something that can be done later, beside that it looks good and will allow the user to identify the issue.

What do you think about 

> "JUnit "+majorversionofjunitbundle+" is incompatible with selected JUnit "+versionweactuallywant+" runtime selected for the launch"

I think we should just state this as is and not add "may" .. from the second picture we maybe like to record the offending resource only once, I assume the same bundle has two requirements that can cause problems.

trancexpress:
I can't construct a bad case anymore; I do see mixed JUnit 5 and 6 dependencies (even `junit-platform-engine`, which was causing problems before) in `JUnitLaunchConfigurationDelegate.preLaunchCheck()`, after the call to `JUnitLaunchRequirements.addRequiredJunitRuntimePlugins()`.

But the JUnit plug-in test launch works fine, both for JUnit 5 and 6.

Maybe some fixes in `SPIBundleClassLoader` are making this also work? Any idea @laeubi ? I'm not sure why we had problems before and now we just don't.

PR #2104, opened 2025-11-05
Remove compilation error resolution proposal for adding Junit 5 bundles

HannesWell:
For junit-5 it's sufficient and recommended to just import the packages needed for compilation. Alternatively it's also possible to require the `junit-jupiter-api` bundle.

Both is already proposed by default and in both cases it's not necessary to require the junit-4 bundle (`org.junit`) too.

As this extra proposal also complicates the code significantly, it is removed with the recommendation to use the other standard proposals in case of missing JUnit-5 dependencies too.

Furthermore the current code seems to be partially incorrect, because actually package names have to be checked while it seems that the code assumed bundle names (which only coincidentally match).

Removing it also avoids the need to handle the JUnit-6 case too.

This was initially proposed in
- https://github.com/eclipse-pde/eclipse.pde/pull/2053#issuecomment-3474463774

With this, the proposals are now (https://github.com/eclipse-pde/eclipse.pde/pull/2053#issuecomment-3474463774 contains the previous state):

In a follow-up PR the proposals should be further enhanced to
- show package-imports at the top
- remove duplicated proposals
- show versions and propose multiple versions if available.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/java/JavaResolutionFactory.java
@@ -185,11 +185,6 @@ public class JavaResolutionFactory {
 			super(project, desc, cu, qualifiedTypeToImport);
 		}
 
-		private RequireBundleManifestChange(IProject project, String desc, CompilationUnit cu,
-				String qualifiedTypeToImport) {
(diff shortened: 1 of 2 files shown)


laeubi:
This sounds reasonable to me also moving package imports to the top would be great.

trancexpress:
Thinking about this, maybe this was done before it was enough to add `junit-jupiter-api` to the dependencies. So at that time likely it made sense to have an extra fix proposal.

PR #2105, opened 2025-11-05
Streamline compile error resolution proposals and strengthen typing

HannesWell:
This currently contains and requires as a prerequisites
- https://github.com/eclipse-pde/eclipse.pde/pull/2104

I already created it to show what kind of simplification is possible with the latter applied.

This will simplify subsequent further enhancements.


HannesWell:
Test failures are unrelated.
Submitting this now as this enables subsequent enhancement in the area of quick-fix proposal adding dependencies.

Issue #2024, opened 2025-10-17
Launching JUnitExecutionTest runs into odd errors


trancexpress:
If I run `org.eclipse.pde.junit.runtime.tests.JUnitExecutionTest` (from the context menu of the Java Editor, as a JUnit plug-in test), I get:

```
org.eclipse.core.runtime.CoreException: Launching failed. Bootstrap code cannot be found.
	at org.eclipse.pde.launching.JUnitLaunchConfigurationDelegate.abort(JUnitLaunchConfigurationDelegate.java:178)
	at org.eclipse.pde.launching.JUnitLaunchConfigurationDelegate.getClasspathAndModulepath(JUnitLaunchConfigurationDelegate.java:444)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.getVMRunnerConfiguration(JUnitLaunchConfigurationDelegate.java:192)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.launch(JUnitLaunchConfigurationDelegate.java:275)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:778)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:689)
	at org.eclipse.debug.internal.ui.DebugUIPlugin.buildAndLaunch(DebugUIPlugin.java:1052)
	at org.eclipse.debug.ui.DebugUITools.buildAndLaunch(DebugUITools.java:1044)
	at org.eclipse.pde.junit.runtime.tests.TestExecutionUtil.launchAndWaitForTermination(TestExecutionUtil.java:126)
	at org.eclipse.pde.junit.runtime.tests.TestExecutionUtil.runTest(TestExecutionUtil.java:81)
	at org.eclipse.pde.junit.runtime.tests.JUnitExecutionTest.executePackage(JUnitExecutionTest.java:111)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.pde.ui.tests.runtime.TestUtils$1.evaluate(TestUtils.java:267)
```

If I use `Restore Defaults` in the `Plugins` tab of the launch configuration, the error is no longer reported.

laeubi:
Is this a general problem? I never seen this before...

We have things like this that add (implicitly) required items

[...]

PR #2116, opened 2025-11-10
Added missing requirement to org.eclipse.equinox.launcher in test bundles

trancexpress:
Launching `org.eclipse.pde.junit.runtime.tests.JUnitRuntimeTests` or
`org.eclipse.pde.ui.tests.launcher.PluginBasedLaunchTest` from Eclipse
results in the following exception:

`org.eclipse.core.runtime.CoreException`: Launching failed. Bootstrap code
cannot be found.

Adding `org.eclipse.equinox.launcher` to the required bundles of
`org.eclipse.pde.junit.runtime.tests` and `org.eclipse.pde.ui.tests` avoids
these errors.

Fixes: #2024


trancexpress:
Seems like `org.eclipse.equinox.internal.simpleconfigurator.utils.SimpleConfiguratorUtils.readConfiguration(URL, URI)` doesn't see the `org.eclipse.equinox.launcher` bundle:

```
➜  ~ cat /tmp/tst2/.metadata/.plugins/org.eclipse.pde.core/pde-junit/org.eclipse.equinox.simpleconfigurator/bundles.info  | grep launcher
junit-platform-launcher,1.13.3,file:/data/eclipses/eclipse_437_20250713/plugins/junit-platform-launcher_1.13.3.jar,4,false
```
```
➜  ~ cat /tmp/tst1/.metadata/.plugins/org.eclipse.pde.core/pde-junit/org.eclipse.equinox.simpleconfigurator/bundles.info | grep launcher
junit-platform-launcher,1.13.2,file:/data/eclipses/eclipse_437_20250703/plugins/junit-platform-launcher_1.13.2.jar,4,false
org.eclipse.equinox.launcher,1.7.0.v20250519-0528,file:/data/eclipses/eclipse_437_20250703/plugins/org.eclipse.equinox.launcher_1.7.0.v20250519-0528.jar,4,false
org.eclipse.equinox.launcher.gtk.linux.x86_64,1.2.1400.v20250331-1702,file:/data/eclipses/eclipse_437_20250703/plugins/org.eclipse.equinox.launcher.gtk.linux.x86_64_1.2.1400.v20250331-1702/,4,false
```

From then on a lot of things are called, but essentially this code doesn't add the launcher bundle because its missing above:

```
	@BeforeClass
	public static void setupProjects() throws Exception {
		Pattern junitRuntimeIDs = Pattern
				.compile("\\.junit\\d*\\.runtime|junit\\..+\\.engine$|org\\.junit\\.platform\\.launcher");
		TargetPlatformUtil.setRunningPlatformSubSetAsTarget(TargetPlatformUtil.class + "_target", b -> {
```

In particular `installedBundles` in `org.eclipse.pde.ui.tests.util.TargetPlatformUtil.addRunningPlatformBundles(Collection<ITargetLocation>, Predicate<Bundle>)` doesn't contain the launcher bundle:

```
	private static List<NameVersionDescriptor> addRunningPlatformBundles(Collection<ITargetLocation> bundleContainers,
			Predicate<Bundle> bundleFilter) {
		Bundle[] installedBundles = FrameworkUtil.getBundle(TargetPlatformUtil.class).getBundleContext().getBundles();
```

Likely I cannot bisect this properly since its not a problem when tests are launched, its a problem when the launch is created...

laeubi:
Alright it then seems that we really need the launcher here as it has to be either in the target or is expected to be present in the runtime. Adding it to the manifest then seems fine, but we should add an `@Reference(some class from the launcher bundle)` to document this requirement and make it not look unused here.

laeubi:
> It does seem odd to have to add to the MANIFEST.MF even though the test code itself doesn't need it.

You showed that the code actually need it or throws exception, this is similar as with classforname ... even if not explicitly required it is implicit.

> should the PDE JUnitLaunchConfigurationDelegate always require org.eclipse.equinox.launcher

Maybe yes, I remember I added equinox there recently as well (what obviously do not work without it)

> Or is LaunchArgumentsHelper.constructClasspath supposed to work also without org.eclipse.equinox.launcher in the target?

That's the interesting question .. one would hope it is not required at all.

iloveeclipse (approved):

PR #2117, opened 2025-11-11
Add expand and collapse options to API Errors/Warnings preferences page

elsazac:
This change introduces `Expand All` and `Collapse All` toolbar options to the API Errors/Warnings preference page under `Settings → Plugin Development Environment → API Errors/Warnings`. These options make it easier to navigate the large number of expandable sections on the page, aligning the behaviour with other preference pages such as Java → Compiler → Errors/Warnings.

https://github.com/user-attachments/assets/9aeb8ca4-bf8d-42ef-b5ec-9d0597fe3489

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
+++ b/apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
@@ -31,15 +31,21 @@ import org.eclipse.pde.api.tools.ui.internal.IApiToolsConstants;
 import org.eclipse.pde.api.tools.ui.internal.IApiToolsHelpContextIds;
 import org.eclipse.pde.api.tools.ui.internal.SWTFactory;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.ScrolledComposite;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Link;
+import org.eclipse.swt.widgets.ToolBar;
+import org.eclipse.swt.widgets.ToolItem;
+import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferencePage;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.dialogs.PreferencesUtil;
+import org.eclipse.ui.forms.widgets.ExpandableComposite;
 import org.eclipse.ui.preferences.IWorkbenchPreferenceContainer;
 
 /**
@@ -98,9 +104,36 @@ public class ApiErrorsWarningsPreferencePage extends PreferencePage implements I
 		super(PreferenceMessages.ApiErrorsWarningsPreferencePage_0);
 	}
 
+	private void expandCollapseItems(Composite parent, boolean expandPage) {
+		for (Control control : parent.getChildren()) {
+			if (control instanceof ExpandableComposite page) {
+				page.setExpanded(expandPage);
+				Control child = page.getClient();
+				if (child != null && !child.isDisposed()) {
+					child.setVisible(expandPage);
+				}
+			} else if (control instanceof Composite comp) {
+				expandCollapseItems(comp, expandPage);
+			}
+		}
+		ScrolledComposite composite = getScrolledComposite(parent);
+		if (composite != null) {
+			composite.setMinSize(parent.computeSize(SWT.DEFAULT, SWT.DEFAULT));
+		}
+	}
+
+	private ScrolledComposite getScrolledComposite(Composite comp) {
+		Composite currentComposite = comp;
+		while (currentComposite != null && !(currentComposite instanceof ScrolledComposite)) {
+			currentComposite = currentComposite.getParent();
+		}
+		return (ScrolledComposite) currentComposite;
+	}
+
 	@Override
 	protected Control createContents(Composite parent) {
 		Composite comp = SWTFactory.createComposite(parent, 1, 1, GridData.FILL_BOTH, 0, 0);
+		((GridLayout) comp.getLayout()).verticalSpacing = 0;
 		link = new Link(comp, SWT.NONE);
 		link.setLayoutData(new GridData(GridData.END, GridData.CENTER, true, false));
 		link.setFont(comp.getFont());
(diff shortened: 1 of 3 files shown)


apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/preferencemessages.properties
```diff
@@ -139,6 +139,9 @@ CONSTRUCTOR_CHANGED_VARARGS_TO_ARRAY=Converted variable argument type to array t
 CONSTRUCTOR_CHANGED_DECREASE_ACCESS=The visibility has been reduced:
 CONSTRUCTOR_REMOVED_TYPE_PARAMETER=A type parameter has been removed:
 
+Preference_Collapse_All= Collapse All
+Preference_Expand_All= Expand All
```
gireeshpunathil:
```suggestion
Preference_Collapse_All= Collapse all preferences
Preference_Expand_All= Expand all preferences
```

elsazac: Done.

gireeshpunathil (approved):

elsazac:
@noopur2507 Could you review this?

merks:
It's convenient and works well:

image

I wonder if it's possible to compress the gap? Between the toolbar buttons and the tab area?

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
```diff
@@ -127,9 +163,33 @@ protected Control createContents(Composite parent) {
...
+		expandItems.addListener(SWT.Selection, e -> expandCollapseItems(comp, true));
+		collapseItems.addListener(SWT.Selection, e -> expandCollapseItems(comp, false));
+		comp.addDisposeListener(e -> {
+			if (!expandImage.isDisposed() && expandImage != null) {
+				expandImage.dispose();
+			}
+			if (!collapseImage.isDisposed() && collapseImage != null) {
+				collapseImage.dispose();
+			}
+		});
```
merks: I think this is probably quite wrong because it's a shared image. Should a shared image be disposed?  Of course the null guard is in the wrong order too...

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
```diff
@@ -98,6 +107,33 @@ public ApiErrorsWarningsPreferencePage() {
...
+					child.setVisible(expandPage);
+				}
+			} else if (control instanceof Composite) {
+				expandCollapseItems((Composite) control, expandPage);
+			}
+		}
+		parent.layout(true, true);
+		ScrolledComposite composite = getScrolledComposite(parent);
+		if (composite != null) {
+			composite.setMinSize(parent.computeSize(SWT.DEFAULT, SWT.DEFAULT));
```
merks, line 121: I'm not sure what this is doing or why it's needed.

merks (changes requested):
I think the image dispose is probably wrong.

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
merks: Actually, given you never assign these the variables are redundant, the are always null, and always cause an NPE.

elsazac: Forgot to update this, I was in fact using an image earlier here, but just now replaced it with the one from IsharedImages. but forgot to update the dispose event.

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
elsazac: after collapsing or expanding the children it changes the size of the content area .So in order to recompute the area I added that. it recomputes the area and make sure that the scroll area updates correctly after the size change.

merks: If it were before the layout call above I would have assume it affected the layout, but it's after so that struck me as odd.

elsazac:
> I wonder if it's possible to compress the gap? Between the toolbar buttons and the tab area?


I could not reduce the gap between the toolbar and the scrolled composite. All the three controls (toolbar, link, scrolled composite) are placed in separate rows, one by one, of the parent. So adjusting indents also did not show any visible change.

As an alternative, I tried moving the toolbar above the link. This makes makes the layout look a bit more compact overall. Let me know if this looks better or not.

image

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
elsazac: Since parent here is not the ScrolledComposite, we have to traverse upwards to find the enclosing scroll container anyway. we need to update it then to minimum size based on the newly computed content size, otherwise even after collapsing, the container allows to scroll extra even if the content is not present.

when i tested the flow again, i think `parent.layout(true, true)` is not strictly required here and the `ScrolledComposite#setMinSize( )` is sufficient enough to correctly update the scrollable area based on the updated content size. so I can remove that.

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
elsazac: removed the image related code.

elsazac:
> I wonder if it's possible to compress the gap? Between the toolbar buttons and the tab area?

@merks I have tried reducing the gap between the toolbar and scroll composite. Hope this looks okay now.

image

merks (approved):
That looks nicer.

elsazac:
@merks I have squashed the commits. Could you merge this?

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
```diff
@@ -98,9 +104,36 @@ public ApiErrorsWarningsPreferencePage() {
...
 
+	private void expandCollapseItems(Composite parent, boolean expandPage) {
+		for (Control control : parent.getChildren()) {
+			if (control instanceof ExpandableComposite page) {
+				page.setExpanded(expandPage);
+				Control child = page.getClient();
+				if (child != null && !child.isDisposed()) {
+					child.setVisible(expandPage);
+				}
+			} else if (control instanceof Composite) {
```
merks: Could you please use a pattern to avoid the cast on the next line?

merks (commented):
Just one last nit. Sorry.

apitools/org.eclipse.pde.api.tools.ui/src/org/eclipse/pde/api/tools/ui/internal/preferences/ApiErrorsWarningsPreferencePage.java
elsazac: Yes, Done.

elsazac:
> Just one last nit. Sorry.

No problem.I have addressed those changes as well.

elsazac:
@merks thanks for reviewing and merging this.

Could you please help merge this as well. https://github.com/eclipse-pde/eclipse.pde/pull/2124 ?
This PR is similar and adds the same option to the Compiler preference page. I have addressed all the requested changes in https://github.com/eclipse-pde/eclipse.pde/pull/2124 as well.

PR #2118, opened 2025-11-11
Remove eclipse32.png because it's not in the build.properties anyway

merks:
https://github.com/eclipse-platform/eclipse.platform/issues/2261


merks:
Build failures are unrelated to the change.

Issue #2125, opened 2025-11-13
Expose SPIBundleClassLoader


trancexpress:
Our own test framework we use to run test so far had a copy of `MultiBundleClassLoader`, now it seems it will have to copy `SPIBundleClassLoader` (and all the other package private classes it uses). Since `SPIBundleClassLoader` is in an internal package, I see no harm to expose the class and make it `public`.

PR #2126, opened 2025-11-13
Make SPIBundleClassLoader public

trancexpress:
Fixes: #2125

### DIFF
--- a/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/SPIBundleClassLoader.java
+++ b/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/SPIBundleClassLoader.java
@@ -31,14 +31,14 @@ import org.osgi.framework.Bundle;
  * The classloader wraps the OSGi provided one but gives access for the JUnit
  * runer to any SPI declared services.
  */
-class SPIBundleClassLoader extends ClassLoader {
+public class SPIBundleClassLoader extends ClassLoader {
 
 	private static final String META_INF_SERVICES = "META-INF/services/"; //$NON-NLS-1$
 	private final List<Bundle> bundles;
 	private final int junitVersion;
 	private final Map<String, List<SPIMapping>> mappings = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


trancexpress:
@laeubi anything against this change?

laeubi:
@trancexpress what is the exact use-case for that? This is really not "public" code in any way it might change in any way or vanish in any way later on and making very special assumptions about the environment.

Maybe it is better to simply *copy* the code to your product then?

trancexpress:
> @trancexpress what is the exact use-case for that?

We have a test framework that runs tests from command line. We have the same problems as PDE, not being able to correctly detect and load the JUnit Jupiter services. So far we have copied `MultiBundleClassLoader`, the next logical step is moving to the new class loader.

> This is really not "public" code in any way it might change in any way or vanish in any way later on and making very special assumptions about the environment.

Its OK if the code changes, we build Eclipse weekly and run our  tests on that. We'll notice and adjust. Usually the code changed for a reason, similar to how the previous class loader was replaced by this one.

> Maybe it is better to simply _copy_ the code to your product then?

Its now 3 classes, we already weren't happy with copying the previous. Generally we avoid copying things from the platform, if we can. We also get more test coverage for the platform that way.

laeubi (changes requested):
> Public doesn't mean API & copy&paste is worst thing ever, I don't need to explain why.
> The change is trivial, minimal and doesn't harm, at least one consumer (Advantest) needs that.

I disagree here. Code should be as local as possible and making it "public for used by other" makes it effectively public API how ever you call it. We already have a whole whack of things that are "internal" but we cannot change it because used everywhere else... of course everyone has a "good reason" to use it leading to [death by thousands cuts](https://en.wikipedia.org/wiki/Lingchi).

> We should simply merge it.

Asking @HannesWell to decide on this.

laeubi:
> Maybe it would be good to ensure the build is green?

It currently fails due to nexus timeout :-\

trancexpress:
I opened https://github.com/eclipse-pde/eclipse.pde/issues/2128 for the flaky tests.

Anything else here or can we merge?

Issue #666, opened 2023-07-12
Random failing UpdateUnitVersionsCommandTests


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/R-4.28-202306050440/testresults/html/org.eclipse.pde.genericeditor.extension.tests_ep428I-unit-win32-java17_win32.win32.x86_64_17.html
```

ID: org.eclipse.fake.7 has the incorrect version. expected:<[1.0.0.v2]> but was:<[0.0.0]>

org.junit.ComparisonFailure: ID: org.eclipse.fake.7 has the incorrect version. expected:<[1.0.0.v2]> but was:<[0.0.0]>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.eclipse.pde.genericeditor.extension.tests.UpdateUnitVersionsCommandTests.confirmVersionUpdates(UpdateUnitVersionsCommandTests.java:92)
at org.eclipse.pde.genericeditor.extension.tests.UpdateUnitVersionsCommandTests.testVersionSort(UpdateUnitVersionsCommandTests.java:55)
```

jukzi:
same in 
https://download.eclipse.org/eclipse/downloads/drops4/I20241118-0600/testresults/html/org.eclipse.pde.genericeditor.extension.tests_ep434I-unit-macosx-aarch64-java17_macosx.cocoa.aarch64_17.html

jukzi:
Enhanced info from https://download.eclipse.org/eclipse/downloads/drops4/I20241120-0600/testresults/html/org.eclipse.pde.genericeditor.extension.tests_ep434I-unit-macosx-aarch64-java17_macosx.cocoa.aarch64_17.html

```
org.junit.ComparisonFailure: ID: org.eclipse.fake.6 has the incorrect version. updatedText=<target>
<locations>
<location>
<unit id="org.eclipse.fake.1" version="0.0.0"/>
<unit id="org.eclipse.fake.2" version="0.0.0"/>
<unit id="org.eclipse.fake.3" version="0.0.0"/>
<unit id="org.eclipse.fake.4" version="0.0.0"/>
<unit id="org.eclipse.fake.5" version="0.0.0"/>
<unit id="org.eclipse.fake.6" version="0.0.0"/>
<unit id="org.eclipse.fake.7" version="0.0.0"/>
<repository location="bundleentry://349.fwk733886212/testing-files/testing-sites/MultipleUnitsConfirmSorting/"/>
</location>

```

[...]

PR #2130, opened 2025-11-14
Add missing bundle so that UpdateUnitVersionsCommandTests doesn't fail when ran from Eclipse

trancexpress:
When running `UpdateUnitVersionsCommandTests` in Eclipse, the test fails. An error is logged before the fail:

```
!ENTRY org.eclipse.pde.genericeditor.extension 2 0 2025-11-14 12:34:50.660
!MESSAGE Cannot invoke "org.eclipse.equinox.internal.p2.repository.Transport.download(java.net.URI, java.io.OutputStream, org.eclipse.core.runtime.IProgressMonitor)" because the return value of "org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.getTransport()" is null
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.equinox.internal.p2.repository.Transport.download(java.net.URI, java.io.OutputStream, org.eclipse.core.runtime.IProgressMonitor)" because the return value of "org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.getTransport()" is null
	at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.handleRemoteIndexFile(AbstractRepositoryManager.java:797)
	at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadIndexFile(AbstractRepositoryManager.java:791)
	at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadRepository(AbstractRepositoryManager.java:727)
	at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.loadRepository(MetadataRepositoryManager.java:107)
	at org.eclipse.equinox.p2.repository.metadata.IMetadataRepositoryManager.loadRepository(IMetadataRepositoryManager.java:101)
	at org.eclipse.pde.internal.genericeditor.target.extension.p2.P2Fetcher.fetchAvailableUnits(P2Fetcher.java:63)
	at org.eclipse.pde.internal.genericeditor.target.extension.model.RepositoryCache.lambda$11(RepositoryCache.java:106)
	at org.eclipse.core.runtime.jobs.Job$2.run(Job.java:187)
```

If I restore defaults in the launch config, this code finds the service for `Transport` in the bundle `org.eclipse.equinox.p2.transport.ecf`: `org.eclipse.osgi.internal.serviceregistry.ServiceRegistry.lookupServiceRegistrations(String, Filter)`

```
				/* services registered under the class name */
				result = publishedServicesByClass.get(clazz);
```

`filter` has contents:

```
(|(p2.agent.service.name=org.eclipse.equinox.internal.p2.repository.Transport)(p2.agent.servicename=org.eclipse.equinox.internal.p2.repository.Transport))
```

Should be defined in this file I guess:

https://github.com/eclipse-equinox/p2/blob/master/bundles/org.eclipse.equinox.p2.transport.ecf/OSGI-INF/ecfTransport.xml

### DIFF
--- a/ui/org.eclipse.pde.genericeditor.extension.tests/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.genericeditor.extension.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for Generic Target Platform Editor
 Bundle-SymbolicName: org.eclipse.pde.genericeditor.extension.tests
-Bundle-Version: 1.3.100.qualifier
+Bundle-Version: 1.3.200.qualifier
 Bundle-Vendor: Eclipse.org
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
--- a/ui/org.eclipse.pde.genericeditor.extension.tests/pom.xml
+++ b/ui/org.eclipse.pde.genericeditor.extension.tests/pom.xml
@@ -18,7 +18,7 @@
     <relativePath>../../</relativePath>
   </parent>
   <artifactId>org.eclipse.pde.genericeditor.extension.tests</artifactId>
-  <version>1.3.100-SNAPSHOT</version>
+  <version>1.3.200-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
 
   <properties>


trancexpress:
This doesn't fix anything in #666, but I can debug at least.

ui/org.eclipse.pde.genericeditor.extension.tests/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,7 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.pde.genericeditor.extension,
  org.eclipse.pde.core,
  org.eclipse.equinox.p2.metadata,
+ org.eclipse.equinox.p2.transport.ecf,
```
laeubi: Requiring a concrete _implementation_ her is likely wrong in many ways.

Instead the provider needs to provide approciate capbilty (service or implementor) and the consumer (where NPE occures) require appropriate requirement (either service or implementor capabilty)


A consumer bunlde like this should not need any of those!

merks: The fact that all the tests pass seems like a good thing.  Future improvement could be saved for the future...

Issue #1863, opened 2025-07-10
Refactoring Participant should remove empty packages from the Manifest


laeubi:
Currently if one moves a type around and the package becomes empty the Manifest editor complains that the exported package does not exists.

It would be good for PDE to register a refactoring participant and remove packages that become empty from the manifest automatically.

As a side-note, _renaming_ a package already updates it in the manifest

laeubi:
This come especially important when other plugins _import_ the package... so depending on the fan-out it could be much work to remove obsolete import/export packages afterwards.

JoseRodriguez26:
All be working on this ! Currently associated Via @Codeday

nburnwal09:
@laeubi 
I am able to reproduce this issue. I get this error when I make a package empty. 
`Package 'foo' does not exist in this plug-in
`
For me this error msg is itself misleading. Do you think we should update the error msg? or provide a quick fix that deletes empty package references from manifest?  Can you please clarify, I didn't exactly get `refactoring participant`.

PR #2136, opened 2025-11-18
Empty package removed from manifest while refactoring

nburnwal09:
Fixes: #1863
Currently if one moves a type around and the package becomes empty the Manifest editor complains that the exported package does not exists.
This fix provides removal of empty package from exported package in Manifest file as part of refactoring participant

https://github.com/user-attachments/assets/0e3760e1-1394-4476-ab89-b441477f132b


nburnwal09:
@laeubi 
Can we proceed with this now?

Issue #2132, opened 2025-11-14
Cannot run JUnit tests with class loading problem


Phillipus:
This is a follow on from https://github.com/eclipse-pde/eclipse.pde/issues/2131

We have a plug-in that uses the GraalVM JS engine and includes jars in its plug-in:

<img width="558" height="309" alt="Image" src="https://github.com/user-attachments/assets/1affd969-b9b0-4614-95af-6f6d37ec29a1" />

Using the latest PDE Junit launch config the GraalVM engine fails to load and so the tests fail.

The GraalVM engine is loaded like this:

```java
ScriptEngine engine = new ScriptEngineManager().getEngineByName("graal.js");
```

This works when running normally but when called from a JUnit test in Eclipse PDE the engine is `null` which means that there might be a classloader problem.

Same problem with JUnit 5 or 6.

iloveeclipse:
Probably bug in org.eclipse.pde.internal.junit.runtime.SPIBundleClassLoader / regression from 24ca18da4fa4af9b866645779f9e720e72c9bfdf / https://github.com/eclipse-pde/eclipse.pde/pull/2071.

@laeubi : could you please check?

Phillipus:
I've created a simple test plug-in (attached below) containing a JUnit test that attempts to load an example `javax.script.ScriptEngine` and `javax.script.ScriptEngineFactory` that is contained in the compiled jar `engine.jar` that is part of the plug-in. This is declared in the jar with a file `META-INF\services\javax.script.ScriptEngineFactory` containing the line `com.example.script.MyScriptEngineFactory`.

The test simply does this:

```java
@Test
public void loadScriptEngine() {
    ScriptEngine engine = new ScriptEngineManager().getEngineByName("my-engine");
    assertNotNull(engine);
}
```

I'm using `Eclipse 4.38 I20251114-1800` as the client and the target for the workspace is `Eclipse SDK 4.38.0.I20251115-1800`.

The ScriptEngine fails to load using this target but loads successfully using an earlier target of `Eclipse SDK 4.37.0.I20250905-0730`.

Steps to reproduce:

1. Ensure the workspace target is set to a recent 4.38 I Build or M release for Eclipse SDK
2. Load the attached project `ScriptEngineLoaderTest`
3. Create a JUnit Test Plug-in Test launch configuration with the following settings:

    On tab "Test":

    - Run a single test
    - Project: ScriptEngineLoaderTest
[...]

PR #2137, opened 2025-11-18
Use Bundle#getResources to discover embedded SPI resources

laeubi:
Currently we use bundle.getEntry but this has some limitations, e.g. fragments and embedded jars.

While the first can be avoided with findEntries, the second can not. As we previously already used bundle.getResources it seems fine to continue using that method.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2132

### DIFF
--- a/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/SPIMapping.java
+++ b/ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/SPIMapping.java
@@ -17,9 +17,9 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.URL;
-import java.util.HashSet;
+import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.osgi.framework.Bundle;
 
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.junit.runtime/src/org/eclipse/pde/internal/junit/runtime/SPIMapping.java
```diff
@@ -28,23 +29,27 @@ final class SPIMapping {
...
+		this.classes = readClasses(urls);
 	}
 
-	private Set<String> readClasses(URL entry) {
-		try (BufferedReader reader = new BufferedReader(new InputStreamReader(entry.openStream()))) {
-			return reader.lines().collect(Collectors.toSet());
-		} catch (IOException e) {
-			return new HashSet<>();
+	private static Set<String> readClasses(Collection<URL> urls) {
+		HashSet<String> result = new LinkedHashSet<>();
```
HannesWell:
```suggestion
		Set<String> result = new LinkedHashSet<>();
```

HannesWell (approved):
Looks reasonable to me.
But due to a recent outage at Github regarding git operations I'm not able to fetch this change to try it locally. But from what I see in the web-UI this is fine.
Thanks for the fix.

I just have two style remarks, that would be nice to have.

PR #2138, opened 2025-11-18
Use Multi-Release Type for Caller implementation

laeubi:
The current implementation for Java 1.8 is rather complex because of only raw string info.

With Java 9+ we can use StackWalker to perform faster and easier, to still retain compatibility with Java 1.8 we use a Multi-Release type here to use that implementation on any 9+ JVM.

See
- https://eclipse.dev/eclipse/markdown/?f=news/4.38/jdt.md#multi-release-jar-compilation-support
- https://github.com/eclipse-tycho/tycho/blob/tycho-4.0.x/RELEASE_NOTES.md#building-multi-release-jars


ui/org.eclipse.pde.junit.runtime/.classpath
```diff
@@ -1,7 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-9">
```
HannesWell: Shouldn't this stay at `1.8` or does it have to be the highest version of all contained source folders?
I assume the general jdt.core settings (with release option enabled) ensure that the main src folder isn't compiling against Java-9?

PR #2142, opened 2025-11-20
Catch IllegalArgumentException when resolving target definition

ptziegler:
When the target editor is opened, PDE tries to delete the old target handle as part of "Bug 545681 - UI Freeze of >800s in TargetEditor when opening editor".

If the profile is not locked (due to whatever reasons), an IllegalArgumentException is thrown, blocking the resolution of this file.


ptziegler:
See https://github.com/eclipse-pde/eclipse.pde/discussions/2141

laeubi (approved):
Sounds reasonable to me

merks:
It's dealing with this case:

image

ptziegler:
Yes. The problem is that PDE forgot that it had already unlocked the profile in the past. So when trying to unlock this already unlocked profile, the exception occurs.

vogella:
I suggest to merge this one

merks:
It looks fine to me as well.

PR #2144, opened 2025-11-21
Mark long-deprecated and unused constants for removal

nburnwal09:
There are few constants in PDE that have been deprecated for a long time and are no longer referenced within the codebase.
This change adds Deprecated tags for those to indicate their planned removal in a future release.


nburnwal09:
Hi @iloveeclipse 
Could you please check this PR when you get time.?

PR #2148, opened 2025-12-02
Updating plug-in dependencies job should provide project information

vogella:
Updating plug-in dependencies can sometimes run relatively long, in its current form it give no indication which project requires a lot of time.

This adds the info, which project is currently updated.


vogella:
This is how it looks like: 

image

PR #2150, opened 2025-12-03
Target platform editor - Only resolve target platform if changed

vogella:
When you leave the Source tab, the editor takes the entire text (XML) from that tab and reconstructs a new internal Target Definition model (ITargetDefinition).

This new model is compared to the current target definition model and if it is the same, the resolving of the target platform is not triggered.


vogella:
I assume is [Tests / Build](https://github.com/eclipse-pde/eclipse.pde/pull/2150/checks?check_run_id=57040379095) — AllPDETests AllLauncherTests PluginBasedLaunchTest.testGetMergedBundleMap_multipleWorkspacePluginVersions_multipleSpecificVersion failed not related.

merks:
I saw this PR and it seems useful for the editor. For targlets we similarly check if the “definition” has changed to determine if we should perform on startup trigger. For manual trigger we always offer to perform because the update site locations’ content might be changed.

PR #2155, opened 2025-12-12
JDT Clean-up Use String.replace() instead of String.replaceAll()

vogella:
Testing this clean-up to enable it by default via
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3517


vogella:
LGTM, lets see what CI thinks

Issue #1945, opened 2025-08-25
Unable to open target editor when a version is malformed


ptziegler:
To reproduce: Set the version of one dependency to "." and open the target editor.

<img width="739" height="558" alt="Image" src="https://github.com/user-attachments/assets/2bdb7de5-c9ae-41ae-83d8-d5f3088dc122" />

PDE Core	3.21.0.v20250813-2022
PDE UI 3.16.200.v20250813-2022

nburnwal09:
@ptziegler 
Could you please tell the concrete steps to recreate this.?
I am unable to reproduce.

ptziegler:
@nburnwal09 
Just take the following target file, as an example:
```
<target name="GEF Classic Development Target Definition" sequenceNumber="1">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<unit id="org.eclipse.license.feature.group" version="x.y.z.v20181016-2210"/>
			<repository location="http://download.eclipse.org/cbi/updates/license"/>
		</location>
	</locations>
</target>
```
Which causes this error when opening the editor:
<img width="1227" height="696" alt="Image" src="https://github.com/user-attachments/assets/97455a55-e1b7-41fd-a77e-eaabecc7da49" />

Issue #1811, opened 2025-06-12
Add an editor for p2.inf


laeubi:
Currently one can do powerful things when using a `p2.inf` but users have to edit it manually.

It would be good to have an editor in PDE to show it along with the manifest (like we already do for `plugin.xml` and others) together with some auto-completion and validation.

merks:
Keep in mind that features, products, and I think a category.xml can have such associated things too.

nburnwal09:
Here is the samples for valud keys considering [here](https://eclipse.dev/eclipse/markdown/?file=eclipse-equinox/p2/master/docs/Customizing_Metadata.md)

| Prefix | Expected Suggestions |
|--------|----------------------|
| `<root>` | `properties`, `provides`, `requires`, `metaRequirements`, `update`, `instructions`, `iu`, `units` |
| `properties.<index>.` | `name`, `value` |
| `provides.<index>.` | `namespace`, `name`, `version` |
| `requires.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple`, `filter`, `min`, `max` |
| `metaRequirements.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple` |
| `update.` | `id`, `range`, `severity`, `description` |
| `instructions.` | `install`, `install.import`, `configure`, `configure.import`, `unconfigure`, `unconfigure.import`, `uninstall`, `uninstall.import` |
| `iu.<index>.` | `id`, `version`, `singleton`, `copyright`, `licenses`, `filter`, `touchpoint`, `update`, `artifacts`, `properties`, `provides`, `requires`, `metaRequirements`, `hostRequirements`, `instructions` |
| `iu.<index>.copyright.` | `location` |
| `iu.<index>.licenses.<index>.` | `location` |
| `iu.<index>.touchpoint.` | `id`, `version` |
| `iu.<index>.update.` | `match`, `id`, `range`, `severity`, `description` |
| `iu.<index>.artifacts.<index>.` | `id`, `version`, `classifier` |
| `iu.<index>.properties.<index>.` | `name`, `value` |
| `iu.<index>.provides.<index>.` | `namespace`, `name`, `version` |
| `iu.<index>.requires.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple`, `filter`, `min`, `max` |
[...]

PR #2157, opened 2025-12-19
Caught the exception when a version is malformed in target file

nburnwal09:
Issue: #1945 
Exception is caught to open the target editor when a version is malformed.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -3219,6 +3219,8 @@ public class PDEUIMessages extends NLS {
 
 	public static String TargetEditor_6;
 
+	public static String TargetEditor_7;
+
 	public static String FeatureImportWizardPage_importHasInvalid;
 
 	public static String ProductInfoSection_plugins;
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
@@ -195,6 +195,10 @@ public class TargetEditor extends FormEditor {
 			setActivePage(fSourceTabIndex);
 			CoreException ce = new CoreException(Status.error(e.getMessage(), e));
 			showError(PDEUIMessages.TargetEditor_5, ce);
+		} catch (IllegalArgumentException e) {
+			setActivePage(fSourceTabIndex);
+			CoreException ce = new CoreException(Status.error(e.getMessage(), e));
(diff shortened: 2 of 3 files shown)


gireeshpunathil (approved):

nburnwal09:
@laeubi @HannesWell 
I have applied the opening of build.properties behaviours to p2.inf in the latest changes.

Also, this PR is for the addition of p2.inf in manifest and feature editor tab just like build.properties.
I will take up the auto-completion and validation part [as suggested](https://github.com/eclipse-pde/eclipse.pde/issues/1811#issue-3139917054) in the follow up PR once this gets approved.

ptziegler:
@nburnwal09 Apologies for the slow response. I played around with the change and noticed that it's not really possible to save the editor, even after the version has been corrected. I assume that's why you initially caught the exception in `loadTargetDefinition()`?

If so, then the catch block is apparently still needed. But instead of converting it to a CoreException, I would rethrow it as is, in order to avoid the side effects described in https://github.com/eclipse-pde/eclipse.pde/pull/2157#discussion_r2710229286

e.g.

```java
private ITargetDefinition loadTargetDefinition() throws CoreException {
	ITargetPlatformService service = getTargetPlatformService();
	try {
		if (fInput instanceof IFileEditorInput) {
			ITargetHandle fileHandle = service.getTarget(((IFileEditorInput) fInput).getFile());
			fTarget = fileHandle.getTargetDefinition();
		} else if (fInput instanceof IURIEditorInput) {
			ITargetHandle externalTarget = service.getTarget(((IURIEditorInput) fInput).getURI());
			fTarget = externalTarget.getTargetDefinition();
		}
	} catch (CoreException | IllegalArgumentException e) {
		fTarget = service.newTarget();
		throw e;
	}
	PlatformUI.getWorkbench().getDisplay().asyncExec(
			() -> TargetEditor.this.getTargetChangedListener().contentsChanged(fTarget, this, true, false));
	return fTarget;
}
```

> It would seem good for the user to know which IU and which version string is causing the problem (if that isn't already diagnosed by an error marker that draw attention to the location).

@merks The problem is that the error itself is thrown by Equinox. So there is little that can be done on the PDE with regards to the error message. The invalid version is shown in the underlying exception, which at least to me is sufficient to find the invalid IU.

image

ptziegler (approved):
Apologies for the slow response time. This looks fine to me now. There is one remaining problem, but I think that exceeds the scope of this PR.

The `IllegalArgumentException` is thrown while constructing the `Definition` page. So even when switching to the `Source` page and correcting the version, the `Definition` page remains broken.

I think one either needs to check the state of the target file when determining the initial page that is openend or one needs to deconstruct this page when the exception is thrown. But as said: I think this is a related but separate issue.

HannesWell (approved):
> I couldn't make much meaning of the CI failures: are they real? I want to land this, but not really sure if the failure is caused by the code or infra flakes.

Sorry for the slow response. The build failure seems to be unrelated (after Ed's restart almost all were green).
I rebased to the latest master for a final verification and then this ready from my POV.
Thanks for this!

PR #2160, opened 2025-12-30
Move bundles to Java 21

akurtakov:
They already require Java 21 transitively.


akurtakov:
New warnings are due to Java 21 API deprecations. Merging as these will have to be handled separately - URL to URI changes are not straightforward.

Issue #2096, opened 2025-11-04
API filters reported as unused with no changes in the plug-in


stephan-herrmann:
Recent jdt.core builds show many new warnings from API tools with no related change in jdt.core:
* last good build on master: https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/875
* first build with new warnings: https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/878/

Most of these warnings/filters may have been wrong all those years, but at least the following warnings should still be reported, i.e., filters should still be needed:
* https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4594/1/apitools/new/source.e013c541-2920-4762-8faf-9cfe063ec3d8/#28
* https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4594/1/apitools/new/source.21c0880c-8010-4cb3-939d-7553561f38c0/#28

This makes me wonder if any of these changes are intended?

stephan-herrmann:
When may we expect to see the fix in CI builds? 
I retriggered one of the affected builds which still shows the 33 new warnings:
* https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4595/2/apitools/new/

I also triggered a fresh master build just now
* https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/880/

By now the fix from #2102 should be visible to our build, shouldn't it?

laeubi:
> When may we expect to see the fix in CI builds?

Whenever the infrastructure is fixed and we get a new ibuild successfully promoted see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6744#note_5320116

stephan-herrmann:
I just noted this build output:
```
API Tools version: 1.3.1000.v20251103-0716
```

while the fix is tagged I20251105-1800 ...

stephan-herrmann:
> > I just noted this build output:
> > ```
> > API Tools version: 1.3.1000.v20251103-0716
> > ```
> >   
> > while the fix is tagged I20251105-1800 ...
> 
> Is a question missing maybe? 

No question implied, I just documented the observation I made before I saw your reply.

>  [...] But since then there was no successful deployment (!) so even though be build and tagged this Tycho can't use it.

[...]

PR #2202, opened 2026-01-21
Fix API tools slow down with multiple source folders sharing output

### LINKED ISSUE #2197: API tools slow down when building SWT for the last few months (regression)
See this video for an example:

https://github.com/eclipse-platform/eclipse.platform.swt/discussions/2885#discussioncomment-15316148

But basically I see `Analyzing API: Checking API used by [package]` repeating package names over and over again.

AFAICT this is a regression caused by 215a618a3c9fc64c5673bccf379d97b40f7d22b5 (which was a fixup for 02fa16884eee468e9b4c4aea96566488de4aea33)

By reverting 215a618a3c9fc64c5673bccf379d97b40f7d22b5 the SWT build runs quickly again in the IDE.

jonahgraham:
_This PR was originally https://github.com/eclipse-pde/eclipse.pde/pull/2198 but it seems to have disappeared (404) - the original PR had passed ECA_

The fix in commit 215a618 for issue #2096 introduced an O(n^2) time complexity regression when projects have multiple source folders sharing the same output location (like SWT with 26 source roots).

The problem was that each source folder caused a new CompositeApiTypeContainer to be created, wrapping all previous containers plus a new one. This resulted in exponential duplication of container visits during API analysis.

The fix modifies ProjectTypeContainer to support multiple IPackageFragmentRoots, allowing a single container to discover packages from all source folders that share the same output location. This avoids creating composite containers and maintains O(n) complexity.

Fixes #2197

### DIFF
--- a/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/compatibility/ProjectTypeContainerTests.java
+++ b/apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/builder/tests/compatibility/ProjectTypeContainerTests.java
@@ -261,7 +261,9 @@ public class ProjectTypeContainerTests extends CompatibilityTest {
 		IApiTypeContainer[] containers = component.getApiTypeContainers();
 		assertEquals("Wrong number of API type containers", 1, containers.length); //$NON-NLS-1$
 		IApiTypeContainer container = containers[0];
-		assertEquals("Should be a composite type container", IApiTypeContainer.COMPOSITE, //$NON-NLS-1$
+		// A single ProjectTypeContainer handles multiple source roots sharing
+		// the same output location by tracking all package fragment roots
+		assertEquals("Should be a folder type container", IApiTypeContainer.FOLDER, //$NON-NLS-1$
 				container.getContainerType());
 
 		String[] packageNames = container.getPackageNames();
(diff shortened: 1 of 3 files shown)


laeubi:
> However I am hoping that @laeubi can review this to make sure that the test itself still addresses the issue raised in https://github.com/eclipse-pde/eclipse.pde/issues/2096

The bad thing with API tools is that it is quite hard to test things without real examples. So my suggestion would be to try it out and maybe re-iterate on this as we are quite early in the release there is enough time to fix things as they are discovered so better sooner than later...

jonahgraham:
macOS test failed with:

Caused by: java.io.IOException: Server returned HTTP code: 400 for URL https://download.eclipse.org/eclipse/updates/4.38/R-4.38-202512010920/content.xml.xz

I raised https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7067 as I have seen this error in a few other builds in the last couple of days.

laeubi:
> I raised https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7067 as I have seen this error in a few other builds in the last couple of days.

See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6783 :-\

iloveeclipse:
Is it OK to merge now or not? Would be good to have it in M2 build.

PR #2203, opened 2026-01-21
Disallow IPackageFragmentRoot from being null within ProjectTypeContainers

jonahgraham:
In all current use cases of this internal class the value passed in was non-null anyway, so there is no functional change with this commit, it simply prevents future changes from trying to pass in a null here

An extra complication is that in 02fa16884eee468e9b4c4aea96566488de4aea33 the javadoc was updated to allow null IPackageFragmentRoot but the code did not match that.

Fixes https://github.com/eclipse-pde/eclipse.pde/pull/2202#discussion_r2713480707


jonahgraham:
> use cases of this internal class the value passed in was non-null anyway

This is because an NPE would be raised on the following line if `root` was null before the constructor or methods in `ProjectTypeContainer` are called  https://github.com/eclipse-pde/eclipse.pde/blob/19e78d8b432d2ff44fc9b9a0dc804c446f6928cd/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ProjectComponent.java#L391

Issue #1432, opened 2024-10-07
Display unit ids for IU Target locations


vogella:
Would be helpful to see the id in the overview tab of the target editor:

![image](https://github.com/user-attachments/assets/a6bfa269-78aa-4b3d-912a-4c61eeaa3201)

Otherwise there is not direct mapping of the displayed entry and the source tab entry.

![image](https://github.com/user-attachments/assets/1df17e5a-71e3-4b4c-b341-b273d841d86a)

laeubi:
> Otherwise there is not direct mapping of the displayed entry and the source tab entry.

Why should one need a direct mapping between source and UI?
The representation of the source is more driven by technical reasons where the UI is for making it more visual to the user, therefore it shows the name:

vogella:
I think the user who modifies the source tab content is the same user who works on the overview tab. The "Show location content" show the desired technical name but it also show the whole content so again it is not possible to see easily which entry maps to which in the UI to the source (especially as the sorting in the overview is not the same as in the source tab).

laeubi:
But why do this user looks at the UI and then heads over to the source to ... ?

e.g. I can edit / delete / ... things in the UI and in the source, why should I want to switch between those?

vogella:
>  why should I want to switch between those?

Activation / Reload is typical done via the Overview tab and if something goes wrong I see the error here. I typically then switch to the source tab to work on the issue. Having the information which ID is cause the issue would make the error resolution easier and consistent.

vogella:
> Maybe one should the better show error marked in the source tab to improve the situation.

That would another feature request, this one is about having the ID available in both main views. The ID could be additional information as the version, the name is also important.

laeubi:
I think the less disruptive option would be to simply have a tooltip that shows the (Unit) ID

vogella:
Tooltip would an improvement but still would make it harder to identify the real ID in case of lots of entries you would have to check one by one.

laeubi:
I think the Screenshot should go into the PR... if find this quite verbose and it might cut of the version part on smaller screens bit will leave this to @HannesWell to decide. 

Given that it then might be useful to show the name as well in the "Show Content" case to be consistent.

PR #2208, opened 2026-02-05
Always show Installable Unit (IU) ID in target editor

vogella:
Modified StyledBundleLabelProvider to always display the technical ID of Installable Units in the Target Platform Editor's Definition and Content tabs. If a descriptive name exists and differs from the ID, both are shown (Name [ID]), otherwise just the ID is displayed.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/1432


HannesWell (commented):
This is a difficult one.
On one hand I understand the desire to see the ID and it looks good.
On the other hand the concern from Christoph from https://github.com/eclipse-pde/eclipse.pde/issues/1432#issuecomment-3853274136 is also valid.

I have to think about this.
Do others have opinions?

merks:
Yes, on the downside it's very verbose. It could be shown in a hover instead...  It's also significant to note that the ID available directly in the source, whereas the name and the version are not.  But the association between the ID in the source versus the arbitrary name is also not shown anywhere currently...

PR #2210, opened 2026-02-05
add an editor for p2.inf

nburnwal09:
Added an editor for p2.inf in PDE along with auto-completion support
issue: #1811

https://github.com/user-attachments/assets/8b688e1b-748f-4d2a-9d24-cdb9c0a92aac

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/PDEProject.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/PDEProject.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2024 IBM Corporation and others.
+ * Copyright (c) 2010, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -258,6 +258,19 @@ public class PDEProject {
 	 */
 	public static IFolder getMetaInf(IProject project) {
 		return getBundleRelativeFolder(project, IPath.fromOSString(ICoreConstants.MANIFEST_FOLDER_NAME));
+	}
+
+	/**
+	 * Returns the resource in the specified project corresponding to its
+	 * <code>p2.inf</code> file (for feature projects). Feature projects have
+	 * the p2.inf file in the project root, not in META-INF.
+	 *
+	 * @param project
+	 *            project
+	 * @return <code>p2.inf</code> file that may or may not exist
+	 */
+	public static IFile getFeatureP2Inf(IProject project) {
+		return getBundleRelativeFile(project, ICoreConstants.P2_INF_FEATURE_PATH);
 	}
 
 	/**
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -52,6 +52,8 @@ import org.eclipse.pde.internal.ui.editor.build.BuildPage;
 import org.eclipse.pde.internal.ui.editor.build.BuildSourcePage;
 import org.eclipse.pde.internal.ui.editor.context.InputContext;
 import org.eclipse.pde.internal.ui.editor.context.InputContextManager;
+import org.eclipse.pde.internal.ui.editor.p2inf.P2InfInputContext;
+import org.eclipse.pde.internal.ui.editor.p2inf.P2InfSourcePage;
 import org.eclipse.swt.SWTError;
 import org.eclipse.swt.dnd.RTFTransfer;
 import org.eclipse.swt.dnd.TextTransfer;
@@ -172,8 +174,14 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 			FileEditorInput in = new FileEditorInput(buildFile);
 			manager.putContext(in, new BuildInputContext(this, in, file == buildFile));
 		}
+		IFile p2InfFile = PDEProject.getFeatureP2Inf(project);
+		if (p2InfFile != null && p2InfFile.exists()) {
+			FileEditorInput in = new FileEditorInput(p2InfFile);
+			manager.putContext(in, new P2InfInputContext(this, in, false));
+		}
 		manager.monitorFile(featureFile);
 		manager.monitorFile(buildFile);
+		manager.monitorFile(p2InfFile);
 	}
 
 	@Override
@@ -199,6 +207,11 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 			if (!fInputContextManager.hasContext(BuildInputContext.CONTEXT_ID)) {
 				IEditorInput in = new FileEditorInput(file);
 				fInputContextManager.putContext(in, new BuildInputContext(this, in, false));
+			}
+		} else if (name.equals(ICoreConstants.P2_INF_FILENAME)) {
+			if (!fInputContextManager.hasContext(P2InfInputContext.CONTEXT_ID)) {
+				IEditorInput in = new FileEditorInput(file);
+				fInputContextManager.putContext(in, new P2InfInputContext(this, in, false));
 			}
 		}
 	}
@@ -254,6 +267,12 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 				IEditorInput in = new FileStoreEditorInput(store);
 				manager.putContext(in, new BuildInputContext(this, in, file == buildFile));
 			}
+			File p2InfFile = new File(file.getParentFile(), ICoreConstants.P2_INF_FILENAME);
+			if (p2InfFile.exists()) {
+				IFileStore store = EFS.getStore(p2InfFile.toURI());
+				IEditorInput in = new FileStoreEditorInput(store);
+				manager.putContext(in, new P2InfInputContext(this, in, false));
+			}
 		} catch (CoreException e) {
 			PDEPlugin.logException(e);
 		}
@@ -298,6 +317,7 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 		}
 		addSourcePage(FeatureInputContext.CONTEXT_ID);
 		addSourcePage(BuildInputContext.CONTEXT_ID);
+		addSourcePage(P2InfInputContext.CONTEXT_ID);
 	}
 
 	@Override
@@ -322,6 +342,9 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 		}
 		if (contextId.equals(BuildInputContext.CONTEXT_ID)) {
 			return new BuildSourcePage(editor, title, name);
+		}
+		if (contextId.equals(P2InfInputContext.CONTEXT_ID)) {
+			return new P2InfSourcePage(editor, contextId, title);
 		}
 		return super.createSourcePage(editor, title, name, contextId);
 	}
@@ -426,9 +449,17 @@ public class FeatureEditor extends MultiSourceEditor implements IShowEditorInput
 	public void showEditorInput(IEditorInput editorInput) {
 		String name = editorInput.getName();
 		if (name.equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR)) {
-			setActivePage(0);
-		} else {
-			setActivePage(getPageCount() - 3);
+			setActivePage(FeatureFormPage.PAGE_ID);
+		} else if (name.equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+			IFormPage uiPage = findPage(BuildPage.PAGE_ID);
+			if (uiPage != null) {
+				setActivePage(uiPage.getId());
+			}
+		} else if (name.equals(ICoreConstants.P2_INF_FILENAME)) {
+			IFormPage page = findPage(P2InfInputContext.CONTEXT_ID);
+			if (page != null) {
+				setActivePage(page.getId());
+			}
 		}
 	}
 
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2005, 2015 IBM Corporation and others.
+ *  Copyright (c) 2005, 2026 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -45,11 +45,18 @@ public class FeatureEditorMatchingStrategy implements IEditorMatchingStrategy {
 			// build.properties matches with editors that have a feature.xml file
 			// as their input and that feature.xml is at the root
 			if (inputFile.getName().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR)) {
-				if (currInputFile.getName().equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+				if (currInputFile.getName().equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)
+						|| currInputFile.getName().equals(ICoreConstants.P2_INF_FILENAME)) {
 					return inputFile.getProjectRelativePath().toString().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
 				}
 				return inputFile.equals(currInputFile);
 			} else if (inputFile.getName().equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+				if (currInputFile.getName().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR))
+					return currInputFile.getProjectRelativePath().toString().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
+				return inputFile.equals(currInputFile);
+			} else if (inputFile.getName().equals(ICoreConstants.P2_INF_FILENAME)) {
+				// p2.inf should match the feature editor when the current editor
+				// input is the feature.xml at the project root
 				if (currInputFile.getName().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR)) {
 					return currInputFile.getProjectRelativePath().toString().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
 				}
@@ -60,5 +67,4 @@ public class FeatureEditorMatchingStrategy implements IEditorMatchingStrategy {
 			return false;
 		}
 	}
-
 }
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
@@ -1,0 +1,319 @@
+/*******************************************************************************
+ * Copyright (c) 2026 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.ui.editor.p2inf;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.contentassist.ICompletionProposal;
+import org.eclipse.pde.internal.ui.PDEPlugin;
+import org.eclipse.pde.internal.ui.editor.PDESourcePage;
+import org.eclipse.pde.internal.ui.editor.contentassist.TypeCompletionProposal;
+import org.eclipse.pde.internal.ui.editor.contentassist.TypePackageCompletionProcessor;
+
+/**
+ * Content-assist processor for {@code p2.inf} files. Suggestions are driven by
+ * a statically built key-segment tree that mirrors the p2.inf property-file
+ * grammar
+ */
+public class P2InfContentAssistProcessor extends TypePackageCompletionProcessor {
+
+	private static final String KEY_PROVIDES = "provides"; //$NON-NLS-1$
+	private static final String KEY_REQUIRES = "requires"; //$NON-NLS-1$
+	private static final String KEY_META_REQUIREMENTS = "metaRequirements"; //$NON-NLS-1$
+	private static final String KEY_PROPERTIES = "properties"; //$NON-NLS-1$
+	private static final String KEY_UPDATE = "update"; //$NON-NLS-1$
+	private static final String KEY_INSTRUCTIONS = "instructions"; //$NON-NLS-1$
+	private static final String KEY_IU = "iu"; //$NON-NLS-1$
+	private static final String KEY_UNITS = "units"; //$NON-NLS-1$
+	private static final String KEY_ARTIFACTS = "artifacts"; //$NON-NLS-1$
+	private static final String KEY_CLASSIFIER = "classifier"; //$NON-NLS-1$
+	private static final String KEY_COPYRIGHT = "copyright"; //$NON-NLS-1$
+	private static final String KEY_DESCRIPTION = "description"; //$NON-NLS-1$
+	private static final String KEY_FILTER = "filter"; //$NON-NLS-1$
+	private static final String KEY_GREEDY = "greedy"; //$NON-NLS-1$
+	private static final String KEY_HOST_REQUIREMENTS = "hostRequirements"; //$NON-NLS-1$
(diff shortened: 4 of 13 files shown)


merks (commented):
It looks generally okay but I don't have time for a deeper review.

nburnwal09:
Hi @noopur2507 
Could you please review this PR? For convenience, I have merged the auto-completion changes from [PR #2261](https://github.com/eclipse-pde/eclipse.pde/pull/2261) into this PR so that everything can be reviewed together.

vogella:
@nburnwal09 can you rebase and fix the 2 Maven issues if they are relevant? Afterwards I can review early next week.

nburnwal09:
> @nburnwal09 can you rebase and fix the 2 Maven issues if they are relevant? Afterwards I can review early next week.

@vogella 
Sure will do it. Thanks for addressing.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+import org.eclipse.pde.internal.ui.editor.text.ArgumentRule;
+import org.eclipse.pde.internal.ui.editor.text.BasePDEScanner;
+import org.eclipse.pde.internal.ui.editor.text.ChangeAwareSourceViewerConfiguration;
+import org.eclipse.pde.internal.ui.editor.text.IColorManager;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+
+public class P2infViewerConfiguration extends ChangeAwareSourceViewerConfiguration {
+
+	protected static String PROPERTIES_FILE_PARTITIONING = "___pf_partitioning"; //$NON-NLS-1$
```
vogella: Looks like a copy of BuildSourceViewerConfiguration. Can you share common code between the two?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+	protected static String COMMENT = "__pf_comment"; //$NON-NLS-1$
+	protected static String PROPERTY_VALUE = "__pf_roperty_value"; //$NON-NLS-1$
```
vogella: Looks like a typo (p missing)

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
```diff
@@ -0,0 +1,173 @@
...
+					proposalText = value + node.key;
+				} else {
+					proposalText = value + node.key + "."; //$NON-NLS-1$
+				}
+				completions
+						.add(new TypeCompletionProposal(proposalText, null, proposalText, lineStart, value.length()));
+			}
+			return completions.toArray(ICompletionProposal[]::new);
+		} catch (BadLocationException e) {
+			e.printStackTrace();
```
vogella: Should this be better logged via ILog.of(..).error()?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2infViewerConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+import org.eclipse.pde.internal.ui.PDEPluginImages;
+import org.eclipse.pde.internal.ui.editor.PDESourcePage;
+import org.eclipse.pde.internal.ui.editor.text.ArgumentRule;
+import org.eclipse.pde.internal.ui.editor.text.BasePDEScanner;
+import org.eclipse.pde.internal.ui.editor.text.ChangeAwareSourceViewerConfiguration;
+import org.eclipse.pde.internal.ui.editor.text.IColorManager;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+
+public class P2infViewerConfiguration extends ChangeAwareSourceViewerConfiguration {
```
vogella: P2infViewerConfiguration -> P2InfViewerConfiguration (capital I like in the other classes)

vogella (commented):
Not yet tested, just some initial feedback.

vogella:
> @nburnwal09 can you rebase and fix the 2 Maven issues if they are relevant? Afterwards I can review early next week.

I see no changes and my "early this week" time frame is over. If it is still open I may be able to review after the summer vacation in ~ 3-4 weeks.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
nburnwal09: Agreed. I kept it separate because p2.inf needs its own content assist behavior. Even though most of the viewer configuration is currently very similar to build.properties.  
For this change, I kept the new editor behavior isolated and avoid a larger refactoring in the same PR.

A good follow-up would be to extract the shared “properties-like source viewer configuration” into a common base class, with BuildSourceViewerConfiguration and P2InfViewerConfiguration as concrete subclasses.

nburnwal09:
> I see no changes and my "early this week" time frame is over. If it is still open I may be able to review after the summer vacation in ~ 3-4 weeks.

Sure please review whenever you are back
Sorry I missed to address these comments. I am updating it now.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
nburnwal09: Yes, but this value is effectively used as an internal partition identifier, so changing it will brake the existing presentation/coloring wiring.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
nburnwal09: Updated to `PDEPlugin.log(e)`, which internally logs via `ILog.get().error()`

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2infViewerConfiguration.java
nburnwal09: Done

merks:
@nburnwal09 

I can spare some time for reviews to keep this moving forward for the current release cycle.  Is it ready for a review at this point?

nburnwal09:
> @nburnwal09
> 
> I can spare some time for reviews to keep this moving forward for the current release cycle. Is it ready for a review at this point?

@merks 
Thank you. Yes, it's ready for review!
Just one thing here: the content assist is functional and context-aware but is in the initial phase. Suggestions are not filtered based on the letters being getting typed. I will address that part in a follow-up PR.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
```diff
@@ -426,9 +449,18 @@ protected boolean isPatchEditor() {
...
-		} else {
-			setActivePage(getPageCount() - 3);
+			setActivePage(FeatureFormPage.PAGE_ID);
+		} else if (name.equalsIgnoreCase(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+			IFormPage uiPage = findPage(BuildPage.PAGE_ID);
+			if (uiPage != null) {
+				setActivePage(uiPage.getId());
+			}
+
+		} else if (name.equalsIgnoreCase(ICoreConstants.P2_INF_FILENAME)) {
```
merks: I wonder about the ignore case here.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
```diff
@@ -426,9 +449,18 @@ protected boolean isPatchEditor() {
 	public void showEditorInput(IEditorInput editorInput) {
 		String name = editorInput.getName();
 		if (name.equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR)) {
-			setActivePage(0);
-		} else {
-			setActivePage(getPageCount() - 3);
+			setActivePage(FeatureFormPage.PAGE_ID);
+		} else if (name.equalsIgnoreCase(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
```
merks: I'm not sure we should be allow any old case.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
```diff
@@ -45,11 +45,18 @@ public boolean matches(IEditorReference editorRef, IEditorInput input) {
 			// build.properties matches with editors that have a feature.xml file
 			// as their input and that feature.xml is at the root
 			if (inputFile.getName().equals(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR)) {
-				if (currInputFile.getName().equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+				if (currInputFile.getName().equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)
```
merks, line 48: Here it's not case sensitive...

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
```diff
@@ -0,0 +1,174 @@
...
+			Messages.P2InfHeader_greedy, Messages.P2InfHeader_optional, Messages.P2InfHeader_multiple,
+			Messages.P2InfHeader_filter };
+
+	public P2InfContentAssistProcessor(PDESourcePage sourcePage) {
+		fSourcePage = sourcePage;
+		buildSuggestionTree();
+	}
+
+	private static class SuggestionNode {
+		String key;
```
merks: Is it intentional that everything is package protected?  Maybe the fields shoudld be provide and the methods should be public?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	private List<SuggestionNode> getSuggestions(String line) {
+		if (line.isEmpty()) {
+			return root.children;
+		}
+		String[] tokens = line.split("\\."); //$NON-NLS-1$
+		SuggestionNode current = root;
+		for (String token : tokens) {
+			if (token.isEmpty())
```
merks: Personally I prefer that there always be braces  for nested content.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfInputContext.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	@Override
+	public String getId() {
+		return CONTEXT_ID;
+	}
+
+	@Override
+	protected void addTextEditOperation(ArrayList<TextEdit> ops, IModelChangedEvent event) {
+
```
merks: The blank line is kind of pointless.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+
+		@Override
+		protected void initialize() {
+			setDefaultReturnToken(new Token(createTextAttribute(fProperty)));
+		}
+	}
+
+	public class PropertyValueScanner extends AbstractJavaScanner {
+
+		public class AssignmentDetector implements IWordDetector {
```
merks: I didn't check if these nested classes could be static...  But maybe...

merks (commented):
It mostly looks fine.  I have a few comments and questions.  I'll want to test it, but will try to find time for that latter.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
nburnwal09: Updated both equalsIgnoreCase calls to equals()

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
nburnwal09: updated

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
nburnwal09: Yes, you are right. I have updated the same in respective places to equals()

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfContentAssistProcessor.java
nburnwal09: Updated

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/P2InfViewerConfiguration.java
nburnwal09: the nested class `AssignmentDetector` accesses fDocument and fOffset, which are protected instance fields from JFace's RuleBasedScanner. If made static, these fields will become unreachable
It was referenced from BuildSourceViewerConfiguration which has the same pattern.

merks: Makes sense then.

merks:
It's cool!

image

image

I don't think this is correct though:

image

This isn't either:

image

Before I look at the code, can you make sure that for each case, it shows only, the extension values that are applicable for the given `xyz.n.` prefix?

nburnwal09:
> It's cool!
> 
> image image
> I don't think this is correct though:
> 
> image
> This isn't either:
> 
> image
> Before I look at the code, can you make sure that for each case, it shows only, the extension values that are applicable for the given `xyz.n.` prefix?

I have taken reference of the pattern from here: https://eclipse.dev/eclipse/markdown/?file=eclipse-equinox/p2/master/docs/Customizing_Metadata.md

Yes, I am seeing properties one is incorrect but others are correct as per this page. I will check and correct it.

merks:
Yes, best to double check them all because units is seemed incomplete (unless I clipped it short):

image

nburnwal09:
@merks 
I have corrected the suggestions. Here is the samples that I tested it considering [here](https://eclipse.dev/eclipse/markdown/?file=eclipse-equinox/p2/master/docs/Customizing_Metadata.md)
Kindly verify it whenever you get sometime. 

| Prefix | Expected Suggestions |
|--------|----------------------|
| `<root>` | `properties`, `provides`, `requires`, `metaRequirements`, `update`, `instructions`, `iu`, `units` |
| `properties.<index>.` | `name`, `value` |
| `provides.<index>.` | `namespace`, `name`, `version` |
| `requires.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple`, `filter`, `min`, `max` |
| `metaRequirements.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple` |
| `update.` | `id`, `range`, `severity`, `description` |
| `instructions.` | `install`, `install.import`, `configure`, `configure.import`, `unconfigure`, `unconfigure.import`, `uninstall`, `uninstall.import` |
| `iu.<index>.` | `id`, `version`, `singleton`, `copyright`, `licenses`, `filter`, `touchpoint`, `update`, `artifacts`, `properties`, `provides`, `requires`, `metaRequirements`, `hostRequirements`, `instructions` |
| `iu.<index>.copyright.` | `location` |
| `iu.<index>.licenses.<index>.` | `location` |
| `iu.<index>.touchpoint.` | `id`, `version` |
| `iu.<index>.update.` | `match`, `id`, `range`, `severity`, `description` |
| `iu.<index>.artifacts.<index>.` | `id`, `version`, `classifier` |
| `iu.<index>.properties.<index>.` | `name`, `value` |
| `iu.<index>.provides.<index>.` | `namespace`, `name`, `version` |
| `iu.<index>.requires.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple`, `filter`, `min`, `max` |
| `iu.<index>.metaRequirements.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple` |
| `iu.<index>.hostRequirements.<index>.` | `namespace`, `name`, `range`, `greedy`, `optional`, `multiple` |
| `iu.<index>.instructions.` | `install`, `install.import`, `configure`, `configure.import`, `unconfigure`, `unconfigure.import`, `uninstall`, `uninstall.import` |
| `units.<index>.` | `id`, `version`, `singleton`, `copyright`, `licenses`, `filter`, `touchpoint`, `update`, `artifacts`, `properties`, `provides`, `requires`, `metaRequirements`, `hostRequirements`, `instructions` |
| `units.<index>.copyright.` | `location` |
| `units.<index>.licenses.<index>.` | `location` |
| `units.<index>.touchpoint.` | `id`, `version` |
| `units.<index>.update.` | `match`, `id`, `range`, `severity`, `description` |
| `units.<index>.artifacts.<index>.` | `id`, `version`, `classifier` |
| `units.<index>.properties.<index>.` | `name`, `value` |
| `units.<index>.provides.<index>.` | `namespace`, `name`, `version` |
| `units.<index>.requires.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple`, `filter`, `min`, `max` |
| `units.<index>.metaRequirements.<index>.` | `namespace`, `name`, `range`, `matchExp`, `greedy`, `optional`, `multiple` |
| `units.<index>.hostRequirements.<index>.` | `namespace`, `name`, `range`, `greedy`, `optional`, `multiple` |
| `units.<index>.instructions.` | `install`, `install.import`, `configure`, `configure.import`, `unconfigure`, `unconfigure.import`, `uninstall`, `uninstall.import` |

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
```diff
@@ -200,6 +208,11 @@ public void monitoredFileAdded(IFile file) {
 				IEditorInput in = new FileEditorInput(file);
 				fInputContextManager.putContext(in, new BuildInputContext(this, in, false));
 			}
+		} else if (name.equalsIgnoreCase(ICoreConstants.P2_INF_FILENAME)) {
```
merks: This should not ignore case.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditor.java
```diff
@@ -426,9 +449,18 @@ protected boolean isPatchEditor() {
...
-			setActivePage(0);
-		} else {
-			setActivePage(getPageCount() - 3);
+			setActivePage(FeatureFormPage.PAGE_ID);
+		} else if (name.equals(ICoreConstants.BUILD_FILENAME_DESCRIPTOR)) {
+			IFormPage uiPage = findPage(BuildPage.PAGE_ID);
+			if (uiPage != null) {
+				setActivePage(uiPage.getId());
+			}
+
```
merks: This blank like is pointless.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/messages.properties
```diff
@@ -0,0 +1,43 @@
+###############################################################################
```
merks: I think this entire design for these string to be translated is completely questionable.  These cannot and should not be translated because they're part of the p2 inf model and not merely textual content to be translated.

merks (changes requested):
I should have noticed previous the use of translation strings for what should be just be "keywords" of the p2.inf "language".

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/p2inf/messages.properties
nburnwal09: @merks 
Sorry for the delayed response. I couldn't exactly understand this then. Now I do.
yes, these are p2 model keys — translating them might break p2.inf files. What's the preferred way to define them for reuse in the content assist? Plain static final String constants in the processor, or a separate constants class?
What do you suggest?

merks: I think they are only used in P2InfContentAssistProcessor  so could be defined there.  Whatever is most convenient is fine...

nburnwal09: @merks 
I have updated and defined the keys in P2InfContentAssistProcessor, verified the changes as per above list.
Can you please review it?

merks:
That's much better.  

Some nits though:

image

image


The logic could do a better job determining if there is a trailing `.` or not and adjust the proposal accordingly. 

In the end though, this is super helpful and convenient!!

nburnwal09:
> That's much better.
> 
> Some nits though:
> 
> image image
> The logic could do a better job determining if there is a trailing `.` or not and adjust the proposal accordingly.
> 
> In the end though, this is super helpful and convenient!!

@merks 
Yeah sure made the change. Kindly review

vogella:
P2InfViewerConfiguration is still a verbatim copy of BuildSourceViewerConfiguration. This should be a shared base class (or BuildSourceViewerConfiguration parameterized with a content-assist processor), not ~290 duplicated lines that will drift.

ManifestEditor. monitoredFileAdded switched from equalsIgnoreCase to equals which looks wrong.

Also PDEProject.getBundleP2Inf(...) is AFAICS never used.

merks:
> ManifestEditor. monitoredFileAdded switched from equalsIgnoreCase to equals which looks wrong.

Why should any of the files be case insensitive?

vogella:
> > ManifestEditor. monitoredFileAdded switched from equalsIgnoreCase to equals which looks wrong.
> 
> Why should any of the files be case insensitive?

The point is that this should not be part of a p2 editor change, if it makes sense this should be moved to a separate clean-up.

merks:
> The point is that this should not be part of a p2 editor change, if it makes sense this should be moved to a separate clean-up.

Maybe we could let is slide in this case because when adding new stuff that's inconsistent with old questionable case-insensitive comparisons,  the result is kind of is bogus inconsistent mishmash.  

It would seem better to focus on substantial things like good behavior, and avoiding to copy code that could be refactored for reuse as you mentioned for P2InfViewerConfiguration.

vogella:
> Maybe we could let is slide in this case because when adding new stuff that's inconsistent with old questionable case-insensitive comparisons, the result is kind of is bogus inconsistent mishmash.

I leave the decision to @nburnwal09 and you, my feedback is not intended to block progress here.

nburnwal09:
@merks @vogella 

I have updated the following
1. Extracted protected `createContentAssistProcessor()` factory method in `BuildSourceViewerConfiguration` so subclasses can override just the processor, keeping everything else shared.
2. Extended `BuildSourceViewerConfiguration` in `P2InfViewerConfiguration` and replaced the entire 290-line duplicate with a 20-line subclass
3. Removed `getBundleP2Inf` in `PDEProject` as it is never used.

Kindly review the changes.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/project/PDEProject.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2024 IBM Corporation and others.
+ * Copyright (c) 2010, 2024, 2026 IBM Corporation and others.
```
merks: I think one is just supposed to change the upper bound not keep adding more years.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2005, 2015 IBM Corporation and others.
+ *  Copyright (c) 2005, 2015, 2026 IBM Corporation and others.
```
merks: Same here.  Please check what other files do, but I think we don't want a list here that can grow and grow...

merks (approved):
It looks good.  Just  the copyright date updates I think are not done properly.

merks:
When I get back later, I will test it in my IDE.  But it looks good so far.  Thanks for all the updates!!

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/feature/FeatureEditorMatchingStrategy.java
nburnwal09: Made the changes.

merks:
It's cool.  Let's land this.  Any further improvements can be done later.

I assume you plan to write an N&N entry!

nburnwal09:
> I assume you plan to write an N&N entry!

Yeah, was eager to do that :D

PR #2213, opened 2026-02-11
Add missing dependency for PDE Runtime to PDE Core

HeikoKlare:
PDE Runtime has a transitive dependency to PDE Core via PDE UI. However, that dependency is optional and some relevant functionality (like the Plug-in Selection Spy) work without the PDE UI dependency but require PDE Core. In an installation that contains PDE Core but not PDE UI, the Plug-in Selection Spy will not work as the PDE Core dependency is not resolved (as it needs to be resolved transitively via the non present PDE UI dependency), thus leading to NoClassDefFoundErrors for required classes from PDE Core:
```
[2026/02/11 10:42:47.417 D060556 #000000 Heap:8192 1688 1225 Meta:626] "Equinox Log Thread - Equinox Container: 50ae596c-3e11-46e5-8b16-686223814ada,6,main" java.lang.NoClassDefFoundError: org/eclipse/pde/internal/core/util/ManifestUtils
	at org.eclipse.pde.internal.runtime.spy.SpyFormToolkit.generatePluginDetailsText(SpyFormToolkit.java:288)
	at org.eclipse.pde.internal.runtime.spy.sections.ActivePartSection.build(ActivePartSection.java:117)
	at org.eclipse.pde.internal.runtime.spy.dialogs.SpyDialog.createDialogArea(SpyDialog.java:117)
	at org.eclipse.pde.internal.runtime.spy.dialogs.SpyDialog.createContents(SpyDialog.java:86)
```

This change adds the missing dependency. With the change, you will be able to use the Plug-in Selection Spy from PDE Runtime without having PDE UI present in your installation.

### DIFF
--- a/ui/org.eclipse.pde.runtime/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.runtime/META-INF/MANIFEST.MF
@@ -14,6 +14,7 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.jdt.core;bundle-version="[3.36.0,4.0.0)";resolution:=optional,
  org.eclipse.core.resources;bundle-version="[3.20.0,4.0.0)";resolution:=optional,
  org.eclipse.jdt.ui;bundle-version="[3.31.0,4.0.0)";resolution:=optional,
+ org.eclipse.pde.core;bundle-version="[3.21.200,4.0.0)";resolution:=optional,
  org.eclipse.pde.ui;bundle-version="[3.14.200,4.0.0)";resolution:=optional,
  org.eclipse.help;bundle-version="[3.10.200,4.0.0)";resolution:=optional
 Export-Package: org.eclipse.pde.internal.runtime;x-internal:=true,


HeikoKlare:
> This looks wrong here. Either it is required or not, having it optional seem completely of in this case and the CNF proofs that the code actually is not handling this correctly.

I am not sure about the original intent of having these dependencies optional (it seems to be like that for decades). I guess that a reason (or at least a benefit) is that there are parts of the functionality of the bundle that you can use without those optional dependencies (as the one explained in the PR description). It would definitely be better to have to properly split up into different bundles that have proper dependencies. Making the bundle gracefully handle missing dependencies is probably quite complicated as all PDE dependencies are optional, so most code would need to deal with missing dependencies.
Maybe this becomes less a problem if the added dependency to PDE Core is not made optional.

HannesWell:
I think we all agree that the current design is far from optimal and ideally should be refactored (although I'm not yet sure how the target should look like as pde.runtime seem to be a bag of a lot of stuff). But that will probably a greater effort and some existing users probably won't be happy about it if it's split up.
Therefore adding `org.eclipse.pde.core` as optional dependency looks like a pragmatic solution and a good compromise, given the current situation:
1. `o.e.pde.runtime` requires `o.e.pde.ui` optionally.
2. `o.e.pde.ui` requires and **re-exports** `o.e.pde.core`

So pde.runtime already requires pde.core optionally transitively. Repeating that again won't change any existing use-case and will allow to use it without pde.ui but pde.core present. Without the former one loses only the ability to open a Manifest in an Editor as far as I can tell (one can simply remove the pde.ui dependency).

Making the dependency to pde.core mandatory could probably increase the dependency closure of pde.runtime more than existing users maybe want. 

Probably it would be good as an immediate action to refactor the two calls to `org.eclipse.pde.internal.ui.editor.plugin.ManifestEditor` into a separate class to ensure the rest of the current callers can work properly without pde.ui and already fail at class loading.

The same should be done for the call to `ManifestUtils.getSourceReferences()`, which seems to be the only class used from pde.core.
But since there is no other user of that method (and it was only introduced recently, we could simply move that into pde.runtime. Then the dependency for pde.core shouldn't be necessary at all.


> > Why I'm mentioning that is, that resolves usually do _as much as possible_ to include optional dependencies and only oit them if there is a conflict. SO you problem might be a different one here and adding another optional dependency not help much.
> 
> My understanding is definitely more "naive" here, but at a high level it is as follows. Let's assume I have PDE Core and PDE Runtime available. Currently, the resolver will for PDE Runtime not consider PDE Core at all, because there is no direct dependency to PDE Core and it cannot know about the transitive dependency via PDE UI because PDE UI is not present at all. With the PDE Core dependency explicitly being added, the resolver will properly consider it. It should not be relevant whether it's declared optional or not as it should be possible to include it without any conflicts anyway.

That seems pretty accurate. With the situation as described above, the resolvers behavior should be correct.

laeubi:
> My understanding is definitely more "naive" here, but at a high level it is as follows. Let's assume I have PDE Core and PDE Runtime available. Currently, the resolver will for PDE Runtime not consider PDE Core at all, because there is no direct dependency to PDE Core and it cannot know about the transitive dependency via PDE UI because PDE UI is not present at all.

I see, the issue is now (as also highlighted by @HannesWell ) that `o.e.pde.ui` requires **and re-exports** `o.e.pde.core` so we are using reexported packages through an optional dependency chain here!

I therefore I agree that

> adding `org.eclipse.pde.core` as optional dependency looks like a pragmatic solution and a good compromise

is the right thing to do, **BUT** we should either add a bug (unless you plan to fix it immediately), that PDE gives an ERROR (ManifestChecker seems a good place for this) if:

1. A bundle is required with `optional`
2. This bundle has `reexports`
3. And this reexported dependencies are not _also required_ by this bundle requiring the other one

Because it can obviously not work to use a reexport through an optional dependency chain and this is really nasty and hard to track.

HeikoKlare:
> Making the dependency to pde.core mandatory could probably increase the dependency closure of pde.runtime more than existing users maybe want.

Agreed, I have changed the dependency back to be optional. With that, this PR currently does not change anything in terms of potentially pulling in more dependencies but just ensures that some more functionality works without PDE UI being present.

> Probably it would be good as an immediate action to refactor the two calls to `org.eclipse.pde.internal.ui.editor.plugin.ManifestEditor` into a separate class to ensure the rest of the current callers can work properly without pde.ui and already fail at class loading.

If I understand correctly, you want to avoid that other functionality does not work because there is some access to `ManifestEditor` in the same class, is that correct? But won't the JIT just compile methods that are actually called? I tested with executing code in the classes that refer to `ManifestEditor` without PDE UI being present and it works fine, unless I actually call one of the referencing methods. But those methods only consist of a call to `ManifestEditor` anyway, so you cannot avoid that they will fail with a NoClassDefFoundError when missing PDE UI.
So I wonder if moving the calls to `ManifestEditor` into a separate class really brings a benefit.

> The same should be done for the call to `ManifestUtils.getSourceReferences()`, which seems to be the only class used from pde.core. But since there is no other user of that method (and it was only introduced recently, we could simply move that into pde.runtime. Then the dependency for pde.core shouldn't be necessary at all.

At least the classes `PDECore`, `SearchablePluginsManager`, `IPluginModelBase` and `PluginRegistry` from PDE Core are also used, so just moving `ManifestUtils.getSourceReferences()` to PDE Runtime will not make it completely get rid of the PDE Core dependency. And actually the `getSourceReferences()` functionality is well-placed in `ManifestUtils` (and also accesses constants of PDE Core), isn't it? Still I can provide a PR for moving that method if you want.

Issue #2215, opened 2026-02-18
Disable DetectVMInstallationsJob for JUnit Plug-in Test launches


trancexpress:
The JDT debug job `DetectVMInstallationsJob` is disabled when tests run in CI jobs. However, its not disabled when running tests from Eclipse. As a result, tests behave differently between the CI environment and executing from Eclipse.

For an example, see: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/864#issuecomment-3920991481

PR #2216, opened 2026-02-18
Disable DetectVMInstallationsJob for JUnit Plug-in Test launches

trancexpress:
This change disables the job `org.eclipse.jdt.internal.launching.DetectVMInstallationsJob`, via the respective system property, for JUnit Plug-in Test launches.

This is done to reduce differences between the CI environment and launching from Eclipse, simplifying debugging of tests failing in the CI environment.

Fixes: #2215

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitWorkbenchLaunchShortcut.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitWorkbenchLaunchShortcut.java
@@ -46,6 +46,8 @@ import org.eclipse.pde.launching.PDESourcePathProvider;
  */
 public class JUnitWorkbenchLaunchShortcut extends JUnitLaunchShortcut {
 
+	private static final String DETECT_VM_INSTALLATIONS_DISABLED = "-DDetectVMInstallationsJob.disabled"; //$NON-NLS-1$
+
 	@Override
 	protected String getLaunchConfigurationTypeId() {
 		return "org.eclipse.pde.ui.JunitLaunchConfig"; //$NON-NLS-1$
@@ -112,9 +114,13 @@ public class JUnitWorkbenchLaunchShortcut extends JUnitLaunchShortcut {
 
 		// VM arguments
 		String vmArgs = LaunchArgumentsHelper.getInitialVMArguments();
-		if (vmArgs.length() > 0) {
-			configuration.setAttribute(IJavaLaunchConfigurationConstants.ATTR_VM_ARGUMENTS, vmArgs);
+		if (!vmArgs.contains(DETECT_VM_INSTALLATIONS_DISABLED)) {
+			if (!vmArgs.isEmpty()) {
+				vmArgs += " "; //$NON-NLS-1$
+			}
+			vmArgs += DETECT_VM_INSTALLATIONS_DISABLED + "=true"; //$NON-NLS-1$
 		}
+		configuration.setAttribute(IJavaLaunchConfigurationConstants.ATTR_VM_ARGUMENTS, vmArgs);
 		configuration.setAttribute(IPDELauncherConstants.BOOTSTRAP_ENTRIES, ""); //$NON-NLS-1$
 
 		// configuration attributes


trancexpress:
@iloveeclipse can you request a Copilot review? I'm curious about the feedback, but I'm not a committer and so I cannot touch the reviewers section.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/ui/launcher/JUnitWorkbenchLaunchShortcut.java
```diff
@@ -46,6 +46,9 @@
  */
 public class JUnitWorkbenchLaunchShortcut extends JUnitLaunchShortcut {
 
+	private static final String DETECT_VM_INSTALLATIONS_DISABLED_PROPERTY = "-DDetectVMInstallationsJob.disabled"; //$NON-NLS-1$
+	private static final String DETECT_VM_INSTALLATIONS_DISABLED = DETECT_VM_INSTALLATIONS_DISABLED_PROPERTY + "=true"; //$NON-NLS-1$
```
HannesWell: Not sure if this separate constant is necessary or not.
From my POV this could also be inlined.

laeubi (changes requested):
I don't think this change is good in its current form.

> This is done to reduce differences between the CI environment and launching from Eclipse

The CI environment is detected by the `DetectVMInstallationsJob` itself and I strongly believe it should possibly detect junit launches as well instead of expect it to happen by the launcher.

Not every Plugin Launch is launch an Eclipse IDE nor does it requires JDT to be on the classpath at all. Also there might be other launcher scenarios where this is then not picked up at all. Even worse, if we bake in project specific properties into PDE the become de-facto API and its very hard to change.

Also how many of these properties that might influence a launch should we include? What projects are eligible for such deep integration with PDE?

Just as some ideas for the `DetectVMInstallationsJob`:

1. It could look for the program arguments, a usual launch include these where I see multiple that could be used e.g. `testLoaderClass`, `application`, `testpluginname`:
`Command-line arguments:  -os linux -ws gtk -arch x86_64 -consoleLog -version 3 -port 36681 -testLoaderClass org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader -loaderpluginname org.eclipse.jdt.junit4.runtime -classNames org.eclipse.jdt.core.tests.builder.MultiReleaseTests -application org.eclipse.pde.junit.runtime.uitestapplication -product org.eclipse.sdk.ide -data ... -dev ... -os linux -ws gtk -arch x86_64 -consoleLog -testpluginname org.eclipse.jdt.core.tests.builder`
2.  Looking at the stacktrace of the current thread, it will include `NonUIThreadTestApplication` or `UITestApplication` (possibly not only when launched from PDE!)
3. Even the System properties already contain something like `eclipse.application=org.eclipse.pde.junit.runtime.uitestapplication`

So I think there are many better alternatives that do not need a change in PDE or other tooling...

laeubi:
> `DetectVMInstallationsJob` checks if ENV variable `CI=true` is set. It doesn't magically detect the CI environment, it uses hints set from outside.

I know as I added support from it, still it is something that most (all?) CI server set by default. So the job can simply adapt to other common ways (e.g. properties already set by such launches).

> Also a fair point. Though if PDE is not knowledgeable of DetectVMInstallationsJob, then DetectVMInstallationsJob must become knowledgeable of what PDE is doing. So the problem is only inverted.

The point is that `DetectVMInstallationsJob` might care about where it is running, but PDE should not care much about what code it executes in a test. As said, tomorrow there will be `WhatEverElseJob` and `JustARandomOtherJob` and soon we add more and more "exceptions"... so for me its clear to have such things implemented at the one responsible for the thing.

> E.g. in our product we disable the job with a preference. In other products the job might be a welcome change to Eclipse and tests can already be relying on the job.

Last time we have discussed about the topic (when the CI check was introduced) there where similar issues raised. Because of this the job [here explicitly check](https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/a3bf1ad921b07398b599c140f1aec0d21256b835/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java#L324-L331) if the property is unset and then only takes action. Another reason for me to implement it there, as it is much more flexible and can take other things into account where PDE can only set it unconditionally.

So user have always the choice to actively set the property to `false` in their test to enable it.

> So we need to also know we are running a test for Eclipse code... I'll have to discuss this with my team. Maybe its more simple to make tests more robust, if they have an issue with multiple JVMs installed

That's for sure, apart from that a rule or junit extension might be used as well to specifically perform such setup/cleanup things in test requiring it specifically.

PR #2217, opened 2026-02-19
Show launch bundles button title case correction

nburnwal09:
Corrected the title case of the button `Show Launch Bundles` in launch configuration as suggested here:
https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/491#pullrequestreview-3818213676

image


nburnwal09:
@merks @gireeshpunathil 
This is a small UI change and part of an [N&N PR](https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/491). It would be great if this one gets landed.

Issue #2220, opened 2026-02-24
Target editor no longer shows unresolvable IUs


ptziegler:
In the target definition for our product, I'm currently referencing the 2026-03 staging repository to make sure our product works with the upcoming Eclipse release. In this definition, the versions of 3rd-party dependencies are locked. So every now and then, I need to update the versions in case they have been updated in the Orbit repository.

Up until the 2025-12 release, I would get an error in the target editor itself, listing all IUs that can no longer be found in the update site. However, this list no longer shows up. I can still get the list of problematic IUs from the error log, but this is obviously less convenient.

4.38

<img width="1208" height="698" alt="Image" src="https://github.com/user-attachments/assets/70fbe1ba-9b25-4784-a966-590620f440f2" />

4.39 RC1

<img width="1211" height="727" alt="Image" src="https://github.com/user-attachments/assets/605146a4-2682-40ee-b746-50e3437a2ddf" />

iloveeclipse:
@ptziegler : if you bisect to the commit caused this, we may fix it for RC2 planned for tomorrow.

PR #2221, opened 2026-02-24
Fix: Target editor no longer shows unresolvable IUs #2220

ptziegler:
The change 7444c3fc23d1b0a6793ff03df2c1f3c7c016488e wrongfully inverted the isOK() check, causing unresolvable IUs to no longer show up in the target editor.

Closes https://github.com/eclipse-pde/eclipse.pde/issues/2220


laeubi (approved):
Good catch!

PR #2223, opened 2026-02-26
Fix classpath updates for bnd projects

laeubi:
With the previous performance optimization the support for adding a classpath container to a plain bnd project now always produces an empty result.

Instead of only check for model, we now go through a chain of checks to find if we should possibly update the project.

### DIFF
--- a/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.core; singleton:=true
-Bundle-Version: 3.21.200.qualifier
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -201,6 +204,20 @@ void add(IProject project, IClasspathContainer classpathContainer) {
 
 	}
 
+	private static boolean isPdeContainerProject(IProject project, IPluginModelBase model) {
```
iloveeclipse: Should it be named `isBndContainerProject`?

laeubi: > `BndProjectManager.getBndProject(project)` checks only `BndProject.isBndProject()` and that one only checks bnd project nature.

It checks a bit more but I think for this purpose it is enough to only check for project nature itself... simplified that.

Issue #2227, opened 2026-02-27
Cannot run ApiToolsPluginTestSuite from Eclipse


trancexpress:
Launch `ApiToolsPluginTestSuite` from Eclipse, there are many errors like:

```
!ENTRY ZipEditor 4 0 2026-02-27 09:09:38.769
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: ZipEditor [6]
  Unresolved requirement: Require-Bundle: org.eclipse.ui
    -> Bundle-SymbolicName: org.eclipse.ui; bundle-version="3.208.0.v20251219-1043"; singleton:="true"
       org.eclipse.ui [384]
         Unresolved requirement: Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=21))"
         Unresolved requirement: Require-Capability: eclipse.swt; filter:="(image.format=svg)"
           -> Provide-Capability: eclipse.swt; image.format="svg"; version:Version="1.0.0"
              org.eclipse.swt.svg [344]
                Unresolved requirement: Fragment-Host: org.eclipse.swt
                  -> Bundle-SymbolicName: org.eclipse.swt; bundle-version="3.133.0.qualifier"; singleton:="true"
                     org.eclipse.swt [342]
                       Unresolved requirement: Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=21))"
         Unresolved requirement: Require-Bundle: org.eclipse.jface; bundle-version="[3.34.0,4.0.0)"; visibility:="reexport"
           -> Bundle-SymbolicName: org.eclipse.jface; bundle-version="3.39.0.qualifier"; singleton:="true"
              org.eclipse.jface [280]
                Unresolved requirement: Require-Bundle: org.eclipse.swt; bundle-version="[3.126.0,4.0.0)"; visibility:="reexport"
                  -> Bundle-SymbolicName: org.eclipse.swt; bundle-version="3.133.0.qualifier"; singleton:="true"
         Unresolved requirement: Require-Bundle: org.eclipse.swt; bundle-version="[3.133.0,4.0.0)"; visibility:="reexport"
           -> Bundle-SymbolicName: org.eclipse.swt; bundle-version="3.133.0.qualifier"; singleton:="true"
[...]

PR #2228, opened 2026-02-27
Update JRE in shared launch for main org.eclipse.pde.api.tools.tests test suite

trancexpress:
Fixes: #2227


merks:
Yes, it looks good.  Moreover, everything actually installed and used in an IDE will be running with Java 21 so it makes sense to test with that as well.

PR #2239, opened 2026-03-03
Adapt to removal of o.e.swt.tools.base Plug-in

HannesWell:
See https://github.com/eclipse-platform/eclipse.platform.swt/pull/3116


akurtakov:
Please don't install SWT tools when spies are installed. "SWT tools" are useful only when developing SWT and one changes native bindings and the include builders and views that are better not widely installed, "SWT spies" are useful for many people fighting leaks and etc. 
Plugins swt.tools.base was just the view category lately so they don't depend on each other, just create new "SWT Spies" category and use it here would be better.

HannesWell:
Test failures are unrelated.

PR #2242, opened 2026-03-04
Move from Full-qualified access to imports in Fix FeatureBasedLaunchTest

vogella:
This fixes the error if the MANIFEST.MF dependencies is clean-up.


laeubi (changes requested):
Please use reference annotation instead of cluttering the code with unused variables!

PR #2250, opened 2026-03-08
Complete PdeProjectJar bin.includes and output prefix handling

laeubi:
Make PdeProjectJar bin.includes-driven instead of blindly iterating output.* entries. Handle dot entry with fallback to JDT default output folder for Java projects, package non-dot libraries as embedded inner jars via JarResource, and include regular files/folders from project.


laeubi:
Only Timeout warning for P2 download

Issue #2244, opened 2026-03-05
Too many dependencies in PDE container org.eclipse.pde.core.requiredPlugins


trancexpress:
Regression from: https://github.com/eclipse-pde/eclipse.pde/pull/2218

An example problem: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2830

There is an abnormal number of plug-ins on the classpath now. E.g. `org.eclipse.jdt.debug.tests` had ~25 entries in the container before. It now has ~230.

E.g. looking at one plug-in that has no place on the classpath of `org.eclipse.jdt.debug.tests`, but is now: `org.apache.ant`

Its required by:

```
org.eclipse.platform 4.40.0.v20260304-2033
org.eclipse.swtbot.eclipse.test.junit 4.3.0.202510091338
org.eclipse.ant.launching 1.4.1100.v20260304-1512
org.eclipse.ant.tests.ui 3.12.500.qualifier
org.eclipse.ant.ui 3.10.500.v20260304-1512
org.eclipse.jdt.apt.core 3.8.600.v20241001-0914
org.eclipse.test 3.6.500.v20260105-1638
org.eclipse.text.quicksearch 1.3.400.v20260117-1027
ZipEditor 1.1.9.202308271843
```

None of these re-export their dependency (`visibility:=reexport`) to `org.apache.ant`, so why is `org.eclipse.jdt.debug.tests` now able to use code from `org.apache.ant`? I can e.g. make this change:

```
@@ -51,7 +51,7 @@ import org.junit.rules.TestName;
  * Tests the Open from Clipboard action.
  */
 public class OpenFromClipboardTests {
-
+       org.apache.tools.ant.taskdefs.Ant a;
        /*
         * Copy of constants from OpenFromClipboardAction
         */
```

It will compile in Eclipse, but it should not. See OSGI specification:

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.requirebundle

```
The following directives can be used in the Require-Bundle header:

visibility - If the value is private (default), then all visible packages from the required bundles are not re-exported.
If the value is reexport then bundles that require this bundle will transitively have access to these required bundle's exported packages.
That is, if bundle A requires bundle B, and bundle B requires bundle C with visibility:=reexport
then bundle A will have access to all bundle C's exported packages as if bundle A had required bundle C.
```

merks:
[...]

PR #2251, opened 2026-03-09
Don't include transitive optional dependencies to PDE container

iloveeclipse:
Since ~20 years PDE never allowed not imported/required bundles be on the classpath of a bundle.

From PDE point of view, bundle code should only see references to the code from other bundles on project classpath if:

- the bundle requires another bundle
- or the bundle imports a package from other bundle

Transitive dependencies are only added to the classpath if the required bundle reexports one of its dependencies.

This has a simple rationale:

1) Developers can't compile code in the IDE which would be not working at runtime, saving time for development by showing the errors before starting application.
2) The IDE can quickly recompile one line code change in a workspace with hundreds of bundles, because the classpath only has the dependencies mentioned above.

With 89b00be1a4bf44132b61e46227c5803c5cc6ad6e above expectations are broken. All transitively used projects are now on the classpath.

With `org.eclipse.jdt.core.compiler.problem.forbiddenReferenc` set to `ignore` or `warning` code that can compile (but uses illegal references) will not work at runtime.

With this option set to `error` the code will not compile, but compilation will suffer from all transitive dependencies added to the bundle classpath, incremental build is significantly slower now.

With 89b00be1a4bf44132b61e46227c5803c5cc6ad6e also optional transitive dependencies are added transitively, increasing the problem impact.

Optional dependencies can be re-exported, but they are unlikely can be part of the bundle own API.

As long as proper solution for the problem introduced by 89b00be1a4bf44132b61e46227c5803c5cc6ad6e is identified, don't add optional transitive dependencies to the classpath.

This partly fixes the problem by slightly reducing set of projects added to the classpath.

Also added regression test for RequiredPluginsClasspathContainer

Shows the regression caused by
https://github.com/eclipse-pde/eclipse.pde/pull/2218 / 89b00be1a4bf44132b61e46227c5803c5cc6ad6e

The test will fail on master after
89b00be1a4bf44132b61e46227c5803c5cc6ad6e and pass before this commit.

See https://github.com/eclipse-pde/eclipse.pde/issues/2244

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -678,8 +678,7 @@ class RequiredPluginsClasspathContainer {
 	 */
 	private void addTransitiveDependenciesWithForbiddenAccess(Set<BundleDescription> added,
 			List<IClasspathEntry> entries) throws CoreException {
-		Set<BundleDescription> closure = DependencyManager.findRequirementsClosure(added,
-				INCLUDE_OPTIONAL_DEPENDENCIES);
+		Set<BundleDescription> closure = DependencyManager.findRequirementsClosure(added);
 		String systemBundleBSN = TargetPlatformHelper.getPDEState().getSystemBundle();
 		Iterator<BundleDescription> transitiveDeps = closure.stream()
 				.filter(desc -> !desc.getSymbolicName().equals(systemBundleBSN))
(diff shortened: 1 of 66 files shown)


laeubi (changes requested):
As explained **multiple** times now by **multilple** users, your assumption that something is not working as expected is wrong.

Optional dependencies *must be included* and setting `org.eclipse.jdt.core.compiler.problem.forbiddenReferenc` to something else than error is wrong when working with PDE projects, so whatever your test expsects is invalid because the premise is wrong.

What happens here with PDE is **exactly what happens at runtime**, the bundle will have all transitive dependencies in the classloader chain but is only allowed to load those that are imported. This does not mean the are not accessible (but not by the classlaoder of the bundle).

iloveeclipse:
> > Can API be dependent on optional dependency?
> 
> Sure why not? - I'm also not sure if its clear but a type can be API but not imported (yet) what results in the same restriction error.

I just want to know if there are any real world example where a bundle with optional dependency includes this dependency as part of the API? How this API can be stable then if it should also work **without** this optional bundle?

I'm pretty sure this can't be API, not in the sense we handle API in Eclipse. 
Do you have ane reference project that provides optional dependency as part of the API?

laeubi:
I think this can be closed here?

HannesWell:
> I think this can be closed here?

Yes it can, since it won't be overall helpful. Still, thanks for the discussion.

PR #2256, opened 2026-03-12
Do not follow generic requirements when computing the classpath

laeubi:
OSGi uses a generic Provide-Capability/Require-Capability model that
allows to declare requirements beyond the classic Import-Package and
Require-Bunlde. A common example is a bundle that provides a DS
component and needs an extender to actually construct the component and
inject the services. Another is a logging API that needs a provider.

These are especially needed because in such scenario a provider/consumer
do not know each other and should not depend directly (only loosely
coupled). Currently such providers are still pulled in even though they
are not contribute to the compilation classpath as they are completely
invisible at runtime - there is a wiring but no classloader access
granted.

This now adds a new option to the DependencyManager to exclude generic
requirements from the computation of the dependency closure and the
RequiredPluginsClasspathContainer is using that option to avoid follow
such generic wiring contracts.

Includes the testcase from PR https://github.com/eclipse-pde/eclipse.pde/pull/2254

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -683,8 +683,7 @@ class RequiredPluginsClasspathContainer {
 	 */
 	private void addTransitiveDependenciesWithForbiddenAccess(Set<BundleDescription> added,
 			List<IClasspathEntry> entries) throws CoreException {
-		Set<BundleDescription> closure = DependencyManager.findRequirementsClosure(added,
-				INCLUDE_OPTIONAL_DEPENDENCIES);
+		Collection<BundleDescription> closure = ClasspathComputer.collectBuildRelevantDependencies(added);
 		String systemBundleBSN = TargetPlatformHelper.getPDEState().getSystemBundle();
 		Iterator<BundleDescription> transitiveDeps = closure.stream()
 				.filter(desc -> !desc.getSymbolicName().equals(systemBundleBSN))
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/MinimalState.java
@@ -271,7 +271,7 @@ public class MinimalState {
 		// now we need to check all bundles if there is any fragment. If a
 		// fragment should be resolved its host and all already attached
 		// fragments must be re-resolved again
-		for (BundleDescription bundleDescription : bundles) {
+		for (BundleDescription bundleDescription : bundles.toArray(BundleDescription[]::new)) {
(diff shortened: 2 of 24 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -684,7 +685,7 @@ private void addImplicitDependencies(BundleDescription desc, Set<BundleDescripti
 	private void addTransitiveDependenciesWithForbiddenAccess(Set<BundleDescription> added,
 			List<IClasspathEntry> entries) throws CoreException {
 		Set<BundleDescription> closure = DependencyManager.findRequirementsClosure(added,
-				INCLUDE_OPTIONAL_DEPENDENCIES);
+				INCLUDE_OPTIONAL_DEPENDENCIES, EXCLUDE_REQUIRED_CAPABILITY);
```
HannesWell: In fact we already have a facility in PDE that computes what you want to obtain here:
```suggestion
		Set<BundleDescription> closure = BuildDependencyCollector.collectBuildRelevantDependencies(added)
```
It's currently an own implementation, but I have just opened a PR that migrates it to use the `DependencyManager` (after the change was just on my computer for a long time):
- https://github.com/eclipse-pde/eclipse.pde/pull/2257

In fact you should already be able to use that code now, even without my PR.
Then you don't have to modify `DependencyManager` at all and introduce a new constant for it.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
```diff
@@ -242,7 +252,19 @@ public static Set<BundleDescription> findRequirementsClosure(Collection<BundleDe
...
+					if (!BundleNamespace.BUNDLE_NAMESPACE.equals(namespace)
+							&& !PackageNamespace.PACKAGE_NAMESPACE.equals(namespace)
+							&& !HostNamespace.HOST_NAMESPACE.equals(namespace)
+							&& !ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE.equals(namespace)) {
+						// If it is not one of the framework namespaces it is a
+						// generic one see https://docs.osgi.org/specification/osgi.core/8.0.0/framework.namespaces.html
+						continue;
+					}
+				}
+				BundleRevision provider = capability.getRevision();
```
HannesWell: For that particular change I have a counter proposal in:
- https://github.com/eclipse-pde/eclipse.pde/pull/2257

This doesn't require a new constant within the `DependencyManager`, which is also quite specific to PDE respectively the build time requirement. In general I would like to keep `DependencyManager` generic in the hope that it could be up-streamed to OSGi in some form at some day.

Plus I'm not sure if it's necessary to follow EE-namespace wires. See also
- https://github.com/eclipse-pde/eclipse.pde/pull/2257#discussion_r2928179990

PR #2257, opened 2026-03-12
Replace BuildDependencyCollector by DependencyManager

HannesWell:
This reduces logically duplicated code.
It also considers the OSGi System bundle for workspace build order, if it's in the same workspace.

This also fixes the build the build order computation in case the an Eclipse project in a workspace is not named like its corresponding bundle, i.e. if a project's name differers from the specified Bundle-SymbolicName.


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/BuildDependencyCollector.java
```diff
@@ -13,76 +13,39 @@
...
-		if (dependency != null && fDependencies.add(dependency)) {
-			fQueue.add(dependency);
-		}
+		Set<BundleDescription> requirementsClosure = DependencyManager.findRequirementsClosure(roots, BUILD_RELEVANT_NAMESPACES,
+				DependencyManager.Options.INCLUDE_OPTIONAL_DEPENDENCIES);
+		// Add system-bundle to ensure prober build order if it's in the
+		// workspace
+		String systemBundleId = PluginModelManager.getInstance().getSystemBundleId();
+		IPluginModelBase systemBundle = PluginRegistry.findModel(systemBundleId);
+		requirementsClosure.add(systemBundle.getBundleDescription());
```
HannesWell: After thinking again about this, I'm not sure anymore if this is actually necessary.
Because if the system-bundle is not an explicit dependency of a project, i.e. by being required as bundle or one of the provided (mainly OSGi) packages is imported, it's actually also not necessary to build the system-bundle before it in the workspace.
Just because the project requires the EE capability provided by the System-Bundle doesn't mean it's needed for compliation.
That's handled differently in the Java/JDT world (i.e. by the corresponding `.classpath` file entry).

And both callers also worked without that before.

PR #2270, opened 2026-03-23
Add Java projects to classpath container org.eclipse.pde.core.externaJavaSearch

### LINKED ISSUE #2269: Navigation problems in IDEApplication from External Plug-in Libraries project
If I have `org.eclipse.swt.gtk.linux.x86_64` imported from the SWT repo, I see navigation problems in `IDEApplication`, when opened from the External Plug-in Libraries project.

If I go to `IDEApplication.start()` and I try to open `Display`, it opens a class in the JDK, not in SWT. Also I cannot Ctrl+left click `checkInstanceLocation` in this line, to navigate to the definition of the method:

```
			Object instanceLocationCheck = checkInstanceLocation(shell, appContext.getArguments());
```

Sometimes I see also broken syntax highlighting. This is not the only class in which I run into such problems as of late.

See screen recording:

https://github.com/user-attachments/assets/e4a00210-d0c9-4d9a-9ad5-ebef11d4b53b

I see this in e.g. a new workspace with:

```
Eclipse SDK
Version: 2026-06 (4.40)
Build id: I20260321-1800
```

To have the External Plug-in Libraries project, Ctrl+3 and use: "Add All Plug-ins to Java Workspace Scope"

Importing `org.eclipse.swt.gtk.linux.x86_64` from the running platform also causes the issue.

trancexpress:
The classpath container `org.eclipse.pde.core.externalJavaSearch` skips adding plug-ins if a respective project is found in the workspace. This can result in unexpected behavior when browsing code from the container, such as wrong syntax highlighting, navigation and search results.

This change adjusts the classpath container to contain also projects found in the workspace.

Fixes: #2269

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SearchablePluginsManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/SearchablePluginsManager.java
@@ -183,10 +183,16 @@ public class SearchablePluginsManager implements IFileAdapterFactory, IPluginMod
 		for (String id : plugins) {
 			ModelEntry entry = PluginRegistry.findEntry(id);
 			if (entry != null) {
-				boolean addModel = Arrays.stream(entry.getWorkspaceModels())
+				List<IProject> javaProjects = Arrays.stream(entry.getWorkspaceModels())
 						.map(IPluginModelBase::getUnderlyingResource).map(IResource::getProject)
-						.noneMatch(PluginProject::isJavaProject);
+						.filter(PluginProject::isJavaProject).collect(Collectors.toList());
+				boolean addModel = javaProjects.isEmpty();
 				if (!addModel) {
+					for (IProject project : javaProjects) {
+						IPath path = project.getFullPath();
+						IClasspathEntry projectEntry = JavaCore.newProjectEntry(path);
+						result.add(projectEntry);
+					}
 					continue;
 				}
 				IPluginModelBase[] models = entry.getExternalModels();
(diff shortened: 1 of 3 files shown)


trancexpress:
> From the naming and from the code it looks like it is intentional... So I think you should do some research what was the reason for this.

The project is used to be able to browse target platform code, search in it and so on. I'm sure its intentional, otherwise you woyld e.g. get double search results - one for the project in the workspace, one for the External Plug-in Libraries project.

I'll see if I can find something in the git history, maybe there is a Bugzilla ticket.

trancexpress:
@iloveeclipse can you take a look here? Whenever I try to use the external plug-ins project, half the time it doesn't work:

image

All the non-highlighted text is useless, cannot use call hierarchy, navigation and so on.

The only way to work with Eclipse code seems to be to import sources...

iloveeclipse:
> This change adjusts the classpath container to contain also projects found in the workspace.

Just from the name alone **external** java search means, the original idea was to reference only sources that are **not** in the workspace. So the proposal to add the projects **in** the workspace to that container doesn't make sense for me, it just does the opposite of what it should be done. See the javadoc of the `SearchablePluginsManager` class.

If this PR fixes the problem, it is likely it fixes some bug in a different place, like indexer not properly triggered after dependencies update etc.

iloveeclipse (changes requested):
I believe this is the wrong place to fix the issue. I would propose to close the PR.

trancexpress:
> Just from the name alone **external** java search means, the original idea was to reference only sources that are **not** in the workspace. So the proposal to add the projects **in** the workspace to that container doesn't make sense for me, it just does the opposite of what it should be done. See the javadoc of the `SearchablePluginsManager` class.
> 
> If this PR fixes the problem, it is likely it fixes some bug in a different place, like indexer not properly triggered after dependencies update etc.

The problem is explained here: https://github.com/eclipse-pde/eclipse.pde/issues/2269#issuecomment-4112463107

For anything not found in the container, there is a 2nd search with relaxed criteria. But for things found in the container (and they are searched by string-search), that type is used for whichever functionality. So navigation stops working.

trancexpress:
> I believe this is the wrong place to fix the issue. I would propose to close the PR.

Sure.

PR #2288, opened 2026-04-14
Update Color constructors to remove deprecated Device parameter

vogella:
This PR updates Color constructors to remove the deprecated Device parameter, following the changes in SWT (https://github.com/eclipse-platform/eclipse.platform.swt/issues/3232). It also removes the now unused imports of Display and RGB in the modified files.


vogella:
Copilot needs an update, it continue to claims that colors must be disposed. All feedback is invalid.

Issue #2290, opened 2026-04-16
Auto-remove empty packages from Export-Package when all classes are deleted


nburnwal09:
When all Java classes within a package are deleted through refactoring operations, the package remains listed in the Export-Package header of the MANIFEST.MF (if already added). This creates an inconsistency where the plugin exports a package that no longer exists or is empty, which can lead to invalid OSGi bundle metadata, build warnings or errors, potential runtime issues for consumers of the exported package

Expected Behavior:
When all classes in a package are deleted via delete operation, the package(having no java files or any other resources) should be automatically removed from the Export-Package header in MANIFEST.MF if it was previously exported.

https://github.com/user-attachments/assets/1d56a3d1-d142-4cf1-8466-f9f1d9347511

nburnwal09:
I am working on it

merks:
I thought I'd mentioned it; you mention all classes but I assume you are also consider account other things like resources.

nburnwal09:
> I thought I'd mentioned it; you mention all classes but I assume you are also consider account other things like resources.

Yeah you are right. I didn’t account for other resources; I’ll update the issue to include that.

PR #2292, opened 2026-04-17
Handle deletion scenario for cleaning up empty exported packages

nburnwal09:
Enhancement added for handling class/type deletion scenario.
When all classes in a package are deleted via delete operation, the package(having no java files or any other resources) should be automatically removed from the Export-Package header in MANIFEST.MF if it was previously exported.

Addressing #2290

https://github.com/user-attachments/assets/00d77e9d-12f1-4757-9341-9fc233371e0a

https://github.com/user-attachments/assets/a5a37d59-4dc1-49c3-a28a-91e94d418164

### DIFF
--- a/ui/org.eclipse.pde.ui/plugin.xml
+++ b/ui/org.eclipse.pde.ui/plugin.xml
@@ -1226,6 +1226,38 @@
       </renameParticipant>
    </extension>
    <extension
+         point="org.eclipse.ltk.core.refactoring.deleteParticipants">
+      <deleteParticipant
+            class="org.eclipse.pde.internal.ui.refactoring.ManifestTypeDeleteParticipant"
+            id="org.eclipse.pde.ui.manifestTypeDeleteParticipant"
+            name="%delete.type.participant">
+         <enablement>
+            <with
+                  variable="element">
+               <instanceof
+                     value="org.eclipse.jdt.core.IType">
+               </instanceof>
+            </with>
+         </enablement>
+      </deleteParticipant>
+   </extension>
+   <extension
+          point="org.eclipse.ltk.core.refactoring.deleteParticipants">
+       <deleteParticipant
+            class="org.eclipse.pde.internal.ui.refactoring.ManifestPackageDeleteParticipant"
+            id="org.eclipse.pde.ui.manifestPackageDeleteParticipant"
+            name="%delete.package.participant">
+         <enablement>
+            <with
+                  variable="element">
+               <instanceof
+                     value="org.eclipse.jdt.core.IPackageFragment">
+               </instanceof>
+            </with>
+         </enablement>
+       </deleteParticipant>
+   </extension>
+   <extension
          point="org.eclipse.ltk.core.refactoring.moveParticipants">
       <moveParticipant
             class="org.eclipse.pde.internal.ui.refactoring.ManifestTypeMoveParticipant"
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
@@ -380,6 +380,8 @@ MainPreferencePage_test_plugin_pattern_note=This pattern is matched against the 
 ManifestEditorContributor_externStringsActionName=Externalize Strings...
 
 ManifestTypeRenameParticipant_composite=Rename classes referenced in plug-in manifest files
+ManifestTypeDeleteParticipant_composite=Remove packages emptied by class deletion from plug-in manifest
+ManifestPackageDeleteParticipant_packageDelete=Delete packages referenced in plug-in manifest
 
 ManifestEditor_MatchSection_optional = Optional
 ManifestEditor_MatchSection_reexport = Re-export the dependency
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/BundleManifestChange.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/BundleManifestChange.java
@@ -15,6 +15,8 @@
 package org.eclipse.pde.internal.ui.refactoring;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 
 import org.eclipse.core.filebuffers.FileBuffers;
 import org.eclipse.core.filebuffers.ITextFileBufferManager;
@@ -69,8 +71,12 @@ public class BundleManifestChange {
 		return null;
 	}
 
-	public static Change createEmptyPackageChange(IFile file, String packageName, IProgressMonitor monitor)
-			throws CoreException {
+	public static Change createDeletePackagesChange(IFile file, Collection<IPackageFragment> packages,
+			IProgressMonitor monitor) throws CoreException {
+		List<String> packageNames = packages.stream().map(IPackageFragment::getElementName).toList();
+		if (packageNames.isEmpty()) {
+			return null;
+		}
 		try {
 			Bundle bundle = getBundle(file, monitor);
 			if (bundle == null) {
@@ -82,8 +88,8 @@ public class BundleManifestChange {
 			bundle.getModel().addModelChangedListener(listener);
 
 			BasePackageHeader header = (BasePackageHeader) bundle.getManifestHeader(Constants.EXPORT_PACKAGE);
-			if (header != null && header.hasPackage(packageName)) {
-				removePackage(header, packageName);
+			if (header != null) {
+				packageNames.stream().filter(header::hasPackage).forEach(name -> removePackage(header, name));
 			}
 			TextEdit[] operations = listener.getTextOperations();
 			if (operations.length > 0) {
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
@@ -1,0 +1,81 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 IBM Corporation and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.ui.refactoring;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.participants.RefactoringArguments;
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEPlugin;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+/**
+ * Handles package deletion in PDE plugin projects. Updates the bundle's
+ * MANIFEST.MF after deleting packages.
+ */
+public class ManifestPackageDeleteParticipant extends PDEDeleteParticipant {
+
+	private IProject fProject;
+	private Set<IPackageFragment> fPackages = new HashSet<>();
+
+	@Override
+	protected boolean initialize(Object element) {
+		IPackageFragment fragment = (IPackageFragment) element;
+		try {
+			if (!fragment.containsJavaResources()) {
+				return false;
+			}
+		} catch (JavaModelException e) {
+			PDEPlugin.logException(e);
+		}
+		fProject = fragment.getJavaProject().getProject();
+		if (WorkspaceModelManager.isPluginProject(fProject)) {
+			fPackages.add(fragment);
+			return true;
+		}
+		return false;
+	}
+
+	@Override
+	public void addElement(Object element, RefactoringArguments arguments) {
+		fPackages.add((IPackageFragment) element);
+	}
+
+	@Override
+	public String getName() {
+		return PDEUIMessages.ManifestPackageDeleteParticipant_packageDelete;
+	}
+
+	@Override
+	protected void addChanges(CompositeChange result, IProgressMonitor pm) throws CoreException {
+		IFile file = PDEProject.getManifest(fProject);
+		if (file.exists()) {
+			Change change = BundleManifestChange.createDeletePackagesChange(file, fPackages, pm);
+			if (change != null) {
+				result.add(change);
+			}
+		}
+	}
+
+}
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 IBM Corporation and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.ui.refactoring;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.participants.RefactoringArguments;
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+/**
+ * Handles class deletion in PDE plugin projects. Updates the bundle's MANIFEST.MF
+ * when a package becomes empty after deleting types.
+ */
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
+
+	private IProject fProject;
+	private Set<IType> fTypes = new LinkedHashSet<>();
+
+	@Override
+	protected boolean initialize(Object element) {
+		IType type = (IType) element;
+		IJavaProject javaProject = type.getJavaProject();
+		if (javaProject != null) {
+			fProject = javaProject.getProject();
+			if (WorkspaceModelManager.isPluginProject(fProject)) {
+				fTypes.add(type);
+				return true;
+			}
+		}
+		return false;
+	}
+
+	@Override
+	public void addElement(Object element, RefactoringArguments arguments) {
+		IType type = (IType) element;
+		fTypes.add(type);
+	}
+
+	@Override
+	public String getName() {
+		return PDEUIMessages.ManifestTypeDeleteParticipant_composite;
+	}
+
+	@Override
+	protected void addChanges(CompositeChange result, IProgressMonitor pm) throws CoreException {
+		IFile file = PDEProject.getManifest(fProject);
+		if (!file.exists()) {
+			return;
+		}
+		// Group deleted types by package and collect their compilation units
+		Map<IPackageFragment, Set<ICompilationUnit>> deletedCUsByPackage = new HashMap<>();
+		for (IType type : fTypes) {
+			IPackageFragment pkg = type.getPackageFragment();
+			ICompilationUnit cu = type.getCompilationUnit();
+			if (cu != null) {
+				deletedCUsByPackage.computeIfAbsent(pkg, k -> new HashSet<>()).add(cu);
+			}
+		}
+		// Check each package to see if it becomes empty after deletion
+		List<IPackageFragment> emptiedPackages = deletedCUsByPackage.entrySet().stream().filter(e -> {
+			IPackageFragment pkg = e.getKey();
+			Set<ICompilationUnit> deletedCUs = e.getValue();
+			try {
+				return willPackageBeEmpty(pkg, deletedCUs);
+			} catch (CoreException ex) {
+				return false;
+			}
+		}).map(Map.Entry::getKey).toList();
+
+		Change change = BundleManifestChange.createDeletePackagesChange(file, emptiedPackages, pm);
+		if (change != null) {
+			result.add(change);
+		}
+
+	}
+
+	/**
+	 * Checks if a package will be empty after deleting the specified compilation units.
+	 *
+	 * @param pkg the package to check
+	 * @param deletedCUs the compilation units being deleted
+	 * @return true if the package will be empty after deletion
+	 * @throws CoreException if an error occurs accessing package contents
+	 */
+	private boolean willPackageBeEmpty(IPackageFragment pkg, Set<ICompilationUnit> deletedCUs) throws CoreException {
+		// Check for non-Java resources (properties files, XML files, etc.)
+		Object[] nonJavaResources = pkg.getNonJavaResources();
+		if (nonJavaResources != null && nonJavaResources.length > 0) {
+			return false;
+		}
+		// Check if any compilation unit in the package is NOT being deleted
+		ICompilationUnit[] compilationUnits = pkg.getCompilationUnits();
+		for (ICompilationUnit cu : compilationUnits) {
+			if (!deletedCUs.contains(cu)) {
+				// This compilation unit is not being deleted, so package is not
+				// empty
+				return false;
+			}
+		}
+		return true;
+	}
+
+}
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
@@ -1,0 +1,43 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 IBM Corporation and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.pde.internal.ui.refactoring;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.RefactoringStatus;
+import org.eclipse.ltk.core.refactoring.participants.CheckConditionsContext;
+import org.eclipse.ltk.core.refactoring.participants.DeleteParticipant;
+import org.eclipse.ltk.core.refactoring.participants.ISharableParticipant;
+
+public abstract class PDEDeleteParticipant extends DeleteParticipant implements ISharableParticipant {
+
+	@Override
+	public RefactoringStatus checkConditions(IProgressMonitor pm, CheckConditionsContext context)
+			throws OperationCanceledException {
+		return new RefactoringStatus();
+	}
+
+	@Override
+	public Change createChange(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		CompositeChange result = new CompositeChange(getName());
+		addChanges(result, pm);
+		return result.getChildren().length == 0 ? null : result;
+	}
+
+	protected abstract void addChanges(CompositeChange result, IProgressMonitor pm) throws CoreException;
+
+}
--- a/ui/org.eclipse.pde.ui/plugin.properties
+++ b/ui/org.eclipse.pde.ui/plugin.properties
@@ -158,6 +158,8 @@ rename.type.participant = Manifest Rename Type Participant
 rename.package.participant = Manifest Rename Package Participant
 move.type.participant = Manifest Move Type Participant
 move.package.participant = Manifest Move Package Participant
+delete.type.participant = Manifest Delete Type Participant
+delete.package.participant = Manifest Delete Package Participant
 
 queryParticipant.name.0 = PDE Java Search Participant
 new.profile.name = Target Definition
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/PDEUIMessages.java
@@ -2389,6 +2389,10 @@ public class PDEUIMessages extends NLS {
 
 	public static String ManifestTypeRenameParticipant_composite;
 
+	public static String ManifestTypeDeleteParticipant_composite;
+	
+	public static String ManifestPackageDeleteParticipant_packageDelete;
+
 	public static String LauncherPage_title;
 
 	public static String WindowImagesSection_16;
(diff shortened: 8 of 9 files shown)


nburnwal09:
This is an enhancement to the previous task mentioned [here](https://github.com/eclipse-pde/eclipse.pde/pull/2136#issuecomment-3546264903)

gireeshpunathil (approved):

nburnwal09:
@laeubi 
Kindly review this PR. 
Addressed the earlier mentioned scenario [here](https://github.com/eclipse-pde/eclipse.pde/pull/2136#issuecomment-3546264903)

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+
+			try {
+				if (willPackageBeEmpty(pkg, deletedTypes)) {
+					Change change = BundleManifestChange.createEmptyPackageChange(file, pkg.getElementName(), pm);
+					if (change != null) {
+						result.add(change);
+					}
+				}
+			} catch (CoreException e) {
+				e.printStackTrace();
```
noopur2507: Printing the stack trace instead of proper logging or handling the issue can cause the action to fail silently.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.ltk.core.refactoring.Change;
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
```
noopur2507, line 42: Javadoc can be added for this class to explain its purpose, when it is triggered, etc.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+		}
+		// Check the underlying folder for any OTHER files
+		IResource resource = pkg.getCorrespondingResource();
+		if (resource instanceof IFolder folder) {
+			IResource[] members = folder.members();
+			for (IResource member : members) {
+				if (member instanceof IFile memberFile) {
+					String fileName = memberFile.getName();
+					String extension = memberFile.getFileExtension();
+					if ("class".equals(extension)) { //$NON-NLS-1$
```
noopur2507: Please check whether any constants exist in the PDE code for the file extensions that can be reused here, or create them at the class level instead of hardcoding them here.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,64 @@
...
+	protected HashMap<Object, String> fElements;
+
+	@Override
+	public RefactoringStatus checkConditions(IProgressMonitor pm, CheckConditionsContext context)
+			throws OperationCanceledException {
+		return new RefactoringStatus();
+	}
+
+	@Override
+	public void addElement(Object element, RefactoringArguments arguments) {
```
noopur2507: Can the "element" be of a type other than IType and IPackageFragment? That should be handled in the 'else' case (probably via logging).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/pderesources.properties
```diff
@@ -380,6 +380,7 @@ MainPreferencePage_test_plugin_pattern_note=This pattern is matched against the
 ManifestEditorContributor_externStringsActionName=Externalize Strings...
 
 ManifestTypeRenameParticipant_composite=Rename classes referenced in plug-in manifest files
+ManifestTypeDeleteParticipant_composite=Delete classes referenced in plug-in manifest files
```
noopur2507: Check the usage of this message and update it, as it's talking about deleting class references from the manifest, which isn't the case in this PR.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,64 @@
...
+		if (element instanceof IType type) {
+			fElements.put(element, type.getFullyQualifiedName());
```
noopur2507: It is better to initialize "fElements" at declaration time to prevent any potential NPE here. Also, initializing it to empty in ManifestTypeDeleteParticipant#initialize will break the sharing behavior of ISharableParticipant in my view.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
+
+	@Override
+	protected boolean initialize(Object element) {
+		if (element instanceof IType type) {
+			IJavaProject javaProject = (IJavaProject) type.getAncestor(IJavaElement.JAVA_PROJECT);
```
noopur2507: "getAncestor" can return null.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+			IProject project = javaProject.getProject();
+			if (WorkspaceModelManager.isPluginProject(project)) {
+				fProject = javaProject.getProject();
```
noopur2507: "javaProject.getProject()" is called twice. Here and at line 45.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+				fElements = new HashMap<>();
```
noopur2507: See PDEDeleteParticipant, line 46 comment.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	 * Checks if a package will be empty after deleting the specified types.
+	 * @param pkg the package to check
+	 * @param deletedTypes the types being deleted
+	 * @return true if the package will be empty after deletion
+	 * @throws CoreException if an error occurs accessing package contents
+	 */
+	private boolean willPackageBeEmpty(IPackageFragment pkg, List<IType> deletedTypes) throws CoreException {
+		IJavaElement[] javaChildren = pkg.getChildren();
+		if (javaChildren.length > deletedTypes.size()) {
```
noopur2507: I think the entire logic here assumes 1:1 mapping between IType and compilation units, which is not the case in JDT. For inner classes, multiple types exist in the same compilation unit. Deleting one inner class shouldn't mark the entire .java file as deleted. The logic should be updated to handle this.

noopur2507 (changes requested):

nburnwal09:
@noopur2507 
I have addressed all of your comments. Kindly review.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,172 @@
...
+ */
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
+
+	private static final String JAVA_EXTENSION = "java"; //$NON-NLS-1$
+	private static final String CLASS_EXTENSION = "class"; //$NON-NLS-1$
+
+	@Override
+	protected boolean initialize(Object element) {
+		if (element instanceof IType type) {
+			IJavaProject javaProject = (IJavaProject) type.getAncestor(IJavaElement.JAVA_PROJECT);
```
HannesWell:
```suggestion
			IJavaProject javaProject = type.getJavaProject();
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,70 @@
...
+	public Change createChange(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		CompositeChange result = new CompositeChange(getName());
+		addChange(result, pm);
+		return (result.getChildren().length == 0) ? null : result;
+	}
+
+	/**
+	 * @throws CoreException may be thrown by overrides
+	 */
+	protected abstract void addChange(CompositeChange result, IProgressMonitor pm) throws CoreException;
```
HannesWell:
```suggestion
	protected abstract void addChanges(CompositeChange result, IProgressMonitor pm) throws CoreException;
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,70 @@
...
+		if (element instanceof IType type) {
+			fElements.put(element, type.getFullyQualifiedName());
+		} else if (element instanceof IPackageFragment pkg) {
+			fElements.put(element, pkg.getElementName());
+		} else {
+			// Log unexpected element types for debugging
+			PDEPlugin.log(Status.warning("Unexpected element type in PDEDeleteParticipant.addElement: " //$NON-NLS-1$
+					+ (element != null ? element.getClass().getName() : "null"))); //$NON-NLS-1$
+		}
+	}
```
HannesWell: I assume that the `ManifestTypeDeleteParticipant` participant introduced in this PR doesn't handle the case of element being an `IPackageFragment` ?
Therefore I wonder if it wouldn't be better to implement `addElement()` entirely in `ManifestTypeDeleteParticipant` and later in `ManifestPackageDeleteParticipant`?
Then the fElements field
```
protected HashMap<Object, String> fElements = new HashMap<>();
```
could also be moved to `ManifestTypeDeleteParticipant` and could become just
```
private final Map<IType, String> fElements = new HashMap<>();
```
This would also avoid the need to cast multiple times.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,172 @@
...
+					return false;
+				}
+				// Check if all types in this compilation unit are being deleted
+				IType[] allTypes = cu.getTypes();
+				if (allTypes.length > deletedFromCU.size()) {
+					// Some types in this compilation unit are NOT being deleted, so package is not empty
+					return false;
+				}
+			}
+		}
```
HannesWell: Would it be sufficient to collect the deleted `ICompilationUnits` per package already in `addChanges()` and pass that Set/List directly to `willPackageBeEmpty()`?
Then this part could just check if there is a CU in the package that is not deleted and wouldn't have to handle CUs and Types multiple times.

This of course expects this handler isn't triggered e.g. by deleting inner classes from the source code or outline. Can you tell if this can happens?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,172 @@
...
+					// Skip all Java source files (already checked via
+					// compilation units above)
+					if (JAVA_EXTENSION.equals(extension)) {
+						continue;
+					}
+					// Any other file means package is not empty
+					return false;
+				}
+			}
+		}
```
HannesWell:
```suggestion
		if (pkg.getCorrespondingResource() instanceof IFolder folder) {
			// Check if any other file (except the already checked source and
			// compiled Java class files) are contained in the packages folder.
			return Arrays.stream(folder.members()).filter(IFile.class::isInstance).map(IFile.class::cast)
					.map(IFile::getFileExtension).allMatch(HANDLED_FILE_EXTENSTIONS::contains);
		}
```
This assumes the to CLASS_EXTENSION and JAVA_EXTENSION fields are replaced with a single field
```
	private static final Set<String> HANDLED_FILE_EXTENSTIONS = Set.of("java", "class"); //$NON-NLS-1$ //$NON-NLS-2$
```
at the top of the class.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,172 @@
...
+			if (element instanceof IType type) {
+				IPackageFragment pkg = type.getPackageFragment();
+				deletedByPackage.computeIfAbsent(pkg, k -> new ArrayList<>()).add(type);
+			}
+		});
+		// Check each package to see if it becomes empty after deletion
+		for (Map.Entry<IPackageFragment, List<IType>> entry : deletedByPackage.entrySet()) {
+			IPackageFragment pkg = entry.getKey();
+			List<IType> deletedTypes = entry.getValue();
+
```
HannesWell:
```suggestion
		deletedByPackage.forEach((pkg, deletedTypes) -> {
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,70 @@
...
+	@Override
+	public Change createChange(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		CompositeChange result = new CompositeChange(getName());
+		addChange(result, pm);
+		return (result.getChildren().length == 0) ? null : result;
+	}
+
+	/**
+	 * @throws CoreException may be thrown by overrides
+	 */
```
HannesWell: IMO this doc is not necessary as an interested reader will find it out easily.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,70 @@
...
+			PDEPlugin.log(Status.warning("Unexpected element type in PDEDeleteParticipant.addElement: " //$NON-NLS-1$
+					+ (element != null ? element.getClass().getName() : "null"))); //$NON-NLS-1$
+		}
+	}
+
+	@Override
+	public Change createChange(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		CompositeChange result = new CompositeChange(getName());
+		addChange(result, pm);
+		return (result.getChildren().length == 0) ? null : result;
```
HannesWell:
```suggestion
		return result.getChildren().length == 0 ? null : result;
```

HannesWell (commented):
Thank you @nburnwal09 for this. 
In general it looks good and works well.

But I have multiple comments below to stream-line the code.
Furthermore please make sure to properly format all new files and edited lines.

Additionally, do you think it makes sense to include the extra change from https://github.com/eclipse-pde/eclipse.pde/pull/2312 here too?
Then the changes could be reviewed together and more holistically.

nburnwal09:
> Additionally, do you think it makes sense to include the extra change from #2312 here too? Then the changes could be reviewed together and more holistically.

I realised about package deletion part later(after raising this PR). Also as that is for package deletion (although the goal of removing the package from manifest is the same), I thought of taking that as a different commit.
But if you suggest, I will merge the change here so that it can be reviewed together.

HannesWell:
> But if you suggest, I will merge the change here so that it can be reviewed together.

In this case I think it's similar and related enough to be together in one change.
Of course if you strongly prefer to keep it separated, that's also fine for me. It's just a suggestion from my side. :)

noopur2507 (approved):

nburnwal09:
> > But if you suggest, I will merge the change here so that it can be reviewed together.
> 
> In this case I think it's similar and related enough to be together in one change. Of course if you strongly prefer to keep it separated, that's also fine for me. It's just a suggestion from my side. :)

Sure @HannesWell  I will merge the changes in this PR. 
Then I will make suggested updates. Thank you.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
nburnwal09: @HannesWell 
yes, this participant can be triggered by deleting inner classes directly.
From plugin.xml (lines 1230-1239), added these:
```
<deleteParticipant
    class="org.eclipse.pde.internal.ui.refactoring.ManifestTypeDeleteParticipant"
    id="org.eclipse.pde.ui.manifestTypeDeleteParticipant"
    name="%delete.type.participant">
    <enablement>
        <with
            variable="element">
                <instanceof
                    value="org.eclipse.jdt.core.IType">
                </instanceof>
         </with>
    </enablement>
</deleteParticipant>
```
      
The participant is registered for any IType.

nburnwal09:
@HannesWell 
I have merged the package deletion part and have addressed all your suggestions. 
Kindly review them.

ui/org.eclipse.pde.ui/plugin.xml
```diff
@@ -1225,6 +1225,38 @@
...
+		<enablement>
+			<with
+				variable="element">
+			<instanceof
+				value="org.eclipse.jdt.core.IPackageFragment">
+			</instanceof>
+			</with>
+		</enablement>
+		</deleteParticipant>
+	</extension>
```
HannesWell: Please use consistent white-space here, the majority of the file uses four spaces per indentation level.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/PDEDeleteParticipant.java
```diff
@@ -0,0 +1,46 @@
...
+import org.eclipse.ltk.core.refactoring.CompositeChange;
+import org.eclipse.ltk.core.refactoring.RefactoringStatus;
+import org.eclipse.ltk.core.refactoring.participants.CheckConditionsContext;
+import org.eclipse.ltk.core.refactoring.participants.DeleteParticipant;
+import org.eclipse.ltk.core.refactoring.participants.ISharableParticipant;
+
+public abstract class PDEDeleteParticipant extends DeleteParticipant implements ISharableParticipant {
+
+	protected IProject fProject;
+
```
HannesWell:
```suggestion
```
The project isn't referenced in this class and therefore should only be defined in the classes that use it.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEPlugin;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+/**
+ * Handles package deletion in PDE plugin projects. Updates the bundle's
+ * MANIFEST.MF after deleting packages.
+ *
+ * @since 3.16
```
HannesWell:
```suggestion
```
Internal class, no `@since` tag necessary.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class ManifestPackageDeleteParticipant extends PDEDeleteParticipant {
+
+	protected HashMap<IPackageFragment, String> fElements = new HashMap<>();
```
HannesWell:
```suggestion
	private IProject fProject;
	private Set<IPackageFragment> fPackages = new HashSet<>();
```
Please only use interfaces of collections (if possible) and the values are actually never used, so a Set is sufficient.
Also restrict the visibility as far as possible (same for the project moved here from https://github.com/eclipse-pde/eclipse.pde/pull/2292/changes#r3212857100).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+				IProject project = javaProject.getProject();
+				if (WorkspaceModelManager.isPluginProject(project)) {
+					fProject = javaProject.getProject();
+					fElements.put(fragment, fragment.getElementName());
+					return true;
+				}
+			}
+		} catch (JavaModelException e) {
+			PDEPlugin.logException(e);
+		}
```
HannesWell, line 51: Please restrict the scope of the try-catch-block further to just wrap the `containsJavaResources()` call above:
```
			try {
				if (!fragment.containsJavaResources()) {
					return false;
				}
			} catch (JavaModelException e) {
				PDEPlugin.logException(e);
			}
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+ * @since 3.16
+ */
+public class ManifestPackageDeleteParticipant extends PDEDeleteParticipant {
+
+	protected HashMap<IPackageFragment, String> fElements = new HashMap<>();
+
+	@Override
+	protected boolean initialize(Object element) {
+		try {
+			if (element instanceof IPackageFragment fragment) {
```
HannesWell: Due to the configuration in the plugin.xml, I think this method will always only be called with a `IPackageFragment`. So like in the `addElement()` method, you could simply cast the received `element` to `IPackageFragment` and avoid the if.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestPackageDeleteParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+				if (!fragment.containsJavaResources()) {
+					return false;
+				}
+				IJavaProject javaProject = (IJavaProject) fragment.getAncestor(IJavaElement.JAVA_PROJECT);
+				IProject project = javaProject.getProject();
+				if (WorkspaceModelManager.isPluginProject(project)) {
+					fProject = javaProject.getProject();
```
HannesWell:
```suggestion
				fProject = fragment.getJavaProject().getProject();
				if (WorkspaceModelManager.isPluginProject(project)) {
```
You can `getJavaProject()` here again and can initialize the `fProject` directly.
I don't think it's a problem if it's initialized in case this participant is eventually discarded.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/BundleManifestChange.java
```diff
@@ -101,6 +103,46 @@ public static Change createEmptyPackageChange(IFile file, String packageName, IP
 		return null;
 	}
 
+	public static Change createMultiplePackageDeleteChange(IFile file, Object[] elements, IProgressMonitor monitor)
+			throws CoreException {
+		List<String> packageNames = Arrays.stream(elements).filter(IPackageFragment.class::isInstance)
+				.map(e -> ((IPackageFragment) e).getElementName()).toList();
```
HannesWell:
```suggestion
	public static Change createDeletePackagesChange(IFile file, Collection<IPackageFragment> packages,
			IProgressMonitor monitor) throws CoreException {
		List<String> packageNames = packages.stream().map(IPackageFragment::getElementName).toList();
```
The code becomes simpler here and at the caller, if simply a `Collection<IPackageFragment>` is accepted.
In general, in new code please don't hesitate to use all the available modern Java features whenever suitable. There is no need to match the Java style from the time when PDE incepted.
And the name could be a little bit more compact as suggested.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/BundleManifestChange.java
```diff
@@ -101,6 +103,46 @@ public static Change createEmptyPackageChange(IFile file, String packageName, IP
...
+		if (packageNames.isEmpty()) {
+			return null;
+		}
+		if (packageNames.size() == 1) {
+			return createEmptyPackageChange(file, packageNames.get(0), monitor);
+		}
```
HannesWell: Since the new method is very similar to the existing `createEmptyPackageChange()` and just can handle multiple packages, it seems better to me to delete the existing `createEmptyPackageChange()` now so it's effectively extended to handle multiple packages at once.
The existing callers could then simply pass the single package using `List.of(pkg)`.
Alternatively the existing method could be called unchanged in a loop in `ManifestPackageDeleteParticipant.addChanges()`, but I think that's not efficient since the MANIFEST.MF file is then read multiple times. Therefore I'd stay with extending the method as suggested above.

```suggestion
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+import org.eclipse.pde.internal.core.WorkspaceModelManager;
+import org.eclipse.pde.internal.core.project.PDEProject;
+import org.eclipse.pde.internal.ui.PDEPlugin;
+import org.eclipse.pde.internal.ui.PDEUIMessages;
+
+/**
+ * Handles class deletion in PDE plugin projects. Updates the bundle's MANIFEST.MF
+ * when a package becomes empty after deleting types.
+ *
+ * @since 3.16
```
HannesWell: `@since` again not necessary.
```suggestion
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
+
+	private static final Set<String> HANDLED_FILE_EXTENSTIONS = Set.of("java", "class"); //$NON-NLS-1$ //$NON-NLS-2$
+	protected HashMap<IType, String> fElements = new HashMap<>();
```
HannesWell:
```suggestion
	private IProject fProject;
	private Set<IType> fTypes = new LinkedHashSet<>();
```
The `Map`'s value are again not used and a Set allows using a simple for-loop below.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+
+	@Override
+	protected boolean initialize(Object element) {
+		if (element instanceof IType type) {
+			IJavaProject javaProject = type.getJavaProject();
+			if (javaProject != null) {
+				IProject project = javaProject.getProject();
+				if (WorkspaceModelManager.isPluginProject(project)) {
+					fProject = project;
+					fElements.put(type, type.getFullyQualifiedName());
```
HannesWell:
```suggestion
				fProject = javaProject.getProject();
				if (WorkspaceModelManager.isPluginProject(fProject)) {
					fTypes.add(type);
```
An extra variable for the project is again not necessary.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+ * @since 3.16
+ */
+public class ManifestTypeDeleteParticipant extends PDEDeleteParticipant {
+
+	private static final Set<String> HANDLED_FILE_EXTENSTIONS = Set.of("java", "class"); //$NON-NLS-1$ //$NON-NLS-2$
+	protected HashMap<IType, String> fElements = new HashMap<>();
+
+	@Override
+	protected boolean initialize(Object element) {
+		if (element instanceof IType type) {
```
HannesWell: Like in https://github.com/eclipse-pde/eclipse.pde/pull/2292/changes#r3212865792, I don't think the `instanceof` check is necessary and we can again assume that the received element is an `IType`.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+				if (willPackageBeEmpty(pkg, deletedCUs)) {
+					Change change = BundleManifestChange.createEmptyPackageChange(file, pkg.getElementName(), pm);
+					if (change != null) {
+						result.add(change);
+					}
+				}
+			} catch (CoreException e) {
+				PDEPlugin.logException(e);
+			}
+		});
```
HannesWell: Couldn't we bulk the package removal here again?

```suggestion
		List<IPackageFragment> emptiedPackages = deletedCUsByPackage.entrySet().stream().filter(e -> {
			IPackageFragment pkg = e.getKey();
			Set<ICompilationUnit> deletedCUs = e.getValue();
			try {
				return willPackageBeEmpty(pkg, deletedCUs);
			} catch (CoreException ex) {
				return false;
			}
		}).map(Entry::getKey).toList();

		Change change = BundleManifestChange.createDeletePackagesChange(file, emptiedPackages, pm);
		if (change != null) {
			result.add(change);
		}
```

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+		// Check for non-Java resources (properties files, XML files, etc.)
+		Object[] nonJavaResources = pkg.getNonJavaResources();
+		if (nonJavaResources != null && nonJavaResources.length > 0) {
+			return false;
+		}
+
+		// Check if any compilation unit in the package is NOT being deleted
+		IJavaElement[] javaChildren = pkg.getChildren();
+		for (IJavaElement child : javaChildren) {
+			if (child instanceof ICompilationUnit cu) {
```
HannesWell:
```suggestion
		ICompilationUnit[] compilationUnits = pkg.getCompilationUnits();
		for (ICompilationUnit cu : compilationUnits) {
```
This would save the instanceof check here, though it has a little overhead due to the filtering done in JDT, but I think that's not a problem in this case.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/ManifestTypeDeleteParticipant.java
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+
+		// Check the underlying folder for any other files (non-Java resources at file level)
+		if (pkg.getCorrespondingResource() instanceof IFolder folder) {
+			// Check if any other file (except the already checked source and
+			// compiled Java class files) are contained in the packages folder.
+			return Arrays.stream(folder.members()).filter(IFile.class::isInstance).map(IFile.class::cast)
+					.map(IFile::getFileExtension).allMatch(HANDLED_FILE_EXTENSTIONS::contains);
+		}
```
HannesWell: With the check of `pkg.getNonJavaResources()` in the beginning, I wonder now if this makes any sense? Isn't this already covered by the former?

HannesWell (commented):
Thank you @nburnwal09 for the updates, they look good.
I made a second pass through this considering the now included `ManifestPackageDeleteParticipant` and also looked again at the previous changes.
Please see the comments below.
With them resolved, I expect this to be ready.

Please squash all changes into one commit named, e.g.:
`Handle class and package deletion for removing empty exported packages`.

Furthermore, just a nitpick, please add new lines at the end of the new files. Github/git shows annoying markers at files without newline at the end.

nburnwal09:
@HannesWell 
Sorry for the delay. I was occupied with a hackathon at my company and couldn’t focus on this earlier.
I have now addressed all the comments. Kindly review the changes at your convenience.

HannesWell:
> Sorry for the delay. I was occupied with a hackathon at my company and couldn’t focus on this earlier.
> I have now addressed all the comments. Kindly review the changes at your convenience.

No problem and thanks for the update. I plan to review it later today.

@MohananRahul could you please hold-off M3 promotion planed for today? I'd like to have this (and maybe another change) included. I can run the promotion myself later today or tomorrow morning (if an extra I-build is not reasonable).

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/refactoring/BundleManifestChange.java
HannesWell: Just pushed an update where `BundleManifestChange.createEmptyPackageChange()` is removed and the existing caller is replaced by the new method. So effectively the existing method is extended to handle multiple packages.

HannesWell (approved):
Thank you for all the updates. This is a great enhancement and I'm looking forward to have it.
It looks very good now and is ready for submission (just in time for M3).

PR #2293, opened 2026-04-17
Fix error with search in XMI-tab of E4WorkbenchModelEditor

r-mennig:
Currently there is an issue with the search functionality inside of the XMI tab of the E4WorkbenchModelEditor: When searching for a specific text, the search "lags" behind the last entered character and thus only highlights the previously entered text.

SearchLag

This is caused by the search highlighting being recalculated on a keyPressed-Event. This event happens before the text value of the search field is recalculated. In order to fix this issue, the search highlighting should instead react to modification events of the underlying text field (in cases in which the entered character causes the text field to update).

How to reproduce/test:
1. Create a new application model (or navigate to any existing application model)
2. Inside the E4WorkbenchModelEditor, select the XMI tab
3. Select the search field and search for any text shown in the XMI (e.g. "addons") -> The problem happens, the highlighting is always one character behind the search text

I've tried adding some tests for this PR, but I could not find any test plugin for org.eclipse.e4.tools.emf.ui.

image

Should I create a follow-up PR adding a new test plugin with some tests for this PR?


vogella:
LGTM, thanks

vogella:
Follow-up PR with tests would be great.

PR #2296, opened 2026-04-19
Append source location to resolved bundles in target editor

vogella:
## Summary

- Adds an optional `ITargetDefinition` context to `StyledBundleLabelProvider`. When set, each resolved `TargetBundle` label is suffixed with a short description of its originating `ITargetLocation` (directory path, feature id, or first IU repository URI), so users can tell apart entries that share the same symbolic name and version but come from different locations.
- Wires `TargetContentsGroup` (the Content tab of the target-definition editor) to supply that context whenever a target is loaded or cancelled, so the suffix follows the current editor input.
- Adds JUnit coverage for the label provider in a new `StyledBundleLabelProviderTests` class, added to `AllTargetTests`.

Answers the common user question: "I see this plug-in twice on the Content tab — where is this coming from?"

## Test plan
- [x] `AllTargetTests` suite passes locally (`mvn -pl ui/org.eclipse.pde.ui.tests -am -P javac -Dtycho.surefire.useUIHarness=true -Dtest=AllTargetTests verify`) — 127 tests, 0 failures, 0 errors.
- [x] New `StyledBundleLabelProviderTests` covers: no suffix without context, suffix present with context, distinct suffixes for bundles resolved from two different directory locations, context clearing, and preservation of the base symbolic-name label.
- [ ] Manual smoke in the target editor — open a target with two locations that contribute the same bundle, verify each row shows its originating location.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/AllTargetTests.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/AllTargetTests.java
@@ -25,7 +25,8 @@ import org.junit.platform.suite.api.Suite;
 		TargetDefinitionResolutionTests.class, //
 		TargetDefinitionFeatureResolutionTests.class, //
 		IUBundleContainerTests.class, //
-		ProfileContainerTests.class })
+		ProfileContainerTests.class, //
+		StyledBundleLabelProviderTests.class })
 public class AllTargetTests {
 
 }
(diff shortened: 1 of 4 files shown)


HannesWell:
> Answers the common user question: \"I see this plug-in twice on the Content tab — where is this coming from?\

Can't this question also be answered by using the 'sort by location' feature of the contents tab?

Could you please provide a screenshot of the new UI.

vogella:
> > Answers the common user question: "I see this plug-in twice on the Content tab — where is this coming from?\
> 
> Can't this question also be answered by using the 'sort by location' feature of the contents tab?

Definitely similar but you loose the plug-ins are close together option.
 
> Could you please provide a screenshot of the new UI.

image

vogella:
Maybe the label is a bit to much...

@HannesWell would it be OK to activate the grouping by default?

vogella:
Abandoned in favor of https://github.com/eclipse-pde/eclipse.pde/pull/2301

PR #2297, opened 2026-04-19
Fixes corruption of e4xmi files when class is renamed

p-O-q:
The corruption happens in a setup with a top level Maven project and plug-in projects in sub directories (eg bundles). If the name of a Java handler class is renamed in the Java file 
- the e4xmi file is found twice with
  - full workspace path
  - plug-in project path

This leads to 2 changes in one e4xmi file with the same text, offset and length.  
If the old and new names are of different length the corruption occurs.  

This fix skips the file if any of the paths (workspace or project) are already included for renaming.

### DIFF
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
@@ -15,8 +15,11 @@
  ******************************************************************************/
 package org.eclipse.e4.tools.emf.editor3x.refactorparticipants;
 
+import java.net.URI;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
@@ -57,11 +60,22 @@ class RefactorParticipantDelegate {
 		final FileTextSearchScope scope = FileTextSearchScope.newWorkspaceScope(filenames, false);
 
 		final Map<IFile, TextFileChange> changes = new HashMap<>();
+		final Set<ChangeAttributes> uniqueChangeAttributes = new HashSet<>();
 		final TextSearchRequestor searchRequestor = new TextSearchRequestor() {
 
 			@Override
 			public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws CoreException {
 				final IFile file = matchAccess.getFile();
+
+				final ChangeAttributes changeAttributes = new ChangeAttributes(
+						file.getLocationURI(),
+						pModel.getNewTextCurrentIndex(),
+						matchAccess.getMatchOffset(),
+						matchAccess.getMatchLength());
+				if (!uniqueChangeAttributes.add(changeAttributes)) {
+					return false;
+				}
+
 				TextFileChange change = changes.get(file);
 
 				if (change == null) {
@@ -97,12 +111,13 @@ class RefactorParticipantDelegate {
 				}
 
 				final ReplaceEdit edit = new ReplaceEdit(
-						matchAccess.getMatchOffset(),
-						matchAccess.getMatchLength(),
-						pModel.getNewTextCurrentIndex());
+						changeAttributes.offset(),
+						changeAttributes.length(),
+						changeAttributes.text());
 				change.addEdit(edit);
 				change.addTextEditGroup(new TextEditGroup(E4_MODEL_CHANGES,
 						edit));
+
 				return true;
 			}
 		};
@@ -130,5 +145,18 @@ class RefactorParticipantDelegate {
 			result.add(c);
 		}
 		return result;
+	}
+
+	/**
+	 * This record contains attributes for a single change. A change has to be
+	 * unique.
+	 * <p>
+	 * The {@code location} is an {@link IFile#getLocationURI()}, which may be
+	 * {@code null} for files without a local file-system location. A
+	 * {@code null} location is treated like any other value: two records with
+	 * {@code null} locations and otherwise equal fields are considered
+	 * duplicates. For e4xmi files on disk this case does not arise in practice.
+	 */
+	private static record ChangeAttributes(URI location, String text, int offset, int length) {
 	}
 }
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/META-INF/MANIFEST.MF
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.e4.tools.emf.editor3x;singleton:=true
-Bundle-Version: 4.10.200.qualifier
+Bundle-Version: 4.10.300.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Require-Bundle: org.eclipse.ui;bundle-version="3.6.0",
  org.eclipse.core.runtime;bundle-version="3.29.0",


p-O-q:
I'm not sure that I understand why the commit got "unverified"...

merks:
Each push does a build and because you're not contributed before, someone needs to approve the workflows which I've just done and did the first time too.

p-O-q:
@merks Thank you for your approval. Is there anything I can do to help?

merks:
Might it be better to check something like the `org.eclipse.core.resources.IResource.getLocationURI()` being the same?   Can you give an example of the two different IFile instances that you're seeing in this scenario?  I don't really have a good mental picture of how this arises...

p-O-q:
@merks Thank you for your effort!

I try to explain using screenshots of a debugger session.

The Eclipse instance the pictures are taken from is the instance that has started a platform ide.  
The platform ide instance has opened the provided/attached Maven project and is processing the rename operation.  
The class SaveHandler is renamed to SavHandler. 

The 
```
return false;
```
in line 66 (part of my fix) is commented out. So we see the original behavior.  

The renaming operation is stopped at the breakpoint in line 146 where the changes are created.

Both pictures show the value of the variable "changes" created in line 59:
```
final Map<IFile, TextFileChange> changes = new HashMap<>();
```

The first picture highlights the first key value of type IFile: it is the project path of the e4xmi file.
The second picture highlights the second key value of type IFile: it is the workspace path of the e4xmi file.

The content of both ReplaceEdit's are equal (text, offset, length).

Later when the ReplaceEdit's are applied the first ReplaceEdit changes the e4xmi file
- from offset 3902 the next 82 characters will be replaced with 'bundleclass://de...handlers.SavHandler'

Now the e4xmi file is changed correctly.

At this point the second ReplaceEdit does exactly the same:
- from offset 3902 the next 82 characters will be replaced with 'bundleclass://de...handlers.SavHandler'

The string to replace has already been replaced and is now shorter (81 chars).
But 82 chars were replaced.
Now the e4xmi file is corrupted.


That also means both paths (workspace and project) are resolved correctly. In this case it does not matter
which of the paths are used for renaming. So in my fix the first path (and ReplaceEdit) that comes in for an e3xmi file will be used.

Using IFile (IResource).getLocationURI() returns the absolute path. For both e4xmi files the absolute has to be the same. So, yes, comparing two URI's is smarter. Thank you :) I'll update today.

Pls don't hesitate to ask for more information. Thank you in advance.


e4xmi-issue-2297-1-of-2
e4xmi-issue-2297-2-of-2

p-O-q:
@merks The code looks much better now (don't know why I used a match var). Thank you.

Could I explain the situation in a bit more detail?

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -105,6 +108,19 @@ public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws Core
...
+			 * Tests whether the given file is already included in the {@code changes} map.
+			 *
+			 * @param file
+			 *            a file to be tested
+			 * @return true if the file is already included in the {@code changes} map
+			 *         otherwise false
+			 */
+			private boolean isFileAlreadyIncluded(IFile file) {
+				return changes.entrySet().stream().anyMatch(
+						e -> e.getKey().getLocationURI().equals(file.getLocationURI()));
```
merks: It seems it's possible for `getLocationURI()` to return null, so one might use Objects.equals but then I"m not sure we'd want null to match null.

Probably better to call file.getLocationURI once before line 121, check for null, returning false in that case, and use the non-null URI on the left side of equals test.

merks (changes requested):

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
merks: I think it should be more like this:
```
			private boolean isFileAlreadyIncluded(IFile file) {
				URI locationURI = file.getLocationURI();
				return locationURI != null
						&& changes.entrySet().stream().anyMatch(e -> locationURI.equals(e.getKey().getLocationURI()));
			}
```

p-O-q: If the call file.getLocationURI returns null (and the file is not omitted) the issue may occur again.
Is file.getLocationURI the right method to call?
There seem to be a lot of possibilities the return value may be null.

In contrast file.getFullPath() never returns null. The code doesn't look that nice b/c the project or the workspace path may come first and one is a substring of the other. However, there are no null values that need to be taken into account.

What do you think?

merks: I tried to determine when `getLocationURI` returns null, but there are many layers involved.  I've never seen it actually return null.  It might happen for example if the project doesn't exist:
```
	public URI getLocationURI() {
		IProject project = getProject();
		if (project != null && !project.exists()) {
			return null;
		}
```
So that's not  really likely to happen here in this situation; arguably it cannot happen here.  I think trying to write to such a file would fail anyway...

Your testing seems to be focused on "does this IFile represent the same underlying file as this other IFile" so I think the locationURI of the IFile is the best semantic match for that.  Two "is prefix" tests seems confusing to me.  Also, moving the getLocationURI out of the loop is a performance enhancement...

p-O-q: Ok, afais returning false if getLocationURI returns null is the current/old behavior. You are right with the best semantic match and - for sure - with moving getLocationURI out of the loop.
Thank you.

merks (approved):
It looks correct to me.

merks:
@laeubi @HannesWell 

Would at least one of you like to have a quick look?

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -105,6 +108,22 @@ public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws Core
...
+			 * Tests whether the given file is already included in the
+			 * {@code changes} map.
+			 *
+			 * @param file
+			 *            a file to be tested
+			 * @return true if the file has a location URI and is already
+			 *         included in the {@code changes} map otherwise false
+			 */
+			private boolean isFileAlreadyIncluded(IFile file) {
+				final URI uri = file.getLocationURI();
```
merks: I guess it's obvious that you missed adding the import.

merks (changes requested):

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
merks: Others are likely to complain that final isn't needed.  It's implicitly final and if it weren't the lamba would complain that it's not final.

p-O-q: Yes, my bad. Take your time to review...

p-O-q:
@merks Pls let us wait until tomorrow.
I found something I'd like to analyze that could be related to my changes.
I'll get back to you.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
p-O-q: > Others are likely to complain that final isn't needed. It's implicitly final and if it weren't the lamba would complain that it's not final.

You are right. In the projects I've worked on we usually enabled the setting to add final wherever possible. 
So a var not being final is something that attracts attention.
If you like I'll remove final. What fits better?

p-O-q:
@merks The fix does not work for project renaming. Project renaming expects a series of changes for the same e4xmi file. 

The fix should ensure that all TextEdit instances are applied to one e4xmi file only once.
That means for a physical e4xmi file all the TextEdit (ReplaceEdit) instance have to be unique (text, length, offset).

Maybe even the type of the TextFileChange 'change' has to be considered.

What do you think?

merks:
I think it would be good to share a reproducible sample. It sounds like the think gathering the resources to change should be more careful about dealing with changes to the same underlying file system file.    Like the Search view has "Show Only Most Nested Match"for a similar purpose.

p-O-q:
@merks Pls take a look at issue #2298. There is a sample project attached and a short description how to reproduce.
Sorry, I missed to tell you.

> It sounds like the think gathering the resources to change should be more careful about dealing with changes to the same underlying file system file

If I'm right the tuple (physical file, TextFileChange, ReplaceEdit) has to be unique before creating the CompositeChange result. I'm not sure about the TextFileChange. Without considering the type of TextFileChange the attributes (file, text, offset, length) have to be unique.

The way the refactoring process works
- first gather all changes (file, text, offset, length)
- second apply all gathered changes

implies that none of the changes may be applied twice.

Does that sound reasonable?

merks:
I think so. FYI I’m at OCX this weekend so my time is limited next week. Perhaps have a look at how search filters “duplicate” search results.

p-O-q:
Using the four attributes (file, text, offset, length) to find present changes seems to work.
The renaming of a class works and now renaming a plug-in project results in the same amount of changes as before.

@laeubi @HannesWell what do you think?

(The renaming of a RCP plug-in project does not work. But that's another story.)

p-O-q:
@merks 
> I’m at OCX

Have a good time

laeubi:
It also happens to me almost always, maybe because of hierarchical projects or similar. So thanks for spotting the root cause here!

p-O-q:
@merks The PR overview says "Changes requested". Are you still requesting changes or haven't you had a chance to look yet (b/c of OCX)?

merks (commented):
I’m not sure how to clear my review from my phone. 

Please remove final keyword from all locations. 

I likely don’t have time to review until next week. 

@laeubi might have time sooner. I trust his judgment so if he approves, that’s sufficient.

p-O-q:
@merks  Ok, I'll remove the final keyword from locations and push again.

p-O-q:
@merks the "presentFileLocation" variable is not "effectively final". So I didn't remove the final keyword b/c I want it to be final.

("effectively final" not me rule! ;)

p-O-q:
@laeubi Could you pls trigger the build again? I don't know how to trigger a build without pushing.
 I don't think removing a final keyword would break the build.
Do you have time to review this PR?

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -103,8 +110,40 @@ public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws Core
...
+			 * {@code changes} map.
+			 *
+			 * @param file
+			 *            the file that will be changed
+			 * @param ma
+			 *            provides offset and length for the change
+			 * @return true if the change is already present and the location of
+			 *         the file is not null otherwise false
+			 */
+			private boolean isChangeAlreadyPresent(IFile file, TextSearchMatchAccess ma) {
```
vogella: Alternative design: key the outer `changes` map by `URI` (or keep a parallel `Set<URI>` of already-seen locations). Dedupe then becomes an O(1) lookup and this helper disappears entirely. Not a blocker — refactoring is not a hot path — but it would simplify the code noticeably.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -103,8 +110,40 @@ public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws Core
...
+				URI newFileLocation = file.getLocationURI();
```
vogella: When `getLocationURI()` returns `null`, this short-circuits to `false` and the caller will add a second edit. For e4xmi files on disk that cannot happen in practice, but a one-line comment stating why `null` is treated as "not duplicate" would make the intent explicit for future readers.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -103,8 +110,40 @@ public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws Core
...
+				return newFileLocation != null && changes.entrySet().stream()
+						.anyMatch(e -> {
+							final URI presentFileLocation = e.getValue().getFile().getLocationURI();
+							for (final TextEdit te : e.getValue().getEdit().getChildren()) {
+								if (te instanceof final ReplaceEdit re) {
+									if (re.getLength() == ma.getMatchLength()
+											&& re.getOffset() == ma.getMatchOffset()
+											&& Objects.equals(re.getText(), pModel.getNewTextCurrentIndex())
```
vogella: `pModel.getNewTextCurrentIndex()` is invoked once per visited entry inside the lambda. Hoisting it (and the `ma.getMatchOffset()` / `ma.getMatchLength()` values) into locals above `anyMatch` avoids the repeated calls and makes the predicate easier to read.

vogella (commented):
A few small suggestions — nothing blocking. The fix itself is correct.

vogella (changes requested):

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
p-O-q: @vogella Thank you for your comments. If I understand this one correctly and use the location as key I still need to iterate over the values. One file may have several changes. Even one line theoretically may have multiple (not overlapping) changes being handled in order from high to low offset.
Would it be a real advantage?

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
p-O-q: Ok

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
p-O-q: I don't have too much knowlege about "Compilerbau" but wouldn't the compiler optimize it that way under the hood?
I'll change that. Thank you!

merks: > I don't have too much knowlege about "Compilerbau" but wouldn't the compiler optimize it that way under the hood? I'll change that. Thank you!

No, the jit compiler will not generally determine that calling the same method on the same object twice will return the same result.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
vogella: By using a Set with a simple record like record MatchLocation(IFile file, int offset, int length), we get O(1) deduplication. This handles your concern about multiple changes in one file (since each unique offset/length is a distinct entry in the set) while completely removing the need for the isChangeAlreadyPresent helper and its associated loops. It makes the code more robust and idiomatic for Java 21.

p-O-q:
@vogella the idea with the set is really smart and avoids useless handling of null for getLocationURI call. Additionally it results in less code. Thank you.

vogella:
Follow-up on my earlier review comment about documenting the `null` URI behavior — here is a minimal addition you can append to the existing Javadoc on the `ChangeAttributes` record:

```java
	 * <p>
	 * The {@code location} is an {@link IFile#getLocationURI()}, which may be
	 * {@code null} for files without a local file-system location. A
	 * {@code null} location is treated like any other value: two records with
	 * {@code null} locations and otherwise equal fields are considered
	 * duplicates. For e4xmi files on disk this case does not arise in
	 * practice.
```

Feel free to take it verbatim or tweak the wording.

p-O-q:
@vogella I don't think that I can explain it better :)

vogella (approved):
LGTM, thanks. Go from my side

p-O-q:
@merks Is there anything I can do for this PR to be merged?

merks:
I’ll look when I arrive at my destination later in the day.

e4tools/bundles/org.eclipse.e4.tools.emf.editor3x/src/org/eclipse/e4/tools/emf/editor3x/refactorparticipants/RefactorParticipantDelegate.java
```diff
@@ -57,11 +60,23 @@ public static CompositeChange createChange(IProgressMonitor pProgressMonitor, fi
...
 			@Override
 			public boolean acceptPatternMatch(TextSearchMatchAccess matchAccess) throws CoreException {
 				final IFile file = matchAccess.getFile();
+
+				final ChangeAttributes changeAttributes = new ChangeAttributes(
+						file.getLocationURI(),
+						pModel.getNewTextCurrentIndex(),
+						matchAccess.getMatchOffset(),
+						matchAccess.getMatchLength());
+				if (uniqueChangeAttributes.contains(changeAttributes)) {
```
merks: This could be written as `if (!uniqueChangeAttributes.add(changeAttributes)) { return false; }` which is more efficient.

merks (changes requested):

merks:
@p-O-q 

Given there is no test, I'm kind of taking your word for it that this works correctly.  Based on previous comments I believe you have carefully confirmed this is fully functional..

p-O-q:
> @p-O-q
> 
> Given there is no test, I'm kind of taking your word for it that this works correctly. Based on previous comments I believe you have carefully confirmed this is fully functional..

Given the project setup described there is only one scenario in which the old code worked correctly. It worked well if the old and new text was of same length. In all other scenario changing the e4xmi file  twice resulted in a corrupt file.

merks:
I trust your judgment and your manual testing.  When I get back from dinner I hope to merge it.

merks (approved):
Looks good. I just want to see green builds now so I will update the branch to rebase it on master.

merks:
@p-O-q thanks for your patience and contribution.

p-O-q:
Thanks to all who contributed to this issue and increased the code quality.
@merks Thank you very much for the merge at this late hour.

PR #2301, opened 2026-04-22
Group target content by Location by default

vogella:
## Summary

- In the Content tab of the target-definition editor, default the "Group by" combo to **Location** instead of **None**.
- Makes it immediately obvious where each resolved bundle originates when a target has multiple locations, without adding any extra label text.

Alternative to #2296 (which suffixed the location to every bundle label). Based on review feedback that "sort/group by location" already answers the "where is this plug-in coming from?" question.

## Test plan
- [ ] Open a `.target` file in the editor, switch to the Content tab, verify that bundles are grouped by their originating location by default.
- [ ] Verify the "Group by" combo can still be switched to `None` / `File Path` and back.
- [ ] Verify feature mode still disables the grouping combo as before.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
@@ -209,7 +209,7 @@ public class TargetContentsGroup {
 	 * @param toolkit toolkit to create controls with
 	 */
 	protected void createFormContents(Composite parent, FormToolkit toolkit) {
-		fGrouping = GROUP_BY_NONE;
+		fGrouping = GROUP_BY_CONTAINER;
 
 		Composite comp = toolkit.createComposite(parent);
 		GridLayout layout = new GridLayout(2, false);
@@ -236,7 +236,7 @@ public class TargetContentsGroup {
 	 * @param parent parent composite
 	 */
 	protected void createDialogContents(Composite parent) {
-		fGrouping = GROUP_BY_NONE;
+		fGrouping = GROUP_BY_CONTAINER;
 
 		Composite comp = SWTFactory.createComposite(parent, 2, 1, GridData.FILL_BOTH, 0, 0);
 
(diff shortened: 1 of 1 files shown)


HannesWell:
> Makes it immediately obvious where each resolved bundle originates when a target has multiple locations, without adding any extra label text.

I'm not sure it's wanted in more cases to have the grouping enabled than it's wanted to have the list just sorted alphabetically.
Surely it can be interesting and relevant to group items by location to find the source, but I'm not sure it's desired instantly most of the time.
Du you think the grouping options are to difficult to discover?

On the other hand I agree that it can be perceived as strange that there can be multiple entries without any grouping. Maybe duplicates should be filtered in that case.

Nevertheless I'd like to gather more opinions on this whole matter. 
What's the opinion of others on this?

merks:
Showing duplicates with no way to distinguish them almost looks like a bug. Maybe they have different versions, but that’s not displayed (if I recall correctly). Better if the flat list did not show duplicates. I personally prefer to see the flat list because it’s mostly irrelevant (to me) where it comes from and grouping doesn’t provide a simple way to see everything.

vogella:
> Du you think the grouping options are to difficult to discover?

Yes it is hard to discover. I discovered it two weeks ago and all my clients were not aware of it (or never used it to find the dups). And I only discovered it, because I was looking at the code trying to implement it. :-o  

> Better if the flat list did not show duplicates.

No. The clients I'm working with want to see dups and frequently clean up the different places they come from, filtering them out would make their life much harder.

I have a local patch which implements "Show only duplicates" I think it not yet pushed.

To clarify: duplicates mean for me "same plug-in in the same version coming from different places" and "same plug-in in a different version". The second is something I like to avoid in RCP and the first also feels like it should be avoided.

merks:
Seems quite some differences of opinion and even different use cases for the purpose of the list. 

I would generally expect that if you use planner mode and have more that one location, you will very likely have duplicates (3rd party bundles) and can’t actually avoid them (because planner mode pulls in 3rd party dependencies transitively from each site). 

I forget, does it currently show the version?

vogella:
> I forget, does it currently show the version?

It? The entries in the tab? Yes

vogella:
Any objections against grouping by default? My understanding is that nobody had a strong opinion against it and it would help clients to identify potential issues. At the moment is looks like the UI is  buggy, as the same entries are listed multiple times if present.

merks:
As I said above

> I personally prefer to see the flat list because it’s mostly irrelevant (to me) where it comes from and grouping doesn’t provide a simple way to see everything.

@HannesWell said

> I'm not sure it's wanted in more cases to have the grouping enabled than it's wanted to have the list just sorted alphabetically.

So let me strengthen that to no, do not change the default. 

---

I already mentioned about duplicates:

> Better if the flat list did not show duplicates

Also @HannesWell 

> On the other hand I agree that it can be perceived as strange that there can be multiple entries without any grouping. Maybe duplicates should be filtered in that case.

So I think we all agree that showing duplicates (same ID and same version) is less than ideal.

PR #2302, opened 2026-04-22
Enable copying entries from the target editor Content tab

vogella:
## Summary

- Wire the existing `CopyTreeSelectionAction` into the Content tab of the target-definition editor (`TargetContentsGroup`), mirroring what is already done on the Locations tab (`TargetLocationsGroup`).
- Entries can now be copied from the Content tab via the context menu or **Ctrl+C**, with support for multi-selection and expanded children (parent/child dedup is handled by the action).

## Test plan
- [ ] Open a `.target` file and switch to the Content tab.
- [ ] Select one or more bundles and verify the context menu shows **Copy** and copies the selection to the clipboard.
- [ ] With a selection active, press **Ctrl+C** and paste — verify the selected entries are on the clipboard.
- [ ] Expand a group, select the group, copy, and verify children appear indented in the copied text.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
@@ -73,6 +73,8 @@ import org.eclipse.pde.internal.ui.shared.CachedCheckboxTreeViewer;
 import org.eclipse.pde.internal.ui.shared.FilteredCheckboxTree;
 import org.eclipse.pde.internal.ui.wizards.target.TargetDefinitionContentPage;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.KeyAdapter;
+import org.eclipse.swt.events.KeyEvent;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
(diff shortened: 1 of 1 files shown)


ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
```diff
@@ -309,13 +311,24 @@ public int compare(Viewer viewer, Object e1, Object e2) {
 
 		});
 
+		CopyTreeSelectionAction copySelectionAction = new CopyTreeSelectionAction(fTree.getTree());
+		fTree.getTree().addKeyListener(new KeyAdapter() {
+			@Override
+			public void keyPressed(KeyEvent e) {
+				if (e.keyCode == 'c' && (e.stateMask & SWT.CTRL) != 0) {
```
jonahgraham, line 318: This seems questionable. Copy isn't always CTRL C and sometimes CTRL C is bound to other commands. Why is this bound with a key press?

jonahgraham (changes requested):

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetContentsGroup.java
vogella: The pattern was copied from TargetLocationsGroup.initializeTree() (line 267), which is what the PR copied from:                                                                                                                                                
   
````
                                                                                                                                                                                                                                                
  TargetLocationsGroup.java:267-276:                                                                                                                                                                                                            
  tree.addKeyListener(new KeyAdapter() {      
      @Override                                                                                                                                                                                                                                 
      public void keyPressed(KeyEvent e) {    
          if (e.keyCode == SWT.DEL && fRemoveButton.getEnabled()) {                                                                                                                                                                             
              handleRemove();                                                                                                                                                                                                                   
          } else if (e.keyCode == 'c' && (e.stateMask & SWT.CTRL) != 0) {                                                                                                                                                                       
              fCopySelectionAction.run();                                                                                                                                                                                                       
          }                                                                                                                                                                                                                                     
      }                                                                                                                                                                                                                                         
  });      
````

We can change it but than I suggest we change both places. Can you please open a new issue for that?

vogella: Or just let me know that you want both places fixed, I can push a PR without issue, no need for that overhead.

merks: Should MOD1/2/3 be used instead?  Would this also kick in for ctrl+alt+c as well?  Is it really key down not up that should invoke the action?  I would have expected binding to the global actions would be required to use the user-configured binding.

vogella: The old and exiting code of the target editor hard coded that. See above. I anyhow want to improve multiple things here, so I can also improve this.

PR #2305, opened 2026-04-22
Prefer native Combo over CCombo in PDE form parts

### LINKED ISSUE #2304: Prefer the usage Combo in PDE instead of CCombo
CCombo looks strange (on Linux) and according to its Javadoc (CCombo.java:31-38) it is intended to support a flexible height.

````
CCombo was written to work around certain limitations in the native combo box. 
Specifically, on win32, the height of a CCombo can be set; attempts to set the height of a Combo are ignored. 
CCombo can be used anywhere that having the increased flexibility is more important than getting native L&F, but the decision should not be taken lightly. 
There is no strict requirement that CCombo look or behave the same as the native combo box.  
````

AFAICS the height support is not needed in the PDE editors:

Image

Image

vogella:
## Summary
- `ComboPart` (PDE UI and DS UI copies) and `ComboViewerPart` always create a native `Combo` instead of branching to `CCombo` when the `FormToolkit` is borderless.
- Dead `instanceof Combo`/`(CCombo)` downcast branches removed from `ArgumentsSection` and `EnvironmentSection`; unused `CCombo`/`Control` imports dropped.
- `CCombo` is kept only where structurally required — Table/Tree cell editors in `AbstractPluginBlock`, `RepositoryReferenceSection`, `UpdatesSection` and `PluginConfigurationSection`.

`CCombo`'s primary advantage over native `Combo` is a settable height (per its Javadoc), which PDE form editors don't need. Native `Combo` renders consistently with current OS themes on Linux/GTK and macOS, where `CCombo` looks dated.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2304

## Test plan
- [x] Open a target-definition editor → **Environment** tab: OS/WS/Arch/NL dropdowns render as native combos and still select/persist values.
- [x] Open a product editor → **Launching** tab: architecture dropdown in *Program and VM Arguments* section still switches per-arch arguments.
- [x] Open a product editor → **Overview** tab → extension ID combo (`ExtensionIdComboPart`) still adds/removes IDs correctly.
- [x] Open a DS component (`.xml`): **Options** section → Configuration Policy dropdown; *Add/Edit Reference* dialog → Cardinality/Policy dropdowns; *Add/Edit Property* dialog → Type dropdown — all render as native combos.
- [x] Launcher *Plug-ins* tab: Start Level / Auto-Start cell editors still open (still `CCombo`, intentionally).

### DIFF
--- a/ds/org.eclipse.pde.ds.ui/src/org/eclipse/pde/internal/ds/ui/parts/ComboPart.java
+++ b/ds/org.eclipse.pde.ds.ui/src/org/eclipse/pde/internal/ds/ui/parts/ComboPart.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008 Code 9 Corporation and others.
+ * Copyright (c) 2008, 2026 Code 9 Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -15,8 +15,6 @@
  *******************************************************************************/
 package org.eclipse.pde.internal.ds.ui.parts;
 
-import org.eclipse.swt.SWT;
-import org.eclipse.swt.custom.CCombo;
 import org.eclipse.swt.events.ModifyListener;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.widgets.Combo;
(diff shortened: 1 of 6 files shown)


vogella:
Planning to merge after the verification build passes

PR #2306, opened 2026-04-22
Wire target editor Copy through workbench binding service

vogella:
## Summary

Addresses review feedback from @jonahgraham and @merks on #2302.

Both `TargetContentsGroup` and `TargetLocationsGroup` had a `KeyAdapter` that hard-coded `e.keyCode == 'c' && (e.stateMask & SWT.CTRL) != 0` to run the copy action. That check:

- is wrong on macOS (should be Cmd+C, not Ctrl+C),
- fires for any modifier combination that has Ctrl set (e.g. **Ctrl+Alt+C**, Ctrl+Shift+C, ...),
- bypasses any user-configured keybinding.

`CopyTreeSelectionAction` already declares `setActionDefinitionId(ActionFactory.COPY.getCommandId())`, so the proper wiring is to register it as the editor's global `ActionFactory.COPY` handler while the tree has focus, and let the platform binding service deliver the correct keystroke.

### Changes

- Drop the `KeyAdapter` Ctrl+C branch in both groups (the `SWT.DEL` remove branch in `TargetLocationsGroup` stays).
- Promote the copy action to a field in `TargetContentsGroup`.
- Add `hookGlobalActions(IActionBars)` on both groups — on tree focus gained install the COPY handler, on focus lost clear it, so the binding doesn't hijack Copy from other widgets in the editor.
- Call `hookGlobalActions(...)` from `ContentSection` and `LocationsSection` with `fEditor.getEditorSite().getActionBars()`.

### Tradeoff

The wizard dialog path (`TargetDefinitionContentPage`, using `createInDialog`) has no `IActionBars` and therefore loses the Ctrl+C keystroke on the tree. **Right-click → Copy still works** there, since the context menu is wired identically. A follow-up can add a dialog-scoped binding if that path turns out to matter.

## Test plan

- [ ] Open a `.target` file → Content tab. Select one or more bundles, press the platform Copy shortcut (Ctrl+C on Linux/Windows, Cmd+C on macOS), paste into an editor — selection copies.
- [ ] Same on the Locations tab.
- [ ] With focus in a combo/text field in the editor, press Ctrl+C — does not trigger the tree copy action.
- [ ] Remap `org.eclipse.ui.edit.copy` in **Preferences → General → Keys**, focus the tree, use the new binding — works.
- [ ] Right-click → Copy still works on both tabs (editor and wizard).
- [ ] In `TargetLocationsGroup`, pressing `Delete` on a selected location still triggers remove.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
@@ -64,6 +64,8 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Tree;
+import org.eclipse.ui.IActionBars;
+import org.eclipse.ui.actions.ActionFactory;
 import org.eclipse.ui.forms.widgets.FormToolkit;
 import org.eclipse.ui.progress.UIJob;
 
@@ -269,11 +271,19 @@ public class TargetLocationsGroup {
 			public void keyPressed(KeyEvent e) {
 				if (e.keyCode == SWT.DEL && fRemoveButton.getEnabled()) {
 					handleRemove();
-				} else if (e.keyCode == 'c' && (e.stateMask & SWT.CTRL) != 0) {
-					fCopySelectionAction.run();
 				}
 			}
 		});
+	}
+
+	/**
+	 * Installs the copy action as the {@link ActionFactory#COPY global copy
+	 * handler} for the hosting editor while the tree has focus, so that the
+	 * platform binding service delivers the platform-correct Copy keystroke
+	 * (Ctrl+C / Cmd+C / user-remapped binding).
+	 */
+	public void hookGlobalActions(IActionBars bars) {
+		CopyTreeSelectionAction.hookAsGlobalCopyHandler(fTreeViewer.getTree(), fCopySelectionAction, bars);
 	}
 
 	/**
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/ContentSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/ContentSection.java
@@ -67,6 +67,7 @@ public class ContentSection extends SectionPart {
 		client.setLayoutData(new GridData(GridData.FILL_BOTH | GridData.GRAB_VERTICAL));
 
 		fContentGroup = TargetContentsGroup.createInForm(client, toolkit);
+		fContentGroup.hookGlobalActions(fEditor.getEditorSite().getActionBars());
 		fEditor.getTargetChangedListener().setContentTree(fContentGroup);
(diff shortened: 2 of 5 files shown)


vogella:
@jonahgraham and @merks FYI

jonahgraham:
Thanks @vogella for working on this some more. I don't have access to the three referenced platforms in your test plan to try this out on (I'm on my phone at OCX). Have you found it now works on the three platforms?

vogella:
> Thanks @vogella for working on this some more. I don't have access to the three referenced platforms in your test plan to try this out on (I'm on my phone at OCX). Have you found it now works on the three platforms?

No, I only have access to Linux at the moment. I can leave this open in case you are interested in testing this on the other.

jonahgraham:
If this works well on Linux, I would be fine with pushing it forward based on my code inspection. 

FWIW I find the size of comment not very concise (the temptation to ask ai to summarize is strong). Perhaps as you use ai more you get used to just asking the ai to summarize too. 🤷‍♂️

vogella:
> FWIW I find the size of comment not very concise (the temptation to ask ai to summarize is strong). Perhaps as you use ai more you get used to just asking the ai to summarize too. 🤷‍♂️

I try to make them shorter again. I used to have very short PR descriptions and relied on the commit message. But the rules for PR description were recently changed (at least for draft PR) to be more self-containted without the commit messages. I forgot in which repo it was document but I remember is was a looooooong text describing what must be in the PR description.

But thanks for the suggestion, I try to make PR shorter again in the future.

vogella:
@merks as you suggested this change, is it now ok for you?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
```diff
@@ -269,13 +273,37 @@ private void initializeTree(Tree tree) {
...
+	/**
+	 * Installs the copy action as the {@link ActionFactory#COPY global copy
+	 * handler} for the hosting editor while the tree has focus, so that the
+	 * platform binding service delivers the platform-correct Copy keystroke
+	 * (Ctrl+C / Cmd+C / user-remapped binding).
+	 */
+	public void hookGlobalActions(IActionBars bars) {
+		if (bars == null || fCopySelectionAction == null) {
+			return;
+		}
```
merks: Is there some circumstance why/when this would be called in this incomplete state?   I mean you might argue better safe than sorry, but it's also a case of being quietly dysfunctional.

merks (commented):
I tested it and it works well.  It's unfortunate that there are two copies of this method.  The null guards seem pointless because it's never called with nulls and should never have nulls.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/shared/target/TargetLocationsGroup.java
vogella: I think this can be removed, will be done.

vogella:
> I tested it and it works well. It's unfortunate that there are two copies of this method. The null guards seem pointless because it's never called with nulls and should never have nulls.

I put the code into a helper method

merks (commented):
That’s nicer. Just an unused import warning if I read it correctly on my iPhone.

merks:
Please squash to a single commit.

PR #2312, opened 2026-04-24
Handle package deletion scenario for cleaning up exported packages

nburnwal09:
Enhancement added for handling package deletion scenario.
When a package is deleted via delete operation, the package should be automatically removed from the Export-Package header in MANIFEST.MF if it was previously exported.

https://github.com/user-attachments/assets/4df5b5af-eba1-41a7-a24f-9ee9d6820408


nburnwal09:
This is a follow up of the PR: https://github.com/eclipse-pde/eclipse.pde/pull/2292

HannesWell:
This is now included in
- https://github.com/eclipse-pde/eclipse.pde/pull/2292

and therefore can be closed.
Thank you for it, nevertheless.

PR #2313, opened 2026-04-24
Make transitive dependency classpath order deterministic

vogella:
When two bundles share a symbolic name in the target platform (e.g. both `jakarta.annotation-api` 1.3.5 and 2.1.1 shipped by the SDK side by side), the sort used to emit classpath entries only compared symbolic names. `Stream.sorted` is stable, so the source order — coming from a `HashSet` iteration that varies between JVM runs — leaked into the classpath, which made the "Updating plug-in dependencies" job rewrite Plug-in Dependencies containers on every IDE restart.

Adding a `HIGHER_VERSION_FIRST` tiebreaker makes the order stable across restarts and matches the preference PDE's resolver selection policy already uses. After this, restarting twice against the same target reports `updated 0 projects` instead of dozens.

Includes a Mockito-based unit test that exercises the comparator with two bundles sharing a symbolic name in ascending and descending source order and asserts the higher-version-first result is independent of input order.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
@@ -11,6 +11,7 @@
  *  Contributors:
  *     IBM Corporation - initial API and implementation
  *     Christoph Läubrich - add bnd support
+ *     Lars Vogel <Lars.Vogel@vogella.com> - deterministic classpath order for duplicate bundle symbolic names
  *******************************************************************************/
 package org.eclipse.pde.internal.core;
 
@@ -97,6 +98,11 @@ class RequiredPluginsClasspathContainer {
 
 	private static final Comparator<BundleDescription> BUNDLE_VERSION = Comparator
 			.comparing(BundleDescription::getVersion);
+
+	private static final Comparator<BundleDescription> SYMBOLIC_NAME_THEN_HIGHER_VERSION = Comparator
+			.comparing(BundleDescription::getSymbolicName)
+			.thenComparing(Comparator.comparing(BundleDescription::getVersion).reversed())
+			.thenComparingLong(BundleDescription::getBundleId);
 
 	private final IPluginModelBase fModel;
 	private final IBuild fBuild;
@@ -640,8 +646,11 @@ class RequiredPluginsClasspathContainer {
 	private static List<BundleDescription> collectRequirements(List<BundleDescription> roots) {
 		var closure = DependencyManager.findRequirementsClosure(roots, INCLUDE_OPTIONAL_DEPENDENCIES);
 		String systemBundleBSN = TargetPlatformHelper.getPDEState().getSystemBundle();
+		// Tie-break by higher version, then bundle id, so the classpath stays stable
+		// across IDE restarts when multiple bundles share a symbolic name
 		return closure.stream().filter(b -> !b.getSymbolicName().equals(systemBundleBSN))
-				.sorted(Comparator.comparing(BundleDescription::getSymbolicName)).toList();
+				.sorted(SYMBOLIC_NAME_THEN_HIGHER_VERSION)
+				.toList();
 	}
 
 	private void addSecondaryDependencies(BundleDescription desc, Set<BundleDescription> added,
@@ -685,7 +694,7 @@ class RequiredPluginsClasspathContainer {
 		String systemBundleBSN = TargetPlatformHelper.getPDEState().getSystemBundle();
 		Iterator<BundleDescription> transitiveDeps = closure.stream()
 				.filter(desc -> !desc.getSymbolicName().equals(systemBundleBSN))
-				.sorted(Comparator.comparing(BundleDescription::getSymbolicName)).iterator();
+				.sorted(SYMBOLIC_NAME_THEN_HIGHER_VERSION).iterator();
 		while (transitiveDeps.hasNext()) {
 			BundleDescription desc = transitiveDeps.next();
 			if (added.add(desc)) {


merks:
This sounds good. 👍

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -641,7 +651,8 @@ private static List<BundleDescription> collectRequirements(List<BundleDescriptio
 		var closure = DependencyManager.findRequirementsClosure(roots, INCLUDE_OPTIONAL_DEPENDENCIES);
 		String systemBundleBSN = TargetPlatformHelper.getPDEState().getSystemBundle();
 		return closure.stream().filter(b -> !b.getSymbolicName().equals(systemBundleBSN))
-				.sorted(Comparator.comparing(BundleDescription::getSymbolicName)).toList();
+				.sorted(SYMBOLIC_NAME_THEN_HIGHER_VERSION)
```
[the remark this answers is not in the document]
vogella: ack

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Lars Vogel <Lars.Vogel@vogella.com> - initial API and implementation
```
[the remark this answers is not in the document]
vogella: ack

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -98,6 +99,15 @@ class RequiredPluginsClasspathContainer {
...
 			.comparing(BundleDescription::getVersion);
 
+	/**
+	 * Order for transitive dependency classpath entries: symbolic name first,
+	 * higher version first as tiebreaker so the classpath is deterministic
+	 * when two bundles share a symbolic name (e.g. two versions of
+	 * {@code jakarta.annotation-api} coexist in the target).
+	 */
+	static final Comparator<BundleDescription> SYMBOLIC_NAME_THEN_HIGHER_VERSION = Comparator
+			.comparing(BundleDescription::getSymbolicName).thenComparing(BUNDLE_VERSION.reversed());
```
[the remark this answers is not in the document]
vogella: ack

merks:
Is it possible to complete these activities in a branch on your fork instead of a broadcasting PR?

vogella:
> Is it possible to complete these activities in a branch on your fork instead of a broadcasting PR?

No, as I'm reacting to feedback from the repeated Copilot reviewer which I assume @iloveeclipse triggers.

iloveeclipse:
> which I assume @iloveeclipse triggers.

Correct. So what you see is just a usual review process, which happened to be "automated" in some kind via AI. Unfortunately I have no time to do that in person.

vogella:
> Correct. So what you see is just a usual review process, which happened to be "automated" in some kind via AI. Unfortunately I have no time to do that in person.

And the Copilot feedback was good and helped improving the change.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -98,6 +99,19 @@ class RequiredPluginsClasspathContainer {
...
 
+	/**
+	 * Order for transitive dependency classpath entries: symbolic name first,
+	 * higher version next, and bundle id last as a final tiebreaker so the
+	 * ordering is a total order even when two different bundles share the same
+	 * symbolic name and version (e.g. duplicates from different suppliers in
+	 * the target). Without the bundle-id tiebreaker, {@code Stream.sorted}
+	 * would fall back to the (non-deterministic) source iteration order from
+	 * a {@code HashSet}.
+	 */
```
HannesWell: IMO it's not necessary to explain the usage again here, especially in that detail. 
The reasoning could be added at the caller and if a comment is necessary here, it should just explain the comparator (if not clear from the code).

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+	@SuppressWarnings("unchecked")
+	public static void loadComparator() throws Exception {
+		Class<?> container = Class.forName("org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer");
+		Field field = container.getDeclaredField("SYMBOLIC_NAME_THEN_HIGHER_VERSION");
+		field.setAccessible(true);
+		comparator = (Comparator<BundleDescription>) field.get(null);
+	}
+
+	@Test
+	public void higherVersionFirstForSameSymbolicName() {
```
HannesWell: This test case could be added to the existing `RequiredPluginsClasspathContainerTest`. An extra class for that seems heavy weight.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+	private static Comparator<BundleDescription> comparator;
+
+	@BeforeClass
+	@SuppressWarnings("unchecked")
+	public static void loadComparator() throws Exception {
+		Class<?> container = Class.forName("org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer");
+		Field field = container.getDeclaredField("SYMBOLIC_NAME_THEN_HIGHER_VERSION");
+		field.setAccessible(true);
+		comparator = (Comparator<BundleDescription>) field.get(null);
+	}
```
HannesWell: The field could be made accessible for tests (packe private).
Adding such hacks is IMO not good.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
vogella: This bug exists since a really long time and its "Re-calculation of the bundle class path" affects a lot of plug-in developers". Basically everyone who uses platform as target platform must have experienced this bug. Most of my clients see a full rebuild at start of Eclipse, finally this will be fixed. So I think a bit of documentation why this new code is added is good, as the problem was so long in the code. But I simplify the description a bit.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
vogella: ack

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
vogella: ack

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
```diff
@@ -76,7 +77,7 @@
 import aQute.bnd.build.Workspace;
 import aQute.bnd.osgi.Constants;
 
-class RequiredPluginsClasspathContainer {
+public class RequiredPluginsClasspathContainer {
```
laeubi: Please don't make this class public, it is not intended to be used outside this package!

laeubi (dismissed):
In general it is actually wrong that PDE sorts the bundles but we historically do this so it does not make things worse here.

laeubi:
> Adding a HIGHER_VERSION_FIRST tiebreaker makes the order stable across restarts and matches the preference PDE's resolver selection policy already uses.

I just wanted to make clear that this assumption is over-simplified if not wrong! The OSGi resolver only chooses the higher version *if they are both valid providers* and then based on bundle id. Also this is not a _preference_ but a matter of class-space consistency (what PDE here not checks and is usually irrelevant for compile time).

Also it depend if you use import-package or require bundle and so on, so actually PDE should not reorder things at all and the collection of transitive closure must use a Linked set and ordering by symbolic name in the first place is wrong. Anyways this is current-state of the art and ordering *additionally* by version won't harm in the majority of cases here.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/RequiredPluginsClasspathContainer.java
vogella: @HannesWell requested that for the test

laeubi: He requested to make the comparator package private, not this class public. In any case I doubt the test brings much anyways even if it would test the comparator it does not guarantee it is actually ever used.

So instead of testing internal behavior we can either assume that this is trivial enough not needing a test or you should write a proper test that actually only uses public API and test a *user-story* instead.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/classpath/RequiredPluginsClasspathContainerComparatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+ * IDE restarts when multiple bundles share a symbolic name.
+ * <p>
+ * The production comparator lives on a package-private class and cannot be
+ * referenced from this test bundle. The expected definition is therefore
+ * duplicated here so the contract is asserted independently. Keep the two in
+ * sync.
+ */
+public class RequiredPluginsClasspathContainerComparatorTest {
+
+	private static final Comparator<BundleDescription> EXPECTED_ORDER = Comparator
```
laeubi: Now this test is testing a comparator that is never used in production code what is complete nonsense.

Either came up with a proper test that uses public API (like other tests do) or consider removing it al together.

HannesWell:
I agree with Christoph in https://github.com/eclipse-pde/eclipse.pde/pull/2313#discussion_r3141806603, that we can consider comparators simple enough to not test them. It's close to testing getters.
Then everything should be simple and private.

HannesWell (approved):
Thanks for the updates. Code-wise this looks good to me.

Considering https://github.com/eclipse-pde/eclipse.pde/pull/2313#issuecomment-4314605872, @laeubi are you also fine?

vogella:
> Thanks for the updates. Code-wise this looks good to me.
> 
> Considering [#2313 (comment)](https://github.com/eclipse-pde/eclipse.pde/pull/2313#issuecomment-4314605872), @laeubi are you also fine?

@laeubi any additional concerns?

laeubi (approved):
It looks fine to me

vogella:
Tested in I-Build, good news is that the updating bundle classpath reports now zero necessary updates, bad news is that JDT still performs a full build (maybe by design I did not look at the JDT side of things).

PR #2318, opened 2026-04-28
Add counter to update classpath message

fedejeanne:
When updating the classpath of the plugins, the message now includes additional information about how many out of how many plugins have been processed (e.g. 10/321), useful for large workspaces.

## Screenshots
Here's the progress monitor for 2 plugins: `plugin1` and `plugin2`

image

---

image

---

image

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
@@ -45,6 +45,7 @@ import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.osgi.util.NLS;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.internal.core.natures.BndProject;
 import org.eclipse.pde.internal.core.natures.PluginProject;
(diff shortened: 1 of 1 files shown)


vogella:
LGTM

fedejeanne:
@HannesWell ?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -114,13 +115,14 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
-			for (UpdateRequest req : requests) {
+			for (int i = 0; i < requests.size(); i++) {
+				UpdateRequest req = requests.get(i);
 				if (monitor.isCanceled()) {
 					break;
 				}
 				IProject project = req.project();
 				if (project.exists() && project.isOpen()) {
-					monitor.subTask(project.getName());
+					monitor.subTask(NLS.bind("({0}/{1}): {2}", i + 1, count, project.getName())); //$NON-NLS-1$
```
HannesWell: Having two colons follow each other doesn't look ideal:
Image
What about appending the progress (maybe with then adjusted argument order for the label):
```suggestion
					monitor.subTask(NLS.bind("{2} ({0}/{1})", i + 1, count, project.getName())); //$NON-NLS-1$
```
or separating it using a dash (if you want to avoid jumping numbers in the progress bar).
```suggestion
					monitor.subTask(NLS.bind("({0}/{1}) - {2}", i + 1, count, project.getName())); //$NON-NLS-1$
```
Or something completely different?

HannesWell (commented):
I really like the idea, but I think the exact label is not yet ideal.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell: > Or something completely different?

If you'd remove the colon after `dependencies`, this would be perfectly fine for me. I think the preceding text reads better without it anyways, in case nothing is appended (colon is not a fullstop).

Edit:
It seems like the first colon comes from the `SubMonitor`. If you want to achieve this, you maybe have to reset the `taskName` and just report progress instead of subtask.

HannesWell: If you'd introduce a separate variable `String messageTemplate = PDECoreMessages.PluginModelManager_1 + " ({0}/{1}): {2}"; //$NON-NLS-1$`, ideally outside the loop, then you could do just:
```suggestion
					monitor.setTaskName(NLS.bind(messageTemplate, i + 1, count, project.getName()));
```
which then looks good IMO:

grafik

I wonder if such 'numeric progress' is something that's wanted more often and should be generalized and backed into the SubMonitor API, with some kind of 'counted subtasks' that are based on split?

fedejeanne: Done, thank you for the suggestion!

HannesWell (approved):
Thanks for the update. Looks very good now.

Test failures are unrelated.

PR #2319, opened 2026-04-28
Add unit tests for TargetResolveRule

vogella:
Adds focused unit tests for the per-target-handle scheduling rule introduced in #2311. The rule was a private nested record in `TargetEditor`; this PR extracts it to a top-level `TargetResolveRule` in the same internal package with a public `forHandle(ITargetHandle)` factory, and grants `org.eclipse.pde.ui.tests` x-friend access so the test bundle can reach it. The package stays internal — no public API is added.

The new tests pin the contract: same memento serializes, distinct mementos stay independent, foreign rules never conflict, and the `CoreException` fallback keeps the same handle instance serialized while still distinguishing different broken handles.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/AllTargetTests.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/target/AllTargetTests.java
@@ -25,7 +25,8 @@ import org.junit.platform.suite.api.Suite;
 		TargetDefinitionResolutionTests.class, //
 		TargetDefinitionFeatureResolutionTests.class, //
 		IUBundleContainerTests.class, //
-		ProfileContainerTests.class })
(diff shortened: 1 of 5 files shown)


vogella:
@HannesWell and @laeubi you had concerns in another PR of making something accessable just for testing. ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetResolveRule.java is using internal so in my opinion we should be fine here? WDYT?

HannesWell:
> you had concerns in another PR of making something accessable just for testing. ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetResolveRule.java is using internal so in my opinion we should be fine here? WDYT?

To be frank, I think the rule is simple enough to not having to test it. To me it seems that this mainly tests the JVM as there is only little own complexity in it.

PR #2320, opened 2026-04-28
Added test for E4WorkbenchModelEditor XMI tab search

r-mennig:
Follow-up PR for https://github.com/eclipse-pde/eclipse.pde/pull/2293 adding a new test fragment for **org.eclipse.e4.tools.emf.ui** in order to test the search functionality of the XMI tab of the E4WorkbenchModelEditor.

The added regression test case asserts that the search highlighting reacts correctly to new key inputs into the search field and considers the full text of the search field for highlighting. The test data used for this is  a simple project with a default application.

### DIFF
--- a/org.eclipse.e4.tools.emf.ui.tests/src/org/eclipse/e4/tools/emf/ui/internal/common/component/tabs/tests/XmiTabTest.java
+++ b/org.eclipse.e4.tools.emf.ui.tests/src/org/eclipse/e4/tools/emf/ui/internal/common/component/tabs/tests/XmiTabTest.java
@@ -1,0 +1,131 @@
+package org.eclipse.e4.tools.emf.ui.internal.common.component.tabs.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.IOException;
+import java.net.URL;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectDescription;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.eclipse.e4.tools.emf.ui.internal.common.component.tabs.XmiTab;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.CTabFolder;
+import org.eclipse.swt.custom.StyledText;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Text;
+import org.eclipse.swt.widgets.Widget;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.ide.IDE;
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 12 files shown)


org.eclipse.e4.tools.emf.ui.tests/src/org/eclipse/e4/tools/emf/ui/internal/common/component/tabs/tests/XmiTabTest.java
```diff
@@ -0,0 +1,131 @@
...
+			throw new IllegalArgumentException(path + " does not exist");
+		}
+		IPath projectFile = IPath.fromPortableString(entry.getPath()).append(IProjectDescription.DESCRIPTION_FILE_NAME);
+		IWorkspace workspace = ResourcesPlugin.getWorkspace();
+		IProjectDescription projectDescription = workspace.loadProjectDescription(projectFile);
+		IProject project = workspace.getRoot().getProject(projectDescription.getName());
+		project.create(projectDescription, null);
+		project.open(null);
+		return project;
+	}
```
r-mennig, line 130: These helper methods could be replaced by existing utility methods (e.g. org.eclipse.pde.ui.tests.util.ProjectUtils), but I wasn't sure if depending on org.eclipse.pde.ui.tests would introduce some problems here.

merks:
@laeubi @HannesWell 

Given there are no tests for the model editor, I assume it does make sense to add a new test bundle for that.  Or would it be better to add it to this bundle?

image

HannesWell:
Sorry for joining the topic so late, but in general I wonder if the special search bar could be replaced with the general search overlay we have for editors since about a year? 
Integrating that into the E4 editor would hopefully save code and would unify the UX.
Or is the search in the E4 editor more specialized/can do more?

If it should stay, I think it's fine to put the tests into the existing e4.tools.tests.

r-mennig:
@HannesWell @merks I've tried looking into this. 

Replacing the special search bar of the XMI tab with the generic find/replace logic should be fine as I could not find any differing behavior between the current search/highlighting logic and the generic one.

I've started the rework of the tab in this PR: https://github.com/r-mennig/eclipse.pde/pull/1

What I couldn't get working however is having the search being shown as an overlay (instead of a dialog) since this is currently only implemented for instances of org.eclipse.ui.texteditor.StatusTextEditor (see org.eclipse.ui.texteditor.FindReplaceAction.shouldUseOverlay)

image

Maybe this check could be replaced by a less restrictive interface which would be implementable by the XMI tab?

HannesWell:
> Maybe this check could be replaced by a less restrictive interface which would be implementable by the XMI tab?

From my POV there is no restriction on that, but I'm also not very much familiar with that specific part of the code.
@HeikoKlare was more involved in the implementation of the overlay sear, can you advice?

HeikoKlare:
If I remember correctly, with started with this kind of restriction because we had to properly integrate into the actual editor, which we initially only implemented for the `StatusTextEditor`. Most of the integration does actually not depend on a `StatusTextEditor` but on anything that can be adapted to an `ITextViewer`.

The crucial point, however, seems to be the disablement/enablement of actions for the workbench part into which the find/replace overlay is embedded. The workbench takes care of enabling the appropriate actions for the currently active part, but since the find/replace overlay is not a distinct part, the actions of the underlying editor (including copy/paste, undo/redo and the like) remain active. We have a very hacky implementation (using reflection, copied from JDT's breadcrumb implementation) that deactivates the editor's actions as long as the find/replace overlay has focus to work around this. One would have to extent this to all kinds of editors.
To me it seems as if this is not a proper design yet: probably the find/replace overlay needs to act as a separate workbench part that is realized as an overlay of and liked to the according editor. Then workbench functionality for ensuring proper action enablement should apply for the find/replace overlay without any manual hacks as well.
Maybe something like this goes into the right direction: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2093

We will try to follow up on this, as a proper solution could enable us to embed the find/replace functionality into many more kinds of views/editors.

PR #2321, opened 2026-04-28
Fixed scrolling inside the XMI tab of the E4WorkbenchModelEditor

### LINKED ISSUE #2299: Scrolling not possible in XMI tab of E4WorkbenchModelEditor
The XMI tab of E4WorkbenchModelEditor does currently not allow scrolling inside the source viewer (despite showing scrollbars)

Image

How to reproduce/test:

- Create a new application model (or navigate to any existing application model)
- Inside the E4WorkbenchModelEditor, select the XMI tab and try scrolling horizontally (resize the editor if necessary) -> Scrolling is not possible, the editor is disabled

The reason for this seems to be the following code in org.eclipse.e4.tools.emf.ui.internal.common.component.tabs.XmiTab.postConstruct():

Image
Image

The description of the preference states that it forces a read-only XMI tab, but isn't this already achieved by making the source viewer non-editable? Disabling the text widget also disables the scrollbars of the viewer which should not be necessary here.

Additional finding: After disabling this preference and reopening the editor, the viewer is still read-only and not scrollable. Is this the correct behavior? If so, the description of the preference seems a little misleading.

r-mennig:
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2299

This PR removes the disablement of the source viewer in the XMI tab of the E4WorkbenchModelEditor. This allows scrolling inside the viewer while still keeping the contents unmodifiable.

This also enables copying of the XMI contents which might be useful e.g. for copying elementIds or other parts.

There's currently no test plugin for the XMI tab, but if https://github.com/eclipse-pde/eclipse.pde/pull/2320 gets merged, I can create a follow-up PR adding tests for the scrollbar too.

### DIFF
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/internal/common/component/tabs/XmiTab.java
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/internal/common/component/tabs/XmiTab.java
@@ -20,6 +20,7 @@ import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.eclipse.e4.core.di.annotations.Optional;
 import org.eclipse.e4.core.services.nls.Translation;
+import org.eclipse.e4.tools.emf.ui.common.ModelEditorPreferences;
 import org.eclipse.e4.tools.emf.ui.internal.Messages;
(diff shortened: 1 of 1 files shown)


merks:
I think the warning is this one which is an infrastructure issue not related to the PR:

image

merks (approved):
Nice to use `ModelEditorPreferences.TAB_FORM_SEARCH_SHOW` instead of a copy of the string literal!

merks:
I restarted two of the jobs to get more green verifications.

merks:
Thanks @r-mennig ❣️

PR #2322, opened 2026-04-28
Add PR check enforcing single functional commit

vogella:
Adds a new `check-commit-count` job to `pr-checks.yml` that fails when a pull request contains more than one functional commit. A single trailing `Version bump...` commit is still allowed so release-engineering PRs pass unchanged.

This makes the project's existing single-commit expectation visible in CI rather than relying on reviewers to spot it, and points contributors at `git rebase -i` / `git commit --amend` in the failure message.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,3 +19,26 @@ jobs:
     with:
       botName: Eclipse PDE Bot
       botMail: pde-bot@eclipse.org
+  check-commit-count:
+    name: Check commit count
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify PR has at most one functional commit
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          subjects=$(gh api --paginate "repos/$REPO/pulls/$PR/commits" \
+            --jq '.[].commit.message | split("\n")[0]')
+          total=$(printf '%s\n' "$subjects" | sed '/^$/d' | wc -l)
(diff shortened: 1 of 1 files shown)


vogella:
@HannesWell you have experience with GH checks. WDYT?

laeubi:
@vogella  please **test it on your fork** instead of rely on others to test things for you and we all have many things to do (including @HannesWell ) so we appreciate you like to help, but work should be proven to work and you should be confident it works. If you think you have no enough expertise to validate the work its better to maybe delay it or get familiar with it (what is quite easy with asking the AI to summarize the basics).

Also try to align with preliminary work:

1.You can create a PAT for this limited to your fork repo. Then create some example PRs in your fork to show it works as expected for you and others to get confidence 
2. We usually share workflows here https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows#readme so it can be used / referenced across platform repositories - you have examples in this file and also here it is best to be checked in your fork that it works to call the workflow
3. Make sure it works across forks (e.g. you can ask someone to fork your fork and open a PR) some actions might not run otherwise or are not allowed when a user is not also a committer.

In general I think a check should be at best self-explaining... adding a comment is probably to much noise here and not "standard" (except the version check).

vogella:
Ok, I leave it to others to consider adding this.

HannesWell:
> you have experience with GH checks. WDYT?

While I agree to the rule, there are sometimes exceptions for good reasons.
Furthermore we hopefully can participate in the preview of _Stacked PRs_ in GH soon:
https://github.github.com/gh-stack/

In general I think it's common best practice for GH PRs (as far as I know), so I hope that many are already familiar with this rule, but we can also document it. I hope that new contributors then learn it quickly. 
I don't plan to spend time on such check at the moment and it's probably difficult to make it perfect. But if somebody has a close to perfect solution, I'm open for it.

PR #2333, opened 2026-05-13
Fix NPE in API problems comparator

iloveeclipse:
IApiProblem::getTypeName() can return null.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3817


HannesWell (approved):
Makes sense. `getTypeName()` says it may return null.
Probably there is no clear answer, if null names should better be first or last, but it definitively should be handled.
Furthermore it would be good to know why the type-name is null in the first place in that case?

But all of that can be investigated later.
Thanks for this fix.

iloveeclipse:
Build is OK, except "known" API errors due https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3814

PR #2337, opened 2026-05-17
Add WildcardCompatibilityTests reproducing false-positive API delta for ? extends Object vs ?

laeubi:


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/comparator/ClassFileComparator.java
```diff
@@ -2070,11 +2070,34 @@ private String getMethodDisplayName(IApiMethod method, IApiType type) {
...
+	 * <p>
+	 * The pattern {@code +Ljava/lang/Object;} is unambiguous: it specifically
+	 * matches the direct {@code ? extends Object} bound and cannot be confused
+	 * with array types ({@code +[Ljava/lang/Object;}) or other class names.
+	 */
+	private static String normalizeWildcardExtendsObject(String signature) {
+		if (signature == null || signature.indexOf('+') < 0) {
+			return signature;
+		}
+		return signature.replace("+Ljava/lang/Object;", "*"); //$NON-NLS-1$ //$NON-NLS-2$
```
merks, line 2098: Are you sure about the *?  I didn’t see that while debugging if I recall correctly?

merks:
@HannesWell 

I think we should merge this.  

I have confirmed that this fixes the problems that I suppressed in p2:

image

HannesWell:
> I think we should merge this.

Yes that's suitable, thank you for the fix.

PR #2338, opened 2026-05-18
Serialize LoadTargetDefinitionJob per target handle

vogella:
Clicking "Reload Target Platform" on the editor toolbar cancels the in-flight load and immediately schedules a new one, but `Job.cancel` only flips a flag, so the cancelled job keeps running until its next monitor check. Repeated clicks therefore pile up "Load Target Platform (Cancel Requested)" entries in the Progress view, and two loads of the same target can race on the p2 profile lock/unlock pair inside `P2TargetUtils.deleteProfile`.

Commit 37f88d7f16 addressed the same pattern for the editor's resolve job but did not cover the explicit Reload action, which also calls `target.resolve()` internally. This change promotes the per-target-handle scheduling rule from `TargetEditor` into a shared `TargetResolveSchedulingRule` and applies it to both `LoadTargetDefinitionJob` and the editor's resolve job, so loads and resolves of the same target serialize while operations on different targets remain independent.

Planned for 4.41.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target/LoadTargetDefinitionJob.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target/LoadTargetDefinitionJob.java
@@ -32,6 +32,7 @@ import org.eclipse.pde.internal.core.PDECore;
 import org.eclipse.pde.internal.core.PDEPreferencesManager;
 import org.eclipse.pde.internal.core.target.Messages;
 import org.eclipse.pde.internal.core.target.TargetPlatformService;
+import org.eclipse.pde.internal.core.target.TargetResolveSchedulingRule;
 
 /**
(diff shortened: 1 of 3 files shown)


merks:
@vogella 

Please try to maintain a single commit based on master.  Never do a merge commit.

HannesWell (commented):
Can you please condense respectively generalize the javadoc of TargetResolveSchedulingRule and the new comment you added in the caller of it.
The current comments are too specific and IMO too verbose.
And if you want to keep references to issues, please use full URLs. That improves stability and accessibility of the reference in case the repo is moved/merged in the future.

vogella:
Javadoc is NOW (edited) shorter and the issue is linked with the long URL.

vogella:
Let me know if you have additional concerns.

vogella:
Planning to merge today

PR #2339, opened 2026-05-18
Error on JDT configuration ignoring OSGi restrictions

HannesWell:
As described in https://github.com/eclipse-pde/eclipse.pde/issues/2244#issuecomment-4013458419, a value of the `org.eclipse.jdt.core.compiler.problem.forbiddenReference` preference other than `error` effectively disables the enforcement of almost all OSGi access rules. This potentially leads to classloading errors at runtime because access rule violations are not detected anymore at compile-time/in the IDE.

With this change, users are now informed about such misconfiguration of PDE projects, by an error on the `build.properties` file.
Alternatively the error marker could be added to the `org.eclipse.jdt.core.prefs` file of a project, but I think this is unusual.

For those that really want to ignore OSGi access rules at compile-time, I'm considering to add yet another preference to degrade this check to warning, info or to ignore. It's currently missing, but I plan to add it this evening.

I know we're now late in the development cycle, but in the context of https://github.com/eclipse-pde/eclipse.pde/pull/2218, I think it's important to have this for the upcoming release.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
@@ -162,6 +162,10 @@ public class BuildErrorReporter extends ErrorReporter implements IBuildPropertie
 	protected int fSrcLibSeverity;
 	protected int fOutputLibSeverity;
 	protected int fEncodingSeverity;
+	protected int fJDCTClasspathMissconfigurationSeverity = Boolean
+			.getBoolean("pde.acceptDiscardedForbiddenAccessRules") //$NON-NLS-1$
+					? CompilerFlags.WARNING
+					: CompilerFlags.ERROR;
 
 	public BuildErrorReporter(IFile buildFile) {
 		super(buildFile);
@@ -1107,6 +1111,13 @@ public class BuildErrorReporter extends ErrorReporter implements IBuildPropertie
 			} else if (useJavaProjectSettings != null) {
 				String message = NLS.bind(PDECoreMessages.BuildErrorReporter_buildEntryInvalidWhenNoProjectSettings, PROPERTY_PROJECT_SETTINGS);
 				prepareCompilerError(PROPERTY_PROJECT_SETTINGS, null, message, PDEMarkerFactory.B_REMOVAL);
+			}
+
+			String forbiddenReferenceSeverity = options.get(JavaCore.COMPILER_PB_FORBIDDEN_REFERENCE);
+			if (forbiddenReferenceSeverity != null && !forbiddenReferenceSeverity.equals(JavaCore.ERROR)) {
+				String message = NLS.bind(PDECoreMessages.BuildErrorReporter_classpathAccessRulesDiscarded);
+				prepareError(null, null, message, PDEMarkerFactory.B_CLASSPATH_ACCESS_RULES,
+						fJDCTClasspathMissconfigurationSeverity, null, PDEMarkerFactory.CAT_OTHER);
 			}
 		}
 	}
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECoreMessages.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/PDECoreMessages.java
@@ -261,6 +261,8 @@ public class PDECoreMessages extends NLS {
 
 	public static String BuildErrorReporter_BuildEntryNotRequiredMatchesDefault;
 
+	public static String BuildErrorReporter_classpathAccessRulesDiscarded;
+
 	public static String BuildErrorReporter_sourceMissing;
 
 	public static String BuildErrorReporter_srcIncludesSourceFolder;
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/PDEMarkerFactory.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/PDEMarkerFactory.java
@@ -91,6 +91,7 @@ public class PDEMarkerFactory {
 	public static final int B_REMOVAL = 0x2005;
 	public static final int B_REPLACE = 0x2006;
 	public static final int B_JAVA_ADDDITION = 0x2007;
+	public static final int B_CLASSPATH_ACCESS_RULES = 0x2008;
 
 	// plugin.xml fixes
 	public static final int P_ILLEGAL_XML_NODE = 0x3001;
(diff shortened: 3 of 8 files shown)


ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
```diff
@@ -1108,6 +1110,14 @@ private void validateJavaCompilerSettings(IBuildEntry useJavaProjectSettings) {
 				String message = NLS.bind(PDECoreMessages.BuildErrorReporter_buildEntryInvalidWhenNoProjectSettings, PROPERTY_PROJECT_SETTINGS);
 				prepareCompilerError(PROPERTY_PROJECT_SETTINGS, null, message, PDEMarkerFactory.B_REMOVAL);
 			}
+
+			String forbiddenReferenceSeverity = options.get(JavaCore.COMPILER_PB_FORBIDDEN_REFERENCE);
+			if (forbiddenReferenceSeverity != null && !forbiddenReferenceSeverity.equals("error")) { //$NON-NLS-1$
```
merks: Maybe better to use org.eclipse.jdt.core.JavaCore.ERROR.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/correction/EnforceStrictAccessRules.java
```diff
@@ -0,0 +1,55 @@
+/*******************************************************************************
+ *  Copyright (c) 2026, 2026 Hannes Wellmann and others.
```
merks: The range is a bit strange.

merks (approved):
Just nits.  Looks fine.  Would be good for the current cycle given the PDE changes to the classpath.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
```diff
@@ -162,6 +162,8 @@ public void addAttributes(Map<String, String> attributes) {
 	protected int fSrcLibSeverity;
 	protected int fOutputLibSeverity;
 	protected int fEncodingSeverity;
+	// TODO: Make configurable?
```
laeubi: Should be removed - I also don't think it must be configurable unless there is a strong demand / usecase for this.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
```diff
@@ -1108,6 +1110,14 @@ private void validateJavaCompilerSettings(IBuildEntry useJavaProjectSettings) {
 				String message = NLS.bind(PDECoreMessages.BuildErrorReporter_buildEntryInvalidWhenNoProjectSettings, PROPERTY_PROJECT_SETTINGS);
 				prepareCompilerError(PROPERTY_PROJECT_SETTINGS, null, message, PDEMarkerFactory.B_REMOVAL);
 			}
+
+			String forbiddenReferenceSeverity = options.get(JavaCore.COMPILER_PB_FORBIDDEN_REFERENCE);
+			if (forbiddenReferenceSeverity != null && !forbiddenReferenceSeverity.equals("error")) { //$NON-NLS-1$
+				String message = NLS.bind(PDECoreMessages.BuildErrorReporter_classpathAccessRulesDiscarded);
+				String compilerKey = null; // TODO: Set this?
```
laeubi: Another TODO to be removed, I would simply inline the null unless we have a demand for it or can fill it with an usefull value.

laeubi (approved):
I'm a bit doubtful this brings much value but it would also not harm.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
HannesWell: Absolutely. In fact I searched for such constant but didn't found it.
Thanks for the hint.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
HannesWell: Generally I also expect that it's not necessary and any every case of existing configuration of `forbiddenReference!=error` can be done better.
But since I'm not 100% sure and we are late in the dev-cycle I added a System-property so that at least one can degrade it to a warning using `-Dpde.acceptDiscardedForbiddenAccessRules=true`, if really necessary.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/BuildErrorReporter.java
HannesWell: Yes, I checked it and the value is not used (the TODO was to remind about this pending task).

HannesWell:
> I'm a bit doubtful this brings much value but it would also not harm.

Usually not, but since we saw also in the Eclipse SDK some projects being configured with `compiler.problem.forbiddenReference!=error` I assume there are some more such project out there. And since that configuration is now even more dangerous, users should be informed about it.

Thanks for your reviews.

I'll submit this once the build is ready to have this included in tonight's I-build.

tdo-code:
I just upgraded to Eclipse 2026-06 (4.40.0) and got hit with this error on all of my plugins projects. What exactly do I need to do to make them go away? The IDE provides no meaningful information on the error, and Google doesn't help either.

merks:
Global search `org.eclipse.jdt.core.compiler.problem.forbiddenReference=` (and replace) to ensure that every  one is set to org.eclipse.jdt.core.prefs is set to `org.eclipse.jdt.core.compiler.problem.forbiddenReference=error`

image

laeubi:
> The IDE provides no meaningful information on the error, and Google doesn't help either.

There is a quickfix that do exactly that what is proposed by @merks

tdo-code:
@merks @laeubi Thank you, I've made the changes manually (and also tried the quickfix, which made the same changes)

image

However, the error doesn't go away in Eclipse. 

image

image


I did a full refresh, full clean, restarted Eclipse, and it's still the same.

Is there a way to disable this error, or change its severity to warning?

laeubi:
I noticed sometimes Eclipse has a problem on markers in a hierarchical project tree. In such case you can select the errors and press `DEL` to remove them manually.

By the way from your screenshot you are embedding third party libraries what is generally a bad style and should be avoided especially as jackson is already an OSGi bundle...

merks:
Yes, I had similar nasty experience eliminating these.  I tried a number of things, actually deleting the errors, doing a restart, then doing a full clean build.  The did eventually go away.  I got the sense the JDT was caching the older preference values and not using my changes at all.  😢

laeubi:
Its just a guess but I think because the build.properties file never changes during this, the validation is simply not kicked off again and can't remove the stale markers.

tdo-code:
I deleted them manually and they seem to be gone, for now.

What I don't understand though, is that the existing setting is

```
org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
```

With that, shouldn't the issue have warning severity in Eclipse? Why did it become an error, and then force me to change this setting (which I don't really care about) to error?

This is especially annoying since I import a bunch of plugins that I don't have control over the source code into my workspace (i.e. I can't make an MR to change this setting in the org.eclipse.jdt.core.prefs file for these plugins). So I can only make the changes locally, but there are a few dozens of them, and now doing git operations on these plugins would be a pain.

This really should be configurable as a global setting from Eclipse.

merks:
The org.eclipse.jdt.core.compiler.problem.forbiddenReference preference can be set globally or locally in the project, but once a project sets it locally, then only the project can change it locally. 

Anyway, it is kind of annoying and I too was confused by the behavior trying to fix it.

Another annoying this is that as soon as you want to set one project-specific preference locally, the entire set of preference is copied and locked in place based on the current global state.   Sometimes I literally go in and delete everything except the one preference that I want to be "special" for the project; but it's not always obvious which preference key that is.

PR #2343, opened 2026-05-21
Added Find support in Plug-ins View

### LINKED ISSUE #2341: Plug-ins View Missing Find/Search Support
The Plug-ins view does not provide any Find/Search action. Standard shortcuts like Cmd+F also do not work in this view, making it difficult to quickly locate plugins in large lists. A dedicated search/find support in the toolbar, menu, or keybinding would improve usability.

Image

nburnwal09:
Fixes: #2341
Added plugin level find support in Plug-ins view

https://github.com/user-attachments/assets/fb4c72ad-a84e-45e0-b63f-49b1dd4b77d7

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/plugins/PluginsView.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/plugins/PluginsView.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


SougandhS:
Cool change ! 
Thanks for working on this 👍

nburnwal09:
@HannesWell
Can you please review this PR?

noopur2507 (changes requested):
Hi Neha, I tried the patch and have a couple of observations:

* I notice some lag in both the filter field and the tree when typing or deleting filter text quickly.
* When filter text is present, Cmd+A does not seem to select the entire text in the filter field, which makes it difficult to clear the filter by selecting all and deleting.

Could you please take a look?

nburnwal09:
> Hi Neha, I tried the patch and have a couple of observations:
> 
> * I notice some lag in both the filter field and the tree when typing or deleting filter text quickly.
> * When filter text is present, Cmd+A does not seem to select the entire text in the filter field, which makes it difficult to clear the filter by selecting all and deleting.
> 
> Could you please take a look?

Sure Noopur. I will check.

nburnwal09:
Hi @noopur2507 
I have addressed both lagging and cmd+A option. 
Kindly review it.

noopur2507 (approved):
Working fine now.

vogella:
For new code I suggest to use the e4 FilteredTree which is a reworked version of the old FilteredTree, should be more a less a drop-in replacement (except the, true, true parameters).

nburnwal09:
> For new code I suggest to use the e4 FilteredTree which is a reworked version of the old FilteredTree, should be more a less a drop-in replacement (except the, true, true parameters).

Hi @vogella 
Yeah that's correct.
I have updated the changes with e4 FilteredTree. Can you pls review it now?

noopur2507 (approved):
Looks good to me.

PR #2348, opened 2026-05-28
 Fix: Update name-relative fields even when project name is empty

vlakn:
### Problem

When the project name field was left empty (or otherwise invalid), the feature name and version fields were not being updated automatically. This broke the "auto-fill" behaviour that derives name-relative field values from the project name during the initial wizard state.

before

### Fix

Moved the `updateNameRelativeFields()` call to before the early-return guard, so that name-relative fields are always refreshed when `fUpdateName` is true, regardless of whether the rest of the page is currently valid.

after

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/wizards/AbstractFeatureSpecPageTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/wizards/AbstractFeatureSpecPageTest.java
@@ -1,0 +1,140 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ******************************************************************************/
+package org.eclipse.pde.ui.tests.wizards;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.eclipse.jface.dialogs.IDialogSettings;
+import org.eclipse.pde.internal.ui.wizards.feature.AbstractFeatureSpecPage;
+import org.eclipse.pde.internal.ui.wizards.feature.FeatureData;
+import org.eclipse.swt.events.ModifyListener;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+import org.junit.After;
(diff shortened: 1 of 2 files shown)


ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/wizards/AbstractFeatureSpecPageTest.java
```diff
@@ -0,0 +1,148 @@
...
+	private static final String FEATURE_ID = "test.feature.id"; //$NON-NLS-1$
+	private static final String PROJECT_NAME = "test.project.name"; //$NON-NLS-1$
+
+	private Shell fShell;
+	private Display fDisplay;
+
+	@Before
+	public void setUp() {
+		fDisplay = Display.getDefault();
+		fDisplay.syncExec(() -> fShell = new Shell(fDisplay));
```
r-mennig: Calling Display.syncExec() here and further below shouldn't be necessary since the test is already running in the UI thread.

r-mennig (approved):
LGTM 👍

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/wizards/AbstractFeatureSpecPageTest.java
vlakn: Thank you for your review! I removed the unnecessary Display.syncExec() calls.

HeikoKlare (approved):
The change also looks conceptually sound and works as expected for me. Thank you, @vlakn!
Would anyone mind processing/merging this PR, so that we have the fix in the upcoming release?

vogella:
@HeikoKlare you are not yet a PDE committer?

Merging on behalf of Heikos review, I did not review or test this as I trust Heiko

HeikoKlare:
Thank you for merging, @vogella!

> you are not yet a PDE committer?

No, I think I have only done like 15 to 20 commits to PDE yet, so probably not sufficient for becoming committer yet.

akurtakov:
I personally see committership not only as some metric but also as a number of other things like involvement, ease to discuss things with, overall care for the project and so on. By all these means you fully qualify for committer (in my book) if you're interested in it.

vogella:
> Thank you for merging, @vogella!
> 
> > you are not yet a PDE committer?
> 
> No, I think I have only done like 15 to 20 commits to PDE yet, so probably not sufficient for becoming committer yet.

The PMC decided some time ago that commits in other top-level projects also count towards a nomination, as with the reduced activity across Eclipse these days, it is hard to collect a large number of contributions in one isolated project. And 15 commits is a lot for PDE IMHO. So unless you'd rather not, I'd be happy to propose you as a committer.

HeikoKlare:
Thank you for your feedback. I would of course be very happy to be nominated as a committer.

HeikoKlare:
@vogella @akurtakov @HannesWell thank you again for the trust, the nomination, and the approval for becoming committer to PDE

PR #2352, opened 2026-06-02
Make CSS spy selector handling API-version agnostic

vogella:
The CSS spy referenced `org.w3c.css.sac.SelectorList` and `Selector` directly, which breaks once eclipse.platform.ui [#4051](https://github.com/eclipse-platform/eclipse.platform.ui/pull/4051) switches `CSSEngine.parseSelectors()` and `matches()` to an internal `Selectors` AST.

Since both the old and new selector list types expose the same `getLength()`/`item(int)` and round-trip cleanly through `engine.matches(...)`, the spy no longer needs to name the type at all. The parsed selectors are now held in `var` locals and the recursive search takes a `Predicate<CSSStylableElement>` instead of a typed selector-list parameter (which also drops the always-null `pseudo` argument). The result compiles against both the current and the post-#4051 target platform, so it can merge independently of the platform change.

### DIFF
--- a/ui/org.eclipse.pde.spy.css/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.pde.spy.css/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %name
 Bundle-SymbolicName: org.eclipse.pde.spy.css;singleton:=true
-Bundle-Version: 0.14.100.qualifier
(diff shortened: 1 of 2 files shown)


laeubi:
I'm not sure this really works. This might make it compile time agnostic but I doubt it works in runtime the way indicated here..

vogella:
Works well (tested in a custom aggregator build with both PR integrated, this one and https://github.com/eclipse-platform/eclipse.platform.ui/pull/4051)

Screenshot From 2026-06-09 16-55-15

PR #2353, opened 2026-06-03
Guard against null target when setting the load scheduling rule

vogella:
`LoadTargetDefinitionJob.load` accepts a null target to represent the empty target platform, but the per-handle scheduling rule was looked up via `target.getHandle()` before that case was handled. Calling `load` with a null target therefore threw a NullPointerException, which `TargetPlatformPreferencePage` can trigger.

The rule is now only set when a target is present; the empty target platform has no handle to serialize on.


vogella:
A bit defensive, but it matches the old pattern and better safe than NPE......

Issue #1030, opened 2023-12-27
[target editor] Allow to make a target "offline"


laeubi:
Currently open a target can take some time because probably remote sites need to be contacted (e.g. p2 or maven repositories) but in many cases a target actually does not change very often and is rather static.

Because of this, there should be a new feature "make offline", that freezes the current state of this target by simply copy everything to a dedicated location (while "copy" might actually just be a reference to an already local file) and the target from there on is only resolved using this local content.

For some operations e.g. update, edit or add new items then the user should be asked to "go online" and the target becomes an online one again.

vogella:
I think we should rather only resolve on user request never automatically

laeubi:
> I think we should rather only resolve on user request never automatically

That doesn't really help you except you always ever only use one target file.

PR #2354, opened 2026-06-03
Reload active target platform when its .target file changes

vogella:
Automatically reloads the active target platform when its backing `.target` file changes, so editing the file (in the editor, via a refactoring, or a `git pull`) no longer requires a manual "Reload Target Platform".

Workspace files are watched through resource deltas; a modification timestamp recorded on each load also catches edits made while the IDE was closed, including external `file:` targets that deltas cannot observe. Reloads are coalesced behind a short delay so a burst of changes triggers a single load rather than repeatedly cancelling an in-flight resolve.

Adds tests covering the backing-file resolution and the auto-reload behaviour.


laeubi:
I found people are often wanting the opposite: Opening a target platform (e.g. without internet connection) can easily break your workspace as it leaves the active target in an somehow intermediate state.

This has been acknowledged here even by yourself: https://github.com/eclipse-pde/eclipse.pde/issues/1030#issuecomment-1870097320

> I think we should rather only resolve on user request never automatically

So what changes your mind and have you considered to make the behavior at least configurable? Especially on larger workspace a change in target change would mean update and possibly recompile a lot of projects, I'm not sure "a short delay so a burst of changes triggers a single load" is sufficient here either.

As this is a major change in behavior at least this has to be discussed with a broader audience!

merks:
Indeed.  Also, given the fact that any resolution failure leaves a corrupted target platform state, the whole workspace could easily become unusable because of a refresh.  I've not reviewed to see how this deals with saving a *.target file; that too triggers a delta...

Issue #2358, opened 2026-06-09
JUnitSuiteExecutionTest fails in I-build


trancexpress:
See e.g.:  https://download.eclipse.org/eclipse/downloads/drops4/I20260608-1800/testresults/html/org.eclipse.pde.junit.runtime.tests_ep441I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
	No tests found with test runner 'JUnit 3'.

org.eclipse.core.runtime.CoreException: No tests found with test runner 'JUnit 3'.
at org.eclipse.pde.launching.JUnitLaunchConfigurationDelegate.abort(JUnitLaunchConfigurationDelegate.java:179)
at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.evaluateTests(JUnitLaunchConfigurationDelegate.java:398)
at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.getVMRunnerConfiguration(JUnitLaunchConfigurationDelegate.java:160)
at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.launch(JUnitLaunchConfigurationDelegate.java:287)
at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:778)
at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:689)
at org.eclipse.debug.internal.ui.DebugUIPlugin.buildAndLaunch(DebugUIPlugin.java:1051)
at org.eclipse.debug.ui.DebugUITools.buildAndLaunch(DebugUITools.java:1044)
at org.eclipse.pde.junit.runtime.tests.TestExecutionUtil.launchAndWaitForTermination(TestExecutionUtil.java:146)
at org.eclipse.pde.junit.runtime.tests.TestExecutionUtil.runTest(TestExecutionUtil.java:98)
at org.eclipse.pde.junit.runtime.tests.JUnitSuiteExecutionTest.executePackage(JUnitSuiteExecutionTest.java:87)
```

There are some other types of failures too: https://download.eclipse.org/eclipse/downloads/drops4/I20260608-1800/testresults/html/org.eclipse.pde.junit.runtime.tests_ep441I-unit-macosx-aarch64-java21_macosx.cocoa.aarch64_21.html

```
	test completed with Error

java.lang.AssertionError: test completed with Error
[...]

PR #2362, opened 2026-06-10
Delete target p2 profile only when a reload is forced

vogella:
The resolve job in the target editor's TargetChangedListener deleted the p2 profile before every resolve it triggered, including opening the editor on a not-yet-resolved target or adding and editing a location. This forced a full re-provisioning with repository metadata loads and artifact downloads even when the profile was still in sync with the target definition, which made these actions take minutes for large remote targets.

With this change the profile is only deleted when a forced resolve is requested. Plain resolves go through P2TargetUtils.synchronize, which reuses a valid profile offline and still recreates it whenever the target content actually changed. The explicit Reload and Update actions keep their cache-busting behavior, which Update relies on so that IUs declared with an unspecific version (0.0.0) pick up newer versions from the repository.

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
@@ -707,11 +707,15 @@ public class TargetEditor extends FormEditor {
 				Job resolveJob = new Job(NLS.bind(PDEUIMessages.TargetEditor_1, name)) {
 					@Override
 					protected IStatus run(IProgressMonitor monitor) {
-						// delete profile
-						SafeRunner.run(() -> {
-							P2TargetUtils.forceCheckTarget(getTarget());
-							P2TargetUtils.deleteProfile(getTarget().getHandle());
-						});
+						if (forceResolve) {
+							// delete the profile so the target is provisioned from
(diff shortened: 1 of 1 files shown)


vogella:
Test errors seem to come from: https://github.com/eclipse-pde/eclipse.pde/issues/2358

vogella:
if no one raises concerns I would like to merge this today, this does not throw away the target resolution on editor open which saves me on my "slow Eclipse" ~20 min wait time (reloading the target file, updating plug-in dependencies and a potential compile run afterwards). 

It is basically not possible to use Eclipse for me at this client without this patch, that is why I'm currently running a patched version of Eclipse and this one is a big saver. Before that I did not dare to touch the target file inside Eclipse and used Code to see what is included in the target file.

merks:
I'm often scared to open that editor and scared to leave it open when I shut down my IDE such that it's the first thing opened on startup.  It's always bothered me that a failed resolve, for whatever reason, leaves one with nothing but a mess; that's why Oomph's targlets support roll-back on failure, not that this help PDE implement the same thing.  😞 

All that being said, I'm not really sure the overall impact of this change, but it sounds okay in principle...

vogella:
Thanks for the feedback @merks I read it as agreement to this change. Also Copilot did not find anything to improve.

merks:
Yes, and I did actually test the PR in a runtime workspace and it seemed to behavior reasonably!

Issue #2361, opened 2026-06-10
Updating plug-in dependencies take very long (10 min in 2026-03, > 30 min in 2026-06 without -Dpde.addTransitiveDependenciesWithForbiddenAccess=false)


vogella:
If I reload the target platform the plug-in dependencies frequently gets updated.

In one of my client implementations updating the plug-in dependencies takes very long, I would guess 10 min for approx 635 plug-ins and unbearable long (stopped the IDE after waiting more than 30 mins) with 2026-06 and  without the -Dpde.addTransitiveDependenciesWithForbiddenAccess=false flag. With the flag it is comparable to 2026-03 but still really long.

<img width="311" height="72" alt="Image" src="https://github.com/user-attachments/assets/4120e9ec-05ba-420b-8f94-7c283f5d5566" />

Note: it sometimes reports more plug-ins so instead of 635, it sometimes claims that we have > 2000 plug-ins.

As this is really time consuming, would it be possible to update the plug-in dependencies in parallel? Similar to the Maven / Tycho parallel build I assume that plug-ins on the same "dependency level" should be possible to update in parallel.

laeubi:
Tycho and PDE working completely differently. In any case you can take a look at `UpdateClasspathsJob` and check if it helps to parallelize things. Of course it would be interesting to check *what* takes long and as everything is centralized it should be quite easy to profile. Having shown "wrong" number at least sounds suspicious to me, but it could be that multiple updates are enqueued from different places.

vogella:
https://github.com/eclipse-pde/eclipse.pde/pull/2362 should help with the case that the profile is still in sync. Currently even a "Reload" on an unchanged profile does a full update. And the end user does not necessary know if the target platform has changed after a checkout (and svn update like in the case for my client), hence they tend to press Reload.

vogella:
This is fixed with https://github.com/eclipse-pde/eclipse.pde/commit/8fbf102c4d505e495654ff4464667245eb8de5a0

PR #2363, opened 2026-06-10
Deduplicate queued classpath container update requests

vogella:
A target platform reload triggers classpath updates from several listeners, so the work queue of UpdateClasspathsJob frequently contains the same project multiple times. Each occurrence ran the full classpath computation again and inflated the project count in the progress message, e.g. reporting more than 2000 plug-ins for a workspace with 635 plug-ins (see #2361).

This change drains the queue into one request per project before processing. A request without a saved container state forces a container update and therefore wins over requests whose update may be skipped when the computed entries match the saved state. This reduces redundant classpath computations and makes the reported progress count match the actual number of projects. Includes unit tests for the deduplication logic.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
@@ -107,11 +107,7 @@ public class ClasspathContainerState {
 			Map<IJavaProject, IClasspathContainer> updateProjects = new LinkedHashMap<>();
 			Map<IProject, IStatus> errorsPerProject = new LinkedHashMap<>();
 
-			java.util.List<UpdateRequest> requests = new java.util.ArrayList<>();
-			UpdateRequest request;
-			while ((request = workQueue.poll()) != null) {
-				requests.add(request);
-			}
+			List<UpdateRequest> requests = drainRequests(workQueue);
 			int count = requests.size();
 			monitor.setWorkRemaining(count * 2);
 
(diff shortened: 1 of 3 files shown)


laeubi:
@vogella it would really be helpful to give other a chance to review changes to central places of the code or explicitly ask for a review. For example one might question if multiple listeners should actually trigger the same update.
Also why `UpdateRequest` is now a **public** type?
This again seem to be to ad some fake testing that actually does not test the behavior of the code and instead test some thing in isolation when there is no guarantee this is actually ever called from the code that is claimed to be tested...

vogella:
UpdateRequest became public only so the test can construct requests, and I agree that is not good. I am preparing a follow-up that deduplicates at enqueue time (a map keyed by project instead of deduplicating while draining the queue). That makes UpdateRequest private again, removes the drainRequests helper, and the isolated tests go away with it.

Whether multiple listeners should trigger the same update at all is a good question. Dedup in the job is a safety net either way since enqueues are concurrent, but if a listener is redundant on target reload, fixing it at the source would be better. I will look into it and report back.

laeubi:
In general de-duplication can only work if listeners submit the same project in very fast way, and I think avoiding such updates should be investigated first. Because the job runs at the background, it can happen that a project needs further updates any time later on so one needs to be careful to not miss an update.

I also want to share this artice here that gives good tips for reviewing AI generated code: https://docs.github.com/en/copilot/tutorials/review-ai-generated-code

So if you are think something is maybe not good, it should be fixed first, this is nothing critical that requires a fix immediately and we have plenty of time to come to a good conclusion!

I agree writing a good test can be tricky but that's often the case that the reproducer/test is the most effort. From your referenced issue, I would have expsected a test doing something like:

1. creating some PDE projects and wait until everything has settled
2. setting up a global job listener
3. setting a new target platform
4. observing that only a certain amount of updates happen during the job run,what can be quite tricky but maybe with a special progress monitor for jobs - thats something the AI can possibly help to scrape the code for possible options and then one can decide what seems best.

PR #2365, opened 2026-06-11
Set BREE of pde.doc.user bundle to stabilize generated javadoc output

HannesWell:
Avoids warnings like
```
[WARNING] useJDK=BREE configured, but no BREE is set in bundle. Fail back to currently running execution environment (JavaSE-21)
```
and comparator errors because the javadoc is generated with Java-25 in the context of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3891


akurtakov:
This is not a java project so this warning looks wrong to me.

HannesWell:
> This is not a java project so this warning looks wrong to me.

Yes. But this should also ensure that the Java-21 is used to build the Java doc. And unfortunately the used JDK is recorded in the output and makes a difference.

I checked again the logs and the mentioned warnings probably only occurs for other bundles without a BREE, but not this one.
Still the stabilization of the generated Javadoc should happen.

PR #2366, opened 2026-06-11
Deduplicate classpath update requests at enqueue time

vogella:
Follow-up to #2363, addressing the review feedback from @laeubi. The work queue of UpdateClasspathsJob is now a map keyed by project, so requests for the same project are merged when they are added rather than while draining the queue. The deduplication is thereby enforced by the data structure itself. This lets UpdateRequest become private again and removes the drainRequests helper that was public only for testing, together with the isolated unit tests for it.


HannesWell:
@laeubi do you have remarks on this, since you requested the changes in https://github.com/eclipse-pde/eclipse.pde/pull/2363#issuecomment-4683253721?

vogella:
Any additional feedback here?

vogella:
Planning to merge tomorrow

merks:
@vogella 

FYI, the m1 build is tonight's build.  Looked at the changes and they look correct and safe, so I would suggest merging this now.

vogella:
> @vogella
> 
> FYI, the m1 build is tonight's build. Looked at the changes and they look correct and safe, so I would suggest merging this now.

Tomorrow already? Thanks for the information.

PR #2369, opened 2026-06-12
Share dependency closure computation across classpath updates

vogella:
When the classpath containers of many projects are updated in one run of UpdateClasspathsJob, the transitive dependency closure used for the forbidden-access entries was computed from scratch for every project, although the dependency graphs of the projects in a workspace overlap to a large degree. This change introduces a cache that memoizes the build-relevant closure per bundle and composes a project's closure as the union of the per-bundle closures, so each subgraph is traversed once per job run instead of once per dependent project. The composition is exact because the build-relevant namespaces include Fragment-Host, so a fragment always pulls its host into the same closure. The cache is scoped to a single job run, which makes invalidation unnecessary. Contributes to #2361.

Note: this builds on #2366, the first commit here is that PR and this should be rebased once it is merged.


HannesWell:
> This change introduces a cache that memoizes the build-relevant closure per bundle and composes a project's closure as the union of the per-bundle closures, so each subgraph is traversed once per job run instead of once per dependent project.

Can you please benchmarks comparing the runtimes before and after this change?
From my experience that particular code to compute the dependency closure of a bundle is quite fast. So I'm in doubt this helps significantly. Since caching is always difficult and often a source of bugs I'd rather not add it if it doesn't help significantly.
I assume that there are other places or better strategies (like the suggested parallelization) to achieve a greater improvement.

vogella:
@HannesWell I like your thinking that you also like the parallel update better. I have this already locally but assumed that would be even more controversal. :-) I will polish that one and provide it as PR.

My "classpath update is ultra slow so that using Eclipse is basically impossible" is unfortunately on a client side which I'm not allowed to do anything but to look at the code. But fortunately I'm already mitigating to custom IDE builds with custom update sides and may be able to report delta times next week based on this change.

vogella:
Today I have the opportunity to work again for the client with the slow PDE dependencies update.

Wall time for updating the classpath entries WITH the -Dpde.addTransitiveDependenciesWithForbiddenAccess=false flag. Without it I stopped the process last week after 30 mins at around 27% percentage, I do not have time to test that again.

(Press Reload in the target platform and watch then the progress monitor says: Updating plug-in dependencies until this entry is replaced with compiling).

Range: ~ 4min to almost 7 min

I know the range is huge but it is a v-machine, so I assume it is highly affected by the usage of other users.

vogella:
With https://github.com/eclipse-pde/eclipse.pde/pull/2374 I could measure "correctly".

HannesWell:
> Range: ~ 4min to almost 7 min

Do you mean with just this change applied the runtime dropped from much over 30min to ~ 4min to almost 7 min?

vogella:
> Do you mean with just this change applied the runtime dropped from much over 30min to ~ 4min to almost 7 min?

No, this was the range with -Dpde.addTransitiveDependenciesWithForbiddenAccess=false and no patches applied. 30 min was the time without the above parameter, and I canceled the update before it finished. The number for this patch and the parallel together are in https://github.com/eclipse-pde/eclipse.pde/pull/2375. I did not have the luxary to test this patch in isolation, the client pressed me to work on a production issue.  I only patched my Eclipse as otherwise the work as almost impossible to do.

vogella:
Dropping this cache.

@r-mennig measured the parallel classpath update (#2375) with and without this cache on a workspace of 1858 projects, and the cache made no measurable difference:

- `addTransitiveDependenciesWithForbiddenAccess=true`: 19s with the cache vs 17-19s without it
- `addTransitiveDependenciesWithForbiddenAccess=false`: 10s vs 9-10s (the cache is never consulted on this path)

Since the parallel run is held constant in both measurements, the difference between them isolates the cache's contribution, and it is within measurement noise even in the forbidden-access case where it should help most. Toggling the flag forces all projects to recompute in a single job run, which is the best case for a per-bundle memoization, so a lighter partial reload would benefit even less. The full speedup came from computing the per-project containers in parallel, not from sharing the closure.

Given that the cache adds a concurrent, run-scoped structure (more complexity and surface for subtle bugs) without a demonstrated gain, it is not worth keeping. Happy to revisit if a workspace turns up where parallel + cache clearly beats parallel-only.

PR #2373, opened 2026-06-16
Improve handling of unresolved bundles for feature-launches #2367

### LINKED ISSUE #2367: Launch Configuration doesn't respect bundle version in feature.xml
For context, our workspace contains two products. One product is using Jersey 2.x, the other 3.x. With the 2026-06, this has changed to 2.x and 4.x. But since then, we are no longer able to start the 2.x-based product, as it now pulls in the 4.x artifacts.

<details>
<summary>Target Platform</summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="target-platform">
	<locations>
		<location path="${eclipse_home}" type="Profile"/>
		<location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" missingManifest="generate" type="Maven">
			<dependencies>
				<dependency>
					<groupId>jakarta.xml.bind</groupId>
					<artifactId>jakarta.xml.bind-api</artifactId>
					<version>2.3.3</version>
					<type>jar</type>
				</dependency>
				<dependency>
					<groupId>org.glassfish.jersey.core</groupId>
					<artifactId>jersey-server</artifactId>
					<version>2.48</version>
					<type>jar</type>
				</dependency>
				<dependency>
					<groupId>org.glassfish.jersey.core</groupId>
					<artifactId>jersey-server</artifactId>
					<version>3.1.11</version>
					<type>jar</type>
				</dependency>
			</dependencies>
		</location>
	</locations>
</target>
```
</details>

<details>
<summary>Feature</summary>

```xml
<feature
      id="Jersey"
      label="Dependencies"
      version="1.0.0.qualifier">

   <description url="http://www.example.com/description">
      [Enter Feature Description here.]
   </description>

   <copyright url="http://www.example.com/copyright">
      [Enter Copyright Description here.]
   </copyright>

   <license url="http://www.example.com/license">
      [Enter License Description here.]
   </license>

   <plugin
         id="org.glassfish.jersey.core.jersey-client"
         version="2.48.0"/>

   <plugin
         id="org.glassfish.jersey.core.jersey-common"
         version="2.48.0"/>

   <plugin
         id="org.glassfish.jersey.core.jersey-server"
         version="2.48.0"/>

   <plugin
         id="jakarta.annotation-api"
         version="1.3.5"/>

   <plugin
         id="jakarta.inject.jakarta.inject-api"
         version="1.0.5"/>

   <plugin
         id="jakarta.ws.rs-api"
         version="2.1.6"/>

   <plugin
         id="jakarta.activation-api"
         version="1.2.2"/>

   <plugin
         id="org.glassfish.hk2.osgi-resource-locator"
         version="1.0.3"/>

   <plugin
         id="jakarta.xml.bind-api"
         version="2.3.3"/>

   <plugin
         id="org.eclipse.equinox.console"
         version="0.0.0"/>

   <plugin
         id="org.apache.felix.gogo.command"
         version="0.0.0"/>

   <plugin
         id="org.apache.felix.gogo.runtime"
         version="0.0.0"/>

   <plugin
         id="org.apache.felix.gogo.shell"
         version="0.0.0"/>

</feature>
```
</details>

<details>
<summary>Launch Configuration</summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<launchConfiguration type="org.eclipse.pde.ui.EquinoxLauncher">
    <setAttribute key="additional_plugins"/>
    <booleanAttribute key="append.args" value="true"/>
    <booleanAttribute key="automaticAdd" value="false"/>
    <booleanAttribute key="automaticIncludeRequirements" value="false"/>
    <booleanAttribute key="automaticValidate" value="true"/>
    <stringAttribute key="bootstrap" value=""/>
    <stringAttribute key="checked" value="[NONE]"/>
    <booleanAttribute key="clearConfig" value="true"/>
    <stringAttribute key="configLocation" value="${workspace_loc}/.metadata/.plugins/org.eclipse.pde.core/Jersey"/>
    <booleanAttribute key="default" value="false"/>
    <booleanAttribute key="default_auto_start" value="true"/>
    <intAttribute key="default_start_level" value="4"/>
    <setAttribute key="deselected_workspace_bundles"/>
    <stringAttribute key="featureDefaultLocation" value="workspace"/>
    <stringAttribute key="featurePluginResolution" value="workspace"/>
    <booleanAttribute key="includeOptional" value="true"/>
    <booleanAttribute key="org.eclipse.debug.core.ATTR_FORCE_SYSTEM_CONSOLE_ENCODING" value="false"/>
    <booleanAttribute key="org.eclipse.jdt.launching.ATTR_ATTR_USE_ARGFILE" value="false"/>
    <booleanAttribute key="org.eclipse.jdt.launching.ATTR_SHOW_CODEDETAILS_IN_EXCEPTION_MESSAGES" value="true"/>
    <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_START_ON_FIRST_THREAD" value="true"/>
    <stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
    <stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value="-os ${target.os} -ws ${target.ws} -arch ${target.arch} -nl ${target.nl} -consoleLog -console"/>
    <stringAttribute key="org.eclipse.jdt.launching.SOURCE_PATH_PROVIDER" value="org.eclipse.pde.ui.workbenchClasspathProvider"/>
    <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true -Declipse.ignoreApp=true -Dosgi.noShutdown=true"/>
    <stringAttribute key="pde.version" value="3.3"/>
    <setAttribute key="selected_features">
        <setEntry value="Jersey:default"/>
    </setAttribute>
    <setAttribute key="selected_target_bundles"/>
    <setAttribute key="selected_workspace_bundles"/>
    <booleanAttribute key="show_selected_only" value="false"/>
    <booleanAttribute key="tracing" value="false"/>
    <booleanAttribute key="useCustomFeatures" value="true"/>
    <booleanAttribute key="useDefaultConfigArea" value="true"/>
</launchConfiguration>
```
</details>

Here the zipped workspace:

[ws-jersey.zip](https://github.com/user-attachments/files/28879819/ws-jersey.zip)

This is the validation error I get when launching the application:

Image

... and then the start fails with:

```
org.osgi.framework.BundleException: Could not resolve module: org.glassfish.jersey.core.jersey-server [10]
  Unresolved requirement: Import-Package: org.glassfish.jersey; version="[2.48.0,3.0.0)"

	at org.eclipse.osgi.container.Module.start(Module.java:495)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:2110)
	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:146)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:2101)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:2043)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:2003)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1915)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:341)
```

Even though jersey-common is set to 2.48.0 in the feature.xml, the newer 4.0.2 is chosen. This doesn't happen when I have Jersey 2.48 and 3.1.11 in the classpath.

ptziegler:
The logic for calculating the included plug-ins for a feature-based launch configuration assigns resolved bundles a higher priority than bundles, that perfectly match the version in a feature.xml.

This means that even though the user configured their application to run with a specific version of a plug-in, PDE might decide to silently update it to its latest version, simply because it conflicts with a different plug-in in the IDE (which might not even be included in the app).

With this change, the logic has been adapted to first check for a version match, before checking the resolution state. So if a version matches perfectly, it will always be chosen first, even if it's not resolved.

Closes https://github.com/eclipse-pde/eclipse.pde/issues/2367

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2022 IBM Corporation and others.
+ * Copyright (c) 2007, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -66,6 +66,7 @@ import org.eclipse.pde.internal.build.BundleHelper;
 import org.eclipse.pde.internal.build.IPDEBuildConstants;
 import org.eclipse.pde.internal.core.DependencyManager;
 import org.eclipse.pde.internal.core.FeatureModelManager;
+import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.core.PDECore;
 import org.eclipse.pde.internal.core.PluginModelManager;
 import org.eclipse.pde.internal.core.TargetPlatformHelper;
@@ -352,10 +353,20 @@ public class BundleLauncherHelper {
 	private static final Predicate<IPluginModelBase> ENABLED_VALID_PLUGIN_FILTER = p -> p.getBundleDescription() != null && p.isEnabled();
 	private static final Function<IPluginModelBase, String> GET_PLUGIN_VERSION = m -> m.getPluginBase().getVersion();
 	private static final Comparator<IPluginModelBase> COMPARE_PLUGIN_RESOLVED = comparing(p -> p.getBundleDescription().isResolved());
+	private static final BiPredicate<IPluginModelBase, Version> EXACT_PLUGIN_VERSION = (e, v) -> Version.parseVersion(GET_PLUGIN_VERSION.apply(e)).equals(v);
 
 	private static IPluginModelBase getIncludedPlugin(String id, String version, String pluginLocation) {
 		List<List<IPluginModelBase>> plugins = getPlugins(id, pluginLocation);
-		return getIncluded(plugins, ENABLED_VALID_PLUGIN_FILTER, GET_PLUGIN_VERSION, COMPARE_PLUGIN_RESOLVED, version);
+		return getIncluded(plugins, ENABLED_VALID_PLUGIN_FILTER, GET_PLUGIN_VERSION, getIncludedPluginComparator(version), version);
+	}
+
+	private static Comparator<IPluginModelBase> getIncludedPluginComparator(String versionString) {
+		if (ICoreConstants.DEFAULT_VERSION.equals(versionString)) {
+			return COMPARE_PLUGIN_RESOLVED;
+		}
+		Version version = Version.parseVersion(versionString);
+		return Comparator.<IPluginModelBase, Boolean> comparing(e -> EXACT_PLUGIN_VERSION.test(e, version)) // false < true
+				.thenComparing(COMPARE_PLUGIN_RESOLVED);
 	}
 
 	private static IPluginModelBase getRequiredPlugin(String id, String version, int versionMatchRule, String pluginLocation) {
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/FeatureBasedLaunchTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/FeatureBasedLaunchTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2019, 2022 Julian Honnen and others.
+ *  Copyright (c) 2019, 2026 Julian Honnen and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -820,6 +820,32 @@ public class FeatureBasedLaunchTest extends AbstractLaunchTest {
 
 			assertGetMergedBundleMap("pluginResolution workspace no match", lc, Set.of());
 		}
+	}
+
+	@Test
+	public void testGetMergedBundleMap_requiredPluginWithSpecificVersion3() throws Throwable {
+		var targetBundles = ofEntries( //
+				bundle("plugin.a", "1.0.0", entry(IMPORT_PACKAGE, "test;version=\"1.0.0\""),
+						entry(EXPORT_PACKAGE, "test;version=\"1.0.0\"")), //
+				bundle("plugin.a", "2.0.0", entry(IMPORT_PACKAGE, "test;version=\"2.0.0\""),
+						entry(EXPORT_PACKAGE, "test;version=\"2.0.0\"")), //
+				bundle("plugin.b", "1.0.0", entry(IMPORT_PACKAGE, "test;version=\"[1.0.0,2.0.0)\"")), //
+				bundle("plugin.b", "2.0.0", entry(IMPORT_PACKAGE, "test;version=\"[2.0.0,3.0.0)\"")));
+
+		var targetFeatures = List.of( //
+				targetFeature("feature.a", "1.0.0", f -> {
+					addIncludedPlugin(f, "plugin.a", "1.0.0");
+					addIncludedPlugin(f, "plugin.b", "1.0.0");
+				}));
+
+		setTargetPlatform(targetBundles, targetFeatures);
+
+		ILaunchConfigurationWorkingCopy lc = createFeatureLaunchConfig();
+		lc.setAttribute(IPDELauncherConstants.FEATURE_DEFAULT_LOCATION, IPDELauncherConstants.LOCATION_WORKSPACE);
+		lc.setAttribute(IPDELauncherConstants.SELECTED_FEATURES, Set.of("feature.a:default"));
+
+		assertGetMergedBundleMap("featureResolution workspace", lc,
+				Set.of(targetBundle("plugin.a", "1.0.0"), targetBundle("plugin.b", "1.0.0")));
 	}
 
 	// --- required/imported feature dependencies ---


ptziegler:
Without this change, the test case would fail with:

```
org.opentest4j.AssertionFailedError: [featureResolution workspace] 
expected: {
plugin.a-1.0.0(e)="default:default",
plugin.b-1.0.0(e)="default:default"
}
 but was: {
plugin.a-1.0.0(e)="default:default",
plugin.a-2.0.0(e)="default:default",
plugin.b-2.0.0(e)="default:default"
}
```

Which is completely different than what's expected when looking at the feature.xml.

laeubi:
I would say that if a specific version is given in the feature this exact version must be used and no higher version is actually valid at all, so for me it looks not only like an ordering problem but that other plugins are actually selected at all.

ptziegler:
> I would say that if a specific version is given in the feature this exact version must be used and no higher version is actually valid at all

That's what I would expect as well. But there is this test case, where a `1.0.0` is set as additional plugin, even though the target definition only contains a `2.0.0` and where the expectation is that this newer version is then selected.

https://github.com/eclipse-pde/eclipse.pde/blob/109a4f552438cac0089211a4f0e15ee73097a92c/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/FeatureBasedLaunchTest.java#L1263-L1284

If I were to filter out all plugins that don't match the specified version, this test case would break. This is not the case when changing the order, as this only comes into play when the specified version really exists.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java
```diff
@@ -352,10 +353,20 @@ private static IFeature getRequiredFeature(String id, String version, int versio
 	private static final Predicate<IPluginModelBase> ENABLED_VALID_PLUGIN_FILTER = p -> p.getBundleDescription() != null && p.isEnabled();
 	private static final Function<IPluginModelBase, String> GET_PLUGIN_VERSION = m -> m.getPluginBase().getVersion();
 	private static final Comparator<IPluginModelBase> COMPARE_PLUGIN_RESOLVED = comparing(p -> p.getBundleDescription().isResolved());
+	private static final BiPredicate<IPluginModelBase, Version> EXACT_PLUGIN_VERSION = (e, v) -> Version.parseVersion(GET_PLUGIN_VERSION.apply(e)).equals(v);
 
 	private static IPluginModelBase getIncludedPlugin(String id, String version, String pluginLocation) {
 		List<List<IPluginModelBase>> plugins = getPlugins(id, pluginLocation);
-		return getIncluded(plugins, ENABLED_VALID_PLUGIN_FILTER, GET_PLUGIN_VERSION, COMPARE_PLUGIN_RESOLVED, version);
+		return getIncluded(plugins, ENABLED_VALID_PLUGIN_FILTER, GET_PLUGIN_VERSION, getIncludedPluginComparator(version), version);
```
ptziegler, line 360: I'm also not quite happy with doing the change here. I'll try to have a look later today to move it down into the `getIncluded(...)` method, so that it's in the same place as the already existing version logic.

laeubi:
> That's what I would expect as well. But there is this test case, where a `1.0.0` is set as additional plugin, even though the target definition only contains a `2.0.0` and where the expectation is that this newer version is then selected.

Where of course the question is if it is actual wanted behavior that is tested or the test just crafted in a way that it passes... At least for P2 an exact version is generated for a feature and Tycho resolves all `0.0.0` to the exact highest version for this reason as well.

ptziegler:
> At least for P2 an exact version is generated for a feature and Tycho resolves all 0.0.0 to the exact highest version for this reason as well.

I also painfully remember that Tycho fails if a plugin has e.g. the version `1.0.0.20260616`, but the feature.xml expects a `1.0.0.20260617`. Meaning the version has to match down to the qualifier, otherwise it's not a valid configuration.

By that logic, the following assertion is also wrong, because it expects the bundle with version `1.0.0.someQualifier` to be matched against the bundle with version `1.0.0`.

https://github.com/eclipse-pde/eclipse.pde/blob/109a4f552438cac0089211a4f0e15ee73097a92c/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/launcher/FeatureBasedLaunchTest.java#L367-L374

@HannesWell I think you initially worked on this as part of [Bug 578005](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578005) - Extend FeatureBasedLaunchTest to fully cover Feature-based launches. Do you remember the use-cases you had in mind when you originally wrote those tests?

laeubi:
If you click on "Versions" in the feature editor tab for included plugins one can see that there is only the option to "Synchronize on build" (e.g. no version given) or to sync it exactly or force the feature version into the plugin so I think its clear also from the UI that this is meant to either be exact or an empty version.

merks:
I don’t recall exactly, but maybe .qualifier is allowed too?

ptziegler:
> I don’t recall exactly, but maybe .qualifier is allowed too?

Yes, .qualifier is valid, too. And also what you get with the "Version Synchronisation" dialog when you select the "Copy versions from plug-in, feature and fragment manifests" option. Tycho also supports this and will replace the qualifier with the timestamp during the build.

For the validation, this doesn't make a difference, as the version still matches perfectly. I'll wait a few days in case someone else knows more about why this was implemented the way it is. Otherwise I'll adjust the filter to discard all artifacts that don't have the expected version.

Issue #215, opened 2022-07-15
[API Tools] should support Declarative Services classpath additions


laeubi:
The PDE(!) DS Plugin allows to set a flag to include the annotations without importing them:

**org.eclipse.pde.ds.annotations.prefs**:
```
classpath=true
dsVersion=V1_3
eclipse.preferences.version=1
enabled=true
generateBundleActivationPolicyLazy=true
path=OSGI-INF
validationErrorLevel=error
validationErrorLevel.missingImplicitUnbindMethod=error
```
but correctly fails execution if one do so see for example: https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-210/4/console

laeubi:
This is actually cause by https://github.com/eclipse/tycho/pull/1246

HannesWell:
> This is actually cause by [eclipse-tycho/tycho#1246](https://github.com/eclipse-tycho/tycho/pull/1246)

Can we then close this as invalid?

PR #2374, opened 2026-06-17
Trace duration of classpath container updates

vogella:
Adds duration tracing to the PDE classpath update job (`UpdateClasspathsJob`), gated behind the existing `org.eclipse.pde.core/classpath` debug option. When enabled, the job traces how long it spent computing the classpath entries, applying the containers to the Java model, and the overall run. This makes it possible to diagnose slow classpath updates from the trace output alone, without attaching a profiler. No behavior changes when tracing is off.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
@@ -27,6 +27,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
@@ -101,6 +102,7 @@ public class ClasspathContainerState {
 
 		@Override
 		protected IStatus run(IProgressMonitor jobMonitor) {
+			long startNanos = PDECore.DEBUG_CLASSPATH ? System.nanoTime() : 0;
 			SubMonitor monitor = SubMonitor.convert(jobMonitor, PDECoreMessages.PluginModelManager_1, 100);
 			PluginModelManager.getInstance().initialize(monitor.split(1));
 			PluginModelManager modelManager = PluginModelManager.getInstance();
@@ -111,6 +113,7 @@ public class ClasspathContainerState {
 			int count = requests.size();
 			monitor.setWorkRemaining(count * 2);
 
+			long computeNanos = PDECore.DEBUG_CLASSPATH ? System.nanoTime() : 0;
 			String messageTemplate = PDECoreMessages.PluginModelManager_1 + " ({0}/{1}): {2}"; //$NON-NLS-1$
 			for (int i = 0; i < requests.size(); i++) {
 				UpdateRequest req = requests.get(i);
@@ -138,6 +141,7 @@ public class ClasspathContainerState {
 				}
 				monitor.worked(1);
 			}
+			traceRuntime("Computed classpath of %2$d project(s) in %1$d ms.", computeNanos, count); //$NON-NLS-1$
 			if (monitor.isCanceled()) {
 				return Status.CANCEL_STATUS;
 			}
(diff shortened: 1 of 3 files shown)


vogella:
AFAICS neither the licence check nor the new maven warnings (None touch pde.core. These appear in PR builds but not in the stale reference build #215, so warnings-ng flags them as "new".) are related to this PR.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -101,6 +101,7 @@ public boolean belongsTo(Object family) {
 
 		@Override
 		protected IStatus run(IProgressMonitor jobMonitor) {
+			long startNanos = PDECore.DEBUG_CLASSPATH ? System.nanoTime() : 0;
```
HannesWell, line 105: Since the output is in milliseconds, wouldn't it be suitable to use milliseconds from the beginning?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -201,6 +219,10 @@ void add(IProject project, IClasspathContainer classpathContainer) {
 			schedule();
 		}
 
+		private static long elapsedMillis(long startNanos) {
+			return (System.nanoTime() - startNanos) / 1_000_000L;
+		}
+
```
HannesWell: The invocation of this uses a repeating pattern, which I suggest to unify to something
```suggestion
		private void traceRuntime(String msg, long start, Object... args) {
			if (PDECore.DEBUG_CLASSPATH) {
				long elapsedMillis = (System.nanoTime() - start) / 1_000_000L;
				Object[] allArgs = Stream.concat(Stream.of(elapsedMillis), Stream.of(args)).toArray();
				PDECore.TRACE.trace(PDECore.KEY_DEBUG_CLASSPATH, String.format(msg, allArgs));
			}
		}
```
Which can then be called with (almost) one-liners:
- `traceRuntime("Computed classpath of %2$d project(s) in %1$d ms.", computeNanos, count); //$NON-NLS-1$`
- traceRuntime("Applied classpath container of %2$d project(s) in %1$d ms.", applyNanos, n); //$NON-NLS-1$
- `traceRuntime("UpdateClasspathsJob finished in %d ms: %d request(s), %d updated, %d error(s).", startNanos, count, updateProjects.size(), errorsPerProject.size());  //$NON-NLS-1$`

HannesWell (changes requested):
This should be useful to gather more performance data.
Besides the mentioned use of milliseconds I think the repeating tracing code can be unified.
Please see my comment below.
With that applied I'm fine with this.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: The traceRuntime helper keeps nanoTime() and only divides to ms at the end, so the durations stay monotonic (immune to wall-clock adjustments) while the output is in ms as before. Let me know if you'd still prefer currentTimeMillis().

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: Applied, thanks. Folded the three trace blocks into your traceRuntime helper and dropped the separate elapsedMillis method.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell: > the durations stay monotonic (immune to wall-clock adjustments)

Have to admit that I haven’t considered that. I'd say that if one modifies a computer's clock it's acceptable to have wrong time recordings in tracing  and `System.currentTimeMillis()` is sufficient (and a bit simpler).

But I'm also fine to keep it, no strict opposition from my side.

HannesWell (approved):
Looks good. Thanks.

vogella:
Build failure seem unrelated.

HannesWell:
> Build failure seem unrelated.

See https://github.com/eclipse-pde/eclipse.pde/pull/2380#issuecomment-4756835159.
Ideally please hold off submitting this until we have a green build again.

HannesWell:
The build is green and this looks good. Lets have this.
Thanks for providing this new tracing.

PR #2375, opened 2026-06-17
Compute classpath container updates in parallel

vogella:
Speeds up the PDE classpath container update when many projects are refreshed in one run, for example after a target platform reload.
The per-project computation only reads the resolved target state and each project's own saved state, so it now runs on a bounded `JobGroup` (one thread less than the number of cores) and the results are applied to the Java model in the existing single `JavaCore.setClasspathContainer` call.
bnd-backed projects and machines with few cores fall back to sequential computation, and the whole thing can be turned off with the new "Update plug-in classpaths in parallel" preference on the PDE main page (on by default).

This PR is now focused on the parallel computation only.
The queue refactor and the shared dependency-closure cache that were previously bundled here will be proposed separately, so the parallel change can be reviewed and measured on its own.

Draft for now. Contributes to https://github.com/eclipse-pde/eclipse.pde/issues/2361


laeubi:
> The base commit is a small refactor of the update queue that the cache and parallel changes build on.

If it is independent it should be made a separate PR so it can eventually already be merged what makes review easier.

>  Parallelization can be disabled with -Dpde.classpath.parallelUpdate=false

I don't think a system property is the right choice here and instead should be a preference that can be set by the user like for other options (e.g. API tools) with PDE.

vogella:
Thanks for the feedback @laeubi. I agree to both points, they will be adressed before I convert that out of draft.

vogella:
Some wall clock numbers from my slow client:

Without this parallelization: Update of classpath between 3-8 mins
With this parallelization (and the included caching): ~ 60 - 90 seconds

HannesWell:
> Without this parallelization: Update of classpath between 3-8 mins
> With this parallelization (and the included caching): ~ 60 - 90 seconds

That's a great improvement.
But as it's involves an important part of PDE and concurrency can lead to subtile bugs this should be reviewed with great care.

And since optimizations usually introduce more complexity and therefore are potentially a source of bugs it should be sorted out what's actually needed and what can stay as it is.
Especially when the classpath is now updated in parallel, I'm not immediately convinced that the change to `Share dependency closure computation across classpath updates` as a significant impact. But I'd change my mind with corresponding numbers proving it's usefulness. That being said, that this discussion is for the PR proposing that feature.
My actual point is that this PR's content must become more focused and shouldn't be a collection of many (not strongly connected) changes.

Without having looked into the details yet, I believe instead of spawning new ThreadPools it would probably be better to use Eclipse Jobs instead? With corresponding SubMonitors we should even get visually nice sub-progress reporting.

r-mennig:
Thanks a lot @vogella for working on this.

We are also experiencing some slowdowns after the latest changes to the classpath calculation so I wanted to share some  numbers for the classpath container update of our 1858 projects.

With -Dpde.addTransitiveDependenciesWithForbiddenAccess=true:
- Current master: 70sec
- With this optimization: 19sec

With -Dpde.addTransitiveDependenciesWithForbiddenAccess=false:

- Current master: 23sec
- With this optimization: 10sec

vogella:
> Thanks a lot @vogella for working on this.
> 
> We are also experiencing some slowdowns after the latest changes to the classpath calculation so I wanted to share some numbers for the classpath container update of our 1858 projects.
> 
> With -Dpde.addTransitiveDependenciesWithForbiddenAccess=true:
> 
> * Current master: 70sec
> * With this optimization: 19sec
> 
> With -Dpde.addTransitiveDependenciesWithForbiddenAccess=false:
> 
> * Current master: 23sec
> * With this optimization: 10sec

Thnaks for the feedback and testing. I plan to make this one independent of the caching commit (as suggsted by the reviewers), would be great if you can test this again, so that we have reliable data for the decision if classpath caching is actually helpful or not.

vogella:
Rebased onto origin/master to include the trace option, now using a JobGroup as suggested by @HannesWell and added a preference as suggested by @laeubi. Also removed the two dependent commits from this PR as suggested to be able to measure this independently.

vogella:
@r-mennig please measure again with this change only.

r-mennig:
@vogella Thanks for the update

I've measured again on our main codebase and found almost no difference compared to the previous implementation: 

With -Dpde.addTransitiveDependenciesWithForbiddenAccess=true: 17-19sec
With -Dpde.addTransitiveDependenciesWithForbiddenAccess=false: 9-10sec

If container updates are required (I've tested this by toggling the pde.addTransitiveDependenciesWithForbiddenAccess in the previous run), the full update times are now:

With -Dpde.addTransitiveDependenciesWithForbiddenAccess=true: 36sec
With -Dpde.addTransitiveDependenciesWithForbiddenAccess=false: 25sec

(I've seen that setting the updated containers is still done sequentially, I've just added those data points here for comparison)

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -104,47 +107,38 @@ public boolean belongsTo(Object family) {
...
-							}
-						} catch (CoreException e) {
-							errorsPerProject.put(project, e.getStatus());
-						}
-					}
-				}
-				monitor.worked(1);
-			}
+			List<ComputedClasspath> computed = computeAll(requests, modelManager, parallel, monitor.split(90));
 			traceRuntime("Computed classpath of %2$d project(s) in %1$d ms.", computeNanos, count); //$NON-NLS-1$
```
r-mennig, line 157: I think the progress monitor task should be updated here. If actual container updates are required, the current progress reporting looks like the last few computations are stuck.

Image

vogella:
@r-mennig feedback reporting should be better.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell, line 138: In a follow-up the monitor could passed into the `computeClasspathEntries()` method and progress there to in order to get even more fine grained progress.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -191,6 +203,54 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
+				List<Job> jobs = new ArrayList<>();
+				for (UpdateRequest request : requests) {
+					if (BndProject.isBndProject(request.project())) {
+						bndProjects.add(request);
+					} else {
+						IProgressMonitor computationMonitor = monitor.slice(1);
+						jobs.add(Job.create(PDECoreMessages.PluginModelManager_1, m -> {
+							classpathComputation.accept(request, computationMonitor);
+							computationMonitor.done();
+						}));
```
HannesWell: In general I think it could be convenient to introduce some structure concurrency like APIs (comparable with the new plain Java APIs in the making), so that one doesn't have to create a sliced monitor and job group manually.
A group of current tasks could even be displayed in the progress view specifically (not not system jobs).

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -106,23 +113,24 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
-			Map<IProject, IStatus> errorsPerProject = new LinkedHashMap<>();
+			Map<IJavaProject, IClasspathContainer> updateProjects = Collections.synchronizedMap(new HashMap<>());
+			Map<IProject, IStatus> errorsPerProject = Collections.synchronizedMap(new LinkedHashMap<>());
 
 			List<UpdateRequest> requests = drainRequests(workQueue);
 			int count = requests.size();
 			monitor.setWorkRemaining(count * 2);
+			Map<IProject, Integer> projectIndex = IntStream.range(0, count)
+					.mapToObj(i -> Map.entry(requests.get(i).project(), i))
+					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
HannesWell: I just created this based on the assumption that the order in which the projects are passed to `setProjectContainers()` respectively `JavaCore.setClasspathContainer` is important.
Can somebody confirm that?
If not, we could just use random order of `updateProjects`.

HannesWell (commented):
Thanks for this Lars.
Generally I couldn't investigate the parallelization in absolute detail and I'm not 100% sure if computing the project class-path is inherently thread-safe, but all explanations I read/found were logical to me. So I assume this is in general sufficiently save.

I conducted a more in depth review and just pushed a second commit where tried to simplify the proposed solution significantly, e.g. by using `IProgressMonitor.slice(int)` and removed things that seemed unnecessary. With the new proposal existing code is also changed less.
Additionally the new approach also parallelizes  the save of the classpath state, which writes a dedicated file per project and I therefore assume is inherently thread-safe.

@vogella please have a look and if you think this is fine, please squash it into your first commit.
I prefer to not have this in separate commits eventually, I just created a new commit for now to ease further review.

Since I was little in time, I only could test this on a small workspace, but there everything looked good.
@r-mennig, if you have some time, could you test this again and also check if the progress-reporting is suitable? Thanks in advance.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -106,23 +113,24 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 			SubMonitor monitor = SubMonitor.convert(jobMonitor, PDECoreMessages.PluginModelManager_1, 100);
 			PluginModelManager.getInstance().initialize(monitor.split(1));
 			PluginModelManager modelManager = PluginModelManager.getInstance();
-			Map<IJavaProject, IClasspathContainer> updateProjects = new LinkedHashMap<>();
-			Map<IProject, IStatus> errorsPerProject = new LinkedHashMap<>();
+			Map<IJavaProject, IClasspathContainer> updateProjects = Collections.synchronizedMap(new HashMap<>());
```
laeubi: Any reason not using a concurrent has map here instead of a synchronized one?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -106,23 +113,24 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
+			Map<IProject, IStatus> errorsPerProject = Collections.synchronizedMap(new LinkedHashMap<>());
```
laeubi, line 115: I found it often more sustainable on the long run to define a state object instead of two maps holding related state. This design also makes it possible to move computation into such state objects what makes the code easier to understand and allow further synchronization to be performed there.

If I see it right one then even has a third map with an index below. Such state object can the also trace the nanos used for computation and so on.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -160,12 +171,13 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
-				for (Entry<IJavaProject, IClasspathContainer> entry : updateProjects.entrySet()) {
-					javaProjects[i] = entry.getKey();
-					container[i] = entry.getValue();
-					i++;
-				}
+				updateProjects.forEach((javaProject, cpContainer) -> {
+					int i = projectIndex.get(javaProject.getProject());
+					javaProjects[i] = javaProject;
+					container[i] = cpContainer;
+				});
```
laeubi: If we have a state object this quite complicated block can be a simple states.stream().map(state->state.container).toArray() call ...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -191,6 +203,54 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 			return overallStatus;
 		}
 
+		private void computePluginClasspaths(List<UpdateRequest> requests,
+				BiConsumer<UpdateRequest, IProgressMonitor> classpathComputation, SubMonitor monitor) {
```
laeubi:
```suggestion
				BiConsumer<UpdateRequest, IProgressMonitor> classpathComputation, IProgressMonitor monitor) {
```

otherwise the convert makes not much sense below...

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
laeubi: Sadly the Jobs API is quite clumsy making it hard to map it consistently to other APIs.
The Jobs API is more designed for *reusable jobs* that can be restarted than for one time tasks (in wich case implementing Executor would make it more conveniently being used with modern Java APIs)

r-mennig:
@HannesWell @vogella I'm getting an index out of bounds with the lastest version if updates are required for some projects but not all of them

```
java.lang.ArrayIndexOutOfBoundsException: Index 1825 out of bounds for length 1816
	at org.eclipse.pde.internal.core.ClasspathContainerState$UpdateClasspathsJob.lambda$4(ClasspathContainerState.java:176)
	at java.base/java.util.HashMap.forEach(HashMap.java:1429)
	at java.base/java.util.Collections$SynchronizedMap.forEach(Collections.java:2956)
	at org.eclipse.pde.internal.core.ClasspathContainerState$UpdateClasspathsJob.run(ClasspathContainerState.java:174)
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
r-mennig: Getting the index from the map of all projects for which updates were requested assumes that an update is actually required for all of them (which might not be the case and leads to out of bounds exceptions otherwise, see https://github.com/eclipse-pde/eclipse.pde/pull/2375#issuecomment-4829834508)

r-mennig (changes requested):

vogella:
@HannesWell your change introduces a few regressions:

AIOOBE in the apply phase. projectIndex maps each project to its position in the full request list (0..count-1), but the apply arrays are sized n = updateProjects.size(). As soon as fewer projects need updating than were requested (some up-to-date, closed, non-PDE, or errored), a surviving project's index can exceed n and javaProjects[i] = … throws ArrayIndexOutOfBoundsException, so the job fails and no classpaths get applied. That's the common case, also reported by @r-mennig. The original i++ counter fill was correct; ordering doesn't matter for JavaCore.setClasspathContainer, so projectIndex is both unnecessary and broken.

Concurrency. The fixup moves monitor.setTaskName(...) and computationMonitor.done() onto the worker threads while the main thread still calls split(). SubMonitor is not thread-safe; the original kept all monitor access on a single thread.

Lost error resilience. The old computeOne caught RuntimeException, logged it, and recorded an error so the batch continued. The new consumer only catches CoreException, so a runtime exception aborts the whole batch (sequential) or is silently dropped (parallel).

Minor. Removing monitor.setWorkRemaining(0) leaves the progress bar near 50% when nothing needs applying, and errorsPerProject.remove(project) is a no-op since each project is processed once.

I fix these issues and will try to keep your improvements and simplifications.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: I will update the PR with this.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: I will update the PR with this.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: @HannesWell fixup removed the per-project record.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: I think this comment is now obsolete with @HannesWell changes, please check again.

HannesWell:
Thanks for testing this Robin, indeed there was an index mismatch.
And thanks Lars for the follow-up fixes.
With Christoph's comments I start to think that overall the code could be simpler if we keep the introduction of the `ComputedClasspath` (or alike state object). Then we could pass a future into the lambda that computes the classpath, potentially concurrently and can maintain the order.

I'll look into that.

> ordering doesn't matter for JavaCore.setClasspathContainer, so projectIndex is both unnecessary and broken.

If you are sure about that? In that case generally simplify the result extraction using streams and would stop handling array indices manually.

And as a more general question: Are you sure that the distinction between an ordinary PDE Plugin project and a 'PDE-BND'- project, i.e. one with automatically generated MANIFEST files is necessary?

Furthermore as a general remark:
I have the impression that the code in this PR and at least parts of your answers in this PR are AI generated. But I'm totally missing corresponding markers. I assumed that the EF project rules mandate noting the use of generative AI, e.g. in commit messages. And also on a human level, I'd like to know if I discuss with a human or a machine. And as you probably know best, while AI can be a very useful tool, it sometimes goes completely off rails. And I don't want to find out by myself where I have to pay attention to an AI going wild and where I can expect human common sense.
If I'm not sure I'll lean to the save side and will question everything suspicious in great detail.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -174,6 +178,8 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 					return e.getStatus();
 				}
 			}
+			// Reach 100 % even when nothing needed to be applied.
+			monitor.setWorkRemaining(0);
```
HannesWell: Since there is only very little work happening after this call, which should be fast and it's the responsibility of the Job framework to complete the monitor passed into this job, this seems unnecessary.
If at all, I think it would be better to call `done()`?
But @vogella you've worked a lot on code using the progress monitor, so I assume you know this best.

vogella: done() calls are only necessary for top-level monitors

vogella:
> If you are sure about that? In that case generally simplify the result extraction using streams and would stop handling array indices manually.



> Thanks for testing this Robin, indeed there was an index mismatch. And thanks Lars for the follow-up fixes. With Christoph's comments I start to think that overall the code could be simpler if we keep the introduction of the `ComputedClasspath` (or alike state object). Then we could pass a future into the lambda that computes the classpath, potentially concurrently and can maintain the order.
> 
> I'll look into that.

Thanks. I assume you will update the PR

> > ordering doesn't matter for JavaCore.setClasspathContainer, so projectIndex is both unnecessary and broken. 
> If you are sure about that? In that case generally simplify the result extraction using streams and would stop handling array indices manually.

AFAICS.

> And as a more general question: Are you sure that the distinction between an ordinary PDE Plugin project and a 'PDE-BND'- project, i.e. one with automatically generated MANIFEST files is necessary?

I have no experience with BND tools so I thought leaving them sequentially is better. If you have a big workspace with them, you can try out to do that out.

> Furthermore as a general remark: I have the impression that the code in this PR and at least parts of your answers in this PR are AI generated. But I'm totally missing corresponding markers. I assumed that the EF project rules mandate noting the use of generative AI, e.g. in commit messages. And also on a human level, I'd like to know if I discuss with a human or a machine. And as you probably know best, while AI can be a very useful tool, it sometimes goes completely off rails. And I don't want to find out by myself where I have to pay attention to an AI going wild and where I can expect human common sense. If I'm not sure I'll lean to the save side and will question everything suspicious in great detail.

Human answer, I typically use Claude code to investigate and confirm findings but try to answer manually.

HannesWell:
> Thanks. I assume you will update the PR

Yes, I'm on it at the moment.

> > And as a more general question: Are you sure that the distinction between an ordinary PDE Plugin project and a 'PDE-BND'- project, i.e. one with automatically generated MANIFEST files is necessary?
> 
> I have no experience with BND tools so I thought leaving them sequentially is better. If you have a big workspace with them, you can try out to do that out.

I talked to Christoph and we both don't see a reason at this part to handle these projects differently.

> Human answer, I typically use Claude code to investigate and confirm findings but try to answer manually.

It's of course fine to share an analysis of whatever AI one uses, but it should be clear that it's AI generated, e.g. by writing `XYZ says...`

r-mennig:
@vogella @HannesWell I've tested the progress reporting again for the current version.

If no updates are needed, the progress jumps from 50% to completion. This should be fixable by reporting the double amount of progress for projects which need no update.

ProgressReportingNoUpdate

If updates are needed, the progress gets stuck on 100% for multiple seconds now. This is fixed by calling `done()` on the monitor after the classpath containers are updated (@HannesWell already proposed this fix in https://github.com/eclipse-pde/eclipse.pde/pull/2375#discussion_r3491320490)

ProgressReportingUpdate

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -160,12 +162,14 @@ protected IStatus run(IProgressMonitor jobMonitor) {
...
-				for (Entry<IJavaProject, IClasspathContainer> entry : updateProjects.entrySet()) {
+				int i = 0;
+				for (var entry : updateProjects.entrySet()) {
 					javaProjects[i] = entry.getKey();
 					container[i] = entry.getValue();
 					i++;
 				}
 				try {
+					monitor.setTaskName(NLS.bind(PDECoreMessages.ClasspathContainerState_applying, n));
 					monitor.setWorkRemaining(n);
```
r-mennig, line 183:
```suggestion
					monitor.done();
```

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -131,16 +132,18 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 									javaProject.getProject());
 							if (!isUpToDate(project, entries, req.container())) {
 								updateProjects.put(javaProject, PDEClasspathContainerSaveHelper.containerOf(entries));
-								errorsPerProject.remove(project);
 								saveState(project, entries);
 							}
```
r-mennig, line 146:
```suggestion
							} else {
							  // report double progress if no update is needed
							  monitor.worked(1);
							}
```

vogella:
> I talked to Christoph and we both don't see a reason at this part to handle these projects differently.

Ok, I assume your update will also include that change.
 
> > Human answer, I typically use Claude code to investigate and confirm findings but try to answer manually.
> 
> It's of course fine to share an analysis of whatever AI one uses, but it should be clear that it's AI generated, e.g. by writing `XYZ says...`

I rarely copy 1:1 the analysis of Claude into a discussion (even though the idea is sometimes tempting).

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: If done () fixes that, I think it is fine to include that. @HannesWell as you are updating the PR, please take this also over.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell: The the `monitor` here is a `SubMonitor` wrapper of the job's monitor, which I assume qualifies as top-level monitor, doesn't it?

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
```diff
@@ -220,12 +276,8 @@ private void traceRuntime(String msg, long start, Object... args) {
...
 
 	/**
-	 * Drains the given queue and deduplicates requests for the same project.
-	 * Multiple requests for one project are queued e.g. when a target platform
-	 * reload triggers classpath updates from several listeners. A request
-	 * without a saved container state forces a container update and therefore
-	 * wins over requests whose update may be skipped when the computed entries
-	 * match the saved state.
+	 * Drains the queue, keeping one request per project. A request without a
+	 * saved container forces an update and so wins over a skippable one.
```
HannesWell: I think this change is something for https://github.com/eclipse-pde/eclipse.pde/pull/2366, respectively would be changed by that.
For this change I suggest to focus on functional changes and apply clean-ups separately.

I'll revert that in my fixup.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
vogella: No, the monitor is passed to the method in the run monitor, that is the top-level monitor. But I would in this case just call done() to update the process fast and maybe at a later point investigate why .setWorkRemaining(0) does not update immediately.  No need to hold this critical performance improvement back due to a done() call.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell: @r-mennig calling done() here doesn't seems correct because two lines below we call
```
setProjectContainers(javaProjects, container, monitor.split(n));
```
And then the monitor should have n ticks left. So the current state looks correct to me.

r-mennig: Sorry, this is indeed the wrong line. I've meant to change the line 182 (the same line you've mentioned in your suggestion), calling done after `setProjectContainers(javaProjects, container, monitor.split(n));`

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathContainerState.java
HannesWell: The suggestion to double the consumed monitor ticks for projects not needing an update sound good, but is difficult to implement with the current monitor splitting.
To make this possible we'd need to reference the top level monitor from the very bottom and of course would need to sync it manually.
If no updates is a very common scenario and the update itself doesn't take too long, we could already assing the submonitor split-off for the Classpath compuations more ticks, e.g. 2/3 of all instead of just one half.

r-mennig: The monitor referenced here is not the `updateMonitor` passed to the `classpathComputation` lambda, but the initial Submonitor we created on method entry. So although only half of the ticks are given to `computePluginClasspaths`, this now consumes an extra tick on the parent monitor whenever an update is skipped (e.g. lets say we request an update for 100 projects, 50 of which need updating, then `classpathComputation` consumes 100 ticks and an additional 50 are consumed directly from within the lambda, leaving 50 ticks remaining for updating the classpath of the 50 other projects)

This is a little hacky, but as far as I can tell it should work.

Another idea could be to pre-split the monitor for setting the updated classpath containers and reporting the progress there 


```
int count = request.size();
int ticks = 2 * count;
Submonitor computeClasspathMonitor = monitor.split(count);
Submonitor setClasspathMonitor = monitor.split(count);
BiConsumer<UpdateRequest, IProgressMonitor> classpathComputation = (req, updateMonitor) -> {
  ...
  if (!isUpToDate(project, entries, req.container())) {
    ...
  } else {
    setClasspathMonitor.worked(1);
  }
  ...
};
computePluginClasspaths(requests, classpathComputation, computeClasspathMonitor); 
...
setProjectContainers(javaProjects, container, setClasspathMonitor);

HannesWell:
@r-mennig I've now tried to use the hacky way which I believe works and worked in my little case.
Hopefully this will be improved subsequently, but as Lars said above we shouldn't hold back this performance improvement many longer just to improve the progress reporting. Of course unless there are clear bugs in it.

With the removal of the special handling of PDE-BND project I could reintroduce the task name setting, which IMO gives nicer progress reports than the generic name from the job group. Furthermore it will later allow even more fine-grained progress reporting within the CP computation.
I hope this is now also synchronized correctly.

Can you please test this again quickly on your larger workspace? Unfortunately I have only very little availability left and can't further help with this much later.

r-mennig:
@HannesWell Thanks for the fixes, progress reporting looks good now in both the update and non-update case

HannesWell:
Thanks for checking! 

@vogella are you fine with this too? 
If yes, do you want to squash that the commits together or should I do it?

HannesWell (approved):
I took the liberty to squash the commits into one commit now since I won't have more time for this.
From my POV this is now ready and can be submitted.
Thanks @vogella for providing this and the discussion.

HeikoKlare:
Thank you, @vogella, for proposing this, and all of you for the collaborative work on this 🏅 
This is a great enhancement for us (the 70 seconds, and sometimes even more, that Robin reported were really painful) as well as for every other IDE user who will benefit from the dramatically increased calculation speed (particularly affecting IDE startup).

vogella:
I did a quick check (not a detailed review) and the changes looks fine for me. BND Tools I cannot judge. This is also ready for merging from my side. Looks like @r-mennig still has a change request?

Thanks to @HannesWell and @r-mennig for the cooperation on this. 

> the 70 seconds, and sometimes even more

My record in updating the classpath is still 8:30 mins with a specific client on a slow v-server and this change is definitely huge improvement.

r-mennig (approved):

r-mennig:
@vogella  My change requests have been addressed (those were about the index out of bounds exceptions which have fixed). This looks good to merge from my side as well.

PR #2376, opened 2026-06-18
Warn about placeholder and part elementId mismatch

akurtakov:
Make the e4xmi editor warn when placecholder and part elementId mismatch. As per
https://github.com/eclipse-platform/eclipse.platform.ui/blob/4e3be89f02f4d30e6f8d0415b798675bd85b9ff9/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/PlaceholderResolver.java#L32 "resolver that expects the id of the placeholder to match the id of a PartDescriptor" .
It is not a problem on initial load as the ref is there but if one tries to dynamically reset/load/etc. manipulate e4 model the code at https://github.com/eclipse-platform/eclipse.platform.ui/blob/4e3be89f02f4d30e6f8d0415b798675bd85b9ff9/bundles/org.eclipse.e4.ui.workbench/src/org/eclipse/e4/ui/internal/workbench/PlaceholderResolver.java#L53 will fail if the elementId is not matching.

### DIFF
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.ui/META-INF/MANIFEST.MF
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.e4.tools.emf.ui;singleton:=true
-Bundle-Version: 4.9.100.qualifier
+Bundle-Version: 4.9.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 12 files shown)


merks (approved):
Looks sound to me.  Good to see a test for it.  I just wondered about the explicit use/need for final on some of the variables, but fine either way.

akurtakov:
> Looks sound to me. Good to see a test for it. I just wondered about the explicit use/need for final on some of the variables, but fine either way.

It matches the existing code https://github.com/eclipse-pde/eclipse.pde/blob/8d27d0c4809df4196253a5c170708abb99662cd5/e4tools/bundles/org.eclipse.e4.tools.emf.ui/src/org/eclipse/e4/tools/emf/ui/internal/common/component/PlaceholderEditor.java#L92 and later explicitly setting final.

akurtakov:
Merging as the test failures on Mac are unrelated.

PR #2379, opened 2026-06-19
Revert "Add Select Required and Remove Unresolved actions to validation dialog"

vogella:
My runtime Eclipse behaves very strange, this is in preparation in case I think a real issue. I now run a full aggregator build to see if my runtime setup is strange or if this commit is not behaving correct. 

Reverts eclipse-pde/eclipse.pde#2351


vogella:
False alarm, local aggregator works fine

PR #2380, opened 2026-06-19
Document running single test bundles and classes in AGENTS.md

vogella:
Adds a short 'Running Specific Tests Locally' section to AGENTS.md.

It documents how to run a single test bundle with `-pl`, a single test class with `-Dtest=<ClassName>` (which overrides the bundle's default suite), how to build a source bundle together with its test bundle, and the Wayland-safe headless invocation. The commands were verified locally against `org.eclipse.pde.ui.tests`, where `-Dtest` correctly ran a single class instead of the full `AllPDETests` suite. This makes local test iteration much faster than running the whole suite.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,32 @@ mvn clean verify -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 
 **Maven Config (`.mvn/`):** Tycho 4.0.13 extension, `-Pbuild-individual-bundles`, `-DtrimStackTrace=false`
 
+### Running Specific Tests Locally
+
+Always use `mvn verify` (NOT `mvn test`): Tycho binds tests to the `integration-test` phase, so `mvn test` runs nothing.
+
+**One test bundle** via `-pl` (artifact selector):
+```bash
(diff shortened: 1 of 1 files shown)


vogella:
@HannesWell pinging you as PDE project lead, no idea by the PR validation is failing but it is definitely not cause by this PR.

HannesWell:
> no idea by the PR validation is failing but it is definitely not cause by this PR.

I assume it's because in I-builds the Javadoc in the pde.doc bundle is built with Java-25 while in the verification build it's build with Java-21. And that JDK's number is recorded in the generated files and also leads to other slightly different content.
So we we either have to run the PR verification builds with java-25 too or should make sure that the pde.doc bundle is always built with java-21.
I'm not in favor of the former since we generally not allow Java-25 yet and expect the latter to be implementable with a corresponding configuration of the `maven-toolchains-plugin`.
I'll look into it over the weekend as time permits.

akurtakov:
Run the PR verification build is inevitable - moving to Java 25 will be allowed for December release thus figuring build/releng things earlier could only benefit us.

PR #2384, opened 2026-06-25
Fix Templates wizard description showing white in dark theme

vogella:
The description panel of the New Plug-in Project **Templates** page renders white in the dark theme.

Two causes in `FormBrowser`:

- Colors came from `FormToolkit`, which derives them from OS system colors (`SWT.COLOR_LIST_BACKGROUND`/`FOREGROUND`) that stay light when only the Eclipse theme is dark. Now uses the CSS-themed parent colors.
- A disabled `FormText` paints with `SWT.COLOR_WIDGET_BACKGROUND`, ignoring its background — and the panel starts disabled (template checkbox unchecked). The `FormText` is now kept enabled with only its text greyed, so the background stays themed while input remains blocked via the disabled container.

Result: the description panel follows the theme in both enabled and disabled states; light theme is unchanged.


vogella:
Before:

image


After:

image

PR #2385, opened 2026-06-26
Fix CSS Spy for the e4 engine's SAC removal

vogella:
After the e4 CSS engine dropped SAC, the CSS Spy's scratch pad still caught `org.w3c.css.sac.CSSParseException`, which is no longer thrown and is no longer shipped.
A syntax error therefore threw the engine's unchecked `CssParseException` uncaught instead of showing the error. This catches it generically; the line and column still appear because they are part of the exception message.

It also repairs the spy's matched-rule lookup, which tested the engine with `instanceof DocumentCSS`. The engine only implements `CSSEngine` now, so that test was always false and the rule-source column stayed empty; it uses `getDocumentCSS()` instead.

Both changes use only API present on the current platform, so the bundle builds against the released target (verified with `mvn verify`).

### DIFF
--- a/ui/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css/CSSScratchPadPart.java
+++ b/ui/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css/CSSScratchPadPart.java
@@ -36,7 +36,6 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Text;
-import org.w3c.css.sac.CSSParseException;
 import org.w3c.dom.stylesheets.StyleSheet;
 import org.w3c.dom.stylesheets.StyleSheetList;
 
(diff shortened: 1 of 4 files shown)


vogella:
@HannesWell this warning seems unrelated, can you have a look?

<pre><code>Some attempts to read artifact osgi.bundle,org.osgi.service.event.source,1.4.1.202109301733 failed: An error occurred while transferring artifact canonical: osgi.bundle,org.osgi.service.event.source,1.4.1.202109301733 from repository https://download.eclipse.org/eclipse/updates/4.40/R-4.40-202606010713: Retry another mirror: download from https://mirror.ibcp.fr/pub/eclipse/eclipse/updates/4.40/R-4.40-202606010713/plugins/org.osgi.service.event.source_1.4.1.202109301733.jar failed</code></pre>

merks:
> @HannesWell this warning seems unrelated, can you have a look?
> 
> ```
> Some attempts to read artifact osgi.bundle,org.osgi.service.event.source,1.4.1.202109301733 failed: An error occurred while transferring artifact canonical: osgi.bundle,org.osgi.service.event.source,1.4.1.202109301733 from repository https://download.eclipse.org/eclipse/updates/4.40/R-4.40-202606010713: Retry another mirror: download from https://mirror.ibcp.fr/pub/eclipse/eclipse/updates/4.40/R-4.40-202606010713/plugins/org.osgi.service.event.source_1.4.1.202109301733.jar failed
> ```

Sometimes mirrors. Fail. Not a big deal because p2 will try another one. Ignore it.

vogella:
> > @HannesWell this warning seems unrelated, can you have a look?

> Sometimes mirrors. Fail. Not a big deal because p2 will try another one. Ignore it.

Thanks for the feedback. This is not urgent (and also because I run Eclipse primary from a patched build for PDE, as otherwise it is mostly unusable for me due to its classpath performance degregation in the last release) so I leave it open and hope that the warning will be removed in the future or the warning count will be reset.

PR #2387, opened 2026-06-29
Convert the layout spy command into an e4 handler

vogella:
Rewrites LayoutSpyHandler as a dependency-injected e4 handler and contributes its command, handler and M1+M2+M3+F9 key binding through a model fragment instead of the org.eclipse.ui command, handler and binding extension points. This removes the layout spy command's dependency on the 3.x compatibility workbench, so it can be used in plain e4 applications while the existing keystroke and behavior stay the same. The org.eclipse.core.runtime package is now consumed via Require-Bundle, since the equinox.common split-package export cannot satisfy a plain Import-Package once org.eclipse.ui is no longer pulled in transitively.

### DIFF
--- a/ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
@@ -9,9 +9,13 @@ Require-Bundle: org.eclipse.jface.databinding;bundle-version="1.9.0",
  org.eclipse.core.databinding.observable,
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.core.databinding.property;bundle-version="1.6.0",
- org.eclipse.ui;bundle-version="3.208.0"
+ org.eclipse.swt;bundle-version="3.135.0",
+ org.eclipse.jface;bundle-version="3.40.0"
 Import-Package: jakarta.annotation;version="[2.0.0,4.0.0)",
- org.eclipse.core.runtime;version="[3.7.0,4)",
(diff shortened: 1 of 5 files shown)


ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
```diff
@@ -9,9 +9,13 @@ Require-Bundle: org.eclipse.jface.databinding;bundle-version="1.9.0",
  org.eclipse.core.databinding.observable,
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.core.databinding.property;bundle-version="1.6.0",
- org.eclipse.ui;bundle-version="3.208.0"
+ org.eclipse.swt;bundle-version="3.135.0",
+ org.eclipse.jface;bundle-version="3.40.0",
+ org.eclipse.core.runtime;bundle-version="3.29.0"
 Import-Package: jakarta.annotation;version="[2.0.0,4.0.0)",
- org.eclipse.core.runtime;version="[3.7.0,4)",
```
merks, line 14: I think this should work rather than replacing it with a bundle requirement:
```
 org.eclipse.core.runtime;common=split;version="[3.8.0,4.0.0)",
```

merks (changes requested):

ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
vogella: Done

merks (approved):
I've not tested, but looks okay.

PR #2389, opened 2026-06-29
Document the javac profile for testing local bundle changes

vogella:
When running a test bundle together with a changed production bundle, the build must add the `javac` profile. Without it the baseline step replaces the freshly built classes with the released bundle of the same version, so the test silently runs the old code instead of your change. This updates the AGENTS.md instructions accordingly and simplifies the headless example by dropping an environment-specific xvfb invocation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,14 +39,16 @@ mvn verify -pl :org.eclipse.pde.ui.tests -Pbuild-individual-bundles
 mvn verify -pl :org.eclipse.pde.ui.tests -Pbuild-individual-bundles -Dtest=DependencyLoopFinderTest
 ```
 
-**Changed production code too?** Build the source bundle alongside its test bundle:
+**Changed production code too?** Build the source bundle alongside its test bundle, and add the `javac` profile:
 ```bash
-mvn verify -pl :org.eclipse.pde.core,:org.eclipse.pde.ui.tests -Pbuild-individual-bundles
+mvn clean verify -pl :org.eclipse.pde.core,:org.eclipse.pde.ui.tests -Pbuild-individual-bundles,javac
 ```
 
-**Headless (Wayland-safe):** pin the GTK backend to X11 so the UI test harness renders on Xvfb instead of the real compositor:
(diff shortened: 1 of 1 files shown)


laeubi:
> When running a test bundle together with a changed production bundle, the build must add the `javac` profile.

This does not sound right and is maybe more a side-effect. In what way does the `javac` profile influence baseline replacement? Baseline replace only happens when you forget to change the bundle version... Javac was a fallback to proof it *can* work and likely just produces different bytecode.

vogella:
> Baseline replace only happens when you forget to change the bundle version... 

If you test locally a bundle which already had its bundle version increased.

laeubi:
> > Baseline replace only happens when you forget to change the bundle version...
> 
> If you test locally a bundle which already had its bundle version increased.

If versions are properly increased then no baseline replace happens, `javac` profile goes completely in the wrong direction and makes absolutely no sense because it is not the compiler that is involved in this process and the profile servers a completely different purpose, you should actually never need it.

vogella:
ok

laeubi:
> Without it the baseline step replaces the freshly built classes **with the released bundle of the same version**

Your (AI generated) comment even mention it that the version is the same... Same for baseline replace means that only qualifier changed.

vogella:
Without javac profile a local change in the regular bundle is not seen, at least not in my testing. Maybe something wrong somewhere else, I could try it out (change a test and production bundle and run it via mvn, the production bundle is replaced by the baseline for me without the parameter).

laeubi:
> change a test and production bundle and run it via mvn, the production bundle is replaced by the baseline for me 

Again you need to change version of both bundles... Otherwise build verification won't work at all as we nowhere use the `javac` profile,

> without the parameter

It is not a parameter. It is a profile that does specific things to allow showing that `javac` can compile some code (e.g. in case we suspect a bug or inconsistency in `ejc`).

merks:
Maybe this can be used just to skip the baseline check?

image

I appears to be turned on the for the javac profile:

image

Maybe that's a point of confusion?

PR #2390, opened 2026-06-29
Keep repeated values in product program arguments

vogella:
Launching a product copied its program arguments into the run configuration through `concatArgs()`, which de-duplicated each user argument against the growing result list rather than against the initial arguments. A repeated value such as `-target test4 -helper test4` lost its second occurrence and became `-target test4 -helper`. The fix compares each user argument against the initial arguments only, so repeated values are kept while flags already provided by the defaults are still not duplicated. A regression test covers the case.


vogella:
@HannesWell I think this issue was introduced during your work in 79ed2805d6 "Simplify Product LaunchAction and associated classes" (Hannes Wellmann, 2022-08-13), please have a look if you have the time.

vogella:
Planning to merge beginning of next week unless I get feedback on this one, this is real problem for one of my clients.

merks (approved):
@vogella 

This looks reasonable to me and I think it's better to get this in place for m1 so that the testing runway is longer.

vogella:
Thanks @merks.

PR #2392, opened 2026-07-01
Contribute layout spy key binding via a model processor

vogella:
The layout spy contributed its command key binding through an e4 model fragment that imported the `org.eclipse.ui.contexts.dialogAndWindow` binding context.
Because the legacy binding table generated for that context shares the context's element id, the fragment import could resolve to the binding table instead of the context and crash `ModelAssembler.resolveImports` during model assembly, breaking `ResourceHandlerTest` (https://github.com/eclipse-platform/eclipse.platform.ui/issues/4148).
The binding is now contributed by a small e4 model processor that resolves the binding context by object and registers the binding on its own table, matching the existing PDE spy infrastructure.
This removes the ambiguous fragment import and keeps the command usable in plain e4 applications without reintroducing a dependency on org.eclipse.ui.

### DIFF
--- a/ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
+++ b/ui/org.eclipse.tools.layout.spy/META-INF/MANIFEST.MF
@@ -10,7 +10,9 @@ Require-Bundle: org.eclipse.jface.databinding;bundle-version="1.9.0",
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.core.databinding.property;bundle-version="1.6.0",
  org.eclipse.swt;bundle-version="3.135.0",
- org.eclipse.jface;bundle-version="3.40.0"
+ org.eclipse.jface;bundle-version="3.40.0",
+ org.eclipse.e4.ui.model.workbench;bundle-version="[2.4.0,3.0.0)",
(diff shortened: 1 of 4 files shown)


vogella:
@merks do you want to review?

merks:
I wonder, did you see this comment?

https://github.com/eclipse-platform/eclipse.platform.ui/issues/4148#issuecomment-4852072745

I have no expertise in how this type of processing works.   Did you test that the expected key bindings actually do work?  Is the user still able to change the binding?

vogella:
> I wonder, did you see this comment?

Yes I also replied. Short form: key  binding contributions via fragments seems to have a bug, I also created and pushed a test to reproduce this and currently trying to fix that.

> [eclipse-platform/eclipse.platform.ui#4148 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/issues/4148#issuecomment-4852072745)

> I have no expertise in how this type of processing works. Did you test that the expected key bindings actually do work? 

Yes

> Is the user still able to change the binding?

For the key binding changes by the user where is (to my knowledge) no difference if it provided by a processor.  Also looks good in in my runtime Eclipse.

image

merks:
That sounds good then.

PR #2396, opened 2026-07-03
Make CSS spy stylesheet access cascade-agnostic

vogella:
The CSS Spy read the e4 CSS engine's stylesheets through `CSSEngine.getDocumentCSS()` and cast the result to `ExtendedDocumentCSS`, both of which are removed by eclipse-platform/eclipse.platform.ui#4122 when it replaces the W3C computed-style cascade with internal types. This routes the two affected call sites, the matched-rule source column and the scratch pad, through a small reflective `CssEngineCompat` helper that works against both the current DocumentCSS cascade and the reworked internal cascade, the same version-agnostic approach as #2352. The scratch-pad parse is invoked reflectively too, so the reworked `parseStyleSheet` return-type change does not break the precompiled spy at runtime in the mixed-binary window. The one visible behavior change is in the scratch pad on the reworked engine: the scratch stylesheet is appended last instead of first, so on equal specificity it now wins against the theme, which is the more intuitive result. This removes the last cross-repo dependency blocking #4122.

### DIFF
--- a/ui/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css/CSSScratchPadPart.java
+++ b/ui/org.eclipse.pde.spy.css/src/org/eclipse/pde/spy/css/CSSScratchPadPart.java
@@ -14,12 +14,9 @@ import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
 import java.text.MessageFormat;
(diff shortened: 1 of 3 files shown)


vogella:
Rebase vía GH app still does a merge commit, will remove the merge commit later

PR #2404, opened 2026-07-25
Organize manifest: do not remove reexported dependencies

HeikoKlare:
## What it does
Removes the "Organize Manifest" logic that dropped a direct dependency on bundle B whenever some other used bundle A already reexported B, and restores the `org.eclipse.pde.core` dependency that this logic had incorrectly stripped from `org.eclipse.pde.runtime`.

## Why
`org.eclipse.pde.runtime` uses `org.eclipse.pde.core` API directly but only optionally depends on `org.eclipse.pde.ui`, which reexports `org.eclipse.pde.core`. "Organize Manifest" saw `pde.core` as reachable transitively through the reexport and removed the direct, optional dependency — which breaks the runtime plug-in whenever `pde.ui` is not installed.

In an first attempt, we tried to fix this narrowly by keeping the direct dependency only when the reexporting bundle is optional:
- #2403 
 
Discussion on that PR showed the assumption doesn't hold even for mandatory reexports: a reexport is a promise about API visibility, not a substitute for a bundle's own declared dependency, and there's no general way to prove it's safe to drop the direct one. This PR removes the cleanup heuristic altogether instead of trying to special-case it further

Supercedes and thus closes #2403

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
@@ -1,0 +1,186 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 Vector Informatik GmbH and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.pde.ui.tests.search.dependencies;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+
+import java.io.ByteArrayInputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.pde.core.plugin.IPluginImport;
+import org.eclipse.pde.core.plugin.IPluginModelBase;
+import org.eclipse.pde.core.plugin.PluginRegistry;
+import org.eclipse.pde.core.project.IBundleClasspathEntry;
+import org.eclipse.pde.core.project.IBundleProjectDescription;
+import org.eclipse.pde.core.project.IBundleProjectService;
+import org.eclipse.pde.core.project.IPackageExportDescription;
+import org.eclipse.pde.core.project.IRequiredBundleDescription;
+import org.eclipse.pde.internal.core.PDECore;
+import org.eclipse.pde.internal.ui.search.dependencies.GatherUnusedDependenciesOperation;
+import org.eclipse.pde.ui.tests.runtime.TestUtils;
+import org.eclipse.pde.ui.tests.util.ProjectUtils;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+import org.osgi.framework.VersionRange;
+
+public class GatherUnusedDependenciesOperationTest {
+
+	@ClassRule
+	public static final TestRule CLEAR_WORKSPACE = ProjectUtils.DELETE_ALL_WORKSPACE_PROJECTS_BEFORE_AND_AFTER;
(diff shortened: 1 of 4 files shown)


HeikoKlare:
@laeubi does this proposal conform to what you would have expected after the discussion we had in https://github.com/eclipse-pde/eclipse.pde/pull/2403?

laeubi (approved):
Yes it looks sound. Sadly we do not really have a testsuite for this part of the code so we can mostly test it by example only.

HeikoKlare:
I have created a test case (or rather asked an agent to create it) for the problematic unused dependency calculation case. Let me know if you think it makes sense to introduce such a test or if I should remote it again.

laeubi (commented):
Test looks reasonable in general, if the `createJavaSource` would use java text blocks it might become more readable.

Also having dedicated methods like `addExportedPackage(...)`, `addRequiredBundle(...)`, ... could improve readability and future enhancements even though then the description might be modified multiple times, as an alternative one could of course think about letting a Manifest being created directly instead of dedicated edits.

HeikoKlare:
The remarks are absolutely reasonable. I've changed the source code blocks to Java text blocks and introduced `addExportedPackage` and `addRequiredBundle` methods. I also wondered if there is a better alternative for the bundle project service than passing it around to the methods (such as storing it as class field, or retrieving it more locally - and redundantly - in the using methods) but kept it as is for now.

laeubi:
Test code is better readable now already.

> I also wondered if there is a better alternative for the bundle project service than passing it around to the methods (such as storing it as class field, or retrieving it more locally - and redundantly - in the using methods) but kept it as is for now.

As it is static and we do not need to care about performance here and no releasing of the service is needed, retrieving it at the place where it is needed seems reasonable to me, in the end I think what counts most is that we can easily read and enhance such tests, so having a little more code in a builder method seams acceptable.

In that mood, one might also make a method for the build+wait step+organize manifests and some for assertions like `assert(Not)FlaggedAsUnused(...)`

HeikoKlare:
Alright. I have simplied the helper method signatures and aquire the service locally where needed. I have also extracted the build+wait step into a helper method. For everything related to the actual test (executing the operation and asserting the actual outcome), I left the code in the test method for now. Once we add further test methods to the class, we may extract that out in case really needed. Agreed?

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
```diff
@@ -0,0 +1,181 @@
...
+				}
+				""".formatted(packageC));
+
+		buildProjects();
+
+		IPluginModelBase modelB = PluginRegistry.findModel(projectB);
+		assertNotNull("Plug-in model for bundle B not found", modelB);
+
+		GatherUnusedDependenciesOperation operation = new GatherUnusedDependenciesOperation(modelB);
+		operation.run(new NullProgressMonitor());
```
laeubi: I would move all this into `buildProjects();` and let it take the model and return the  `operation.getList()` so this becomes 
```suggestion
		List<..> unused = buildProjects(projectB);
```

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+		boolean bundleCFlaggedAsUnused = operation.getList().stream() //
+				.anyMatch(element -> element instanceof IPluginImport imp && bundleC.equals(imp.getId()));
+		assertFalse(
+				"Direct, used dependency to bundle C must not be flagged as unused just because bundle A reexports it",
+				bundleCFlaggedAsUnused);
```
laeubi: Given we have our unused list now from previous comment, can we then have something like

`assertNotFlaggedUnused("Direct, used dependency to bundle C must not be flagged as unused just because bundle A reexports it", bundleC)`

or alternatively

```suggestion
		assertFalse(
				"Direct, used dependency to bundle C must not be flagged as unused just because bundle A reexports it",
				isUnused(bundleC));
```

That would make is more obvious to the reader without needing to decode the stream operation and is probably also easier in case the assertion fails.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
HeikoKlare: I've extracted this code into a separate method, but did not combine it with `buildProjects()` as those are two semantically independent steps. We now have `buildProjects()` followed by `gatherUnusedDependencies(projectB)`.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
HeikoKlare: Yes, the assertion just reduces to `unusedPlugins.contains(bundleC)` now.

HeikoKlare:
From my side, everything should be addressed, so that this is ready.

HeikoKlare:
Would someone be so kind to merge this? I would appreciate to have this fixed for M3.

laeubi (approved):
Looks all sane to me.

HeikoKlare:
Thank you, Christoph!

PR #2410, opened 2026-07-30
Consider requirements shared with not-included sibling fragments

akurtakov:
When a host and its fragments declare equal requirements, only the first one encountered while attaching the fragments becomes a constraint of the host, all others are discarded. The single resulting wire therefore names only one declaring revision, chosen by fragment attachment order, which is sorted by symbolic name since PDE resolves in development mode. If an earlier sorting sibling fragment declares the same requirement as the launched one, the wire names that sibling and the requirement was dropped from the closure silently.

Also check whether a fragment of the host that is part of the closure declares a requirement matching the wire's capability. This covers Import-Package too and still ignores requirements of fragments that are genuinely not included.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -37,6 +37,7 @@ import org.osgi.framework.Bundle;
 import org.osgi.framework.Constants;
 import org.osgi.framework.Version;
 import org.osgi.framework.namespace.HostNamespace;
+import org.osgi.framework.wiring.BundleCapability;
 import org.osgi.framework.wiring.BundleRequirement;
 import org.osgi.framework.wiring.BundleRevision;
 import org.osgi.framework.wiring.BundleWire;
@@ -265,7 +266,8 @@ public class DependencyManager {
 					: namespaces.stream().map(wiring::getRequiredWires).flatMap(List::stream)::iterator;
 			for (BundleWire wire : requiredWires) {
 				BundleRevision declaringBundle = wire.getRequirement().getRevision();
-				if (declaringBundle != bundle && !closure.contains(declaringBundle)) {
+				if (declaringBundle != bundle && !closure.contains(declaringBundle)
+						&& !isAlsoDeclaredByIncludedFragment(bundle, wire, closure)) {
 					// Requirement is declared by an attached fragment, which is
 					// not included into the closure.
 					continue;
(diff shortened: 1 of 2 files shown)


akurtakov:
The attached test fails without the DependencyManager change and shows the problem this PR is fixing.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
```diff
@@ -281,6 +283,38 @@ public static Set<BundleDescription> findRequirementsClosure(Collection<BundleDe
...
+	 * </p>
+	 */
+	private static boolean isAlsoDeclaredByIncludedFragment(BundleDescription host, BundleWire wire,
+			Set<BundleDescription> closure) {
+		BundleDescription[] fragments = host.getFragments();
+		if (fragments.length == 0) {
+			return false;
+		}
+		BundleCapability capability = wire.getCapability();
+		String namespace = wire.getRequirement().getNamespace();
```
merks: One place I looked it seemed it might be possible for requirement to be null..

https://github.com/eclipse-equinox/equinox/blob/73815d3f63a3fd5e2614138ef5789c371f9ed85b/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/resolver/VersionConstraintImpl.java#L123-L129

But I see lots of callers that assume this method never returns null so I think it's fine.

merks (approved):
I can't say I deeply understand the semantics here, but it looks reasonable and is well-documented.

ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
akurtakov: Fixed with the latest commit.

laeubi:
> I can't say I deeply understand the semantics here, but it looks reasonable and is well-documented.

This is actually a bad basis for merging such central change just right away near M3 and with little time for other to look at :-\

For me I miss a consistent an actual use-case where this becomes important e.g if both declare *the same requirement* why should it matter wich one is chosen afterwards?

For me it looks more that PDE should consider all requirements of selected fragment, also the explanation in the test mixes references to the spec with implementation details that are **not** part of the spec what is confusing - if one would suspect a but there it would not be PDE part to be solved (but resolver). Also nowhere in the spec there is anywhere talking about a "requirements closure" or what must be included in a launch so this is completely misleading here.

akurtakov:
The problem this fixes is when a single test from one of the fragments is executed and it can't resolve the extra dependency as it's been dropped due to being "attached" by the other fragment which is not enabled in the launch configuration. 
Also there are 2 more weeks of M3 (out of 3) so technically we are in the "beginning" of M3. 
Anyway, the effect is what you describe - no requirements of the selected fragment are dropped. 
Feel free to show how to do it better! Considering the amount of open PRs no one is looking at, having a reviewed PR is better than waiting weeks to get really into "end of" M3 and let the bug slip into next release.

laeubi:
> Feel free to show how to do it better! Considering the amount of open PRs no one is looking at, having a reviewed PR is better than waiting weeks to get really into "end of" M3 and let the bug slip into next release.

So what has cause this bug? For me it looks like it is the case for a very long time so nothing that was introduced recently. Also there is a difference between *weeks* and a few hours and I know at least @HannesWell is also generally reviewing PRs in this area.

akurtakov:
I have never said it's new bug, have I? I just finally got to look into the problem. Would you please explain what is this PR breaking as you seem to have big concerns?

HannesWell:
Generally I'm very reluctant to extend the `DependencyManager` to include more artifacts, mainly because it's impossible to exclude any bundle from the being added automatically once it was added. The only workaround is to disable the automatic inclusion of dependencies and define everything manually. Therefore one 'bad' inclusion can ruine the entire runtime and prevent one from using this feature.

The returned dependency closure should only be the bare minimum to make the resulting set resolve and should not include optional dependencies, which fragments are usually also count for.
In the past I've expressed that opinion/goal/concern on each occasion when more bundles should be added to the closure.
If e.g. a fragment is necessary to make an app function, there should be corresponding capabilities and requirements from the host to that fragment (e.g. the SWT case). Just because modeling such dependencies is currently not possible with the current build tooling a special exception was made fore that (SWT) case.

In general, if an optional dependency (in the OSGi/dependency sense) is required for an application, it should be explicitly included into a launch. Either through e.g. a feature or by listing that fragment (or optional dependency) in the launch config.

That being said about the general case, I have to look into this specific case in detail to say if we should have this exception.

> The problem this fixes is when a single test from one of the fragments is executed and it can't resolve the extra dependency as it's been dropped due to being "attached" by the other fragment which is not enabled in the launch configuration.

Does anybody know, if the resolver can be configured to not drop 'duplicated' dependencies if they originate from multiple different (OSGi) resources?

And do you have a publicly available real world case where this happens?

Issue #1617, opened 2025-02-14
Unable to edit target platform source file if text starts with `<`


laeubi:
Typing some text that starts with `<` or even pasting some (incomplete) text-fragments make the source editor going completely wild and at best deleting the inserted text, at worst it delete some other content as well, see video.

https://github.com/user-attachments/assets/350ccd92-3b49-42c5-9aeb-3a22b77ddd9e

@mickaelistria I'm not sure if it is a PDE issue, a Generic Editor issue or a lemminx issue (or feature?) or a combination of both but it makes the target editor hardly usable at the moment.

laeubi:
If I start eclipse with the console I see:

```
Exception in thread "org.eclipse.pde.internal.genericeditor.target.extension.reconciler.folding.TargetPlatformFoldingReconciler" java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at org.eclipse.pde.internal.genericeditor.target.extension.reconciler.folding.FoldingReconcilerStrategy.getNewPositionsOfAnnotations(FoldingReconcilerStrategy.java:181)
	at org.eclipse.pde.internal.genericeditor.target.extension.reconciler.folding.FoldingReconcilerStrategy.initialReconcile(FoldingReconcilerStrategy.java:71)
	at org.eclipse.pde.internal.genericeditor.target.extension.reconciler.folding.FoldingReconcilerStrategy.reconcile(FoldingReconcilerStrategy.java:53)
	at org.eclipse.jface.text.reconciler.Reconciler.process(Reconciler.java:149)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.run(AbstractReconciler.java:207)


```

mhuber-dev:
I encountered the same error (insert new text or paste new content leads to refresh the source code editor and delete the added content).
With other XML-files the behaviour do not occure. I also installed Wild Web Developer, but the error only happens in the source editor of the target platform.

HannesWell:
[...]

PR #2414, opened 2026-08-07
Prevent Target Editor reverts for marker-only resource changes

vlakn:
### Summary

The issue first surfaced with the m2e PDE integration, which triggers marker updates while editing a `.target` definition. These marker updates cause `TargetEditor` to revert the editor, discarding unsaved source-page input.

The behavior is not specific to m2e. Creating a bookmark on the open `.target` file also produces a marker-only resource delta and triggers the same revert.

For example, enter any unsaved XML change in the source page and then create a bookmark on the `.target` file. The Target Editor discards the working copy although the file content remains unchanged.

This change prevents the Target Editor from reverting unsaved source-page edits when the workspace reports a marker-only change for the opened `.target` file.

### Background

Commit 68642a19c8e4173b4ca240e635c80d0cd2cb3b2a extended `TargetEditor.doRevert()` to also revert dirty text-editor pages.

`TargetEditor.InputHandler.resourceChanged()` currently invokes `doRevert()` for every `IResourceDelta.CHANGED` delta on the target file. This includes marker-only deltas, although they do not alter the target definition content.

With the m2e PDE integration installed, workspace refreshes and marker updates trigger such deltas. While editing the XML source page, typing a character such as `<` then discards the unsaved working copy.

### Before
Target_Editor_Before

### After
Target_Editor_After

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
@@ -586,15 +586,15 @@ public class TargetEditor extends FormEditor {
 				if (delta != null) {
 					if (delta.getKind() == IResourceDelta.REMOVED) {
 						TargetEditor.this.close(false);
-					} else if (delta.getKind() == IResourceDelta.CHANGED || delta.getKind() == IResourceDelta.REPLACED) {
-						if (!fSaving) {
-							Display display = getSite().getShell().getDisplay();
-							display.asyncExec(() -> {
-								if (getActivePage() != -1) {
-									TargetEditor.this.doRevert();
-								}
-							});
-						}
+					} else if (delta.getKind() == IResourceDelta.CHANGED
+						&& (delta.getFlags() & (IResourceDelta.CONTENT | IResourceDelta.REPLACED)) != 0
+						&& !fSaving) {
+						Display display = getSite().getShell().getDisplay();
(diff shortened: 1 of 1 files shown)


merks (approved):
That looks like a nice simple change to fix behavior that drives me crazy. 😕

vogella:
Ah, that is that? Drove me also crazy, I always assumed that it is the XML language server.

merks:
Ignore the license failure, it's been flaky again:

> Received an error response from the Eclipse Foundation

vogella (approved):

vogella:
Big thanks @vlakn for this fix

HeikoKlare (approved):
I did some manuel testing of the change (in particular that marker/bookmark changes do not revert changes of the editor content) and found everything to work fine so far.

Awesome to have this fixed. Thank you, @vlakn!

merks:
FYI, I manually test this in a self-hosted launch, first reproducing the problem (with the language server stuff in the target platform), then applying the patch to verify that the reverts don't happen, and then also verifying that if I edit the file via a different editor and save, the revert does happen. 

So this looks great!  🏆 🏅

laeubi:
@vlakn well done on identify the root cause of this, I would not have though that marker make the difference here!

HannesWell:
Can only second what have been said. Thanks for this fix.
I didn't expect the cause of https://github.com/eclipse-pde/eclipse.pde/issues/1617 being in PDE since it didn't happen in the usual workflows in a plain Eclipse SDK and only with m2e (and it's dependencies) installed.

It was probably quite some work to identify the issue, but it paid off. Well done!

PR #2415, opened 2026-08-07
Fix dead REPLACED check in the sample editor resource listener

vogella:
`SampleEditor.InputFileListener` closes the editor when its `sample.properties` input goes away, but the replacement case never worked: `IResourceDelta.REPLACED` is a delta flag, not a delta kind, so `getKind() == IResourceDelta.REPLACED` could never be true. `getKind()` only ever returns `NO_CHANGE`, `ADDED`, `REMOVED`, `CHANGED`, `ADDED_PHANTOM` or `REMOVED_PHANTOM`.

The fix tests `REPLACED` against `getFlags()` on a `CHANGED` delta, which is how `ResourceComparator` actually reports a resource that was deleted and recreated in the same operation. The removal case was already correct and is unchanged.

This is the same class of bug as the one fixed for `TargetEditor` in #2414, found while reviewing it.


vogella:
Nothing in this repo contributes to the samples extension point, there is no sample.properties anywhere in the tree, SampleWizard isn't even registered as a newWizard (it's only ever constructed programmatically by ShowSampleAction), and the update site it bootstraps from is a long-dead http:// URL. In practice the editor only appears if some external contributor still ships that extension point.#

Still worth the fix, to guide users of this sample extension point.

PR #2418, opened 2026-08-11
Convert Runtime Packages section from table to tree

### LINKED ISSUE #1858: [manifest-editor] Show uses of exported packages in a tree
Currently we have this view for exported packages:

Image

as one can see there is a "Calculate Uses" button that would for example result in this:

```
Export-Package: org.eclipse.core.internal.preferences;
 org.eclipse.core.runtime.preferences;version="3.6.0";uses:="org.eclipse.core.runtime,org.osgi.service.prefs"
```

but the packages in the `uses` directive are currently invisible there, so we have two problems:

1. If I press the button it seems "nothing" happens
2. If I want to see the result I have to examine the source of the manifest what is often not very convenient

Instead one should enhance the current view by:

- Making the Table a Tree
- Show as childs of the package any used packages

That way it would be easy to see what packages has uses constraints and what packages are used, in addition to that the Add/Remove buttons might be used to manually add/remove used packages.

elsazac:
This PR changes the Runtime packages section from table to a tree and adds the packages declared in the uses: directive as children of their corresponding parent package. previously, this section displayed only the exported packages in a table. with this PR, each exported package can now be expanded to show the individual packages referenced by its uses: directive. it also supports adding and deleting packages.  individual uses: entries can be deleted directly from the tree and it will be automatically reflected in the manifest. 

Changes :

1. converted table to tree
2. uses directives will appear as children to the parent packages
3. one can remove any uses directive using the `Remove` option

Fixes : https://github.com/eclipse-pde/eclipse.pde/issues/1858

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
@@ -19,9 +19,11 @@
 package org.eclipse.pde.internal.ui.editor.plugin;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.eclipse.core.resources.IProject;
@@ -30,17 +32,20 @@ import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.ui.ISharedImages;
 import org.eclipse.jdt.ui.JavaElementLabelProvider;
+import org.eclipse.jdt.ui.JavaUI;
 import org.eclipse.jdt.ui.actions.FindReferencesAction;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.Separator;
 import org.eclipse.jface.viewers.ILabelProvider;
 import org.eclipse.jface.viewers.ISelection;
-import org.eclipse.jface.viewers.IStructuredContentProvider;
 import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.viewers.ITreeContentProvider;
+import org.eclipse.jface.viewers.LabelProvider;
 import org.eclipse.jface.viewers.StructuredSelection;
-import org.eclipse.jface.viewers.TableViewer;
+import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.jface.viewers.ViewerComparator;
 import org.eclipse.jface.window.Window;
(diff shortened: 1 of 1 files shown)


elsazac:
cc/ @nburnwal09

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
```diff
@@ -502,6 +641,13 @@ public void modelChanged(IModelChangedEvent event) {
 	@Override
 	public void refresh() {
 		fPackageViewer.refresh();
+		if (fHeader != null) {
+			for (ExportPackageObject pkg : fHeader.getPackages()) {
+				if (getUsesPackageNames(pkg).length > 0) {
+					fPackageViewer.expandToLevel(pkg, 1);
```
nburnwal09: The tree auto-expands(as default) all uses: nodes on every refresh, overriding any collapse the user made. It would be better to start collapsed(after refresh) and let users expand on demand.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
```diff
@@ -100,9 +145,67 @@ public Object[] getElements(Object parent) {
...
+			return fDelegate.getText(element);
+		}
+
+		@Override
+		public Image getImage(Object element) {
+			if (element instanceof usesNode) {
+				return JavaUI.getSharedImages().getImage(ISharedImages.IMG_OBJS_PACKAGE);
+			}
+			return fDelegate.getImage(element);
+		}
```
elsazac: If it is a child node, return the eclipse Java package icon from JDT's shared image registry.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
```diff
@@ -386,10 +501,23 @@ private void handleOpenProperties() {
...
 
+	private void removeUsesEntry(usesNode node) {
+		ExportPackageObject pkg = node.parent;
+		String[] current = getUsesPackageNames(pkg);
+		String remaining = Arrays.stream(current)
+				.filter(name -> !name.equals(node.packageName))
+				.collect(java.util.stream.Collectors.joining(",")); //$NON-NLS-1$
+		pkg.setUsesDirective(remaining.isEmpty() ? null : remaining);
+	}
+
```
elsazac: get the parent node and read the full list of packages listed in uses and filter out the one that needs to be removed and then combine them to the manifest format using commas and write back.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
```diff
@@ -64,34 +69,74 @@
...
+		if (raw == null || raw.length == 0) {
+			return new String[0];
+		}
+		return Arrays.stream(raw)
+				.flatMap(s -> Arrays.stream(s.split(","))) //$NON-NLS-1$
+				.map(String::trim)
+				.filter(s -> !s.isEmpty())
+				.toArray(String[]::new);
+	}
+
```
elsazac: turns the string to a stream and then for each element, split it on commas, then flatten all those pieces into one single stream. then removes whitespaces, removes blank strings, and collect back into string.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/plugin/ExportPackageSection.java
elsazac: Done

nburnwal09:
@elsazac Could you please attach the updated UI screenshot here?

elsazac:
Converted the table to tree and the `Calculate Uses` will add directives as child nodes :

image

nburnwal09 (approved):
It looks good to me.

PR #2419, opened 2026-08-11
Externalize string in PDE preferences in ui

elsazac:
This PR makes changes for externalizing one of the entries in the UI preference page. the non externalized string generated warning and hence the string has been externalized.


elsazac:
cc/ @nburnwal09

nburnwal09:
This is a small UI update. lgtm.

merks:
FYI, no need to change this PR or the commit, but for the future, the message could be slightly more description about which string in particular is being externalized.  Picky I know...  But imagine seeing a list of 5 commits just like this one...

PR #2422, opened 2026-08-16
Report a missing layout in the layout spy output

vogella:
The layout section of the layout spy stayed empty for composites that have no layout, both in the details panel and in the copied widget tree. That looks like a broken spy rather than like a statement about the widget, so it now prints `getLayout() == null`, mirroring the `getLayoutData() == null` line right above it. For leaf controls the section is dropped completely instead of repeating "not a Composite" for every label and button in a tree dump.

While in there: the truncation check for grid layouts intersected the parent bounds with themselves instead of with the child bounds, so the intersection always matched and the warnings about missing grab flags could never be shown.


vogella:
https://github.com/eclipse-pde/eclipse.pde/pull/2425

PR #2425, opened 2026-08-17
Force qualifier update to enable PR validation

vogella:


ui/org.eclipse.pde.bnd.ui/forceQualifierUpdate.txt
```diff
@@ -4,3 +4,4 @@ https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/19
 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2995
 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3814
 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3975
+https://github.com/eclipse-pde/eclipse.pde/pull/2422/ fails with baseline error for this plugin
```
laeubi, line 7: You should make sure this is not happening because of missing rebase or compiler changes.

Blindly bumping version (especially without explanation) is not what this check is meant for!

laeubi (changes requested):

laeubi:
For completeness this is the reported difference:

```
[WARNING] MavenProject: org.eclipse.pde:org.eclipse.pde.bnd.ui:1.2.700-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.pde_PR-2422/ui/org.eclipse.pde.bnd.ui/.polyglot.META-INF: baseline and build artifacts have same version but different contents
no-classifier: different
org/eclipse/pde/bnd/ui/plugins/EclipseWorkspaceRepository.class: different
org/eclipse/pde/bnd/ui/templating/ReposTemplateLoader.class: different
```

this looks like a compiler change and should be properly investigated.

PR #2428, opened 2026-08-24
Fix NPE caused by stale BundleWiring in dependency resolution

### LINKED ISSUE #2427: NPE in DependencyManager.findRequirementsClosure
During testing something I suddenly got the following NPE, so far I cannot reproduce and I also did not notice anything breaking in the IDE.

````

java.lang.NullPointerException
      at java.base/java.util.stream.ReferencePipeline$7$1FlatMap.accept(ReferencePipeline.java:288)
      at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
      at java.base/java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1950)
      at java.base/java.util.stream.StreamSpliterators$WrappingSpliterator.lambda$initPartialTraversalState$0(StreamSpliterators.java:297)
      at java.base/java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.fillBuffer(StreamSpliterators.java:206)
      at java.base/java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.doAdvance(StreamSpliterators.java:169)
      at java.base/java.util.stream.StreamSpliterators$WrappingSpliterator.tryAdvance(StreamSpliterators.java:303)
      at java.base/java.util.Spliterators$1Adapter.hasNext(Spliterators.java:669)
      at org.eclipse.pde.internal.core.DependencyManager.findRequirementsClosure(DependencyManager.java:267)
      at org.eclipse.pde.internal.core.ClasspathComputer.collectBuildRelevantDependencies(ClasspathComputer.java:180)
      at org.eclipse.pde.internal.core.DynamicPluginProjectReferences.getDependentProjects(DynamicPluginProjectReferences.java:47)
      at org.eclipse.core.internal.resources.ProjectDescription.computeDynamicReferencesForProject(ProjectDescription.java:1037)
      at org.eclipse.core.internal.resources.ProjectDescription.getAllBuildConfigReferences(ProjectDescription.java:297)
      at org.eclipse.core.internal.resources.ProjectDescription.getAllReferences(ProjectDescription.java:252)
      at org.eclipse.core.internal.resources.Project.getReferencedProjects(Project.java:522)
      at org.eclipse.ui.internal.ide.misc.ProjectReferenceGraph.buildDisjointSet(ProjectReferenceGraph.java:219)
      at org.eclipse.ui.internal.ide.misc.ProjectReferenceGraph.buildComponents(ProjectReferenceGraph.java:192)
      at org.eclipse.ui.internal.ide.misc.ProjectReferenceGraph.rebuild(ProjectReferenceGraph.java:147)
      at org.eclipse.ui.internal.ide.misc.ProjectReferenceGraph$1.run(ProjectReferenceGraph.java:72)

````

### PR DESCRIPTION
**Pull request by nburnwal09**
Guard all affected getRequiredWires() call sites in DependencyManager and ClasspathUtilCore against a null return. In DependencyManager, a null result causes the current bundle to be skipped, consistent with the existing isInUse() guard above. In ClasspathUtilCore, a null result returns an empty stream so no classpath entries are produced for the stale wiring.

The per-namespace stream path in DependencyManager also gains a null filter before flatMap to handle the case where a live wiring has no wires for a specific namespace.

The original crash site is covered by a regression test. The remaining sites
require precise timing to reproduce in a test and are fixed defensively.

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/2427

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Queue;
 import java.util.Set;
 
@@ -251,7 +252,11 @@ public class DependencyManager {
 				// Requirements of a fragment are hosted at the host, which
 				// therefore requires the corresponding wires: OSGi Core spec,
 				// chapter 6.4.1 - Hosted Requirements and Capabilities
-				for (BundleWire hostWire : wiring.getRequiredWires(HostNamespace.HOST_NAMESPACE)) {
+				List<BundleWire> hostWires = wiring.getRequiredWires(HostNamespace.HOST_NAMESPACE);
+				if (hostWires == null) {
+					continue; // wiring became stale
+				}
+				for (BundleWire hostWire : hostWires) {
 					// Temporarily remove this fragment's host from the closure
 					// to ensure it's added again below. In the subsequent
 					// processing this fragment's requirements will then also be
@@ -262,8 +267,17 @@ public class DependencyManager {
 				}
 			}
 
-			Iterable<BundleWire> requiredWires = namespaces.isEmpty() ? wiring.getRequiredWires(null)
-					: namespaces.stream().map(wiring::getRequiredWires).flatMap(List::stream)::iterator;
+			Iterable<BundleWire> requiredWires;
+			if (namespaces.isEmpty()) {
+				List<BundleWire> allWires = wiring.getRequiredWires(null);
+				if (allWires == null) {
+					continue; // wiring became stale
+				}
+				requiredWires = allWires;
+			} else {
+				requiredWires = namespaces.stream().map(wiring::getRequiredWires).filter(Objects::nonNull)
+						.flatMap(List::stream)::iterator;
+			}
 			for (BundleWire wire : requiredWires) {
 				BundleRevision declaringBundle = wire.getRequirement().getRevision();
 				if (declaringBundle != bundle && !closure.contains(declaringBundle)
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathUtilCore.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/ClasspathUtilCore.java
@@ -128,6 +128,9 @@ public class ClasspathUtilCore {
 		}
 
 		List<BundleWire> wires = wiring.getRequiredWires(PackageNamespace.PACKAGE_NAMESPACE);
+		if (wires == null) {
+			return Stream.empty(); // wiring became stale
+		}
 		return wires.stream().filter(wire -> wire.getProviderWiring() != null).map(wire -> {
 			return wire.getProvider();
 		}).distinct().flatMap(provider -> {
@@ -169,6 +172,9 @@ public class ClasspathUtilCore {
 			return Stream.empty();
 		}
 		List<BundleWire> wires = wiring.getRequiredWires(PackageNamespace.PACKAGE_NAMESPACE);
+		if (wires == null) {
+			return Stream.empty(); // wiring became stale
+		}
 		return wires.stream().map(wire -> wire.getProviderWiring().getBundle()).distinct()
 				.filter(b -> b.getBundleId() != 0)
 				.flatMap(b -> classpathEntriesForRuntimeBundle(b, extra).stream());

### CODE REVIEW

**Comment by nburnwal09** [comment]
@vogella 
Kindly review this PR. 
Addressing https://github.com/eclipse-pde/eclipse.pde/issues/2427

**Comment by vogella** [comment]
Thanks! The fix is right, but looks like we have more places like this.

getRequiredWires(String) returns an empty list for a namespace without requirements. It returns null only when the wiring is not in use, and getRequiredWires(null) behaves the same way.

So the wiring went out of use between the isInUse() check at line 230 and the wire lookup at line 266. PDE re-resolves its state on project close/open and target platform changes, which invalidates live wirings, while ProjectReferenceGraph walks the closure in a background job. Nothing serialises the two.

Because it is a stale wiring and not an empty namespace, the same problem exists at DependencyManager lines 254 and 265 and at ClasspathUtilCore lines 130 and 171. Could you update the commit message and cover those too? I have a regression test that reproduces the place you fixed, but the other call sites are hard to reach for a test.

**Comment by vogella** [comment]
See https://github.com/eclipse-pde/eclipse.pde/pull/2429

**Comment by nburnwal09** [comment]
> Thanks! The fix is right, but looks like we have more places like this.
> 
> getRequiredWires(String) returns an empty list for a namespace without requirements. It returns null only when the wiring is not in use, and getRequiredWires(null) behaves the same way.
> 
> So the wiring went out of use between the isInUse() check at line 230 and the wire lookup at line 266. PDE re-resolves its state on project close/open and target platform changes, which invalidates live wirings, while ProjectReferenceGraph walks the closure in a background job. Nothing serialises the two.
> 
> Because it is a stale wiring and not an empty namespace, the same problem exists at DependencyManager lines 254 and 265 and at ClasspathUtilCore lines 130 and 171. Could you update the commit message and cover those too? I have a regression test that reproduces the place you fixed, but the other call sites are hard to reach for a test.

Sure, will check and cover them. Thank you

**Comment by nburnwal09** [comment]
@vogella 
I have updated the PR. Kindly review it.

File: ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/DependencyManager.java [hunk 1]
```diff
@@ -262,8 +267,13 @@ public static Set<BundleDescription> findRequirementsClosure(Collection<BundleDe
...
 
-			Iterable<BundleWire> requiredWires = namespaces.isEmpty() ? wiring.getRequiredWires(null)
-					: namespaces.stream().map(wiring::getRequiredWires).flatMap(List::stream)::iterator;
+			List<BundleWire> allWires = wiring.getRequiredWires(null);
+			if (allWires == null) {
+				continue; // wiring became stale
+			}
+			Iterable<BundleWire> requiredWires = namespaces.isEmpty() ? allWires
+					: namespaces.stream().map(wiring::getRequiredWires).filter(Objects::nonNull)
+							.flatMap(List::stream)::iterator;
```
vogella: `getRequiredWires(null)` builds every wire of every visited bundle and is then discarded unless `namespaces` is empty. Only compute it in that branch:

```suggestion
			Iterable<BundleWire> requiredWires;
			if (namespaces.isEmpty()) {
				List<BundleWire> allWires = wiring.getRequiredWires(null);
				if (allWires == null) {
					continue; // wiring became stale
				}
				requiredWires = allWires;
			} else {
				requiredWires = namespaces.stream().map(wiring::getRequiredWires)
						.filter(Objects::nonNull).flatMap(List::stream)::iterator;
			}
```

nburnwal09: Done

vogella (approved):
LGTM, thanks

HannesWell:
Thanks Neha for the fix. I'll have a look at it.

But the more interesting question to me is why have bundles went out of use. I've never seen the reported issue before and didn't even encounter it now in my daily work.
If it's caused by a recent change, maybe that change is incomplete and misses synchronization.
By just guarding against null, we maybe miss the now in use wiring.

Generally the TP update should be an atomic operation and afterwards none of them should go out of use. Actually wirings of a TP should always be in use it discarded before used by anything else.

vogella:
I have seen it during test execution in client code, maybe humans are not fast enough to trigger that. See my unit test for an example how to trigger it https://github.com/eclipse-pde/eclipse.pde/pull/2429/

HannesWell (changes requested):
I've looked into this in more detail and summarized my findings in https://github.com/eclipse-pde/eclipse.pde/issues/2427#issuecomment-5402299485.
To make it short, adding null checks seems not suitable to me as they hide the actual underlying problem: An inconsistent state and parts of the code that operate on outdated data. Sorry the negative statement, but having a silent inconsistency is IMO worse. At the same time I don't have a good solution for the original problem at hand.

@nburnwal09 did you experience the issue by yourself too? Or did you just want to help Lars with this fix?

nburnwal09:
> @nburnwal09 did you experience the issue by yourself too? Or did you just want to help Lars with this fix?

@HannesWell  
I have not yet faced this issue. I have raised PR to provide the fix for Lars problem.

laeubi:
I agree with @HannesWell that we should be cautious here. Also a test should best be included (e.g. by cherry-pick it from @vogella's branch) what would make it easier to see if it actually fixes the problem and having easier review.

laeubi (commented):
From an architectural point of view it seems better to throw an exception if we found something inconsistent here rather than hide the problem so the caller can possibly decide how to proceed further (e.g. retry, log or ignore).

vogella:
After clicking like a maniac to reproduce this in the IDE I was not able to do this outsite tests. 

Lets close this for now.  We can reopen this is we find a way to reproduce. Feel free to re-open if you disagree.

Thanks @nburnwal09 for the effort.

PR #2429, opened 2026-08-24
Reproduce NPE on stale bundle wiring in findRequirementsClosure

vogella:
Deterministic regression test for #2427.

`findRequirementsClosure` checks `isInUse()` and then reads the wires from a captured local. A state re-resolve in between invalidates the wiring, and `getRequiredWires()` returns `null` once a wiring is out of use. Project close/open and target platform reloads re-resolve the state while `ProjectReferenceGraph` walks the closure in a background job, and nothing serialises the two.

A real second thread would be flaky, so the test injects the re-resolve through the `namespaces` set: the walk builds a fresh stream per bundle and consumes it lazily, so `state.resolve(false)` from inside `next()` lands in the window. Only the scheduling is synthetic. The stack matches #2427 frame for frame.

Fails on master by design, no fix included. #2428 covers the site this test hits, but `DependencyManager` lines 254 and 265 and `ClasspathUtilCore` lines 130 and 171 have the same exposure, hence draft.

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/DependencyManagerTest.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/core/tests/internal/DependencyManagerTest.java
@@ -34,7 +34,9 @@ import static org.osgi.framework.Constants.REQUIRE_CAPABILITY;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.AbstractSet;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


laeubi (commented):
I would like to see call-sites here that show when this happens in production code. This would also satisfy @HannesWell concerns so we can probably check and fix these call sites (or accept it can happen) more than a fully synthetic test-case.

vogella:
After clicking like a maniac to reproduce this in the IDE I was not able to do this outsite tests. Closing.

Issue #2394, opened 2026-07-02
API Tools and sealed classes in the Skia plugin


DenisUngemach:
Hello, I wanted to extend SWT with a Skia Plugin, which requires some modifications to the SWT code as well.

PR: https://github.com/eclipse-platform/eclipse.platform.swt/pull/3231

One of these modifications is that the GC class receives the sealed modifier, with GCExtension extending the GC.

To prevent API tools errors, I had to set some filter values. However, these are declared as unused in the Skia Plugin PR: https://github.com/eclipse-platform/eclipse.platform.swt/pull/3231/checks?check_run_id=83400991229

However, when I remove the specified filters from the code, other errors appear. See this PR: https://github.com/swt-initiative31/skija-canvas/pull/100

But if I remove the sealed modifier and make GCExtension a simple extension of GC (where GC is neither sealed nor final), the build works without the specified filters:

https://github.com/swt-initiative31/skija-canvas/pull/101

This makes me wonder: I would expect that the API tools do not distinguish between extending sealed classes and normal classes. For attribute visibility, there should be no difference.

vogella:
Maybe @HannesWell or @akurtakov can help here? I have no insides how API tooling works.

merks:
FYI, @HannesWell  is on vacation so we should have no expectations from him for next few weeks...

DenisUngemach:
Hello, any updates here?

DenisUngemach:
Probably the reason for the bug is the insufficient java version analysis in APIFileGenerator. 
With this the api_descriptor will be built, and if the java version is too high, it falls back to version 1.3:

The code of resolveCompliance was implemented in 2019, there the most used java version 8, allthough version 12 already existed.

If it falls back here to 1.3 the analysis of sealed classes might just not happen properly. 
Ich will see whether i can provide a proper pull request also with tests, which fixes this issue.


In APIFileGenerator in org.eclipse.pde.api.tools.internal:

	/**
	 * Resolves the compiler compliance based on the BREE entry in the
	 * MANIFEST.MF file
	 *
	 * @return The derived {@link JavaCore#COMPILER_COMPLIANCE} from the BREE in
	 *         the manifest map, or {@link JavaCore#VERSION_1_3} if there is no
	 *         BREE entry in the map or if the BREE entry does not directly map
	 *         to one of {"1.3", "1.4", "1.5", "1.6", "1.7","1.8"}.
	 */
[...]

PR #2430, opened 2026-08-24
APIFileGenerator: update the java version parsing

DenisUngemach:
- This should provide a future prove manifest java version analysis by using the bree and osgi.ee.
- It should solve the issue: https://github.com/eclipse-pde/eclipse.pde/issues/2394 where before this fix, the api description creation is incomplete, because the java version falls back to version 1.3, where sealed classes do not exist.

Some more points:
-I use the JavaCore.getAllJavaSourceVersionsSupportedByCompiler() as reference java versions which are valid.  Currently the oldest version is 1.8. I think staying up to date with JavaCore is the best way to stay in sync and deprecate older versions. 
- If there are multiple java versions in the manifest, i take the older one. 
- Maybe the newer osgi.ee filter option could make it possible to have more complex java version restrictions. I do ignore other restrictions and take the lowest java version i can find, that is valid by JavaCore.supportedVersions.

What do you think?


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
```diff
@@ -0,0 +1,224 @@
...
+	 */
+	private static String fromSingleBree(String eename) {
+		if (eename.startsWith(JAVASE_PREFIX)) {
+			String version = eename.substring(JAVASE_PREFIX.length());
+			if (JavaCore.getAllJavaSourceVersionsSupportedByCompiler().contains(version)) {
+				return version;
+			}
+		}
+		return null;
+	}
```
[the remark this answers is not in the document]
DenisUngemach: Correct, but these java versions only exist for 1.8. So i check these explicitly.

apitools/org.eclipse.pde.api.tools.tests/src/org/eclipse/pde/api/tools/util/tests/ExecutionEnvironmentResolverTest.java
```diff
@@ -0,0 +1,350 @@
...
+	public void testRequireCapabilityTwoEntriesReturnsLowest() throws IOException, BundleException {
+		String manifestContent = "Manifest-Version: 1.0\n" //$NON-NLS-1$
+				+ "Bundle-ManifestVersion: 2\n" //$NON-NLS-1$
+				+ "Bundle-SymbolicName: test.bundle\n" //$NON-NLS-1$
+				+ "Bundle-Version: 1.0.0\n" //$NON-NLS-1$
+				+ "Require-Capability: osgi.ee;filter:=\"(&(osgi.ee=JavaSE)(version=19))\",\n" //$NON-NLS-1$
+				+ " osgi.ee;filter:=\"(&(osgi.ee=JavaSE)(version=12))\"\n"; //$NON-NLS-1$
+		Map<String, String> manifest = parseManifestString(manifestContent);
+		String result = ExecutionEnvironmentResolver.resolveCompliance(manifest);
+		assertEquals("Two osgi.ee capabilities (19 first, 12 second) should resolve to 12", "12", result); //$NON-NLS-1$ //$NON-NLS-2$
```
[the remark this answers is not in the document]
DenisUngemach: Done

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
```diff
@@ -0,0 +1,224 @@
...
+				String filterString = element.getDirective(FILTER_DIRECTIVE);
+				if (filterString == null) {
+					continue;
+				}
+				String version = matchJavaSEVersion(filterString);
+				if (version == null) {
+					continue;
+				}
+				if (lowest == null || JavaCore.compareJavaVersions(version, lowest) < 0) {
+					lowest = version;
```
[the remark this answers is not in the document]
DenisUngemach: Yes the Require-Capability are conjunctive. But also "or filters" in one Require-Capability are possible. So this is fixed now. We check conjunctive over Require-Capability.

merks (approved):
It's very well structured and well documented such that the control flow is really easy to follow.  So to me it looks good.  Also, we're early in the cycle, so this is the best time to make such changes.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
```diff
@@ -0,0 +1,238 @@
...
+
+		return getFallbackJavaVersion();
+	}
+
+	/**
+	 * Returns the lowest supported compliance from a comma-separated BREE header,
+	 * or {@code null} if no supported {@code JavaSE-X} entry is found.
+	 */
+	private static String fromBree(String breeHeader) {
+		String[] entries = breeHeader.split(BREE_SEPARATOR);
```
HannesWell: Why didn't you use `ManifestElement.parseHeader()` and let that handle the splitting?

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
```diff
@@ -0,0 +1,238 @@
...
+		if (bree != null) {
+			String result = fromBree(bree);
+			if (result != null) {
+				return result;
+			}
+		}
+
+		// 2. Modern osgi.ee Require-Capability header
+		String requireCapability = manifestMap.get(Constants.REQUIRE_CAPABILITY);
+		if (requireCapability != null) {
```
HannesWell: Since `ManifestElement.parseHeader`, called in `fromRequireCapability()` can handle null, this null check is not required.

HannesWell (changes requested):
Thank you Denis for this one.
The approach totally makes sense and is fine.

But I think you can reuse, at least parts of the already existing `org.eclipse.pde.internal.core.util.ManifestUtils` and their `parseRequiredEEsFromFilter()` method.
Since that class already has `getRequiredExecutionEnvironments(Resource)` I think it makes sense to add an overload `getRequiredExecutionEnvironments(Map)` do what you did in `ExecutionEnvironmentResolver.resolveCompliance()`.
You can then probably reuse some of the existing methods/fields, especially for parsing EE requirements.

The general strategy could then be to extract all required EEs from the Map and return them as stream in `getRequiredExecutionEnvironments(Map)` . A second method could call the latter and could map the stream to version strings.
In case of esoteric EEs that don't have a direct version or are very old, you could again fall back to Java-1.8 or more generally the lowest supported version by JDT.

> * If there are multiple java versions in the manifest, i take the older one.
> 
> * Maybe the newer osgi.ee filter option could make it possible to have more complex java version restrictions. I do ignore other restrictions and take the lowest java version i can find, that is valid by JavaCore.supportedVersions.

If a bundle requires multiple EEs (execution environments), at least in case the BREE header is used, it means that all listed EEs have to be specified. So you can actually use the highest version, because below the bundle doesn't resolve.
For EE requirements you can theoretically combine them with `and` or `or` and the selection would have to happen accordingly.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
DenisUngemach: Done.

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/ExecutionEnvironmentResolver.java
DenisUngemach: Done.

DenisUngemach:
@HannesWell 
I have now moved the functional methods to ManifestUtils. As a result, the resolveCompliance method in ExecutionEnvironmentResolver simply delegates to ManifestUtils methods, which map EE IDs to the highest Java version found.

As @merks recommended, I also believe that using the highest version for parsing is the best approach. Parsing with too old java versions might fail silently, which leads to incorrect or incomplete api description results.

So using a too old java version is worse than using the latest.

HannesWell:
Thanks for the update. I looked into it and found some things that could be improved. In order to speed up the review of this, I pushed my suggestions so far in a separate commit on this PR's branch, after rebasing on top of master. This is not yet ready and some tests are now failing but I wanted to give earlier feedback to discuss with you further (I don't have much more time today).
Generally I leveraged modern java features at some points, mainly in the tests and tried to generalize the code, e.g. to use the JDT API to derived the minimally supported java version instead of hard coding it (in case it changes in the future and to make it more expressive).
Furthermore I think `ManifestUtils` should be very general, i.e. should not filter any EEs based on JDTs support for them.
That part should be moved into the `ExecutionEnvironmentResolver`.
Since that class is now quiet small I wonder if we should just move the one, two methods it contains back to `APIFileGenerator`? WDYT?

Furthermore it should be checked if there are now redundant test cases. While tests are good, they also need to be maintained and should therefore not be redundant. And from first sight I had the impression that there is some overlap between `ExecutionEnvironmentResolverTest` and `ExecutionEnvironmentTestCase`, but I can be wrong since I didn't look into the details.
Additionally we should also check if some of the comments/doc are needed in the current extend. They are quiet verbose (I assume AI generated) and I'm not sure that level of detail is need. Like for test, doc is good, but again needs to be maintained.

But I'm confident that this can be completed soon. Before submission all the commits on the branch needs to be squashed.

Please let me know what you think, about these points and my suggestions.

> So using a too old java version is worse than using the latest.

Agree.

DenisUngemach:
> Thanks for the update. I looked into it and found some things that could be improved. In order to speed up the review of this, I pushed my suggestions so far in a separate commit on this PR's branch, after rebasing on top of master. This is not yet ready and some tests are now failing but I wanted to give earlier feedback to discuss with you further (I don't have much more time today). Generally I leveraged modern java features at some points, mainly in the tests and tried to generalize the code, e.g. to use the JDT API to derived the minimally supported java version instead of hard coding it (in case it changes in the future and to make it more expressive). 

I used the JavaCore class for this analysis...

> Furthermore I think `ManifestUtils` should be very general, i.e. should not filter any EEs based on JDTs support for them. That part should be moved into the `ExecutionEnvironmentResolver`. 

Agree. Actually there is a check to do so:

```
	
					**.filter(JavaCore::isJavaSourceVersionSupportedByCompiler)**
				
```


> Since that class is now quiet small I wonder if we should just move the one, two methods it contains back to `APIFileGenerator`? WDYT?

I also thought about that. Personally i would rather keep the separated class, it splits the topics and simplifies the testing.
With the reduced tests i doubt, that something goes wrong there.

> Furthermore it should be checked if there are now redundant test cases. While tests are good, they also need to be maintained and should therefore not be redundant. And from first sight I had the impression that there is some overlap between `ExecutionEnvironmentResolverTest` and `ExecutionEnvironmentTestCase`, but I can be wrong since I didn't look into the details. Additionally we should also check if some of the comments/doc are needed in the current extend. They are quiet verbose (I assume AI generated) and I'm not sure that level of detail is need. Like for test, doc is good, but again needs to be maintained.

I have removed java doc and unnecessary tests.

> But I'm confident that this can be completed soon. Before submission all the commits on the branch needs to be squashed.
> 

I have updated it. If this is fine, i will squash it. 

> Please let me know what you think, about these points and my suggestions.
> 
> > So using a too old java version is worse than using the latest.
> 
> Agree.

Thanks for the review. Now the code looks much simpler.

DenisUngemach:
One remark to:
"JavaSE/compact1-1.8", "JavaSE/compact2-1.8", "JavaSE/compact3-1.8"

It seems these versions are no longer supported. Is that intended or not?
If not, i will add them in ManifestUtils to the java versions.

DenisUngemach:
Any update? I added the requested lower border. The rest stays the same.

PR #2431, opened 2026-08-28
Stop the model editor stylesheet from painting over every theme

vogella:
The `css/default.css` of `org.eclipse.e4.tools.emf.ui` is contributed without a `themeid`, so it applies in every theme, third-party ones included. It painted three hard `#fff` areas and a fixed grey gradient over whatever the active theme had chosen, and pinned the section header to 15pt regardless of the font size the user picked. Every other stylesheet in this repo binds itself to `e4_default` or `e4_dark`.

Rather than adding a `themeid` and a dark counterpart, which would still leave third-party themes unstyled, this drops the colours and the font size so the editor simply inherits from the active theme. The `.subSectionHeader` and `.contentContainer` rules go with them: no widget carries those classes, they are never assigned anywhere in the repo. Only `font-weight` remains, which no theme needs to override.

Verified in a running IDE against the built-in Light and Dark themes.

### DIFF
--- a/e4tools/bundles/org.eclipse.e4.tools.emf.ui/css/default.css
+++ b/e4tools/bundles/org.eclipse.e4.tools.emf.ui/css/default.css
@@ -1,27 +1,7 @@
-.headerSectionContainer {
-	background-color: gradient #575757 #101010 100%;
-}
-
+/*
+ * Contributed to every theme, so it must not set colours: the model editor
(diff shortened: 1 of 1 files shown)


vogella:
In the platform themes we solve this by having more specific rules, so for platform themes this change does not result in any difference.

PR #2432, opened 2026-08-28
Make the XMI tab of the model editor readable on dark themes

vogella:
On a dark theme the XMI tab renders dark navy tags and dark green attribute values on a dark background. The five XML syntax colours are hardcoded light values in `OSGI-INF/resources.properties` and reach the reconciler through `IResourcePool`, where no stylesheet can see them, and the source viewer never gets a foreground or background at all.

Tag, attribute value, comment and processing instruction become `org.eclipse.ui.themes` `colorDefinition`s, the way `org.eclipse.pde.genericeditor.extension` already does it in this repo, and the source viewer takes its colours from the Text Editors preferences. Any theme can now recolour the tab, which was impossible before.

Independent of #2431; the two share no file.


vogella:
Before:

01-vorher-xmi-tab-dark

After:

02-nachher-xmi-tab-dark

vogella:
Screenshot From 2026-09-01 08-51-52

PR #2436, opened 2026-08-30
Skip unusable API components instead of failing the baseline restore

vogella:
`ApiBaseline.addApiComponents` added every component to the OSGi state without a guard, so one failing component aborted the whole loop and left `resolveSystemLibrary` and `resolve` unexecuted. The baseline then stayed empty and the user only saw "An internal error occurred during: Performing API Analysis", with nothing pointing at the baseline as the cause.

Each component is now handled individually. A component that cannot be added is skipped and logged with its symbolic name, location and baseline name, so the remaining components still load and the failure is diagnosable from the error log alone. Observed in a workspace where two of 684 bundles threw `IllegalStateException: The bundle belongs to another state` from `State.addBundle` and took the entire baseline down with them.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
+++ b/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
@@ -38,6 +38,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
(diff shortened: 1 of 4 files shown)


apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
```diff
@@ -357,13 +358,19 @@ public void addApiComponents(IApiComponent[] components) throws CoreException {
...
+				if (component.isSourceComponent()) {
+					continue;
+				}
+				BundleDescription description = component.getBundleDescription();
+				getState().addBundle(description);
+				addComponent(component);
+				ees.addAll(component.getExecutionEnvironments());
+			} catch (CoreException | RuntimeException e) {
+				// a single unusable component must not abort the whole baseline
+				ILog.of(ApiBaseline.class).warn(MessageFormat.format(CoreMessages.ApiBaseline_7,
```
HannesWell, line 371: Could be a bit simpler:
```suggestion
				ILog.get().warn(MessageFormat.format(CoreMessages.ApiBaseline_7,
```

apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/model/ApiBaseline.java
```diff
@@ -357,13 +358,19 @@ public void addApiComponents(IApiComponent[] components) throws CoreException {
...
+			try {
+				if (component.isSourceComponent()) {
+					continue;
+				}
+				BundleDescription description = component.getBundleDescription();
+				getState().addBundle(description);
+				addComponent(component);
+				ees.addAll(component.getExecutionEnvironments());
+			} catch (CoreException | RuntimeException e) {
+				// a single unusable component must not abort the whole baseline
```
HannesWell, line 370: I'd formulate this more like
```suggestion
				// Skip erroneous component
```

HannesWell (commented):
Generally this sounds reasonable, but could it be that the state and the baseline remain in an inconsistent state, depending on where the exception is thrown?

laeubi:
> Observed in a workspace where two of 684 bundles threw `IllegalStateException: The bundle belongs to another state` from `State.addBundle` and took the entire baseline down with them.

This usually mean something changed in the meanwhile. Instead of ignoring the error, it would be more feasible to restart reading the baseline again I see no point in creating an incomplete baseline here what will lead to even more hard to track errors then, especially as in this case not the bundle is faulty.

At other places we even check (and possibly copy) the bundle from one state to another.

vogella:
AFAICS @HannesWell approves the changes with minor corrections and @laeubi not. What is the way forward?

laeubi:
@HannesWell **left a comment** (as well as me) - non of us explicitly approved or disproved it. 

So if you can reproduce the problem you should deeper look about if it is e.g. only a transient problem (you can just set the API baseline again) or a permanent one (you get the error every time). Depending on that you should decide what is a better way forward here.

In general we should not try to "fix" vague errors with something that is not obviously fixing a bug and I would assume this falls into the category of not being obvious (it just hides an unknown deeper problem).

vogella:
Thanks for the reply, closing, as I have no way to reliable reproduce this.

PR #2437, opened 2026-08-31
Remove the dead buildDisabled flag from ApiAnalysisBuilder

vogella:
`ApiAnalysisBuilder.buildDisabled` is a private static field that nothing ever assigns, so the `buildDisabled ||` disjunct in `build()` was always false. Its Javadoc described a flag that other tools could set, which has not been reachable for a long time.

The `DISABLE_API_ANALYSIS_BUILDER` preference, checked a few lines above and exposed as "Disable API analysis builder" on the Plug-in Development preference page, already provides that workspace-wide off switch properly. No behaviour change.

### DIFF
--- a/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
+++ b/apitools/org.eclipse.pde.api.tools/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.pde.api.tools;singleton:=true
-Bundle-Version: 1.3.1300.qualifier
+Bundle-Version: 1.3.1400.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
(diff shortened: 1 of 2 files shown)


merks (changes requested):
This is being used despite appearances:

Image

Leaving it does no harm.

vogella:
@merks why are you using the official preference for that?

/instance/org.eclipse.pde.ui /instance/org.eclipse.pde.ui Preferences.MainPage.disableAPIAnalysisBuilder

merks:
I needed to paw through history.  What you're removing here was from 2013 while the preference is from 2018; other stuff from Oomph is between those dates, i.e., 2014.  The bugzilla provided some historical detail that would have been good to have taken into consideration:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=221913

I would indeed be better to use the newer preference, not the the key `org.eclipse.pde.internal.core.ICoreConstants.DISABLE_API_ANALYSIS_BUILDER` is actually API mind you...

I'll add that to my TODO list for the next release cycle.

vogella:
Some background on this from me, I also did not know about this preference. :-) 

I discovered the flag you are using in Oomph while trying to optimize the Eclipse build time.

I notice that the flag is only accesable via reflection and started to implement a UI to be able to turn API builder off. Afterwards I found the preference in the code and saw that we already have a official preference. Afterwards I also found it on the UI. 😂


Screenshot From 2026-08-31 10-11-43

merks:
Yes, and given the Disable API analysis builder, the oomph button for that should be removed.  I will also take that into consideration for further simplification.

vogella:
Add tried to fix it in Oomph https://github.com/eclipse-oomph/oomph/pull/202, AFAIR this is my first Oomph contribution so review it with care.

merks (approved):
The Oomph usage is eliminated in favor of the available preference so this looks fine now.

vogella:
Just to be 100% sure, I can merge this? I don't want to break Oomph.

merks:
I appreciate the caution!  Of course this will only change things for 4.42 and the Oomph nightly build will already contains the changes to use the preference.  So in the end, I think mostly only the bleeding edge SDK 4.42 users might be affected but the worse case there is the API build kicks in early and causes strange problems (which the API build tends to do anyway).

So yes, please feel free to proceed.

PR #2438, opened 2026-08-31
Activate the workspace target atomically with its handle preference

vogella:
LoadTargetDefinitionJob stored the new target definition and then wrote its handle memento to the preference in two steps. getWorkspaceTargetDefinition() treats a target whose handle does not match the stored memento as invalid and silently replaces it with a new default target, so a concurrent caller in that gap, such as a builder asking for TargetPlatform.getOS(), swapped the target being loaded for the running platform.

This is the cause of the sporadic PluginBasedLaunchTest failures on Jenkins (for example on #2432 and master 90cb1d22), where the merged bundle map suddenly contains the whole running platform. Both values are now updated under the lock that guards getWorkspaceTargetDefinition(), and TargetPlatformHelper uses the same path when it re-stores the memento of a re-resolved target.

### DIFF
--- a/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target/LoadTargetDefinitionJob.java
+++ b/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/target/LoadTargetDefinitionJob.java
@@ -29,7 +29,6 @@ import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.pde.internal.core.EclipseHomeInitializer;
 import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.core.PDECore;
-import org.eclipse.pde.internal.core.PDEPreferencesManager;
 import org.eclipse.pde.internal.core.target.Messages;
 import org.eclipse.pde.internal.core.target.TargetPlatformService;
 import org.eclipse.pde.internal.core.target.TargetResolveSchedulingRule;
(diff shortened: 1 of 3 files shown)


laeubi (changes requested):
This now losses some comments from the original code, also both call sites has previously done different things (e.g. one never fired a change event) and the second call site now seem to possibly check the memento twice!

I think this has to be carefully checked so we do not unnecessarily send events that trigger other changes again. It would also be beneficial to return a `boolean` from the new method (and possibly having a parameter for the event sending case) so callers can possibly know if there call resulted in a new target being set.

vogella:
This will fix the flaky test: 1 out of 3 runs failed: testGetMergedBundleMap_automaticallyAddRequirements_requirementSharedWithEarlierSortingSiblingFragment (org.eclipse.pde.ui.tests.launcher.PluginBasedLaunchTest)

vogella:
Thanks, all four points were right, reworked.

The call sites are split again. `setWorkspaceTargetDefinition(target, memento, asyncEvents)` is the activation path for `LoadTargetDefinitionJob` and now returns whether a different target was replaced. `TargetPlatformHelper` uses the new `refreshWorkspaceTargetHandle(target)`, which only re-stores the memento, fires no event and compares the memento once. The lost comments are back on the shared helper.

`TargetPlatformHelper` is still touched because its clear-then-set has the same race: while the preference is `""`, a concurrent `getWorkspaceTargetDefinition()` sees a handle mismatch and silently swaps in a fresh default target.

HannesWell:
I'll have to look into this in details, but in the meantime I wonder if with this the change from https://github.com/eclipse-pde/eclipse.pde/pull/2205 can be reverted?

vogella:
> I'll have to look into this in details, but in the meantime I wonder if with this the change from #2205 can be reverted?

I think once the root cause is fixed, the long waiting in the test can be removed or reduced.

PR #2443, opened 2026-09-01
Refresh feature dependency decorations on target platform changes

vogella:
The Dependencies tab of the feature editor puts an error icon on a required plug-in or feature that cannot be found, and `FeatureImport` looks that up live on every call. The section only listened to plug-in and feature model deltas though, and a target reload fires neither, so the icons kept showing the previous result until the editor was reopened.

It now also listens to `IStateDeltaListener.stateChanged`, which is what a reload does fire, and marks itself stale so the next refresh repaints the table. The practical effect is that fixing a missing dependency and reloading the target clears the icon instead of requiring the editor to be closed and reopened.

This is the feature editor counterpart to #2412, which does the same for the manifest editor. The two are independent.


vogella:
Current behavior before the fix:

feature-target-stale

vogella:
This is with the fix:

feature-target-stale-fixed

PR #2444, opened 2026-09-01
Avoid copying all exported packages in ImportPackageObject.isResolved

vogella:
`State.getExportedPackages()` is not cached: every call takes the state monitor and copies the exports of all resolved bundles into a freshly allocated array. `PDELabelProvider` calls `isResolved()` once per row of the Import-Package table, so a refresh rebuilt that array once per visible row on the UI thread.

Following Ed's review, this now uses `State.getExportedPackages(String)` from eclipse-equinox/equinox#1347 rather than duplicating the walk in PDE. The state filters by package name internally and returns only the exports of the queried package, so PDE just checks the version range.

The new method is `@since 3.25` and the current I-build ships it. An earlier revision also added a versioned `Import-Package: org.eclipse.osgi.service.resolver;version="[1.8.0,2.0.0)"`, but that broke the `org.eclipse.pde.ds.tck` build: its `tycho-eclipse-plugin` `eclipse-build` step starts Tycho's own Equinox (3.24.x, exporting the package at 1.7.0), in which `org.eclipse.pde.core` then no longer resolved. The package therefore keeps arriving unversioned through `org.eclipse.core.runtime`, as before.

Verified with `StaleDependencyResolutionTest`: the imported-package tests pass, including a new one for the version-range and no-version cases.


ui/org.eclipse.pde.core/text/org/eclipse/pde/internal/core/text/bundle/ImportPackageObject.java
```diff
@@ -93,14 +93,32 @@ public void restoreProperty(String propertyName, Object oldValue, Object newValu
...
 		VersionRange versionRange = Utils.parseVersionRange(getVersion());
-		return Arrays.stream(exportedPackages)
-				.filter(p -> p.getName().equals(getName()))
-				.anyMatch(p -> versionRange.includes(p.getVersion()));
+		// State.getExportedPackages() copies the exports of every bundle on each call
+		return isExportedBy(state.getResolvedBundles(), versionRange)
+				|| isExportedBy(state.getRemovalPending(), versionRange);
+	}
+
+	private boolean isExportedBy(BundleDescription[] bundles, VersionRange versionRange) {
```
merks: I'm concerned that this makes things more complicated by duplicating the logic that computes exported packages, hopefully in the same way as what the state does.

@HannesWell @laeubi 

Is my concern of concern?

vogella: I also share the concern. AFAICS moving from the State API to ModuleContainer would solve this properly, but that would be a super big rewrite and I'm not aware of any plan for it.

HannesWell: In general there is a plan for that:
- https://github.com/eclipse-pde/eclipse.pde/issues/1069

But indeed it's currently not progressing significantly, but maybe I find some more time in the future. I still have some PRs pending with some steps towards that goal. But there is much more to go.

So for now it would indeed be simpler the extend the OSGi State implantation.
And yes, we should try avoiding duplicating logic.

HannesWell:
> For an SDK-sized target (731 bundles, ~3200 exported packages) that replaces roughly 9600 reference stores per call with 731.

Can you tell what this means in s, ms, µs per operation?

vogella:
API merged in Equinox, this should build green as of tomorrow, when the API is available in I-Build.

vogella:
> > For an SDK-sized target (731 bundles, ~3200 exported packages) that replaces roughly 9600 reference stores per call with 731.
> 
> Can you tell what this means in s, ms, µs per operation?

Sorry, didn't see that questions, the complete trace in available in https://github.com/eclipse-equinox/equinox/pull/1347. The savings were primary in memory.

PR #2457, opened 2026-09-02
Drop the unused timestamp in ConfigurationParser.processConfig

vogella:
The catch block for an unparsable configuration date stored a current-time value in a local that nothing reads afterwards, because the `Configuration` created there already sets the current date itself. With that gone the parsed value is the local's only use, so it is inlined.

No behaviour change, one local and one throwaway `Date` allocation less.


HannesWell (approved):
Looks good, thanks.

PR #2458, opened 2026-09-02
Remove unused internal classes from PDE and the e4 tools

vogella:
Deletes seven internal classes that nothing references: `AbbreviatedPluginHandler`, `QueryControl`, `ILinkLabelProvider`, `MessageLine`, `PersistenceSwitch`, `Autocomplete` and the e4 tools copy of `IXMLColorConstants`. `AbbreviatedPluginHandler` lost its last caller when the abbreviated plug-in models were removed; the rest were already unreferenced.

Checked with a JDT reference search over the full platform workspace and with a text search across the whole releng aggregator, including plugin.xml, manifests and OSGi component descriptors, so neither reflective registration nor another repository is left pointing at them. A full workspace build before and after reports the same 0 errors and no new warnings.


vogella:
Build errors looks like a CI issue, I check later again

PR #2460, opened 2026-09-02
Offer only execution environments that can still be compiled against

akurtakov:
The execution environment lists in PDE offered everything JDT launching
registers. The compiler
dropped support for anything below Java 1.8 long ago, so those cannot be
developed against and only make the lists harder to use.

JDT launching already provides API for that
IExecutionEnvironmentsManager.getSupportedExecutionEnvironments(), added
in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/452 and used by
its own JRE container and Java project wizard pages.

Add VMUtil.getSupportedExecutionEnvironments() on top of it and use it
wherever an execution environment is offered for selection: the manifest
editor section and its source tab content assist, the product and target
editors, the launch configuration JRE tab, the plug-in, fragment,
library and target definition wizards, and the plug-in conversion.

Environments newer than the latest version the compiler supports are
kept. JDT launching contributes those ahead of the compiler front-end so
PDE duplicates the approach.

Code that has to deal with environments already written into a manifest,
such as the builders and the export operations, keeps using the full
list.


akurtakov:
Unstable test failure , there has been effort to stabilize it in https://github.com/eclipse-pde/eclipse.pde/pull/2205 but seems to still cause problems.

vogella:
https://github.com/eclipse-pde/eclipse.pde/pull/2438 should fix the test.

HannesWell (approved):
That makes sense to me and looks good Thanks.

The Javadoc for the new method is IMO very verbose and over detailed.
And maybe returning a List instead of an Array would be more modern, but that requires more cleanup work. Maybe I look into that later.

akurtakov:
I'll do both proposed changes. It's not smth to do in a hurry.

akurtakov:
Both suggestions applied .

akurtakov:
2 new maven warnings are:
```
Received transient HTTP error for https://download.eclipse.org/eclipse/updates/4.41-I-builds/I20260828-1142/plugins/org.eclipse.pde.spy.core_1.1.700.v20260531-1153.jar, retrying (attempt 1/3) in 5 seconds...
Received transient HTTP error for https://download.eclipse.org/eclipse/updates/4.41-I-builds/I20260828-1142/plugins/org.eclipse.pde.spy.core_1.1.700.v20260531-1153.jar, retrying (attempt 2/3) in 10 seconds...
```
Nothing to do for retry. Merging.

PR #2464, opened 2026-09-03
Remove illegal flags and deprecation warnings from StateTree #2462

### LINKED ISSUE #2462: Invalid tree-style combination in StateTree
This tree uses the `SWT.SINGLE` and `SWT.MULTI` style. I.e. both single- and multi-selection are enabled for this tree. That doesn't make much sense.

https://github.com/eclipse-pde/eclipse.pde/blob/ecacdf5360f87c60f6e4a47a1bd356c5e57ba884/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/target/StateTree.java#L58-L66

ptziegler:
When both the `SWT.MULTI` and `SWT.SINGLE` bits are set for a viewer, the `SWT.SINGLE` one takes priority. This means the `SWT.MULTI` flag can be removed.

Additionally, the viewer only works on instances of `BundleDescription` and `DependencyGroup`, whose implementations don't override the default `equals` and `hashCode` method. Because the same bundle and/or
dependency group should not appear multiple times in the viewer, the "fast hash-lookup" flag can be set, thus avoiding the deprecated `FilteredTree` constructor.

Given that it is not possible to select more than one element, the checks done on the selection can also be simplified from `selection.size() == 1` to `!selection.isEmpty()`.

Closes https://github.com/eclipse-pde/eclipse.pde/issues/2462

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/target/StateTree.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/target/StateTree.java
@@ -55,9 +55,8 @@ import org.osgi.framework.Version;
 
 public class StateTree extends FilteredTree {
 
-	@SuppressWarnings("deprecation")
 	public StateTree(Composite parent) {
-		super(parent, SWT.MULTI | SWT.V_SCROLL | SWT.H_SCROLL | SWT.SINGLE, new PatternFilter(), true);
+		super(parent, SWT.V_SCROLL | SWT.H_SCROLL | SWT.SINGLE, new PatternFilter(), true, true);
 		TreeViewer fTreeViewer = getViewer();
 		fTreeViewer.setContentProvider(new StateContentProvider());
 		fTreeViewer.setLabelProvider(new StateLabelProvider());
@@ -250,7 +249,7 @@ public class StateTree extends FilteredTree {
 
 	protected void handleDoubleClick() {
 		IStructuredSelection selection = getViewer().getStructuredSelection();
-		if (selection.size() == 1) {
+		if (!selection.isEmpty()) {
 			BundleDescription desc = getBundleDescription();
 			if (desc != null) {
 				ManifestEditor.open(desc, false);
@@ -260,7 +259,7 @@ public class StateTree extends FilteredTree {
 
 	public BundleDescription getBundleDescription() {
 		IStructuredSelection selection = getViewer().getStructuredSelection();
-		if (selection.size() == 1) {
+		if (!selection.isEmpty()) {
 			Object obj = selection.getFirstElement();
 			if (obj instanceof BundleSpecification) {
 				obj = ((BundleSpecification) obj).getSupplier();


akurtakov:
This has been in SWT since day 0 I dislike - styles are silently "normalized" (e.g. https://github.com/eclipse-platform/eclipse.platform.swt/blob/b8c95fa653e196068a965261efb560ffbee00f14/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Tree.java#L272 ) . It is probably high time to consider some way to report back in non fatal way instead.

ptziegler:
> It is probably high time to consider some way to report back in non fatal way instead.

I don't remember SWT having many options when it comes to reporting errors without throwing an exception. So I suppose this would have to be something like an error printed to System.err?

laeubi:
> So I suppose this would have to be something like an error printed to System.err?

The `Display` already has

- `setRuntimeExceptionHandler`
- `setErrorHandler`

that allows to plug in different strategies.

I think I already suggested a while back one might enhance this for a warning/debug/whatever Handler for such cases that then can e.g. be plugged into something like the usual Eclipse Logging/SLF4J/System.out/whatever.

laeubi:
> When both the `SWT.MULTI` and `SWT.SINGLE` bits are set for a viewer, the `SWT.SINGLE` one takes priority. This means the `SWT.MULTI` flag can be removed.

Just to be pedantic here, it might be that `SWT.MULTI` is actually intended and this is only a bug so `SWT.SINGLE` should be removed ;-)

ptziegler:
> Just to be pedantic here, it might be that SWT.MULTI is actually intended and this is only a bug so SWT.SINGLE should be removed ;-)

The context menu for an element doesn't appear when more than one element is selected. So multi-selection is completely useless here.

image

Reason being that the `BundleDescription` will always be `null`, then.

https://github.com/eclipse-pde/eclipse.pde/blob/54c463f9c9607d6953d9416614ef290f2a4dc40d/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/views/target/StateTree.java#L261-L275

So one could probably clean up this viewer even further and check whether the selection is empty, rather than of size 1.

ptziegler:
After going through the history: The viewer was initially created with b9966c0f5b79dab1b51749fe02b00ff4e833c082, with the `SWT.MULTI` flag set. This was later changed with acedf131cfc95271ea347f405a5a54de2a193139, likely as part of the fix for [Bug 182215](https://bugs.eclipse.org/bugs/show_bug.cgi?id=182215) - [Plug-in Dependencies view] Initial empty state may be confusing. My guess is as part of this:

> 4. Added a context menu that matches the one found in the traditional dependencies view (minus "focus on...").  

To me, that means the fact that the `SWT.MULTI` flag remains is nothing but an oversight.

ptziegler:
> The Display already has

> - setRuntimeExceptionHandler
> - setErrorHandler

> that allows to plug in different strategies.

You're right. At least with the Eclipse platform, you could forward an `AssertionError` to the error handler, which will then be converted to a `Status` and logged using the `ILog`.

PR #2467, opened 2026-09-03
Stop Smart Import from proposing projects nested inside bundles

### LINKED ISSUE #2466: Avoid imports nestled test projects in PDE
The Smart Import wizard scans a directory tree for every `.project` file and proposes each hit as a project to import.
That includes `.project` files that only exist as test fixtures inside a bundle, for example the ones below `org.eclipse.pde.ui.tests/tests/projects`.
Importing the platform aggregator this way ends up creating hundreds of fixture projects in the workspace, and they are indistinguishable from the real ones in the Project Explorer:

Image

vogella:
The Smart Import wizard scans for every `.project` file and proposes each one, so test fixtures below a bundle or feature (for example under `org.eclipse.pde.ui.tests/tests/projects`) end up imported as real projects. Importing the platform aggregator this way yields hundreds of such fixture projects.

This makes PDE treat a bundle or feature project as a leaf for Smart Import: `BundleProjectConfigurator` and `FeatureProjectConfigurator` drop every proposal that lies strictly below a bundle or feature, while nested projects below plain projects such as an aggregator root are still proposed. Nesting a bundle inside another bundle does build and work, so this is a deliberate PDE decision rather than a rule imposed from outside: the fixture noise is far more common than the nested-bundle layout, and anyone who does want the nested projects can still use the classic Import Projects wizard, which is unchanged.

Covered by a new `NestedProjectSmartImportTest` with bundle, feature and plain-project fixtures. Touched bundles compile locally with Tycho.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2466

### DIFF
--- a/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/AllPDETests.java
+++ b/ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/AllPDETests.java
@@ -42,6 +42,7 @@ import org.eclipse.pde.ui.tests.target.AllTargetTests;
 import org.eclipse.pde.ui.tests.util.PDELabelProviderTest;
 import org.eclipse.pde.ui.tests.views.log.AllLogViewTests;
 import org.eclipse.pde.ui.tests.wizards.AllNewProjectTests;
+import org.eclipse.ui.tests.smartimport.NestedProjectSmartImportTest;
 import org.eclipse.ui.tests.smartimport.ProjectSmartImportTest;
 import org.junit.platform.suite.api.SelectClasses;
 import org.junit.platform.suite.api.Suite;
@@ -75,6 +76,7 @@ import org.junit.platform.suite.api.Suite;
 	BundleErrorReporterTest.class, //
 	AllPDECoreTests.class, //
 	ProjectSmartImportTest.class, //
+	NestedProjectSmartImportTest.class, //
 	GatherUnusedDependenciesOperationTest.class, //
 	PDELabelProviderTest.class, //
 	ProductInfoSectionTest.class, //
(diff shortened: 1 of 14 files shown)


laeubi:
While the change itself sounds reasonable, the rationale is completely odd. This has absolutely nothing to do with the OSGi spec, as the spec only defines how bundles are shaped in the **runtime**. It does not define anything **how these are build** or mandates or forbids any structure.

So a bundle project can legitimate have any structure here, and it would even work to have one bundle `A` that contains bundle `B`+`C `in a subfolder (and you can import and compile it without a problem).

So if we decide to do it that way, please don't make this look like if it where mandated by the spec but is a decision of the PDE project about how we want smart import to behave.

vogella:
@HannesWell do you have feedback?

HannesWell:
> This makes PDE treat a bundle or feature project as a leaf for Smart Import: `BundleProjectConfigurator` and `FeatureProjectConfigurator` drop every proposal that lies strictly below a bundle or feature, while nested projects below plain projects such as an aggregator root are still proposed.

That sounds reasonable and I think your following assessment is correct, that the most of the time it's not wanted to import projects nested within Plug-in project.
But I personally in fact have scenarios where I want to do that occasionally. So I'd be in favor of not making it too hard. At the same time, adding yet another option would further clutter the UI.
Using the plain import wizard is fine for that.
Additionally I wonder if it would still be possible to import nested projects, if I specify the containing Plug-in project as source?
grafik

HannesWell:
While thinking about it further, would it be possible to still show projects nested in a Plug-in project, but uncheck them by default?
Wouldn't that already make the import from large git repos simpler since the nested projects are excluded by default, while also keeping it relatively simple to import nested projects if desired? That way deep nested project would remain visible.

merks:
The default checkmark idea sounds cool. 😎

vogella:
> While thinking about it further, would it be possible to still show projects nested in a Plug-in project, but uncheck them by default? Wouldn't that already make the import from large git repos simpler since the nested projects are excluded by default, while also keeping it relatively simple to import nested projects if desired? That way deep nested project would remain visible.

I think that is currently not supported by the smart import wizard, you can only define exclusions not "unflagged projects".

HannesWell:
> I think that is currently not supported by the smart import wizard, you can only define exclusions not \"unflagged projects\".

Then I suggest to extend the smart import Wizard with that functionality. Maybe that's also useful for other kind of projects, like Maven.
Or is there a reason speaking against it?

PR #2472, opened 2026-09-06
[RelEng] Cleanup github workflows

HannesWell:
Remove unused updateRelease.yml (the called workflow doesn't exist anymore for some time).
Pin third-party action to a commit sha.


HannesWell:
Jenkins build seems to be back on track now, but certainly not due to this change. So it probably was an infrastructure issue in the first place.

PR #2473, opened 2026-09-07
Add --enable-native-access=ALL-UNNAMED to default launch VM arguments

akurtakov:
Add --enable-native-access=ALL-UNNAMED to default launch VM arguments
When the workspace default JRE is Java 24 or newer, new PDE launch
configurations (Eclipse application, OSGi framework, JUnit plug-in
tests) get --enable-native-access=ALL-UNNAMED in their initial VM
arguments. Because it lands in the Arguments tab like the other
defaults, users can edit or remove it per configuration. An existing
--enable-native-access argument in the target platform is respected.

Starting with JDK 24 (JEP 472) the JVM prints warnings like

  WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for
  callers in this module
  WARNING: Restricted methods will be blocked in a future release unless
  native access is enabled

whenever class-path code uses JNI, which every Eclipse application does
through SWT.

Assisted-by: Anthropic Claude Code (claude-fable-5-1)

### DIFF
--- a/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchArgumentsHelper.java
+++ b/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchArgumentsHelper.java
@@ -43,6 +43,7 @@ import org.eclipse.jdt.launching.AbstractVMInstall;
 import org.eclipse.jdt.launching.ExecutionArguments;
 import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.jdt.launching.IVMInstall;
+import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.pde.core.plugin.IPluginModelBase;
 import org.eclipse.pde.core.plugin.ModelEntry;
 import org.eclipse.pde.core.plugin.PluginRegistry;
@@ -59,6 +60,11 @@ import org.eclipse.pde.launching.IPDELauncherConstants;
 import org.osgi.framework.Bundle;
 
 public class LaunchArgumentsHelper {
+
+	/**
+	 * VM argument that enables native access for all code on the class-path.
+	 */
+	public static final String ENABLE_NATIVE_ACCESS_VM_ARGUMENT = "--enable-native-access=ALL-UNNAMED"; //$NON-NLS-1$
 
 	/**
 	 * Returns the location that will be used as the workspace when launching or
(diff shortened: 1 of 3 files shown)


HannesWell (commented):
I understand the desire to add this flag and I assume many users are happy to see less warnings. But I'm not sure if that's wanted by all or at least the vast majority of users. And I cannot see we do something comparable already.
What could a user do, if that flag is not desired?

For launch configurations this might not be too much of a concern, but for built products it should really be the decision of the product builder to allow all native access. Also from a security POV, that should be a deliberative decision. But that's not in question here.

What do the others think about this?

akurtakov:
> I understand the desire to add this flag and I assume many users are happy to see less warnings. But I'm not sure if that's wanted by all or at least the vast majority of users. And I cannot see we do something comparable already. What could a user do, if that flag is not desired?

The flag is put in 
image

and one can delete it from there . Further usage of the launch config will start without the flag. In the product case specifically, if one uses 
image
the parameter will be put by default but one can delete it. Second usage of these links adds it back though.

Actually, PDE does something worse - it adds --add-modules=ALL-SYSTEM always in a way that user can't remove in https://github.com/eclipse-pde/eclipse.pde/blob/523f495f87650156d3548ecbbc1aba9974c6b5c6/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/AbstractPDELaunchConfiguration.java#L182 . This leads to warnings like 
```
WARNING: Using incubator modules: jdk.incubator.vector
``` 
and there are a number of RCP products (e.g. chemclipse) that don't need this setting.

merks:
Does this apply to all launches?  Even to plain old Java applications doing any old thing?

Given it avoids this warning about swt:

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::loadLibrary has been called by org.eclipse.swt.internal.Library in an unnamed module (file:/D:/Users/merks/platform-sdk-4.41/ws/.metadata/.plugins/org.eclipse.pde.core/.install_folders/1788418807689/../../../../../../git/eclipse.platform.swt/binaries/org.eclipse.swt.win32.win32.x86_64/)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```
I expect everyone launching an application with swt will want this flag.  We added it to the sdk product, but I don't think that necessarily helps....

akurtakov:
It is needed even in plain old launches when swt is used. But I'm not adding it there as not every project will be using swt. For PDE launches I would dare to say it's safe assumption that majority of users will be using swt/launcher/filesystem/etc. and all of them will load native libraries.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchArgumentsHelper.java
```diff
@@ -171,9 +177,36 @@ public static String getInitialVMArguments() {
 			}
 		}
 
+		if (isNativeAccessArgumentRequired() && result.indexOf("--enable-native-access") == -1) { //$NON-NLS-1$
```
HannesWell: There is IMO no reason to repeat the bad patterns from above.
```suggestion
		if (isNativeAccessArgumentRequired() && !result.contains("--enable-native-access")) { //$NON-NLS-1$
```

HannesWell (approved):
Thanks for your elaboration in https://github.com/eclipse-pde/eclipse.pde/pull/2473#issuecomment-5579459008, that makes sense and sounds like a good compromise.

The code is fine, I just have one nit-pick.

ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchArgumentsHelper.java
akurtakov: I have decided to even change the previous one for consistency sake.

PR #2475, opened 2026-09-09
Validate public access modifier for DS constructor injection

### LINKED ISSUE #1218: [scr 1.4] constructor for injection must be public

elsazac:
This PR provides a validation check for DS constructor injection for a non-public injectable constructor. when an injectable constructor is found but is not public, a specific error message is reported to clearly indicate that the constructor must be public. 

Fixes: https://github.com/eclipse-pde/eclipse.pde/issues/1218

Before :

https://github.com/user-attachments/assets/0b609e5f-6c45-4437-8f6d-25f08ac93055

After : 

https://github.com/user-attachments/assets/3cdfc06c-7a7e-4d29-979c-187898cc4e9b

### DIFF
--- a/ds/org.eclipse.pde.ds.annotations/META-INF/MANIFEST.MF
+++ b/ds/org.eclipse.pde.ds.annotations/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.pde.ds.annotations;singleton:=true
-Bundle-Version: 1.4.300.qualifier
+Bundle-Version: 1.4.400.qualifier
 Bundle-Activator: org.eclipse.pde.ds.internal.annotations.Activator
 Bundle-Vendor: %Bundle-Vendor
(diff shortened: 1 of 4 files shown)


laeubi:
Its bit unclear here as you are talking about constructor injection but your video seem to suggest that you have not enabled proper DS level (it still complains that `@Activate` is not valid at this place).

elsazac:
Thanks for pointing that out. There was some confusion with the DS level settings in my initial video. I have corrected the setup now and enabled the appropriate DS level to properly show the constructor injection behavior.

Before :

image


After :

https://github.com/user-attachments/assets/af238056-2030-4f2e-b22f-f8a3fb71fb95

elsazac:
@laeubi Does this look okay now?

elsazac:
@nburnwal09 Can you review this?

PR #2477, opened 2026-09-09
[poc] using the Java-24 classfile API instead of ASM

laeubi:
As ASM tends to become cumbersome to maintain, this is a small proof-of-concept using the new [java.lang.classfile](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/classfile/package-summary.html) API instead.


HannesWell:
Thank you Christoph for working on this.
We also discussed about replacing ASM in the PMC call last week and @mpalat suggested that we could also use JDT's disassembler API to read class-files and walk their AST.

This could avoid the restriction that we can only consider language elements known to the running JVM's Java version and could allow to analyze code targetting newer Java versions with constructs unknown to the running Java version.

Of course we would depend on JDT supporting that Java version, but we generally do that in PDE already, so that wouldn't be a new restriction.

What do you think about that and would you be interested in investigate that alternative too?

mpalat:
> Thank you Christoph for working on this. We also discussed about replacing ASM in the PMC call last week and @mpalat suggested that we could also use JDT's disassembler API to read class-files and walk their AST.
> 
> This could avoid the restriction that we can only consider language elements known to the running JVM's Java version and could allow to analyze code targetting newer Java versions with constructors unknown to the running Java version.

We are planning to look at this option of JDT Disassembler and any changes required for the same. In jdt.debug, there is this work https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/994 ongoing on a related note to moving away from ASM.

See also https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5418

PR #2480, opened 2026-09-10
Reuse the platform syntax colors in the target and XMI editors

vogella:
The target definition editor and the XMI tab of the model editor each carried their own color definitions, dark stylesheet and Colors and Fonts category for the same XML syntax coloring. Both now read the shared colors from eclipse-platform/eclipse.platform.ui#4372, so the "Target file editor" and "EMF ModelTooling XMI Tab" entries disappear from the preference page.

CI stays red until the platform change is published, `org.eclipse.ui.editors 3.23.0` cannot resolve before that. The light colors of the target definition editor shift a little since it was the outlier of the three, dark is unchanged, and customizations of the removed ids are not migrated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vogella:
compare-prefs-dark
compare-prefs-light
compare-target-dark
compare-target-light
compare-xmi-dark
compare-xmi-light

PR #2481, opened 2026-09-11
Added a reload icon for the "Reload Target platform"

### LINKED ISSUE #2300: Using a reload icon for the "Reload Target platform"
Would it be ok to use a reload icon in the toolbar of the Target editor instead of text?

Image

We could also use this icon in the code minings for the target platform to make is easier to discover this. Currently most the end users I have seen using the target editor regard this text as a comment.

Image

nburnwal09:
Fixes: #2300
Replaced the text of Reload Target platform" with a refresh icon. The  text (depending on state) will change based on the state.

Before state(Set as active target platform)

Set as active target platform

After state(Reload target platform)
Reload target platform

### DIFF
--- a/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
+++ b/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/editor/targetdefinition/TargetEditor.java
@@ -46,7 +46,6 @@ import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.eclipse.e4.core.services.events.IEventBroker;
 import org.eclipse.jface.action.Action;
-import org.eclipse.jface.action.ControlContribution;
 import org.eclipse.jface.dialogs.ErrorDialog;
 import org.eclipse.jface.dialogs.IMessageProvider;
 import org.eclipse.jface.text.DocumentEvent;
@@ -80,9 +79,7 @@ import org.eclipse.pde.internal.ui.shared.target.ITargetChangedListener;
 import org.eclipse.pde.internal.ui.shared.target.TargetContentsGroup;
 import org.eclipse.pde.internal.ui.shared.target.TargetLocationsGroup;
 import org.eclipse.pde.internal.ui.wizards.exports.TargetDefinitionExportWizard;
-import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.BusyIndicator;
-import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IEditorInput;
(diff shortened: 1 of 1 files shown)


laeubi:
@nburnwal09 please show before / after screenshots and different states!

nburnwal09:
> @nburnwal09 please show before / after screenshots and different states!

@laeubi 
I have added the screenshots in the PR description. Kindly check them.

vogella:
The new activate icon looks IMHO to similar to the Export icon. What about something like the new icon in the active launch configuration?

Screenshot From 2026-09-15 10-46-14

nburnwal09:
@vogella 
I couldn't understand from the screenshot. Which new icon are you referring to?

vogella:
> @vogella I couldn't understand from the screenshot. Which new icon are you referring to?

Sorry, here is an isolated screenshot. 

Screenshot From 2026-09-15 10-52-54

This looks similar to what I get if I google for "Icon for activate"

Screenshot From 2026-09-15 10-53-20

Edit: It is not an icon but the Unicode gear character \u2699 (⚙) appended to the menu label, added in commit c38536352d "Add active launch indicator to launch history menu" (Sougandh S, May 2026) in eclipse.platform

nburnwal09:
> > @vogella I couldn't understand from the screenshot. Which new icon are you referring to?
> 
> Sorry, here is an isolated screenshot.
> 
> Screenshot From 2026-09-15 10-52-54

Yeah, this looks better for activate icon. Will push it

nburnwal09:
@vogella 

Here is the screenshot after updating with Unicode gear and refresh character. But the sizing are not in proportionate with other icons alongside. Also resizing it will not be simple and might lead to unclear images. 
What do you suggest?

image

image

merks:
Mixed emojis and icons looks poor and seems not a good idea.

vogella:
> @vogella
> 
> Here is the screenshot after updating with Unicode gear and refresh character. But the sizing are not in proportionate with other icons alongside. Also resizing it will not be simple and might lead to unclear images. What do you suggest?

Maybe find them as SVG? For example https://fonts.google.com/icons?icon.query=refresh and a similar icon like https://fonts.google.com/icons?selected=Material+Symbols+Outlined:mode_off_on:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=activate&icon.size=24&icon.color=%231f1f1f 

AFAIK the svg you download from this side are free and have been used before in Eclipse code.

merks:
It’s probably good to keep @BeckerWdf in the loop on icons for consistency and for consideration in the dual tone icon pack.

BeckerWdf:
> Mixed emojis and icons looks poor and seems not a good idea.

I absolutely agree. We should not do that.
In discussions on the Open Community Meetup yesterday we hear a lot of opinions about icons in the IDE and that we should get away from designing new dedicated icons for very similar features. Reusing same icons for very similar features would be better - for sure this must not lead to confusion for the user.

BeckerWdf:
> The new activate icon looks IMHO to similar to the Export icon. What about something like the new icon in the active launch configuration?
> 
> Screenshot From 2026-09-15 10-46-14

I am not a big fan of this either: Having two icons in a drop down is a concept we don't have in any other places in the ide. If would find it better e.g. to have a (running) suffix added to the text of the item instead of an icon. Because user first need to lear what the "gear" symbol means - it's much easier to read the text instead. But that's a different dsicussion.

nburnwal09:
@vogella 
I have added the svg icons. Here is the recording. 
It looks a bit different from the icons beside it. Let me know your suggestions.

https://github.com/user-attachments/assets/e707a881-cbf1-4ffe-899d-3f8a9f427f1c

vogella:
Lgtm

BeckerWdf:
Does absolutely not look good to me. 
It looks very odd to me. The icon looks "alien" next to the other icons.
@jasmin261098 what do you think?

BeckerWdf (changes requested):
icon looks alien

iloveeclipse:
> icon looks alien

Fully agree with that. It is mixing two fully different UI themes in same widget.

PR #2482, opened 2026-09-11
Organize manifest: retain dependencies providing own packages

### LINKED ISSUE #2356: "Find unused dependencies" does not properly consider split packages
Let's assume you have bundle `A` exporting package `p` and bundle `B` with a dependency to `A` and classes in package `p` using classes from package `p` of bundle `A`. Then the "find unused dependencies" operation will report the dependency from `B` to `A` being unused. Without having looked at the code yet, I guess this is because the classes of bundle `B` in package `p` do not have any import of classes from bundle `A`, because they are placed in the same package. The class of interested should be this: https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java

The following archive contains projects demonstrating that situation: [demo.zip](https://github.com/user-attachments/files/28643003/demo.zip)

Image

While split packages may be considered problematic in general, this situation occurs for us with a bunch of test bundles that have the test classes placed in the same package as the tested production code classes.

I plan to put this issue into the backlog of my team. But in case anyone has input the issue or a potential solution, please share that, or if anyone wants to pick this issue before we find the time to do so, of course feel free to do so.

### LINKED ISSUE #2357: "Find unused dependencies" does not consider dependencies for using method return types
Let's assume you have bundle `A` exporting a type `T`, bundle `B` exporting a method `m` that returns `T`, and bundle `C` consuming method `m` and executing a method on `T`. Then `C` requires a dependency to bundle `A` to be able to execute that method. However, when using "find unused dependencies" on bundle `C`, the dependency to `A` will be removed.
The class of interested for this functionality should be this: https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java

The following archive contains projects demonstrating that situation: [demo2.zip](https://github.com/user-attachments/files/28643545/demo2.zip)

Before "find unused dependencies":
Image

After removing unused dependencies:
Image

Of course, it's not nice that as a consumer of public methods you potentially need to define additional dependencies for being able to use the return type of the methods. But if I am not mistaken, the only way to avoid this would be some kind of reexport and it's not the idea to do this for every type used in some API, is it?

I plan to put this issue into the backlog of my team. But in case anyone has input the issue or a potential solution, please share that, or if anyone wants to pick this issue before we find the time to do so, of course feel free to do so.

HeikoKlare:
The analysis for unused dependencies, which backs both "Find unused dependencies" and "Organize Manifests", reports a required bundle as unused although the project does use it, whenever the used package is one that the project provides itself as well. Following that proposal breaks the project. This fixes that.

Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2356
Fixes https://github.com/eclipse-pde/eclipse.pde/issues/2357

### Why the usage becomes invisible

Whether a dependency is used is derived from the packages that the project's byte code imports, as computed by bnd. Those imports are by definition the referred packages reduced by the packages that the project provides itself. A reference into a package that the project provides itself, while a required bundle provides that same package as well, is therefore subtracted again as being contained and becomes invisible. The subsequent search of the Java model does not compensate for it either, because that search looks for references to the package, which a reference from within that same package never produces, as it needs no import statement.

This affects any package that is present in more than one bundle, no matter whether it is a split package declared as such, and it occurs regularly for test bundles that place their tests in the packages they test.

### Why this takes the conservative route

I first tried to implement a complete analysis that determines whether something from such a package is actually used from the required bundle. That analysis turned out to become quite complex: the referred types have to be read from the constant pool of the project's classes and matched against the types that the dependency provides, and even then the result stays incomplete, because the compiler inlines compile-time constants and leaves no trace of the type providing them, so the sources would have to be analyzed on top of that.

The conservative rule implemented here is a single condition instead: a required bundle that exports a package which the project provides itself is considered used. That gives up precision, as a genuinely unused dependency of that shape is no longer reported. The constellation should occur rarely, though, and in it retaining one dependency too many is far less harmful than proposing to remove one that is needed, which is what happens today. That does not justify introducing the complexity of a complete analysis for such a seldom case.

### How to test

The archives attached to the linked issues reproduce the situation: import the projects, then invoke "Find unused dependencies" or "Organize Manifests" on the consuming bundle. The required bundle that is actually used is no longer proposed for removal. Both constellations are also reflected by automated tests added with this pull request.


laeubi:
So do I see it correct that you are talking about a split package here? While I think this would be legal to retain I wonder if we not simply should add the ones to the required packages instead of having two sets (and two if cases) if it is not needed elsewhere to separate the two.

Also one would need to carefully check if that case that bundle has an own package `a.b.c` but imports (unnecessary) package `a.b` if it will be removed then? So actually we talking about a very specific case here require bundle+split package what are both have pitfall on their own we should at least make sure it does not break the common case.

HeikoKlare:
Yes, it is effectively a split package we are talking about. Maybe it is also helpful to repeat my statement from the original issue here, that points to the use cases we have where the "Organize Manifest" tooling currently breaks for us:
> While split packages may be considered problematic in general, this situation occurs for us with a bunch of test bundles that have the test classes placed in the same package as the tested production code classes.

So you mean we should simply add the own packages to the required ones? That would indeed have the same result with less code. I just did the split for the sake of clarity, as the computed `requiredPackages` are (other than the `ownPackages`) not an over-estimation of what should be preserved as dependencies. But merging them into a single set would technically be absolutely fine.

> Also one would need to carefully check if that case that bundle has an own package `a.b.c` but imports (unnecessary) package `a.b` if it will be removed then? So actually we talking about a very specific case here require bundle+split package what are both have pitfall on their own we should at least make sure it does not break the common case.

Indeed, that case must not break. The proposed implementation is compatible with that expectation as packages are not hierarchical (though they look like they would be), so `a.b.c` is not conceptually included in `a.b` or the like, but it definitely makes sense to pin that behavior with a test. I have added an according test to the PR.

HeikoKlare:
> I wonder if we not simply should add the ones to the required packages instead of having two sets (and two if cases) if it is not needed elsewhere to separate the two.

I have now integrated the own/split packages into the computed packages, so the change becomes even more simple, and the different characteristics of the two contributions to the set are explained via code comment. @laeubi I hope that now fully addresses your comment?

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java
```diff
@@ -83,16 +83,22 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException, Inte
...
+			if (imports != null) {
+				imports.keySet().stream().map(PackageRef::getFQN).forEach(computedPackages::add);
 			}
+			// A reference into a package that the project provides itself is
+			// invisible in the computed packages, because those are the packages
+			// referred to by the byte code reduced by the ones the project
+			// contains. Detecting such a reference would require analyzing the
+			// sources, so the packages that the project provides itself are
+			// treated as required: that retains dependencies which are in fact
+			// unused but never removes one that is needed.
```
r-mennig, line 100: This is not strictly true. Since the analyzer is looking at the whole bundle, it could potentially track all references to the split package name in the byte code and remove all references declared by the bundle's own package. If no other references remain after this, the split packages from other bundles are not actually used and the dependency could be removed.

I'm not sure if this is a feasible solution, I've just wanted to point this out.

ui/org.eclipse.pde.ui.tests/src/org/eclipse/pde/ui/tests/search/dependencies/GatherUnusedDependenciesOperationTest.java
```diff
@@ -86,6 +98,356 @@ public class UsesC {
...
+				public class B {
+					public String name() {
+						return A.NAME;
+					}
+				}
+				""");
+
+		buildProjects();
+		List<String> unusedPlugins = gatherUnusedDependencies(projectB);
+		assertFalse("Dependency to bundle A must not be flagged as unused although its inlined constant is used",
```
r-mennig, line 198: Typo: although -> since (the same typo is present in other assertions)

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java
r-mennig: I've realized this would not work for cases in which just the inlined constant from the other bundle's split package is used since no reference to the split package name is left in the byte code. So I think you're right here.

ui/org.eclipse.pde.ui/src/org/eclipse/pde/internal/ui/search/dependencies/GatherUnusedDependenciesOperation.java
r-mennig, line 132: If we hit this line (or line 144) and the analyzed package is also contained in this bundle, we know that the package is split across multiple bundles. While we might not know if the contents of the other bundle's split package are actually used, what could be done here is to inform the user of this inaccuracy as a pointer to further manual cleanup.

r-mennig (approved):
With these changes, split packages defined by both the analyzed bundle and imported from (one or more) other bundles are now always considered used. This is  a reasonable assumption in my mind, as it may lead to unused dependencies not being cleaned up, but fixes the previous errors in which actual dependencies were removed. The added unit tests cover all the potential test cases I could think of.

PR #2479, opened 2026-09-10
Add system property to disable framework state dump in RemotePluginTestRunner

trancexpress:
Dumping the state of all bundles in `RemotePluginTestRunner` can result in 1000+ lines in the console, when running a single test, making logs more difficult to read.

This change adds a system property to disable the state dump for cases, where unresolvable bundles is a known problem with no pending solution. The unresolvable bundles are still printed on standard error.

PR #2487, opened 2026-09-16
Migrate org.eclipse.pde.ua.tests to JUnit Jupiter

vlakn:
Replace legacy JUnit 3/4 imports with JUnit Jupiter (org.junit.jupiter.api)
and JUnit Platform Suite (org.junit.platform.suite.api)


HannesWell:
Thanks for providing this. The change generally looks sound.

> 117 files   - 12    117 suites   - 12   35m 50s ⏱️ + 5m 6s
> 3 551 tests ± 0  3 497 ✅ ± 0   54 💤 ±0  0 ❌ ±0 
> 9 402 runs   - 54  9 272 ✅  - 54  130 💤 ±0  0 ❌ ±0

It makes me wonder, why there are 54 test runs less recorded. Has something been removed?
In Jenkins the same number of tests is reported.

HeikoKlare:
> It makes me wonder, why there are 54 test runs less recorded. Has something been removed?

The numbers look quite inconclusive. The report mentions only less test runs without a reduced number of tests. In addition, 12 removed files / test suites are reported, which is even more than the number of touched files. Even with different counting methods for JUnit 5 compared to JUnit 4 due to differences in test suite execution, I cannot relate those numbers. And in the Platform repositories, I am quite sure that we have seen consistent numbers on JUnit 4 to 5 migration.

Maybe we should just rebase the PR on master again to have another run and verify the numbers?

HannesWell:
> Maybe we should just rebase the PR on master again to have another run and verify the numbers?

Yes that seems to have changed things. Now tests are added according to the summary.
I assume we can say that due to the change from Junit 4 to 5 confuses the summary.

HannesWell (approved):

HeikoKlare:
> Now tests are added according to the summary.

That's because the previous master build was not successful and had a wrong number of test runs reported. The actual numbers are still wrong. The bot reported 9272 successful test case runs (the number as before, where -54 were reported). Other PRs still report a number that is higher by 54, such as https://github.com/eclipse-pde/eclipse.pde/pull/2479:
image

That's strange. At least the build logs show that the migrated tests are properly executed.

